PR #20, opened 2024-02-29
EVEREST-726 use internal manifest

oksana-grishchenko:
EVEREST-726 

Now the manifest file is in the same repo as the cli, so there is no need for the logic that differentiate the branches to pick up the manifest from.

### DIFF
--- a/deploy/quickstart-k8s.yaml
+++ b/deploy/quickstart-k8s.yaml
@@ -53,7 +53,7 @@ roleRef:
 subjects:
   - kind: "ServiceAccount"
     name: everest-admin
-    namespace: percona-everest
+    namespace: everest-system
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -47,7 +47,7 @@ type Install struct {
 }
 
 const (
-	everestBackendServiceName = "percona-everest-backend"
+	everestBackendServiceName = "everest"
 	everestOperatorName       = "everest-operator"
 	pxcOperatorName           = "percona-xtradb-cluster-operator"
 	psmdbOperatorName         = "percona-server-mongodb-operator"
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -41,7 +41,7 @@ var (
 	catalogImage string //nolint:gochecknoglobals
 )
 
-// CatalogImage returns a catalog image needed for the build of everestctl
+// CatalogImage returns a catalog image needed for the build of everestctl.
 func CatalogImage() string {
 	catalogImage = devCatalogImage
 	v, err := goversion.NewSemver(Version)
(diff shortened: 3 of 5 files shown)


deploy/quickstart-k8s.yaml
```diff
@@ -53,7 +53,7 @@ roleRef:
 subjects:
   - kind: "ServiceAccount"
     name: everest-admin
-    namespace: percona-everest
+    namespace: everest-system
```
oksana-grishchenko, line 56: Is this SA actually used? It was configured for an unexisting namespace, but nothing is failing bc of this.

pkg/install/install.go
```diff
@@ -47,7 +47,7 @@ type Install struct {
 }
 
 const (
-	everestBackendServiceName = "percona-everest-backend"
+	everestBackendServiceName = "everest"
```
oksana-grishchenko, line 50: This was causing [the restart problem](https://github.com/percona/everest/actions/runs/8108835236/job/22163432926) - when everest is already installed we were trying to restart the pods with the label `app.kubernetes.io/name: percona-everest-backend` which doesn't exist anymore.

in Beta we still use the [quickstart-k8s.yaml](https://github.com/percona/percona-everest-backend/blob/main/deploy/quickstart-k8s.yaml) from the https://github.com/percona/percona-everest-backend main branch, where we use that old labels for the resources, that why the problem didn't show up before we actually started to use quickstart-k8s.yaml from the monorepo with the new labels

pkg/version/version.go
```diff
@@ -28,8 +28,6 @@ import (
 const (
 	devCatalogImage     = "docker.io/perconalab/everest-catalog:latest"
 	releaseCatalogImage = "docker.io/percona/everest-catalog:%s"
-	devManifestURL      = "https://raw.githubusercontent.com/percona/percona-everest-backend/main/deploy/quickstart-k8s.yaml"
-	releaseManifestURL  = "https://raw.githubusercontent.com/percona/percona-everest-backend/v%s/deploy/quickstart-k8s.yaml"
```
recharte, line 32: :raised_hands:

pkg/version/version.go
```diff
@@ -43,10 +41,7 @@ var (
 	catalogImage string //nolint:gochecknoglobals
 )
 
-// CatalogImage returns a catalog image needed for the build of everestctl
```
recharte, line 46: Even though the rest of the comment is wrong, this specific line should not be removed

recharte (approved):

pkg/version/version.go
oksana-grishchenko: And it's not, it can be found below :) for some reason the diff tool considered the string as fully removed and added below

deploy/quickstart-k8s.yaml
oksana-grishchenko: Oh, I see, thank you. So looks like no additional action needed.

PR #22, opened 2024-03-01
EVEREST-726 checks for the default state

oksana-grishchenko:
EVEREST-726 checks for the default state of the main branch

they got missed during the dev pipelines work


recharte (approved):
Nice :smile:

PR #30, opened 2024-03-04
EVEREST-107 Add gatekeeper

oksana-grishchenko:
**Problem** 
Need to set the different sets of checks for FE and API+CLI as required (to block PR merge if smth is failed). 
GitHub doesn't allow to set the required checks conditionally.

**Solution**
Use merge-gatekeeper github action. It fails if any of the configured checks are failed, so it's added to both API+CLI and UI CI and now we can make only the gatekeeper check as required. [Here](https://github.com/percona/everest/actions/runs/8143600237?pr=30) is how it looks like

**Action needed after merge:**
set "Merge Gatekeeper" check as required

### DIFF
--- a/.github/workflows/dev-be-ci.yaml
+++ b/.github/workflows/dev-be-ci.yaml
@@ -447,19 +447,3 @@ jobs:
     secrets: inherit
     with:
       make_target: ${{ matrix.make_target }}
-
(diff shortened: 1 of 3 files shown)


recharte (approved):
This is perfect Oksana, well done :muscle:

PR #58, opened 2024-03-05
[UI] EVEREST-866 Changed modal in backup list to general one

filipmikes1:
EVEREST-866

image


ui/apps/everest/src/modals/restore-db-modal/restore-db-modal.tsx
```diff
@@ -175,6 +179,10 @@ const RestoreDbModal = <T extends FieldValues>({
               {
                 label: Messages.fromPitr,
                 value: BackuptypeValues.fromPitr,
+                disabled:
+                  !!backupName &&
+                  pitrData?.latestBackupName !==
+                    watch(RestoreDbFields.backupName),
```
fabio-silva, line 185: Isn't this the same that comes from props?

fabio-silva (changes requested):

ui/apps/everest/src/modals/restore-db-modal/restore-db-modal.tsx
filipmikes1: No it is not, because users can after change backupName to different one

fabio-silva (approved):

PR #63, opened 2024-03-06
chore: update CI to Node 20

fabio-silva:


.github/workflows/dev-build.yaml
```diff
@@ -19,14 +19,14 @@ jobs:
           lfs: true
           ref: 'main'
 
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v3
         with:
           version: 8
 
       - name: Run with Node 16
```
recharte, line 26:
```suggestion
      - name: Run with Node 20
```

recharte: Or just remove it and name it `Use Node` as we have [here](https://github.com/percona/everest/blob/fb90d046c10383a981885129d402ffe5531bedb1/.github/workflows/dev-fe-ci.yaml#L40)

recharte (approved):

PR #67, opened 2024-03-07
EVEREST-793: everestctl upgrade command

michal-kralik:
Add support for `everestctl upgrade` command which upgrades Everest installation.

EVEREST-793


go.mod
```diff
@@ -2,10 +2,15 @@ module github.com/percona/everest
 
 go 1.22
 
+toolchain go1.22.1
+
+replace github.com/Percona-Lab/percona-version-service => ../percona-version-service
```
michal-kralik, line 7: Waits for https://github.com/Percona-Lab/percona-version-service/pull/197

pkg/version/version.go
```diff
@@ -37,18 +36,39 @@ var (
...
+	if Version == "" {
+		return true
+	}
+
+	v, err := goversion.NewVersion(Version)
+	if err != nil {
+		panic(err)
+	}
+
+	if v.Prerelease() != "" && !strings.HasSuffix(v.Prerelease(), "-upgrade-test") {
```
oksana-grishchenko, line 67: a comment to not forget to delete "-upgrade-test"

go.mod
oksana-grishchenko: The PR is merged, should we delete the replacement? hopefully at least some of the piplelines will be fixed then😅

pkg/version/version.go
```diff
@@ -37,18 +36,39 @@ var (
...
-	return catalogImage
+	return fmt.Sprintf(releaseCatalogImage, v)
+}
+
+// ManifestURL returns a manifest URL to install Everest.
+func ManifestURL(v *goversion.Version) string {
+	if isDevVersion() {
+		return devManifestURL
+	}
+	return fmt.Sprintf(releaseManifestURL, v)
```
oksana-grishchenko, line 54: there is no such constants in the codebase anymore. the latest change was using the embed manifest that pointed out to the current version.

go.mod
michal-kralik: Updated

commands/install.go
```diff
@@ -66,6 +66,7 @@ func initInstallFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP("kubeconfig", "k", "~/.kube/config", "Path to a kubeconfig")
 	cmd.Flags().String("namespaces", "", "Comma-separated namespaces list Percona Everest can manage")
 	cmd.Flags().Bool("skip-wizard", false, "Skip installation wizard")
+	cmd.Flags().String("version-metadata-url", "https://check.percona.com", "URL to retrieve version metadata information from")
```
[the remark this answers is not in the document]
michal-kralik: We can try to find a way around this, but I believe we need it for testing, air-gapped environments and maybe other scenarios.

pkg/version/version.go
michal-kralik: This is here intentionally to support dev releases and at the same time support upgrades from dev releases suffixed with `-upgrade-test`

pkg/version/version.go
michal-kralik: Seems it got lost during migration. Added it back. Thank you for pointing it out.

oksana-grishchenko (commented):
Thanks Michal for the implementation! It wasn't an easy one for sure. 

We would need to fix the CI anyway before merging, right? So there are some more changes expected. I believe it wasn't fixed bc first we wanted to reach an agreement on the base logic, and as for me - it looks fine. I would take another look later when the PR is ready to merge👍

commands/install.go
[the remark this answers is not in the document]
recharte: @oksana-grishchenko we have recently seen a lot of interest from several users to run everest in an air-gapped envrironment. In such environments they'd need to also run a version service in their infra, so we need to be able to configure such endpoint.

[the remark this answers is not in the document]
oksana-grishchenko: Ok, let's have it👍

pkg/kubernetes/kubernetes.go
```diff
@@ -375,7 +374,13 @@ func (k *Kubernetes) applyCSVs(ctx context.Context, resources []unstructured.Uns
 }
 
 // InstallPerconaCatalog installs percona catalog and ensures that packages are available.
-func (k *Kubernetes) InstallPerconaCatalog(ctx context.Context) error {
+func (k *Kubernetes) InstallPerconaCatalog(ctx context.Context, version *goversion.Version) error {
+	if version == nil {
+		// No need to check the version if not provided.
+		k.l.Debug("No version provided to InstallPerconaCatalog")
+		return nil
```
recharte, line 381: Shouldn't this be considered an error?

pkg/kubernetes/kubernetes.go
```diff
@@ -905,32 +910,57 @@ func (k *Kubernetes) ApplyObject(obj runtime.Object) error {
...
 
 // InstallEverest downloads the manifest file and applies it against provisioned k8s cluster.
-func (k *Kubernetes) InstallEverest(ctx context.Context, namespace string) error {
-	data, err := scripts.Manifest()
+func (k *Kubernetes) InstallEverest(ctx context.Context, namespace string, version *goversion.Version) error {
+	if version == nil {
+		// No need to check the version if not provided.
+		k.l.Debug("No version provided to InstallEverest")
+		return nil
+	}
```
recharte, line 918: Shouldn't this be considered an error?

michal-kralik: I had this in place for a purpose. I think it was due to upgrade since the code is so mixed up.
I can't remember why, so just returning error here and we'll handle it elsewhere once it pops up.

pkg/kubernetes/kubernetes.go
michal-kralik: Same as https://github.com/percona/everest/pull/67#discussion_r1525047315

pkg/upgrade/upgrade.go
```diff
@@ -79,138 +89,253 @@ func NewUpgrade(c Config, l *zap.SugaredLogger) (*Upgrade, error) {
...
-	u.l.Info("Upgrading Percona Catalog")
-	if err := u.kubeClient.InstallPerconaCatalog(ctx); err != nil {
+
+	// Start upgrade.
+	if err := u.upgradeOLM(ctx, recVer.OLM); err != nil {
 		return err
 	}
-	u.l.Info("Percona Catalog has been upgraded")
-	u.l.Info("Patching subscriptions")
-	if err := u.patchSubscriptions(ctx); err != nil {
```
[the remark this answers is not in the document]
michal-kralik: Originally the upgrade command served only for telemetry disabling. That's not the purpose of the upgrade command though.
Upgrade will now upgrade Everest. Telemetry is part of install command now.

pkg/upgrade/upgrade.go
```diff
@@ -79,138 +89,253 @@ func NewUpgrade(c Config, l *zap.SugaredLogger) (*Upgrade, error) {
...
+		}
+
+		if currentEverestVersion.GreaterThanOrEqual(ver) {
+			continue
+		}
+
+		if upgradeTo == nil {
+			upgradeTo = ver
+			meta = v
+			continue
```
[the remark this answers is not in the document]
michal-kralik: The order of versions shall not matter.
I added a test for the patch versions not being in order and fixed a bug in the implementation.

pkg/everest/client/client.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// NewEverestFromURL returns a new Everest from a provided URL.
+func NewEverestFromURL(rawURL, everestPwd string) (*Everest, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, errors.Join(err, errors.New("could not parse Everest URL"))
+	}
+	everestCl, err := client.NewClient(
+		u.JoinPath("v1").String(),
```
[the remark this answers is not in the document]
michal-kralik: Sure, I added it as a constant.

pkg/everest/client/client.go
```diff
@@ -0,0 +1,110 @@
...
+
+// NewEverest returns new Everest.
+func NewEverest(everestClient *client.Client) *Everest {
+	return &Everest{
+		cl: everestClient,
+	}
+}
+
+// NewEverestFromURL returns a new Everest from a provided URL.
+func NewEverestFromURL(rawURL, everestPwd string) (*Everest, error) {
```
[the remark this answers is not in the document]
michal-kralik: It was used before when we retrieved version information from Everest API. Now it's not used.
I would personally opt-in for keeping it so we don't have to write it from scratch later. Nobody is going to find it in git history :)

If you guys prefer not having it, it's fine as well. I can remove it.

pkg/version/everest.go
```diff
@@ -0,0 +1,71 @@
...
+
+	cutIndex := i + len(versionSplit)
+	versionTag, _ := strings.CutPrefix(image[cutIndex:], "v")
+
+	v, err := goversion.NewVersion(versionTag)
+	if err != nil {
+		return nil, err
+	}
+
+	return v, nil
```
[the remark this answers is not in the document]
michal-kralik: I planned to cover this with integration tests. It finds the docker image tag from the docker image name.
I don't think unit tests are necessary here.

pkg/kubernetes/client/kubeclient_interface.go
```diff
@@ -137,6 +138,8 @@ type KubeClientConnector interface {
 	ListDatabaseEngines(ctx context.Context, namespace string) (*everestv1alpha1.DatabaseEngineList, error)
 	// GetDatabaseEngine returns database clusters by provided name.
 	GetDatabaseEngine(ctx context.Context, namespace, name string) (*everestv1alpha1.DatabaseEngine, error)
+	// Deployment returns a deployment.
+	Deployment(ctx context.Context, name, namespace string) (*appsv1.Deployment, error)
```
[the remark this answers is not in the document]
michal-kralik: I'm trying to follow [Effective Go](https://go.dev/doc/effective_go#Getters) where it recommends against using `Get` verb in a function name.
Similarly we have 
```
kubernetes.CoreV1().Pods(namespace).List(ctx, options)`
// instead of
kubernetes.GetCoreV1().GetPods(namespace).RunList(ctx, options)
```

I'm happy to rename it to `GetDeployment` if you feel this is not relevant to us.

pkg/kubernetes/kubernetes.go
```diff
@@ -375,7 +374,11 @@ func (k *Kubernetes) applyCSVs(ctx context.Context, resources []unstructured.Uns
 }
 
 // InstallPerconaCatalog installs percona catalog and ensures that packages are available.
-func (k *Kubernetes) InstallPerconaCatalog(ctx context.Context) error {
+func (k *Kubernetes) InstallPerconaCatalog(ctx context.Context, version *goversion.Version) error {
+	if version == nil {
+		return errors.New("no version provided for Perconal catalog installation")
```
recharte, line 379:
```suggestion
		return errors.New("no version provided for Percona catalog installation")
```

recharte (changes requested):
This PR is missing a very important piece which is to approve the everest-operator install plan.

go.mod
```diff
@@ -17,16 +20,16 @@ require (
 	github.com/oapi-codegen/echo-middleware v1.0.1
 	github.com/oapi-codegen/runtime v1.1.1
 	github.com/operator-framework/api v0.22.0
-	github.com/operator-framework/operator-lifecycle-manager v0.26.0
+	github.com/operator-framework/operator-lifecycle-manager v0.27.0
 	github.com/percona/everest-operator v0.6.0-dev1.0.20240322111754-0a077df71aa7
+	github.com/percona/percona-everest-backend v0.8.0
```
recharte, line 25: Is this a leftover?

pkg/version/metadata.go
```diff
@@ -0,0 +1,88 @@
...
+		recVer.OLM = v
+	}
+
+	if catalog, ok := meta.GetRecommended()["catalog"]; ok {
+		v, err := goversion.NewSemver(catalog)
+		if err != nil {
+			return nil, errors.Join(err, fmt.Errorf("invalid catalog version %s", catalog))
+		}
+		recVer.Catalog = v
+	}
```
mayankshah1607, line 85: Should we also add the rest, like PG, PXC, etc.?

mayankshah1607 (commented):
LGTM, just some minor nitpicks/suggestions :)

go.mod
michal-kralik: Yes, removed.

pkg/kubernetes/install_plan.go
```diff
@@ -0,0 +1,124 @@
...
+	err := wait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (bool, error) {
+		k.l.Debug("Looking for install plan")
+		ips, err := k.client.ListInstallPlans(ctx, namespace)
+		if err != nil {
+			k.l.Debugf("Could not list install plans: %s", err)
+			return false, nil
+		}
+
+		for _, i := range ips.Items {
+			i := i
```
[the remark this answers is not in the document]
michal-kralik: Updated

pkg/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,169 @@
...
+					{Version: "0.7.0"},
+					{Version: "0.6.0"},
+					{Version: "0.5.0"},
+				},
+			},
+			wantUpgradeTo: "0.7.0",
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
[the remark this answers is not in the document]
michal-kralik: Updated

pkg/kubernetes/install_plan.go
```diff
@@ -0,0 +1,124 @@
...
+
+	k.l.Debugf("Approving install plan %s/%s", namespace, installPlanName)
+
+	ip.Spec.Approved = true
+	_, err = k.client.UpdateInstallPlan(ctx, namespace, ip)
+	if err != nil {
+		var sErr *apierrors.StatusError
+		if ok := errors.As(err, &sErr); ok && sErr.Status().Reason == metav1.StatusReasonConflict {
+			// The install plan has changed. We retry to get an updated install plan.
+			k.l.Debugf("Retrying install plan update due to a version conflict. Error: %s", err)
```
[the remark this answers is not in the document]
michal-kralik: This code has been moved to a separate file as part of the refactoring. This is the original logic for install plan approval.

I guess the person who wrote it experienced issues with version conflicts and added a retry logic.
The install plan can change a couple times before approval. I think it caused issues.

pkg/upgrade/upgrade.go
```diff
@@ -79,138 +90,292 @@ func NewUpgrade(c Config, l *zap.SugaredLogger) (*Upgrade, error) {
...
+		common.EverestOperatorName, upgradeEverestTo,
+	)
+	if err != nil {
+		return errors.Join(err, errors.New("could not find install plan"))
+	}
+
+	u.l.Infof("Upgrading Everest to %s in namespace %s", upgradeEverestTo, common.SystemNamespace)
+	if err := u.kubeClient.InstallEverest(ctx, common.SystemNamespace, upgradeEverestTo); err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
michal-kralik: The logic for upgrade has been defined as:
1. Upgrade Everest
2. Upgrade Everest operator

In case we switch the order we will be in a situation where we have Everest operator running version `n+1` and Everest running version `n`. During this time Everest may not be compatible with the Everest operator which can cause issues.
In cases where the upgrade fails during this step, Everest may not be usable at all.

Upgrading Everest first helps us avoid this type of issue because Everest version is `n+1` and Everest operator is `n`. In this situation we can ensure Everest is compatible with Everest operator version `n+1` and `n`.

Why do you think we shall do this in a different order?

pkg/version/metadata.go
michal-kralik: At some point we may need to do this.
At the moment the agreement is the Everest and Everest operator are always of the same version.

pkg/kubernetes/install_plan.go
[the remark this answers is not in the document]
recharte: Unless I missed something, we no longer have this retry logic so I'd rephrase this log.

recharte (approved):
Even thought there still a few unresolved comments, I think the PR is already in a state where I'm comfortable with merging so I'm unblocking it.

pkg/install/install.go
```diff
@@ -310,7 +328,11 @@ func (o *Install) provisionEverestOperator(ctx context.Context, recVer *version.
 		v = recVer.EverestOperator.String()
 	}
 
-	if err := o.installOperator(ctx, everestOperatorChannel, common.EverestOperatorName, common.SystemNamespace, v)(); err != nil {
+	ch := everestBackendServiceName
```
mayankshah1607, line 331: Should it be `everestOperatorChannel` instead?

michal-kralik: nice catch, updated.

oksana-grishchenko (approved):
Looks good, I'm happy we're at the stage we can finally move forward with this long-running problem❤️

Is there a way to fix the CI? Currently it fails to provision with
`could not determine the latest Everest version `
I'm afraid if we merge as is, it's going to block both FE and BE future PRs.

mayankshah1607 (approved):

pkg/kubernetes/install_plan.go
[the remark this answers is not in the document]
michal-kralik: There's still retry logic when installing an operator.

PR #96, opened 2024-03-19
EVEREST-726 RC/Release workflow

oksana-grishchenko:
EVEREST-726 RC/Release workflow

The [doc](https://www.notion.so/percona/Everest-pipelines-812733d9b5034722b8be7adc5c8b491c#e6c93743551a418a975663fe14d69261) based on which the pipelines were designed. It was created before the monorepo, so there are some changes in details but the basic principles are the same.

Related PRs:
- https://github.com/percona/everest-operator/pull/308 
- https://github.com/percona/everest-catalog/pull/9


.github/workflows/release.yml
```diff
@@ -0,0 +1,388 @@
...
+        # docker/metadata-action action looks more elegant when being triggered by a GH tag,
+        # however this workflow can't be triggered by a GH tag since there are some changes need to be done
+        # in the codebase prior putting the tag, so the action uses the raw tags
+        with:
+          images: |
+            percona/everest-operator,enable=${{ env.IS_RC == 0 }}
+            perconalab/everest-operator
+          tags: |
+            type=raw,value=${{ env.VERSION }}
+            type=raw,value=latest
```
recharte, line 137: When will we have the latest tag?

.github/workflows/release.yml
```diff
@@ -0,0 +1,388 @@
...
+          push: true
+          tags: ${{ steps.operator_meta.outputs.tags }}
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@0.12.0
+        with:
+          image-ref: ${{ steps.operator_meta.outputs.tags }}
+          format: 'table'
+          exit-code: '1'
+          severity: 'CRITICAL,HIGH'
```
recharte, line 170: Shouldn't we do this before pushing the image?

.github/workflows/release.yml
```diff
@@ -0,0 +1,388 @@
...
+      - name: Everest - setup docker build metadata
+        uses: docker/metadata-action@v5
+        id: everest_meta
+        with:
+          images: |
+            percona/percona-everest,enable=${{ env.IS_RC == 0 }}
+            perconalab/everest
+          tags: |
+            type=raw,value=${{ env.VERSION }}
+            type=raw,value=latest
```
recharte, line 357: When will we use the latest tag?

recharte (approved):
This is great stuff, well done Oksana :clap: :muscle:

.github/workflows/release.yml
oksana-grishchenko: Every time the workflow runs it will push the `latest` tag to the `perconalab/everest-operator` repo. In case it's a release run it also pushes `latest` to `percona/everest-operator`

.github/workflows/release.yml
oksana-grishchenko: The tags that this check scans (`operator_meta` tags - the `everest-operator` images) are already pushed at the previous step.  
The next step below pushes the `bundle_meta` tags (the `everest-operator-bundle` images) and another Trivy scanner runs afterwards to check them.

.github/workflows/release.yml
oksana-grishchenko: Every time the workflow runs it will push the latest tag to the `perconalab/everest` repo. In case it's a release run it also pushes latest to `percona/percona-everest`

.github/workflows/release.yml
recharte: I know that they are already pushed at the previous step, that was my point. If we find a vulnerability shouldn't we block the image from being pushed?

oksana-grishchenko: Ran some tests, it's possible first to build an image, then scan it and then push. Updated the flow for each of 4 docker images the RC build. 
Also realized that Trivy can't scan a list of tags, it needs only a single one as an input. So added an additional step that defines the first image tag from the list we've just built (it doesn't really matter which one since the build itself is the same) for the Trivy scan step.

recharte (approved):

PR #97, opened 2024-03-19
EVEREST-107 add codeowners

recharte:
FE devs will get notified about stuff in the /ui directory (FE devs exclusively, BE devs won't) and the BE devs will be notified about everything else.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+*	@percona/everest-backend-devs
+/ui	@percona/everest-frontend-devs
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/oapi-codegen/runtime v1.1.1
 	github.com/operator-framework/api v0.22.0
 	github.com/operator-framework/operator-lifecycle-manager v0.26.0
-	github.com/percona/everest-operator v0.6.0-dev1.0.20240315123444-d5584130d918
+	github.com/percona/everest-operator v0.6.0-dev1.0.20240319093206-a178f0ebc9cd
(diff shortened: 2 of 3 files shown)


fabio-silva (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+*	@percona/everest-backend-devs
+/ui	@percona/everest-frontend-devs
```
oksana-grishchenko, line 2: how will it work? I believe the FE will be notified only if there are the changes in the /ui folder. Will the BE devs be notified for all the FE changes? If so, is there a way to avoid this?

oksana-grishchenko (approved):

.github/CODEOWNERS
recharte:
```
# These owners will be the default owners for everything in
# the repo. Unless a later match takes precedence,
# @global-owner1 and @global-owner2 will be requested for
# review when someone opens a pull request.
*       @global-owner1 @global-owner2

# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner #This is an inline comment.
```

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

recharte: Sorry I should have added this info to the PR description.

oksana-grishchenko: Thanks for the explanation!

PR #108, opened 2024-03-21
[UI] Login doesn't work immediately after logout

dianabirs:
[EVEREST-921](https://perconadev.atlassian.net/browse/EVEREST-921)

[EVEREST-921]: https://perconadev.atlassian.net/browse/EVEREST-921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/apps/everest/src/contexts/auth/auth.provider.tsx
```diff
@@ -50,7 +50,7 @@ const AuthProvider = ({ children }: { children: ReactNode }) => {
   }, []);
 
   useEffect(() => {
-    if (queryStatus === 'success') {
+    if (queryStatus === 'success' && authStatus === 'loggingIn') {
```
fabio-silva, line 53: This change will work only because `authStatus` was added to the dependencies, which will make the effect run after the `login` function was called because of `setAuthStatus`. But this can easily get out of hand because we also do a `setAuthStatus` inside this effect, which has the potential to trigger an never-ending loop if we change this condition down the line. The bug on our code now is that `queryStatus` was already `success` and it doesn't change.

I would recommend the following:
- Check https://tanstack.com/query/latest/docs/framework/react/guides/queries#fetchstatus
- Add `fetchStatus` to the dependency array
- In the beginning of the effect, for safety reasons:
```
if (fetchStatus === 'fetching') {
    return;
}
```
Because `fetchStatus` will always change as we fetch, the effect will run and the login will succeed normally.

fabio-silva (changes requested):

ui/apps/everest/src/contexts/auth/auth.provider.tsx
```diff
@@ -50,6 +54,7 @@ const AuthProvider = ({ children }: { children: ReactNode }) => {
   }, []);
 
   useEffect(() => {
+    if (fetchStatus === 'fetching') return;
```
fabio-silva, line 57:
```suggestion
    if (fetchStatus === 'fetching') {
        return;
    }
```

fabio-silva (changes requested):

fabio-silva (approved):

PR #161, opened 2024-04-04
Everest 951 schedules in table header

fabio-silva:
https://perconadev.atlassian.net/browse/EVEREST-951
Screenshot 2024-04-10 at 12 46 57
Screenshot 2024-04-10 at 12 46 51


ui/apps/everest/src/pages/db-cluster-details/backups/backups-list/table-header/backups-list-table-header.module.css
```diff
@@ -0,0 +1,13 @@
+.scheduleToggleButton {
```
dianabirs, line 1: I see that this selector is no longer used and there is instead inline styling for this component

fabio-silva: yep, good one!

PR #162, opened 2024-04-05
Everest 946 retention copies

fabio-silva:
https://perconadev.atlassian.net/browse/EVEREST-668
Screenshot 2024-04-05 at 14 27 24


dianabirs:
The `Invalid number` error is not displayed when editing a schedule and introducing a negative retention copies number, so a schedule can be saved with negative retention copies number
 
Screenshot 2024-04-10 at 15 10 38
Screenshot 2024-04-10 at 15 12 38

fabio-silva:
> The `Invalid number` error is not displayed when editing a schedule and introducing a negative retention copies number, so a schedule can be saved with negative retention copies number
> 
> Screenshot 2024-04-10 at 15 10 38 Screenshot 2024-04-10 at 15 12 38

@dianabirs well spotted. In the modal schema I was returning a message instead of adding the issue to zod context 🤦

dianabirs (approved):

PR #163, opened 2024-04-05
EVEREST-964 rate limits

oksana-grishchenko:
EVEREST-964 

Add rate limits middleware

### DIFF
--- a/api/everest.go
+++ b/api/everest.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"golang.org/x/time/rate"
 	"io/fs"
 	"net/http"
 
@@ -64,7 +65,7 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
 	}
 
 	echoServer := echo.New()
-	echoServer.Use(echomiddleware.RateLimiter(echomiddleware.NewRateLimiterMemoryStore(config.APIServerRequestsRateLimit)))
+	echoServer.Use(echomiddleware.RateLimiter(echomiddleware.NewRateLimiterMemoryStore(rate.Limit(c.APIRequestsRateLimit))))
 
 	e := &EverestServer{
 		config:     c,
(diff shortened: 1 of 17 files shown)


api/everest.go
```diff
@@ -63,10 +63,13 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
 		return nil, errors.New("could not get namespace from Kubernetes")
 	}
 
+	echoServer := echo.New()
+	echoServer.Use(echomiddleware.RateLimiter(echomiddleware.NewRateLimiterMemoryStore(20)))
```
oksana-grishchenko, line 67: 20 request per second is used in the echo RateLimiter examples and seems to be a reasonable number. Tried to make usual operations with an Everest deployments - the limit is not exceeded.

recharte: I think we should make this configurable though. We want everest to be able to support hundreds if not thousands of DBs. At that scale 20 req/s might not be enough.

oksana-grishchenko: In the latest version I've put the constants to the build flags to increase the limit at least for the debug builds which we use for the API tests. 
@recharte do you think we should make it configurable from the beginning? Should it be an env variable?

recharte: I think we should. Otherwise, in a few months we'll have customers complaining about their requests being rate limited and we won't be able to help them.

dev/patch-deployment-image.yaml
```diff
@@ -11,3 +11,7 @@ spec:
       containers:
         - name: everest
           image: localhost:5000/perconalab/everest:0.0.0
+          env:
+            - name: API_REQUESTS_RATE_LIMIT
+              value: "200"
+
```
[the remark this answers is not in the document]
recharte: I think it's fine to have just the variable for now. I think we are walking closer and closer to needing a config file for the CLI to accommodate all the options we want to give to users.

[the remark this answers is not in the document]
oksana-grishchenko: > we are walking closer and closer to needing a config file for the CLI

I also have this feeling

PR #174, opened 2024-04-08
EVEREST-816 | add API endpoint for starting operator upgrades

mayankshah1607:
- Provides a new api `PUT /namespaces/{namespace}/database-engines/{name}/operator-version` for starting operator upgrades
- Update the DatabaseEngine schema to see the upgrade status and list of available upgrade versions for the operator

For the complete upgrade feature, we still need to add:
- namespace locking/unlocking ([EVEREST-806](https://perconadev.atlassian.net/browse/EVEREST-806))
- upgrade preconditions API ([EVEREST-815](https://perconadev.atlassian.net/browse/EVEREST-815))

These will land as follow-up PRs

[EVEREST-806]: https://perconadev.atlassian.net/browse/EVEREST-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EVEREST-815]: https://perconadev.atlassian.net/browse/EVEREST-815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/spec/openapi.yml
```diff
@@ -579,6 +579,48 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  '/namespaces/{namespace}/database-engines/{name}/upgrade-operator':
```
oksana-grishchenko, line 582: It doesn't feel natural for the REST API to have a verb in the path. Wdyt about using 
put `/namespaces/{namespace}/database-engines/{name}/operator`?

The meaning remains the same. In the path we have the object we work with - the operator, in the method we have the verb about what action is performed over it - PUT.

api/database_engine.go
```diff
@@ -52,3 +54,33 @@ func (e *EverestServer) UpdateDatabaseEngine(ctx echo.Context, namespace, name s
...
+		return err
+	}
+	// Update annotation to start upgrade.
+	annotations := dbEngine.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	annotations[everestv1alpha1.DatabaseOperatorUpgradeAnnotation] = req.TargetVersion
+	dbEngine.SetAnnotations(annotations)
+	_, err = e.kubeClient.UpdateDatabaseEngine(ctx.Request().Context(), namespace, dbEngine)
```
oksana-grishchenko, line 81: Let's have some kind of validation/sanitation for the `targetVersion`, at least check that only the acceptable symbols are used. Bc I bet the method wouldn't pass the next pen testing when one tries to inject some harmful code into the string fields and does not receive a Bad Request response but a 500 😅

docs/spec/openapi.yml
mayankshah1607: Hmm, I agree about the verb in the path. However `PUT /namespaces/{namespace}/database-engines/{name}/operator` also doesn't seem to indicate that it is meant to "upgrade" the operator version.

How about `PUT /namespaces/{namespace}/database-engines/{name}/operator/version`

oksana-grishchenko: I see your concern. I'm wondering if we're going to have any other actions over the `/operator` and why do we need it as a separate segment in the path. However we want to specify we update _the version_ and not anything else. Hmmm... 
What about `PUT /namespaces/{namespace}/database-engines/{name}/operator-version` then?

api/database_engine.go
mayankshah1607: Done, added semver validation

oksana-grishchenko (approved):

PR #180, opened 2024-04-10
[UI] Everest 924 - multischedules in db Wizard

solovevayaroslavna:
image

global logic of changes:

before: schedule modal form which we used for create/edit schedule was under the common db-details/backups page context which needed a whole dbCluster as context data

after: schedule modal form has own small context (because it contains several nests inside, so as not to do props drilling). We use this modal with one context data in db-details page in backup tab, and with another context data in dbWizard. We use 2 context data because in db-details/backups we use data from cluster, on dbWizard (especially in creation mode) we haven't dbCluster data, so we need to get data from formfields

schedule-form has no changes inside
schedule-form was moved inside schedule-form-dialog from components because it's not used anymore separately, only inside schedule-form-dialog
schedule-form-wrapper- also has no changes of logic
In shedule-form and schedule-form-wrapper just some functions was moved to utils to reusing them in dbWizard (like submit and preparing data from ScheduleFormData interface to Schedule interface)

MongoDb adding schedule in dbWizard:
image

Postgres And MySQL adding schedule in dbWizard:
image

No storageLocation:
image

No schedule:
image


ui/apps/everest/src/components/time-selection/time-selection.constants.ts
```diff
@@ -12,6 +12,23 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+import {
```
solovevayaroslavna, line 16: as we don't need anymore data for time selection in DbWizard all timeselection things were moved to timeselection component

dianabirs (approved):

fabio-silva (approved):

PR #181, opened 2024-04-10
EVEREST-955 | provide an option to clean-up backup storage on cluster/backup deletion

mayankshah1607:


api/database_cluster_backup.go
```diff
@@ -67,11 +74,34 @@ func (e *EverestServer) CreateDatabaseClusterBackup(ctx echo.Context, namespace
...
+	namespace, name string,
+	params DeleteDatabaseClusterBackupParams,
+) error {
+	cleanupStorage := pointer.Get(params.CleanupBackupStorage)
+	// If cleanup is required, we add a finalizer to the backup object.
+	if cleanupStorage {
+		reqCtx := ctx.Request().Context()
+		backup, err := e.kubeClient.GetDatabaseClusterBackup(reqCtx, namespace, name)
+		if err != nil {
+			return err
```
michal-kralik, line 88: Shall we return a more user friendly error?

api/database_cluster_backup.go
```diff
@@ -67,11 +74,34 @@ func (e *EverestServer) CreateDatabaseClusterBackup(ctx echo.Context, namespace
...
+		}
+		if err := e.cleanupBackupStorage(reqCtx, backup); err != nil {
+			return err
```
michal-kralik, line 91: Shall we return a more user friendly error?

api/database_cluster_backup.go
```diff
@@ -67,11 +74,34 @@ func (e *EverestServer) CreateDatabaseClusterBackup(ctx echo.Context, namespace
...
+) error {
+	cleanupStorage := pointer.Get(params.CleanupBackupStorage)
+	// If cleanup is required, we add a finalizer to the backup object.
+	if cleanupStorage {
+		reqCtx := ctx.Request().Context()
+		backup, err := e.kubeClient.GetDatabaseClusterBackup(reqCtx, namespace, name)
+		if err != nil {
+			return err
+		}
+		if err := e.cleanupBackupStorage(reqCtx, backup); err != nil {
```
michal-kralik, line 90: Shall the backups be removed only after the DB has been successfully removed?

mayankshah1607: No, its actually the other way around:

- Add this finalizer on all backups for the DB
- DB deletion is triggered, however there is a finalizer set on it by everest-operator
- DB finalizer will delete all the DBBackups (everest-operator)
- DBBackup finalizer is handled (storage clean-up)
- DB is GC-ed by k8s only after all backups are gone

api/database_cluster_backup.go
mayankshah1607: Updated

michal-kralik (approved):

PR #185, opened 2024-04-11
[UI] EVEREST-967 Last backup column should only show completed backups

dianabirs:
[EVEREST-967](https://perconadev.atlassian.net/browse/EVEREST-967)

[EVEREST-967]: https://perconadev.atlassian.net/browse/EVEREST-967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/apps/everest/src/pages/databases/lastBackup/LastBackup.tsx
+++ b/ui/apps/everest/src/pages/databases/lastBackup/LastBackup.tsx
@@ -4,6 +4,7 @@ import { IconButton, Tooltip, Typography } from '@mui/material';
 import { Messages } from '../dbClusterView.messages';
 import { getLastBackupTimeDiff } from '../DbClusterView.utils';
 import { WarningIcon } from '@percona/ui-lib';
+import { BackupStatus } from 'shared-types/backups.types';
 
 export const LastBackup = ({ dbName, namespace }: LastBackupProps) => {
   const { data: backups = [] } = useDbBackups(dbName!, namespace, {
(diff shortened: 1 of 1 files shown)


ui/apps/everest/src/pages/databases/lastBackup/LastBackup.tsx
```diff
@@ -13,12 +14,15 @@ export const LastBackup = ({ dbName, namespace }: LastBackupProps) => {
 
   const { data: pitrData } = useDbClusterPitr(dbName, namespace);
 
-  const lastBackup = backups[backups.length - 1];
-  const lastBackupDate = lastBackup?.created || new Date();
+  const finishedBackups = backups.filter(
+    (backup) => backup.completed && backup.state === BackupStatus.OK
+  );
+  const lastBackup = finishedBackups[finishedBackups.length - 1];
```
fabio-silva, line 20: this one can break if none was retrieved from filter. Let's do some sanity checking here

fabio-silva (changes requested):

ui/apps/everest/src/pages/databases/lastBackup/LastBackup.tsx
dianabirs: @fabio-silva 
If none was retrieved from the filter the array will be empty and `lastBackup` will be `undefined`. I don't think it can break

fabio-silva: You're right, I mixed up filter with find, which retrieves undefined. Good to go

fabio-silva (approved):

PR #187, opened 2024-04-11
[UI] EVEREST-971 Hide S3 keys

dianabirs:
https://github.com/percona/everest/assets/79999411/f049f628-8790-41ce-98cd-4f6ba781a8dc


fabio-silva:
@dianabirs we should rather create a component around TextInput that shows/hides the text, so that we don't repeat the hidden/visible state.

solovevayaroslavna (approved):

fabio-silva (approved):

PR #200, opened 2024-04-15
EVEREST-976 | allow updating DBEngine version

mayankshah1607:
### Description

This PR adds the ability to edit/update the database server (engine) version. This is required as a part of the components upgrade feature. However, the following restrictions will apply:

- Prevent downgrades
- Prevent major version upgrades (since they're handled differently for different operators)

TODO:
- Allow updating this version from the UI


recharte (approved):
:penguin:

PR #201, opened 2024-04-16
EVEREST-806 | Block modifications to DB when an operator is upgrading

mayankshah1607:


api/database_cluster_backup.go
```diff
@@ -66,11 +66,21 @@ func (e *EverestServer) CreateDatabaseClusterBackup(ctx echo.Context, namespace
...
 		e.l.Error(err)
 		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString(err.Error())})
 	}
+
+	// Check if DB update is locked?
+	if locked, err := e.kubeClient.GetIsDBUpdateLocked(ctx.Request().Context(), namespace, dbb.Spec.DbClusterName); err != nil {
+		return errors.Join(err, errors.New("cannot check DB update lock"))
+	} else if locked {
+		return ctx.JSON(http.StatusPreconditionFailed, Error{
+			Message: pointer.ToString("modifying this resource is temporarily disabled"),
```
oksana-grishchenko, line 81: the message seems to be out of sync with the enclosing method

oksana-grishchenko (commented):
Looks good. 
I'm wondering why only the three methods were restricted and thinking if we should restrict the other operations which changes the resources. On the screenshot I marked green what was updated in this PR and marked red the methods to concider - they also change the CRs but they're not restricted yet (I don't think we should restrict the methods for the `backupStorages` and `monitoringConfigs` that's why they are not in the list). WDYT?

Screenshot 2024-04-23 at 15 16 20

mayankshah1607:
@oksana-grishchenko thanks for the feedback!

I added it to only those 3 routes because my understanding was that we only want to prevent mutating the existing DBs. But you're probably right that we should block any non-GET requests (except for monitoring instances and backup storage).

I've re-implemented this PR now so that we're using a middleware instead of calling this method each time explicitly. Please let me know what you think. :)

michal-kralik (approved):

api/middlewares.go
```diff
@@ -0,0 +1,83 @@
...
+		return true, nil
+	}
+
+	// List of resources that should not be modified during an upgrade.
+	targetResources := []string{
+		"database-clusters",
+		"database-engines",
+		"database-cluster-restores",
+		"database-cluster-backups",
+	}
```
oksana-grishchenko, line 41: should Everest allow to 
put '/namespaces/{namespace}/database-engines/{name}'
put '/namespaces/{namespace}/database-engines/{name}/operator-version' ?

oksana-grishchenko (approved):
Great, thanks for the update! Left one suggestion to conciser. 
Approving the PR though

api/middlewares.go
oksana-grishchenko: oh, there we don't allow it already. resolving

michal-kralik (approved):

michal-kralik (approved):

PR #202, opened 2024-04-16
EVEREST-980 | (Bug fix) Ensure install approves the correct InstallPlan

mayankshah1607:
## Problem

Currently, `install` approves the `InstallPlanRef` from the the Subscription, which may change over time (because of OLM). This means that running `install` twice can result in the installation of an undesired operator version (CSV).

## Solution

- We use the `installedCSV` from the Subscription status
- During the first installation, `installedCSV` is empty, so we take `startingCSV`
- `startingCSV` may be unspecified since it is optional, then we take the `InstallPlanRef`


recharte (approved):
Logic looks good :+1:

PR #191, opened 2024-04-12
EVEREST-950 Update release pipeline

oksana-grishchenko:
EVEREST-950

Release pipeline: update `veneer.yaml` in `everest-catalog`:
- for RC and release - add `perconalab` image to the Fast channel bundles list
- for release - add `percona` image to the Stable channel

No need to keep operator 0.0.0 in the stable channel anymore even for the main branch. To build the latest main version of everest run `make build-cli`. The pipelines are updated to use `make bulid-cli` instead of `go run` 

How to keep the operator images list consistent between the releases:
After the release merge the `everest-catalog` release branch to main (only for `everest-catalog`, other repos release branches shouldn't be merged)

Related PRs:
- https://github.com/percona/everest-catalog/pull/14

PR #203, opened 2024-04-16
EVEREST-950: use make build-cli

michal-kralik:
EVEREST-950
Restores the logic that was temporarily removed from #191

### DIFF
--- a/.github/workflows/dev-fe-ci.yaml
+++ b/.github/workflows/dev-fe-ci.yaml
@@ -154,7 +154,8 @@ jobs:
 
       - name: Run Provisioning
         run: |
-          go run cmd/cli/main.go install -v \
+          make build-cli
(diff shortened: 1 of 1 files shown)


recharte (approved):

.github/workflows/dev-fe-ci.yaml
```diff
@@ -187,7 +188,7 @@ jobs:
 
       - name: Adding pg namespace
         run: |
-          go run cmd/cli/main.go install -v \
+          ./bin/everestctl -v \
```
recharte, line 191:
```suggestion
          ./bin/everestctl install -v \
```

recharte (changes requested):
I missed this in the 1st review :see_no_evil: it's causing the CI to fail

PR #205, opened 2024-04-17
EVEREST-975 | [UI] allow editing database engine version

mayankshah1607:
With https://github.com/percona/everest/pull/200, we now allow editing the database engine version. This PR adds the ability to edit the database engine version from the UI


public/dist/index.html
recharte, line 1: @fabio-silva please help me out here but I don't think this file should have been committed...
I see it was first added in March 21st in e380d8a3f180e278b99da7ec0b7df0f2845fe394 but I think we shouldn't have it committed to the repo because it references a file in the `static` directory which doesn't exist on the repo. The static directory is only generated during the build.

mayankshah1607: Ok, I see that it was added to `.gitignore` only a few days before the file was committed. One possibility is that I didn't have the latest changes from main before creating https://github.com/percona/everest/pull/103

I'll remove this file

mayankshah1607: The CI build fails after removing this file:

```
Error: public/public.go:23:12: pattern dist/*: no matching files found
```

I'm not sure what needs to be done

recharte: I think we need to have the file but it should be empty. At least that's what we had originally.

oksana-grishchenko (approved):

public/dist/index.html
recharte: @mayankshah1607 what about having an empty `index.html` like we had before https://github.com/percona/everest/commit/e380d8a3f180e278b99da7ec0b7df0f2845fe394?

mayankshah1607: Done!

recharte (approved):

ui/apps/everest/src/pages/database-form/database-form-body/steps/first/first-step.tsx
```diff
@@ -278,7 +278,7 @@ export const FirstStep = ({ loadingDefaultsForEdition }: StepProps) => {
           name={DbWizardFormFields.dbVersion}
           label={Messages.labels.dbVersion}
           selectFieldProps={{
-            disabled: mode === 'restoreFromBackup' || mode === 'edit',
+            disabled: mode === 'restoreFromBackup',
```
oksana-grishchenko, line 281: what is this change? Is it indended?

.gitignore
```diff
@@ -11,4 +11,4 @@
 api_tests/node_modules/*
 cmd/cmd
 
-public/
\ No newline at end of file
+public/
```
oksana-grishchenko, line 14: I have concerns regarding ignoring the folder completely, are the dev builds going to work in this case? https://github.com/percona/everest/blob/c1224dc63cc0bab61e8ad7f47597726a4ebadc05/.github/workflows/dev-build.yaml#L37

ui/apps/everest/src/pages/database-form/database-form-body/steps/first/first-step.tsx
mayankshah1607: Yes, that's intended. By removing `mode === 'edit'`, we allow editing the engine version from the UI

.gitignore
mayankshah1607: I'll resolve this one for now since we discussed that we will tackle this issue separately. For now I'm reverting the index.html back to the original (empty) state

fabio-silva (approved):

PR #207, opened 2024-04-18
EVEREST-815 | Implement API endpoint for validating operator upgrade preconditions

mayankshah1607:
### Description

- This PR adds a new API `GET /namespaces/{namespace}/database-engines/{name}/operator-version/preflight` which performs a set of checks for performing the operator upgrade.
- The result contains a list of DBs using the requested engine type, and a pending action/task that needs to be performed
- Adds a new `version_service` package that interfaces with the version service API. This change also refactors some existing code that uses the version service.


docs/spec/openapi.yml
```diff
@@ -584,6 +584,51 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  '/namespaces/{namespace}/database-engines/{name}/operator-version/preflight':
```
oksana-grishchenko, line 587: Could the `preflight` be considered as `status`? I'm trying to find more obvious naming, bc when I look at this method it's hard to say what it does...  The method basically checks the current status of the upgrades and returns this information.

oksana-grishchenko (approved):
Great job! I like the idea of creating the version service client which hides the details behind an interface. And thanks for adding the tests for each status. 

No objections to the implementation, so approving the PR. Only one comment below to consider

docs/spec/openapi.yml
mayankshah1607: > Could the preflight be considered as status?

No, not really 😅 .
The status of the upgrade is obtained from the DBEngine status. This API runs a set of preflight checks before you can actually perform the upgrade, that's why I went with `/operator-version/preflight` (as it returns the result of the preflight checks).

Another alternative I considered is `/check`, but that also sounds confusing - is it checking the status? Is it checking if it can upgrade? etc..

mayankshah1607: For more context, this API gets the data that is displayed on this page: https://www.figma.com/file/g77gIX02jKysAzHzsDA9KW/Beta-Release?type=design&node-id=2686-27881&mode=design&t=vKuJZ56VnqZTgSAf-0

oksana-grishchenko: Ok, I see. Let's stick with the `preflight` then

PR #226, opened 2024-04-22
EVEREST-107: add support for basic template in opm

michal-kralik:
Updated workflows after changing opm template to basic.


.github/workflows/release.yml
```diff
@@ -232,35 +232,26 @@ jobs:
           fi
           git checkout $BRANCH_NAME
 
-          # update veneer file
-          yq -i '.Fast.Bundles += [{"Image": "docker.io/perconalab/everest-operator-bundle:" + strenv(VERSION) }]' catalog/everest-operator/veneer.yaml
-          if [[ ! $env.IS_RC ]]; then
-            yq -i '.Stable.Bundles += [{"Image": "docker.io/percona/everest-operator-bundle:" + strenv(VERSION) }]' catalog/everest-operator/veneer.yaml
+          # if there is something to commit, commit it and add the tag
```
recharte, line 235:
```suggestion
```

recharte, line 235:
```suggestion
          # if the tag doesn't exist yet, create it
```

.github/workflows/release.yml
```diff
@@ -232,35 +232,26 @@ jobs:
...
-            git commit -a -m "update version tag"
-            git push origin $BRANCH_NAME
-          
-            git tag $GH_TAG
-            git push origin $GH_TAG
+          # Check if veneer has the new version listed
+          if grep -q "$VERSION" catalog/everest-operator/catalog.yaml; then
+            echo "catalog/everest-operator/catalog.yaml does not include the version $VERSION"
+            echo "Update the file manually before running the workflow"
+            exit 1
```
recharte, line 249: I'm thinking if can actually update it before running this workflow :thinking: 
In order to generate the new catalog we need the operator bundle image to be available in the public registry and that image is only built and pushed as part of this workflow.

recharte: @oksana-grishchenko what I just said is true?

recharte: As discussed offline:

> the process would be:
> 1. run the release pipeline with 0.10.0-rc3 input
> 2. wait until it exits on the catalog step
> 3. go to the catalog release branch, edit the catalog manually, run the generation manually, commit the changes to the release branch
> 4. run the release pipeline again with 0.10.0-rc3 input
> the pipeline does for the operator in case we run the same tag again - it shouldn’t fail. If there is no changes the only thing it will do is to push a new build to the dockerhub (to the same tags), which is not ideal but we can afford it. And the Michal’s branch already contains the exit logic for the catalog if there is no mention of the version.
> So the process I described above should work just fine.

recharte (approved):

.github/workflows/release.yml
oksana-grishchenko: This process should work fine with this PR:

1. run the release pipeline with `0.10.0-rc3` input
2. wait until it exits on the catalog step
3. go to the catalog release branch, edit the catalog manually, run the generation manually, commit the changes to the release branch
4. run the release pipeline again with `0.10.0-rc3` input

PR #227, opened 2024-04-22
[UI] EVEREST-957 Add checkbox to delete data from storage when deleting a backup

dianabirs:
[EVEREST-957](https://perconadev.atlassian.net/browse/EVEREST-957)

image

[EVEREST-957]: https://perconadev.atlassian.net/browse/EVEREST-957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/apps/everest/src/components/confirm-form-dialog/confirm-form-dialog.tsx
```diff
@@ -41,42 +41,51 @@ interface ConfirmFormDialogProps<T extends ConfirmDialogType>
...
       <DialogContent sx={{ px: 1 }}>{dialogContent}</DialogContent>
-      <TextInput
-        name={ConfirmFormDialogFields.confirmInput}
-        label={inputLabel}
-        labelProps={{ sx: { mt: 0 } }}
-        textFieldProps={{
-          placeholder: inputPlaceholder,
-        }}
-      />
+      {confirmationInput && (
```
solovevayaroslavna, line 79: suggestion: maybe we can create a confirm Form Dialog with different options (like: input/checkbox etc... and each of them will be customizable via boolean and props)? I suppose confirmFormDialog should be universal component and that's why we shouldn't add to this component some meaning sence like "Delete" word in DeleteDataCheckbox, because it can be useful for us in future in some other places where checkbox will be needed not for deletion. What do you think about saving it universal?

dianabirs: merged this branch into `EVEREST-959-ui-delete-data-from-backup-storage-when-deleting-a-db` and making the refactoring [here](https://github.com/percona/everest/pull/215)

PR #235, opened 2024-04-25
EVEREST-1013 | validate preflight checks before starting operator upgrades

mayankshah1607:


api/database_engine.go
```diff
@@ -90,6 +94,20 @@ func (e *EverestServer) UpgradeDatabaseEngineOperator(ctx echo.Context, namespac
 		return err
```
michal-kralik, line 94: Wouldn't this always return `nil`?

api/database_engine.go
```diff
@@ -90,6 +94,20 @@ func (e *EverestServer) UpgradeDatabaseEngineOperator(ctx echo.Context, namespac
...
 	}
 
+	// Validate preflight checks.
+	preflight, err := e.getOperatorUpgradePreflight(ctx.Request().Context(), req.TargetVersion, name, namespace)
```
michal-kralik, line 98: Is the engine/namespace locked at this point? I could not find a good reference for it.

api/database_engine.go
mayankshah1607: If the operator is at the latest version, it will return nil.

If the operator can be upgraded, i.e, a newer InstallPlan is available, `GetPendingUpgrade` will return that

mayankshah1607: In the context of the dbEngine Status, a pending upgrade refers to a newly available version to upgrade to

api/database_engine.go
mayankshah1607: Its not. The lock is put in place only once a DBEngine is upgrading (visible in its status). See reference: https://github.com/percona/everest/pull/201/files#diff-bd0949bd5a83da0c2752e1466c3562542455b2e260def83d578158465b2e59e6R64

Again, there is no explicit locking mechanism for this. Its just an API middleware that checks if there's an upgrade happening in a namespace and blocks requests accordingly..

api/database_engine.go
mayankshah1607: This PR will provide clarity on this: https://github.com/percona/everest-operator/pull/365

michal-kralik: Right, but if `u == nil` then we do `return err`. But the `err` comes from `e.kubeClient.GetDatabaseEngine` which is always `nil` at this point.

mayankshah1607: Ah, yes. You're right, didn't notice the typo. Thanks, fixed it.

api/database_engine.go
mayankshah1607: Based on our Slack chat, we will redo the locking mechanism. This is implemented in: https://github.com/percona/everest/pull/201

api/database_engine.go
```diff
@@ -87,8 +91,22 @@ func (e *EverestServer) UpgradeDatabaseEngineOperator(ctx echo.Context, namespac
...
+	}
+
+	// Validate preflight checks.
+	preflight, err := e.getOperatorUpgradePreflight(ctx.Request().Context(), req.TargetVersion, name, namespace)
+	if err != nil {
 		return err
 	}
+	notReady := slices.ContainsFunc(pointer.Get(preflight.Databases), func(db OperatorUpgradePreflightForDatabase) bool {
+		return pointer.Get(db.PendingTask) != Ready
+	})
```
michal-kralik, line 104: I think it would be great to have this as a separate function, something along the lines `canUpgrade(OperatorUpgradePreflight) bool` so this can be easily reused in other places without having to figure out what needs to be checked.

Or alternatively return a third `bool` return value from `getOperatorUpgradePreflight` which would indicate if you can upgrade.

michal-kralik (approved):
Looks good

api/database_engine.go
mayankshah1607: Thanks, updated

PR #238, opened 2024-04-26
EVEREST-869 | [CLI] install: fallback to the default namespace when none specified

mayankshah1607:


michal-kralik (commented):
Is it possible to implement this on the "flags" level as a default value?
This way, it's visible in `--help` and possibly the wizard as well.

mayankshah1607:
@michal-kralik thanks for the suggestion, thats a better idea.

I've updated the code now to use a default at the flags level. There was also a way to specify default in the wizard (which was empty so far)..

pkg/install/install.go
```diff
@@ -431,8 +436,8 @@ func (o *Install) runWizard() error {
 func (o *Install) runEverestWizard() error {
 	var namespaces string
 	pNamespace := &survey.Input{
-		Message: "Namespaces managed by Everest (comma separated)",
-		Default: namespaces,
+		Message: "Namespaces managed by Everest [comma separated]",
+		Default: DefaultEverestNamespace,
```
michal-kralik, line 440: This will make the default value respect the `--namespaces` flag, if provided.

```suggestion
		Default: o.Config.Namespaces,
```

michal-kralik (approved):

PR #269, opened 2024-05-03
EVEREST-1023 EVEREST-1025 | [feat] User account management

mayankshah1607:
## Description

- Add `everestctl accounts` command for managing user accounts on everest
  - `everestctl accounts create`
  - `everestctl accounts list`
  - `everestctl accounts delete`
- Add a `POST /session` API endpoint that takes as input an existing username and password, and returns a JWT token
  - This JWT token must be provided with the `Authorization` header in order to authenticate a request
- Remove the old tokens mechanism along with the `everestctl token ..` command
- Update UI to take username and password as input

TODO:
- `everestctl accounts update` command for updating passwords


recharte:
> The tests are failing because the CI doesn't use the `quickstart-k8s.yaml` file from the local branch. So I'll have to first create a separate PR to update that, and then merge this PR

Shouldn't we change the CI to use the local  `quickstart-k8s.yaml` instead?

mayankshah1607:
> Shouldn't we change the CI to use the local quickstart-k8s.yaml instead?

I did think of that, but that part is tightly coupled with the upgrade stuff, so I didn't want to touch it in this PR at least. If there are no objects, I can create a separate PR

pkg/kubernetes/client/accounts/accounts.go
```diff
@@ -0,0 +1,288 @@
...
+		computedHash, err := a.computePasswordHash(ctx, password)
+		if err != nil {
+			return err
+		}
+		provided = computedHash
+	}
+
+	if stored != provided {
+		return accounts.ErrIncorrectPassword
+	}
```
oksana-grishchenko, line 277: Potential timing attack. Consider using [`crypto/subtle` ](https://pkg.go.dev/crypto/subtle#ConstantTimeCompare)

oksana-grishchenko (commented):
Great job Mayank! 🚀 
Left some comments to consider

commands/accounts/delete.go
```diff
@@ -0,0 +1,78 @@
...
+
+	accountscli "github.com/percona/everest/pkg/accounts/cli"
+	"github.com/percona/everest/pkg/kubernetes"
+)
+
+// NewDeleteCmd returns a new delete command.
+func NewDeleteCmd(l *zap.SugaredLogger) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "delete",
+		Example: "everestctl accounts delete --username user1 --password $USER_PASS",
```
michal-kralik, line 37: Is it possible to delete a user without a password?

mayankshah1607: Yes, that's right. The reason being that we haven't thought about password recovery mechanism. So if a user forgets the current password, the account is probably lost. So with this PR, anyone with a kubeconfig can manipulate any account, but we're going to mitigate this as follows (in upcoming PRs):
- first we will implement RBAC for the Everest API endpoints
- move the user management behind the Everest API
- Enforce RBAC policies to restrict who can manipulate user accounts

mayankshah1607: And the reason why we're leaving out password recovery out for now is because we won't advertise this feature as "suited for production". We expect that users will integrate with external SSO providers

commands/accounts/create.go
```diff
@@ -0,0 +1,82 @@
...
+	cmd.Flags().StringP("username", "u", "", "Username of the account")
+	cmd.Flags().StringP("password", "p", "", "Password of the account")
+}
+
+func initDeleteViperFlags(cmd *cobra.Command) {
+	viper.BindPFlag("username", cmd.Flags().Lookup("username"))     //nolint:errcheck,gosec
+	viper.BindPFlag("password", cmd.Flags().Lookup("password"))     //nolint:errcheck,gosec
+	viper.BindEnv("kubeconfig")                                     //nolint:errcheck,gosec
+	viper.BindPFlag("kubeconfig", cmd.Flags().Lookup("kubeconfig")) //nolint:errcheck,gosec
+}
```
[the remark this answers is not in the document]
mayankshah1607: Thanks, fixed!

pkg/kubernetes/client/accounts/accounts.go
mayankshah1607: Good idea, updated

api/session.go
```diff
@@ -0,0 +1,90 @@
...
+
+// CreateSession creates a new session.
+func (e *EverestServer) CreateSession(ctx echo.Context) error {
+	var params UserCredentials
+	if err := ctx.Bind(&params); err != nil {
+		return err
+	}
+
+	c := ctx.Request().Context()
+	err := e.sessionMgr.Authenticate(c, *params.Username, *params.Password)
```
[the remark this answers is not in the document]
mayankshah1607: Good idea. Added

pkg/accounts/cli/accounts.go
```diff
@@ -0,0 +1,203 @@
...
+	if username == common.EverestAdminUser && password == "" {
+		return true, ""
+	}
+	if !validateUsername(username) {
+		return false,
+			"Username must contain only letters, numbers, and underscores, and must be at least 3 characters long"
+	}
+	if !validatePassword(password) {
+		return false,
+			"Password must contain only letters, numbers and specific special characters (@#$%^&+=!_), and must be at least 6 characters long"
```
michal-kralik, line 178: Is there any reason we restrict the characters for the password?

michal-kralik (approved):

dianabirs (approved):

commands/accounts/delete.go
```diff
@@ -0,0 +1,78 @@
...
+		Run: func(cmd *cobra.Command, args []string) { //nolint:revive
+			initDeleteViperFlags(cmd)
+
+			kubeconfigPath := viper.GetString("kubeconfig")
+			username := viper.GetString("username")
+			password := viper.GetString("password")
```
recharte, line 43: @mayankshah1607 Didn't we decide we'd be removing the authentication from this command?

api/everest.go
```diff
@@ -120,16 +131,28 @@ func (e *EverestServer) initHTTPServer() error {
...
+	tokenLookup := "header:Authorization:Bearer "
+	tokenLookup = tokenLookup + ",cookie:" + common.EverestTokenCookie
+	return echojwt.WithConfig(echojwt.Config{
+		Skipper: func(c echo.Context) bool {
+			return strings.Contains(c.Request().URL.Path, "session")
+		},
+		SigningKey:  []byte(e.config.JWTSigningKey),
+		TokenLookup: tokenLookup,
+	})
+}
```
recharte, line 154: I don't think you need to change it but I just wanted to clarify something. During the demo, you mentioned that this middleware would not only validate the authenticity of the JWT but it would also validate if the user account exists. I believe this isn't true thought, it seems to me that it only checks if the JWT was signed by everest but it doesn't check the user account at all, am I missing something?

fabio-silva (approved):

api/everest.go
mayankshah1607: No, you're right. We don't check if the user exists when validating the token. I probably had this mixed up with the /session endpoint during the demo..

PR #279, opened 2024-05-07
[UI] EVEREST-1016 Fix logic for "Inactive" status

dianabirs:
[EVEREST-1016](https://perconadev.atlassian.net/browse/EVEREST-1016)

[EVEREST-1016]: https://perconadev.atlassian.net/browse/EVEREST-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/apps/everest/src/pages/databases/DbClusterView.utils.ts
+++ b/ui/apps/everest/src/pages/databases/DbClusterView.utils.ts
@@ -13,10 +13,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { DbClusterStatus } from 'shared-types/dbCluster.types';
+import { DbClusterStatus, Schedule } from 'shared-types/dbCluster.types';
 import { DbClusterForNamespaceResult } from '../../hooks/api/db-clusters/useDbClusters';
 import { Messages } from './dbClusterView.messages';
 import { DbClusterTableElement } from './dbClusterView.types';
+import { Backup, BackupStatus } from 'shared-types/backups.types';
 
 const DB_CLUSTER_STATUS_HUMANIFIED: Record<DbClusterStatus, string> = {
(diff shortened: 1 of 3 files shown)


ui/apps/everest/src/pages/databases/DbClusterView.utils.ts
```diff
@@ -97,3 +98,33 @@ export const getLastBackupTimeDiff = (lastBackup: Date): string => {
...
+    return Messages.lastBackup.inactive;
+  }
+
+  if (lastBackup.state === BackupStatus.IN_PROGRESS) {
+    return Messages.lastBackup.pending;
+  }
+
+  if (lastBackup.state === BackupStatus.UNKNOWN) {
+    return Messages.lastBackup.notStarted;
+  }
```
fabio-silva, line 129: what if `lastBackup.state` is neither IN_PROGRESS or UNKNOWN? Won't this return anything?

fabio-silva (changes requested):

ui/apps/everest/src/pages/databases/DbClusterView.utils.ts
dianabirs: This function only handles the case in which we have no completed and successful backup to display in the Last Backup column.
`FAILED` status is filtered already, so if all of the backups are failed, we return 'Inactive'
And if there is any backup with `OK` status, we are in the first case, of displaying the time difference since the last completed backup ([here](https://github.com/percona/everest/pull/279/files#diff-8c361b4461716a80d07ed2591437fe86737824741ace373d991caf576f54d851L25-R32)), so all cases should be covered

fabio-silva (approved):

PR #282, opened 2024-05-09
EVEREST-979 | Reverse the behaviour of backup storage clean-up

mayankshah1607:
## Description

Earlier, the everest-operator always preserved backup storage by default. Now the behaviour has reversed, meaning that the operator will now delete the storage always by default. In order to preserve the storage, we must set a specific finalizer on the backups.

This change in behaviour has been introduced in order to allow storage clean-up for backups that are deleted due to retention settings.

### DIFF
--- a/api/database_cluster_backup.go
+++ b/api/database_cluster_backup.go
@@ -34,7 +34,7 @@ import (
 const (
 	databaseClusterBackupKind = "databaseclusterbackups"
 
-	storageCleanupFinalizer     = "everest.percona.com/dbb-storage-cleanup"
+	storageProtectionFinalizer  = "everest.percona.com/dbb-storage-protection"
 	foregroundDeletionFinalizer = "foregroundDeletion"
 )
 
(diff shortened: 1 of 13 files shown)


api/database_cluster_backup.go
```diff
@@ -34,7 +34,7 @@ import (
 const (
 	databaseClusterBackupKind = "databaseclusterbackups"
 
-	storageCleanupFinalizer     = "everest.percona.com/dbb-storage-cleanup"
+	storageProtectionFinalizer  = "everest.percona.com/dbb-storage-protection"
```
oksana-grishchenko, line 37: the constant can be available directly from the `everest-operator` source code, the `everest` repo has it as a dependency anyway, wdyt about using it directly? 

It would require first to update the module to
`go get github.com/percona/everest-operator@EVEREST-979`

and then after the merge of the operator PR we would need to update the module to  
`go get github.com/percona/everest-operator@main` before merging this PR

oksana-grishchenko (approved):

api/database_cluster_backup.go
```diff
@@ -100,9 +105,17 @@ func (e *EverestServer) GetDatabaseClusterBackup(ctx echo.Context, namespace, na
...
 	controllerutil.AddFinalizer(backup, foregroundDeletionFinalizer)
 	_, err := e.kubeClient.UpdateDatabaseClusterBackup(ctx, backup)
 	return err
 }
+
+func (e *EverestServer) ensureBackupForegroundDeletion(ctx context.Context, backup *everestv1alpha1.DatabaseClusterBackup) error {
+	if controllerutil.AddFinalizer(backup, foregroundDeletionFinalizer) {
+		_, err := e.kubeClient.UpdateDatabaseClusterBackup(ctx, backup)
+		return err
+	}
```
recharte, line 119: This is causing frequent 500 errors in when deleting a DB cluster.

Here's an example log:
```{"time":"2024-05-13T19:37:07.532657977Z","id":"","remote_ip":"127.0.0.1","host":"localhost:8080","method":"DELETE","uri":"/v1/namespaces/my-special-place/database-clusters/postgresql-p1l","user_agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36","status":500,"error":"Operation cannot be fulfilled on databaseclusterbackups.everest.percona.com \"postgresql-p1l-repo2-full-zkxxv\": the object has been modified; please apply your changes to the latest version and try again\ncould not ensure foreground deletion","latency":16659397,"latency_human":"16.659397ms","bytes_in":0,"bytes_out":36}```

mayankshah1607: Thanks for pointing out, I've wrapped this logic in a retry block, we should now see this happen far less frequently

recharte (approved):

api/database_cluster_backup.go
mayankshah1607: Thanks, updated

PR #287, opened 2024-05-12
[UI]Everest 1004 max width for content container

X6uddy:
https://perconadev.atlassian.net/browse/EVEREST-1004

set the maximum width for the content container to 1280px.

before: 
image
image

after: 
image
image


dianabirs (approved):

solovevayaroslavna:
needs design review

Daninda:
Updated:

https://github.com/percona/everest/assets/126169546/b1e00a70-8210-40cf-a037-ae021a210ee6

fabio-silva (changes requested):
Thanks for you contribution @Daninda .
From this video you posted, I can tell that the page shrinks when a column is removed from the table, which I think should not happen

solovevayaroslavna:
@fabio-silva 
Nuna asked in the ticket to set max width to 1280px, but some tables is bigger then 1280px, so it turned out that a large table was wrapped in a scroll and looks like we cannot limit the width of the table to any width if the screen is larger than the table.
And I see 2 ways
1 - we show by the width of the content, which will inevitably lead to jumps when changing the number of column
2 - we somehow adjust the width of the table to fit the screen, but in this case, the content may spread out on large screens, which Nuna would like to avoid
maybe you have another thought how we can handle it without spreading out on the big screen, without limiting the width to a specific number?

PR #298, opened 2024-05-14
EVEREST-107 fix api tests

oksana-grishchenko:
**Problem:**
1. DB clusters deletion takes longer time since they are deleted in foreground now. It caused issues in the CI.
2. When running in parallel and re-running the api-tests may use the same monitoring config simultaneously which caused errors.

**Solution:**
1. Increased waiting time for the deletion. There are still flaky tests about the exposure but for now I think we can move forward with this solution which unblocks the CI.  
2. Use unique monitoring config for each test case


mayankshah1607 (approved):
LGTM

PR #300, opened 2024-05-14
EVEREST-937: add Everest version to telemetry

michal-kralik:
EVEREST-937

Add Everest version to telemetry


api/telemetry.go
```diff
@@ -119,12 +130,17 @@ func (e *EverestServer) collectMetrics(ctx context.Context, url string) error {
 
 		for _, cl := range clusters.Items {
 			types[string(cl.Spec.Engine.Type)]++
+
+			// Number of backup schedules per DB cluster.
+			metrics = append(metrics, Metric{
+				Key:   fmt.Sprintf(telemetryBackupSchedulesKeyTemplate, string(cl.Spec.Engine.Type), cl.UID),
+				Value: strconv.Itoa(len(cl.Spec.Backup.Schedules)),
+			})
```
oksana-grishchenko, line 138: Time-series databases don't work well with the unique keys. From prometheus [docs](https://prometheus.io/docs/practices/naming/):

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

I think we should look for alternative approaches

michal-kralik: I don't think we have other options. We have only a key and a value and we need to uniquely identify a db.

oksana-grishchenko: I see. I was just wondering if the goal can be achieved without using UIDs. What is the goal of the metric? I mean, why do we collect this information? 
Such design of the keys may also make it hard to configure the dashboards, is Peter (or whoever will configure the dashboards) ok with such a format?

oksana-grishchenko (commented):
> Send telemetry data about DB cluster every time user creates DB cluster

this requirement wasn't addressed, is it supposed to be in a separate PR?

michal-kralik:
> > Send telemetry data about DB cluster every time user creates DB cluster
> 
> this requirement wasn't addressed, is it supposed to be in a separate PR?

I think it's better to create a separate PR, since it's not clear if the current solution doesn't already cover this.

oksana-grishchenko:
> I think it's better to create a separate PR, since it's not clear if the current solution doesn't already cover this.

I think it doesn't cover this. As described in [the slack message](https://perconacorp.slack.com/archives/C0545J2BEJX/p1715758186906069?thread_ts=1715704642.213509&cid=C0545J2BEJX):
> It's about the time when we send the telemetry. Currently it's being sent first after 5min and then every 24h. This leads to the situations when the telemetry doesn't show any dbs, bc dbs were created later than 5 min but the whole deployment is destroyed before 24h. The idea is to send the telemetry once a db is created to fix the problem (edited)

api/telemetry.go
michal-kralik: Per agreement, the metric has been removed.

oksana-grishchenko (approved):

PR #332, opened 2024-05-27
EVEREST-1024 | Add `set-password` command

mayankshah1607:
With this change, users can update Everest account passwords as follows:

```bash
$ everestctl accounts set-password --user admin --new-password newadminpassword
```


commands/accounts/set_password.go
```diff
@@ -0,0 +1,69 @@
...
+				}
+				os.Exit(0)
+			}
+
+			cli := accountscli.New(l)
+			cli.WithAccountManager(k.Accounts())
+			if err != nil {
+				l.Error(err)
+				os.Exit(1)
+			}
```
oksana-grishchenko, line 47: Looks like this block will never be executed, bc the error check above (lines 33-40) will `exit(0)` earlier, and this block checks the same `err`

commands/accounts/set_password.go
```diff
@@ -0,0 +1,69 @@
...
+		},
+	}
+	initSetPwFlags(cmd)
+	return cmd
+}
+
+func initSetPwFlags(cmd *cobra.Command) {
+	cmd.Flags().StringP("username", "u", "", "Username of the account")
+	cmd.Flags().StringP("password", "p", "", "Password of the account")
```
oksana-grishchenko, line 61: Should we specify that the flag stands for "New password of the account"? Otherwise it sounds like the user should provide the current password

oksana-grishchenko (approved):
LGTM overall, left a couple of comments

commands/accounts/set_password.go
mayankshah1607: Thanks for catching, I copied over this file from `create.go` and looks like it had an issue as well. Fixed now

commands/accounts/set_password.go
mayankshah1607: You're right. It might sound confusing. 

I've changed it now so it says "new-password" instead

commands/accounts/set_password.go
```diff
@@ -0,0 +1,65 @@
...
+
+	accountscli "github.com/percona/everest/pkg/accounts/cli"
+	"github.com/percona/everest/pkg/kubernetes"
+)
+
+// NewSetPwCommand returns a new set-password command.
+func NewSetPwCommand(l *zap.SugaredLogger) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "set-password",
+		Example: "everestctl accounts set-password --username user1 --password $USER_PASS",
```
oksana-grishchenko, line 24: needs to be updated since the `password` flag is renamed `new-password`

PR #333, opened 2024-05-27
EVEREST-1087 OIDC config

oksana-grishchenko:
EVEREST-1087 

What's new:
1. Added the CLI `oidc configure` command. 
       Usage:  `everestctl oidc configure --issuer-url=http://url.com --client-id=034rkslld`

2. Added the `everest-settings` ConfigMap to store the Everest settings, including the OIDC config
```
% kubectl get configmap/everest-settings -o yaml -n everest-system
apiVersion: v1
data:
  oidc.config: |
    issuerUrl: http://url.com
    clientId: id2342
kind: ConfigMap
metadata:
    ...
```

-------

**Out of scope:**
An API method to get the Everest Settings. Coming in a separate PR.


pkg/kubernetes/kubernetes_interface.go
```diff
@@ -115,4 +116,8 @@ type KubernetesConnector interface {
 	SetJWTToken(ctx context.Context, token string) error
 	// GetJWTToken returns the JWT token from the everest-jwt secret.
 	GetJWTToken(ctx context.Context) (string, error)
+	// UpdateEverestSettings accepts the full list of Everest settings and updates the settings.
+	UpdateEverestSettings(ctx context.Context, settings common.EverestSettings) error
```
oksana-grishchenko, line 120: Considered an option to have Patch update for the settings - I think for now it's sufficiently to have only the full Update method. Not sure how the Settings are going to evolve and we can always add the Patch update later if we would need it.

pkg/oidc/configure.go
```diff
@@ -0,0 +1,77 @@
...
+func (u *OIDC) Run(ctx context.Context) error {
+	err := u.kubeClient.UpdateEverestSettings(ctx, common.EverestSettings{
+		OIDCIssuerURL: u.config.IssuerURL,
+		OIDCClientID:  u.config.ClientID,
+	})
+	if err != nil {
+		return err
+	}
+
+	err = u.kubeClient.RestartDeployment(ctx, common.PerconaEverestDeploymentName, common.SystemNamespace)
```
mayankshah1607, line 70: I think we should in some way inform the user that we will be restarting the Everest server. Or maybe even better, allow the user to skip the restart and let them do it themselves later.

For example, once the ConfigMap is configured, maybe we can show a wizard, like: `Everest requires a restart for the changes to take effect. Restart now? (y/N): `

And a CLI flag so that they can skip the prompt and always perform a restart, like `-r` or `-f` (for force). Example:
```bash
$ everestctl oidc configure -f ...
```

mayankshah1607 (commented):
Looks good to me, just a few suggestions to consider below.

Another suggestion I had was regarding the ConfigMap. Since the `everest-settings` configMap will contain all Everest related settings (and not just OIDC), what do you think about grouping together all the OIDC settings? This way it might be less confusing when we add other options in the ConfigMap. Here's an example:

```yaml
% kubectl get configmap/everest-settings -o yaml -n everest-system
apiVersion: v1
data:
  oidc.config: |
    clientId: 034rkslld
    issuerURL: http://url.com
  saml.config: |
    ...
kind: ConfigMap
metadata:
  creationTimestamp: "2024-05-27T11:04:19Z"
  name: everest-settings
  namespace: everest-system
  resourceVersion: "19619"
  uid: de6da5e3-11cd-4a99-b77c-aa17d9e7a8a6
```

oksana-grishchenko:
@mayankshah1607 regarding the [comment](https://github.com/percona/everest/pull/333#pullrequestreview-2080871006): 

> what do you think about grouping together all the OIDC settings

It was my first idea too, but then during the implementation I realised that the grouping makes it much more cumbersome to read and write the actual values bc of the format conversion but doesn't bing that much of a value. So for the sake of keeping things simple I used the flat values. 

Let's check for more opinions on the topic, @recharte what data format do you think it would be better to use in the settings ConfigMap?

pkg/oidc/configure.go
oksana-grishchenko: I agree, the wizard option is more user-friendly then the forced restart, thank you for the suggestion. I think then we should also provide a command to restart Everest if the user chooses restart later. 
Let's also involve @recharte to agree on the command design

mayankshah1607:
@oksana-grishchenko I agree that with the grouping, we now need to manipulate a YAML string, so it becomes a bit more tricky. But I think that with a little bit of of code refactoring, we can reduce the complexity.

For example, we're dealing with the same level of complexity with the Everest users Secrets as well, where we need to manipulate a YAML string, but I tried to organise the code such that its not too hard to deal with: https://github.com/percona/everest/blob/main/pkg/kubernetes/client/accounts/accounts.go#L122

pkg/oidc/configure.go
recharte: Before I share my opinion on how to approach the restart option I'd like to ask why do we need to restart the API server?

mayankshah1607: @recharte the JWT middleware which validates the OIDC token is configured once, at the time of initialising the HTTP server. So after you add the OIDC config, the middleware needs to be setup again, for which we need to re-initialise the HTTP server, hence the restart..

For reference, this is the code I'm talking about: https://github.com/percona/everest/pull/324/files#diff-0c59c4dd4c8c53065d0ab7e81ef16c96419f65ba10879efa1c8aecb1e0a73f3dR133-R137

mayankshah1607: Alternatively, to sync the config changes without a restart, the middleware will have to Get the configmap on every request, which will cause a performance overhead. However, we can avoid making a request to k8s each time with the help of an informer. For example, we can setup an informer to watch this ConfigMap and store it in an in-memory structure. Then we read it from memory, instead of making a request to k8s each time.

But for the initial implementation, I felt that taking the restart route would be quicker.

recharte: Thank you, yes, restart is better for now. 

I'd vote for the following:
1. Wizard - We ask for both fields (clientID and issuersURL) and then confirm the restart like Mayank suggested (`Everest requires a restart for the changes to take effect. Restart now? (y/N):`)
2. Headless - We introduce a flag called `--skip-restart`. I think this is clearer than having a `-f` or `-r`. Also, the default (no flag) means that the restart will happen which is what we want.

WDYT?

recharte:
> Let's check for more opinions on the topic, @recharte what data format do you think it would be better to use in the settings ConfigMap?

Looking at the CM alone I do prefer the grouping like Mayank was suggesting. @oksana-grishchenko What do you think about Mayank's suggestion of processing it similarly to what we did with the accounts secret? Do you still think it's too complex?

pkg/oidc/configure.go
mayankshah1607: Sounds good to me!

pkg/common/settings.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// FromMap tries to convert a map the EverestSettings struct.
+func (e *EverestSettings) FromMap(m map[string]string) error {
+	err := yaml.Unmarshal([]byte(m[oidcMapKey]), &e.OIDC)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
oksana-grishchenko, line 36: I'm not really happy about this approach in compare with the flat list because:
1. The hardcoded `oidcMapKey` constant 
2. The fact that we would need to change the `ToMap()` and `FromMap()` methods each time we add a new setting.

The approach with the flat list doesn't have this downsides. Please let me know what approach do you prefer.

mayankshah1607: Hey @oksana-grishchenko 

Thanks for trying this. I agree with your points. But out of curiosity, I took a look at how argocd handles this since we have the exact same use-case.

It seems that in their internal structures, they represent the OIDC config as a raw YAML string, and have additional methods to convert it to a struct and back to raw.

I tried to replicate their approach, and together with the `mapstructure` library, I think it tackles both the issues you mentioned. Here's an example: https://github.com/percona/everest/commit/092a06a805db2892dfb4607b0cffd64be4e37aea

Please let me know what you think :)

oksana-grishchenko: Looks great, this approach is much more flexible, thank you Mayank! Cherry-picked your commit and updated the tests.

pkg/oidc/configure.go
oksana-grishchenko: UPD: 
after the discussion with Product & Design it was decided to keep the forced restart since the oidc configuration happens mostly once in the Everest lifetime and the restart itself only affects the UI and API.

mayankshah1607 (approved):
Thanks for making all the changes, LGTM!

recharte (approved):

PR #335, opened 2024-05-28
EVEREST-1084: components endpoint

michal-kralik:
EVEREST-1084

Add components endpoint


mayankshah1607 (approved):

docs/spec/openapi.yml
```diff
@@ -1635,6 +1674,38 @@ components:
...
+              $ref: '#/components/schemas/DatabaseClusterComponentContainer'
+    DatabaseClusterComponentContainer:
+      type: object
+      properties:
+        name:
+          type: string
+        age:
+          type: string
+        restarts:
+          type: integer
```
oksana-grishchenko, line 1708: On the UI design there are also those "Ready" and "Status" columns, are they going to be figured out by the FE somehow from the existing data?

docs/spec/openapi.yml
```diff
@@ -1635,6 +1674,38 @@ components:
...
+            type: array
+            items:
+              $ref: '#/components/schemas/DatabaseClusterComponentContainer'
+    DatabaseClusterComponentContainer:
+      type: object
+      properties:
+        name:
+          type: string
+        age:
+          type: string
```
recharte, line 1706: Maybe we should use a CreationTimestamp instead of the age, this way the frontend can continuously update the age between requests.

docs/spec/openapi.yml
michal-kralik: They're part of the response in https://github.com/percona/everest/blob/1838b0e702f0d55fd7b9b0173803134930283586/docs/spec/openapi.yml#L1677

docs/spec/openapi.yml
michal-kralik: Updated

PR #344, opened 2024-05-30
EVEREST-1096 | Implement RBAC (based on casbin)

mayankshah1607:


api/rbac/rbac.go
```diff
@@ -0,0 +1,191 @@
...
+			basePath + "/namespaces/:namespace/database-clusters/:name":             "database-clusters",
+			basePath + "/namespaces/:namespace/database-clusters/:name/backups":     "database-clusters",
+			basePath + "/namespaces/:namespace/database-clusters/:name/credentials": "database-clusters",
+			basePath + "/namespaces/:namespace/database-clusters/:name/pitr":        "database-clusters",
+			basePath + "/namespaces/:namespace/database-clusters/:name/restores":    "database-clusters",
+			basePath + "/namespaces/:namespace/database-clusters/:name/components":  "database-clusters",
+			basePath + "/namespaces/:namespace/database-cluster-backups":            "database-cluster-backups",
+			basePath + "/namespaces/:namespace/database-cluster-backups/:name":      "database-cluster-backups",
+			basePath + "/namespaces/:namespace/database-cluster-restores":           "database-cluster-restores",
+			basePath + "/namespaces/:namespace/database-cluster-restores/:name":     "database-cluster-restores",
```
mayankshah1607, line 144: @percona/everest-backend-devs What do you guys think about automating this part, so that we don't have to manually update this code everytime a new path is added?

For example, what I had in mind is - read the swagger spec, parse the URL and extract the resource name out of it and build this map

michal-kralik (approved):

api/rbac/rbac.go
michal-kralik: yes, would be great

oksana-grishchenko (approved):

api/rbac/rbac.go
mayankshah1607: OK, will do in a follow up PR. I feel this PR is already too big and complex already..

pkg/kubernetes/informer/informer.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"go.uber.org/zap"
+	"k8s.io/client-go/rest"
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Informer is a generic wrapper around the controller-runtime cache.
+// It provides a minimastic interface for watching Kubernetes objects
```
recharte, line 31:
```suggestion
// It provides a minimalist interface for watching Kubernetes objects
```

api/rbac/rbac.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		if err := enforcer.LoadPolicy(); err != nil {
+			l.Error("failed to load policy", zap.Error(err))
+		}
+	})
+	if inf.Start(ctx, &corev1.ConfigMap{}) != nil {
+		return errors.Join(err, errors.New("failed to watch RBAC ConfigMap"))
+	}
+	return nil
+}
```
recharte, line 68: This is such an important piece that was missing from the PoC, thanks for adding this hot reload :clap:

api/rbac/rbac.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	issuer, err := claims.GetIssuer()
+	if err != nil {
+		return "", errors.Join(err, errors.New("failed to get issuer from claims"))
+	}
+
+	if issuer == session.SessionManagerClaimsIssuer {
+		return strings.Split(subject, ":")[0], nil
+	}
```
recharte, line 120: It has nothing to do with your PR but this got me questioning the usefulness of the capabilities suffix...

recharte (approved):

api/rbac/rbac.go
mayankshah1607: Hmm yeah. Even I don't know yet.. But we can revisit it later if its not useful.
Resolving for now to merge..

PR #349, opened 2024-05-31
EVEREST-107 Release CI update

oksana-grishchenko:


oksana-grishchenko:
The changes are applied directly

PR #357, opened 2024-06-03
EVEREST-1099 | Get rid of hard-coded URLs from JWT skipper

mayankshah1607:
- Specify a bearer token security scheme that is application for the APIs
- Specify that `/settings` and `/session` will opt out from authentication
- In the middleware code, we parse the swagger spec and build a list of paths that skip authentication and pass that to the skipper func


michal-kralik (approved):

api/everest.go
```diff
@@ -44,6 +44,10 @@ import (
 	"github.com/percona/everest/public"
 )
 
+const (
+	propertySkipEverestAuthnSwagger = "x-everest-skip-authn"
+)
+
```
oksana-grishchenko, line 50: I was wondering if there is a native openapi way to specify the security schema without a need to define the user-defined constant. And there is https://swagger.io/docs/specification/authentication/ 
So theoretically we could define the bearer auth as a default one for the API and then override the security settings to empty for the endpoints which don't require authentication. And then retrieve this information in the code to figure out if the method needs to be skipped.   

Not sure if it's worth to put more effort on the topic, so leaving it up to you, just wanted to bring it to consideration

oksana-grishchenko (approved):
The approach is fine, thanks for the improvement, it's much better than hardcoding the list bc now it's more transparent which endpoint uses authentication and which one doesn't.  

Leaving one thing to consider in the comments. I understand it might require some additional effort, so I'm approving the PR to indicate that I'm fine with the current approach too.

PR #361, opened 2024-06-03
EVEREST-870 Fix wizard

oksana-grishchenko:
EVEREST-870 

**Problem:**
The `namespaces`, `operator.<operator_name>` installation flags are conflicting with the installation wizard when the install command has been run without the `skip-wizard` flag, so that even if the user has provided the values via flags, they've been asked for the values again in the wizard

**Solution:**
As discussed during one of the DS, the cli should give the priority to the flags and if the user has set up the flags do not run the corresponding wizard section.


pkg/output/command.go
```diff
@@ -54,6 +55,10 @@ func PrintError(err error, l *zap.SugaredLogger) {
 	if errors.Is(err, common.ErrExitWithError) {
 		return
 	}
+	if errors.Is(err, terminal.InterruptErr) {
+		l.Info("The command execution is interrupted")
+		return
+	}
```
oksana-grishchenko, line 61: Graceful shutdown for interrupting the wizard, otherwise an ugly stack trace is printed.

mayankshah1607 (approved):

PR #379, opened 2024-06-06
Everest 1102 | Ensure 'managed-by' label on DB namespaces during upgrades

mayankshah1607:


pkg/kubernetes/client/namespace.go
```diff
@@ -26,3 +26,8 @@ func (c *Client) DeleteNamespace(ctx context.Context, name string) error {
 func (c *Client) ListNamespaces(ctx context.Context, opts metav1.ListOptions) (*corev1.NamespaceList, error) {
 	return c.clientset.CoreV1().Namespaces().List(ctx, opts)
 }
+
+// UpdateNamespace updates the given namespace.
+func (c *Client) UpdateNamespace(ctx context.Context, namespace *corev1.Namespace, opts metav1.UpdateOptions) (*corev1.Namespace, error) {
+	return c.clientset.CoreV1().Namespaces().Update(ctx, namespace, opts)
```
oksana-grishchenko, line 32: Would it require to update [the operator permissions](https://github.com/percona/everest-operator/blob/12c6e2898bcb61a1667ac83d655e8710bd96f58e/bundle/manifests/everest-operator.clusterserviceversion.yaml#L174-L179)?

oksana-grishchenko (approved):

pkg/kubernetes/client/namespace.go
oksana-grishchenko: Another thing - I'm a bit concerned about that permission from the users perspective, I mean, I remember we discussed once that the users might not be happy about the fact that Everest operator can edit namespaces in the cluster... @recharte any thoughts?

recharte: > Would it require to update [the operator permissions](https://github.com/percona/everest-operator/blob/12c6e2898bcb61a1667ac83d655e8710bd96f58e/bundle/manifests/everest-operator.clusterserviceversion.yaml#L174-L179)?

Why would it? Nothing changed in the operator yet, it still get the needed info from the env var.

> Another thing - I'm a bit concerned about that permission from the users perspective, I mean, I remember we discussed once that the users might not be happy about the fact that Everest operator can edit namespaces in the cluster... @recharte any thoughts?

Isn't this all just for the CLI uninstall? Will the operator use this label? Even if it does, can't it have only read permissions? Would read permissions for the namespaces be ok?

oksana-grishchenko: Nevermind, that's right, the CLI can already create and delete namespaces and it doesn't affect the operator.

PR #382, opened 2024-06-07
EVEREST-1135 | Everest accounts/auth does not work after upgrading

mayankshah1607:
Currently, Everest accounts and authentication features are installed only upon running `everestctl install`. This PR adds some extra steps during upgrade that ensures that the resources needed for these features to work correctly are installed.


michal-kralik (approved):

pkg/upgrade/upgrade.go
```diff
@@ -158,6 +167,10 @@ func (u *Upgrade) Run(ctx context.Context) error {
 		return errors.Join(err, errors.New("could not find install plan"))
 	}
 
+	if err := u.ensureEverestJWTIfNotExists(ctx); err != nil {
+		return err
+	}
```
recharte, line 172: Should we do this conditionally? The user may use this CLI version to upgrade from 0.9 to 0.10 and in that case we don't want to run this code. We only want to run it if upgrading to 0.11.

pkg/upgrade/upgrade.go
```diff
@@ -171,6 +184,10 @@ func (u *Upgrade) Run(ctx context.Context) error {
 
 	u.l.Infof("Everest has been upgraded to version %s", upgradeEverestTo)
 
+	if err := u.ensureEverestAccountsIfNotExists(ctx); err != nil {
+		return err
+	}
```
recharte, line 189: Should we do this conditionally? The user may use this CLI version to upgrade from 0.9 to 0.10 and in that case we don't want to run this code. We only want to run it if upgrading to 0.11.

pkg/upgrade/upgrade.go
```diff
@@ -222,6 +239,39 @@ func (u *Upgrade) ensureManagedByLabelOnDBNamespaces(ctx context.Context) error
 	return nil
 }
 
+// This is needed only for 0.10 to 0.11 upgrade.
+// TODO: Remove after 0.11.0 release.
```
recharte, line 243: Do we really want to remove this after 0.11? Users may use later versions of the CLI (e.g. 1.0.0) to do an upgrade from 0.10 to 0.11. Maybe we should keep the code (at least for a few more releases) and use it conditionally (see the other comments above)

recharte (commented):
Since we are adding custom upgrade code specific for this version should we also clean up the old token secret?

pkg/upgrade/upgrade.go
mayankshah1607: Ok, I hadn't considered the fact that users can use a newer CLI to install/upgrade older versions..

I've updated the code so that it runs conditionally (based on version), and also removed the TODOs...

oksana-grishchenko (approved):

pkg/common/version_test.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func TestCompareVersions(t *testing.T) {
+	assert.Equal(t, int(0), CompareVersions("1.0.0", "1.0.0"))
+	assert.Equal(t, int(0), CompareVersions("1.0.0", goversion.Must(goversion.NewVersion("1.0.0"))))
+	assert.Equal(t, int(1), CompareVersions("1.0.0", "0.9.0"))
+	assert.Equal(t, int(1), CompareVersions("1.0.0", goversion.Must(goversion.NewVersion("0.9.0"))))
+	assert.Equal(t, int(1), CompareVersions(goversion.Must(goversion.NewVersion("1.0.0")), "0.10.1"))
+	assert.Equal(t, int(-1), CompareVersions("1.0.0", "1.1.0"))
+	assert.Equal(t, int(-1), CompareVersions("1.0.0", goversion.Must(goversion.NewVersion("1.1.0"))))
```
recharte, line 19: Should we add some tests for RC versions or are we assuming that we'll only handle versions with metadata tags?

PR #404, opened 2024-06-13
EVEREST-1033 | install should run with minimum 1 operator

mayankshah1607:


pkg/install/install.go
```diff
@@ -528,13 +534,12 @@ func (o *Install) runInstallWizard() error {
 	if err := survey.AskOne(
 		pOps,
 		&opIndexes,
-		survey.WithValidator(survey.MinItems(1)),
```
recharte, line 531: Why was this removed? It didn't do the job?

oksana-grishchenko (approved):

pkg/install/install.go
mayankshah1607: It did, but it gave a generic error like:
```
Sorry, your reply was invalid: value is too short. Min items is 1
```

Since there was already a check happening explicitly after that, I just used that so that we can give our own error

recharte: Fair :+1:

recharte (approved):

PR #417, opened 2024-06-17
EVEREST-1077 | Improve uninstall logs message flow

mayankshah1607:
- instead of saying "deleting; waiting to be deleted; successfully deleted" we now say "trying to delete; doesn't exist or successfully deleted"
- At the end of uninstall, if no resources are deleted, show that Everest was not installed


pkg/uninstall/uninstall.go
```diff
@@ -274,8 +284,12 @@ func (u *Uninstall) deleteDBs(ctx context.Context) error {
 
 func (u *Uninstall) deleteNamespaces(ctx context.Context, namespaces []string) error {
 	for _, ns := range namespaces {
-		u.l.Infof("Deleting namespace '%s'", ns)
-		if err := u.kubeClient.DeleteNamespace(ctx, ns); client.IgnoreNotFound(err) != nil {
+		u.l.Infof("Trying to deleting namespace '%s'", ns)
```
oksana-grishchenko, line 287:
```suggestion
		u.l.Infof("Trying to delete namespace '%s'", ns)
```

oksana-grishchenko (approved):

PR #425, opened 2024-06-18
[UI] EVEREST-931: design review fix

Daninda:
https://perconadev.atlassian.net/browse/EVEREST-931

Now the edit buttons don't disappear, but become disabled when there is an error

https://github.com/percona/everest/assets/126169546/5b4cbf94-b10d-4f0a-b266-9f037dc40f9b


solovevayaroslavna:
@fabio-silva @dianabirs agreed with Nuna, this change was added as a design review after the merge of 931, and since the change was a low priority, it was made only now

PR #432, opened 2024-06-19
EVEREST-1153 | Prettify CLI screen for install/upgrade/uninstall commands

mayankshah1607:
Attached below is a demo of the changes introduced by this PR for the install command:

https://github.com/percona/everest/assets/30319228/0e97044b-3ac4-4491-9642-8a3305978a08

To switch back to the old logs, use the `-l` flag with install/upgrade/unisntall commands.


commands/install.go
```diff
@@ -73,6 +79,7 @@ func initInstallFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(install.FlagOperatorMongoDB, true, "Install MongoDB operator")
 	cmd.Flags().Bool(install.FlagOperatorPostgresql, true, "Install PostgreSQL operator")
 	cmd.Flags().Bool(install.FlagOperatorXtraDBCluster, true, "Install XtraDB Cluster operator")
+	cmd.Flags().BoolP("logs", "l", false, "If set, logs are printed during the installation process")
```
recharte, line 82: What do you think about re-purposing the `-v` flag for this? I think it's more natural for CLI users to use `-v` for increasing the logging level.
We can for example have `-v` for adding the logs with the `Info` level and `-vv` for `Debug` logs. Or even just use the `Debug` level by default and have just `-v`. WDYT?

mayankshah1607: Yeah, that's a good idea.. updated 👍🏻

recharte (approved):

PR #433, opened 2024-06-19
EVEREST-876: block allowed ns removal

michal-kralik:
EVEREST-876
EVEREST-877

Block `allowedNamespaces` removal if the resources is in use in that namespace.


api/backup_storage.go
```diff
@@ -276,7 +276,7 @@ func (e *EverestServer) UpdateBackupStorage(ctx echo.Context, backupStorageName
 		})
 	}
 
-	params, err := validateUpdateBackupStorageRequest(ctx, bs, secret, namespaces, e.l)
+	params, err := e.validateUpdateBackupStorageRequest(ctx, backupStorageName, bs, secret, namespaces, e.l)
```
oksana-grishchenko, line 279: Just a thought, leaving up to you.  

We could have a universal method to extract the name and the namespace from the resources metadata to avoid passing additional parameters like this. For example, the `nameFromDatabaseCluster` method does a very similar thing, we could change the error strings to be not the databasecluster's specific and reuse it.

oksana-grishchenko (approved):

api/backup_storage.go
michal-kralik: yes, that's a good idea. I may look into it in a separate PR.

PR #461, opened 2024-06-27
EVEREST-1182 fix install of everest operator

recharte:
We introduced this phase check because OLM has a bug where it sometimes creates duplicate InstallPlans for the same CSV and we found a few cases where the duplicate InstallPlan wasn't reconciled correctly and abandoned by OLM. This abandoned InstallPlan was missing the status field meaning it was also missing the necessary plan to install the operator. Approving this InstallPlan would cause the operator to never be installed. By checking the phase we make sure we will be approving an InstallPlan that is actually ready to be approved. See https://github.com/operator-framework/kubectl-operator/issues/13 for more details on a similar issue.

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -599,7 +599,9 @@ func (k *Kubernetes) getTargetInstallPlanName(ctx context.Context, subscription 
 			// InstallPlan that is actually ready to be approved.
 			// See https://github.com/operator-framework/kubectl-operator/issues/13
 			// for more details on a similar issue.
-			if csv == targetCSV && ip.Status.Phase == olmv1alpha1.InstallPlanPhaseRequiresApproval {
+			// We also need to return the InstallPlan if the Phase is Complete
+			// to ensure the idempotency of the InstallOperator function.
+			if csv == targetCSV && (ip.Status.Phase == olmv1alpha1.InstallPlanPhaseRequiresApproval || ip.Status.Phase == olmv1alpha1.InstallPlanPhaseComplete) {
 				return ip.GetName(), nil
 			}
 		}
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -586,7 +586,20 @@ func (k *Kubernetes) getTargetInstallPlanName(ctx context.Context, subscription 
 	}
 	for _, ip := range ipList.Items {
 		for _, csv := range ip.Spec.ClusterServiceVersionNames {
-			if csv == targetCSV {
+			// If the CSV is the one we are looking for and the InstallPlan is
+			// waiting for approval, we return it.
+			// We introduced this phase check because OLM has a bug where it
+			// sometimes creates duplicate InstallPlans for the same CSV and we
+			// found a few cases where the duplicate InstallPlan wasn't
+			// reconciled correctly and abandoned by OLM. This abandoned
+			// InstallPlan was missing the status field meaning it was also
+			// missing the necessary plan to install the operator. Approving
+			// this InstallPlan would cause the operator to never be installed.
+			// By checking the phase we make sure we will be approving an
+			// InstallPlan that is actually ready to be approved.
+			// See https://github.com/operator-framework/kubectl-operator/issues/13
+			// for more details on a similar issue.
+			if csv == targetCSV && ip.Status.Phase == olmv1alpha1.InstallPlanPhaseRequiresApproval {
 				return ip.GetName(), nil
 			}
 		}


oksana-grishchenko (approved):

mayankshah1607 (commented):
Were you able to test if this works? 

My understanding was that in most cases, this function returns at line 581, which returns the InstallPlanRef from the subscription - when this happens, the Subscription still points at the malformed InstallPlan, and takes a while to update..

I realised that the Install flow is a bit complicated, and its hard to figure out just by looking which conditions will be triggered and which execution path our code will take 😓

recharte:
> Were you able to test if this works?
> 
> My understanding was that in most cases, this function returns at line 581, which returns the InstallPlanRef from the subscription - when this happens, the Subscription still points at the malformed InstallPlan, and takes a while to update..
> 
> I realised that the Install flow is a bit complicated, and its hard to figure out just by looking which conditions will be triggered and which execution path our code will take 😓

I did test that the install works correctly for the generic case. I'm still unable to replicate the case where the IP was missing the status but I guess this is as good as we'd get at this stage. I'll do some more tests to see if I can catch that case but it hasn't been easy...

Yes, the install flow is a bit complicated, we should probably refactor it in following releases.

recharte:
> My understanding was that in most cases, this function returns at line 581, which returns the InstallPlanRef from the subscription - when this happens, the Subscription still points at the malformed InstallPlan, and takes a while to update..

Actually, for the everest operator we are setting the startingCSV so it never returns at line 581 because in line 575 we set it.

mayankshah1607:
> Actually, for the everest operator we are setting the startingCSV so it never returns at line 581 because in line 575 we set it.

If you trace back at how the `req` is passed, we never set the `StartingCSV` value there (it is empty).. So it is the same as `""` and 581 gets triggered

recharte:
> > Actually, for the everest operator we are setting the startingCSV so it never returns at line 581 because in line 575 we set it.
> 
> If you trace back how the `req` is passed, we never set the `StartingCSV` value there (it is empty).. So it is the same as `""` and 581 gets triggered

Maybe I'm missing something but I don't think that is true. Let's take a closer look...
The `getTargetInstallPlanName` function is called here https://github.com/percona/everest/blob/e04e8820edb476f270296dec2845c3ac5e78c564/pkg/kubernetes/kubernetes.go#L662`
In this function the `req` is never changed so it comes from the caller function. This call is [here](https://github.com/percona/everest/blob/e04e8820edb476f270296dec2845c3ac5e78c564/pkg/install/install.go#L684)
The `startingCSV` is set [here](https://github.com/percona/everest/blob/e04e8820edb476f270296dec2845c3ac5e78c564/pkg/install/install.go#L660) which comes from [here](https://github.com/percona/everest/blob/e04e8820edb476f270296dec2845c3ac5e78c564/pkg/install/install.go#L648-L651) which in turn takes the `version` param that comes from the caller function.
This is [the call](https://github.com/percona/everest/blob/e04e8820edb476f270296dec2845c3ac5e78c564/pkg/install/install.go#L386) which has the `v` var set as long as it exists in the recommended versions from VS. These recommended versions come from [here](https://github.com/percona/everest/blob/e04e8820edb476f270296dec2845c3ac5e78c564/pkg/install/install.go#L224-L232).

Is the recommended version empty in VS? :see_no_evil:

mayankshah1607:
> Is the recommended version empty in VS? 🙈

If you trace back ever further and see the implementation of [`ReccomendedVersions`](https://github.com/percona/everest/blob/main/pkg/version/metadata.go#L37), you'll see that we never set `EverestOperator` there. So, its `nil`.. Now if you follow along from there, it eventually leads to not setting the StartingCSV 😅

recharte:
> > Is the recommended version empty in VS? 🙈
> 
> If you trace back ever further and see the implementation of [`ReccomendedVersions`](https://github.com/percona/everest/blob/main/pkg/version/metadata.go#L37), you'll see that we never set `EverestOperator` there. So, its `nil`.. Now if you follow along from there, it eventually leads to not setting the StartingCSV 😅

But then how is it set in the Subscription? I always see it set there...

mayankshah1607:
> > > Is the recommended version empty in VS? 🙈
> > 
> > 
> > If you trace back ever further and see the implementation of [`ReccomendedVersions`](https://github.com/percona/everest/blob/main/pkg/version/metadata.go#L37), you'll see that we never set `EverestOperator` there. So, its `nil`.. Now if you follow along from there, it eventually leads to not setting the StartingCSV 😅
> 
> But then how is it set in the Subscription? I always see it set there...

My bad, you're right. I missed this line where we're checking if it is `nil`, and then using the latest version: https://github.com/percona/everest/blob/e04e8820edb476f270296dec2845c3ac5e78c564/pkg/install/install.go#L231

PR #472, opened 2024-07-02
EVEREST-1104 | API endpoint for fetching permissions of the currently logged in user

mayankshah1607:


docs/spec/openapi.yml
```diff
@@ -72,6 +72,85 @@ paths:
...
+                "*/*"
+            ],
+            [
+                "adminrole:role",
+                "database-clusters",
+                "*",
+                "*/*"
+            ]
+          ]
+        }
```
recharte, line 117: Can the FE work directly with this format? Will they need to convert it somehow?

recharte, line 117: Should we hide the info about roles (e.g. replace `adminrole:role` with `alice`)? Can there be any security implication of letting the user know about which roles are assigned to him?

mayankshah1607: Yes, confirmed with @dianabirs that this should work

recharte: I don't think there is but just sparking some discussion :sunglasses:

recharte (approved):

docs/spec/openapi.yml
mayankshah1607: I don't think there should be any security implications. I'm actually thinking quite the opposite - since most users may not have access to the RBAC policy, this API would be the way to let them know which roles they've been assigned

But if it makes more sense, I'm fine with resolving the role names to the user..

recharte: Yeah, I think this is fine :+1:

recharte (approved):

docs/spec/openapi.yml
mayankshah1607: Update: after speaking with @dianabirs , it was decided to resolve the roles to the username, since it would make the FE implementation easier

PR #476, opened 2024-07-04
EVEREST-1096 | remove the hard-coded resource map for enforcer RBAC

mayankshah1607:
This path-resource map, which was earlier hard-coded, was required for enforcing RBAC on the provided API endpoints. The problem with this is that anytime a new API is added, we have to remember to update this map.

With this change, this map is no longer hard-coded, and RBAC may be enforced on API endpoints simply by adding the `x-everest-resource-name` annotation in the swagger spec.


api/rbac/rbac.go
```diff
@@ -121,27 +122,43 @@ func GetUser(c echo.Context) (string, error) {
...
+	skipPaths := []string{}
+	for path, pathItem := range swg.Paths.Map() {
+		parsedPath := parseEndpoint(path)
+		if val, ok := pathItem.Extensions[common.EverestAPIExtnResourceName]; ok {
+			if resourceName, ok := val.(string); ok {
+				resourceMap[parsedPath] = resourceName
+			}
+			continue
+		}
+		skipPaths = append(skipPaths, parsedPath)
```
oksana-grishchenko, line 151: So starting for now we shouldn't forget to add the `x-everest-resource-name` extension to the new endpoints otherwise they won't be authenticated. It's not that obvious so I was thinking if we can make it more prominent somehow

oksana-grishchenko (approved):

api/rbac/rbac.go
mayankshah1607: This extension controls authorisation only, authentication is controlled using the `security: []` property

But yeah, its still possible to forget to add it if we add new APIs. I thought of adding some manual CI checks, but the problem there is same; how would it know which path needs RBAC and which doesn't? 🤔

mayankshah1607: But with the previous approach, we had to remember to modify code, so I still this this is a decent improvement from that 😅 
I'm happy to discuss ideas to further improve it..

recharte: I don't think there's anything we can do about it, we need to manually control this either way.

recharte (approved):

api/rbac/rbac.go
oksana-grishchenko: Makes sense, let's just keep it in mind when adding and reviewing new endpoints 👍

PR #495, opened 2024-07-10
EVEREST-1142 | Add `rbac can` command

mayankshah1607:
This PR adds a command `everestctl settings rbac can SUBJECT ACTION RESOURCE SUBRERESOURCE` command for testing RBAC policies.

Examples:

```shell
$everestctl settings rbac can alice create database-clusters '*'
No

$everestctl settings rbac can admin create database-clusters '*'
Yes
```


recharte (approved):
:penguin:

pkg/rbac/validate_test.go
```diff
@@ -167,3 +167,115 @@ func TestValidateTerms(t *testing.T) {
...
+			},
+			can: false,
+		},
+		{
+			request: []string{
+				"bob",
+				"create",
+				"database-clusters",
+				"dev/*",
+			},
```
oksana-grishchenko, line 255: from that perspective the `all` alias looks a bit confusing since we can use `all` instead of `*` but we can't use `dev/all` instead of `dev/*` 
I'm not saying we should have the `dev/all` I just wanted to highlight inconsistency

oksana-grishchenko (approved):
LGTM, posted a couple of things to consider

pkg/rbac/validate_test.go
mayankshah1607: Hmm, good point. If you put `*` in the command, it misbehaves, probably some shell thing. But that means it should also work with with `namespace/all` too..

oksana-grishchenko: regarding the `namespace/all` - what if we have the namespace named `all`?😅 I even saw Yusaf was using that name

well, it might be not such a big deal after all. We can just add more usage examples and/or describe how to use the "all" alias.

mayankshah1607: > what if we have a resource named `all`?

I won't completely rule out the possibility of this, but I think the likelihood of this is very low, given that `all` for a resource type in Everest sounds very vague and may not make much sense 🙈 

But yes, I shall add more examples..

mayankshah1607: Ok, I realised that this `all` alias creates more confusion and increases complexity as well. So I just removed it, and added a note in the example that the user must put quotes around it, otherwise the shell will misinterpret it.. Will also make sure to add it in the docs

commands/settings/rbac/can.go
```diff
@@ -0,0 +1,111 @@
...
+	cmd := &cobra.Command{
+		Use: "can SUBJECT ACTION RESOURCE SUBRESOURCE",
+		Long: `Test RBAC policy.
+Examples:
+$ everestctl settings rbac can alice read database-clusters all
+
+$ everestctl settings rbac can alice read database-clusters my-namespace/my-cluster
+
+$ everestctl settings rbac can adminrole:role update database-cluster-backups prod-namespace/prod-backup-1
+`,
```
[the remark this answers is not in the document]
mayankshah1607: Added an example

PR #496, opened 2024-07-11
EVEREST-1096 | RBAC improvements

mayankshah1607:
- handle `database-cluster-credentials` as a separate resource type
- allow `monitoring-instances` and `backup-storages` to be accessed by name
- ensure RBAC is installed when upgrading to 1.1.x


pkg/rbac/rbac.go
```diff
@@ -190,18 +190,16 @@ func NewEnforceHandler(basePath string, enforcer *casbin.Enforcer) func(c echo.C
...
-		case "backup-storages", "monitoring-instances":
-			object = "*"
-		case "database-clusters",
-			"database-engines",
-			"database-cluster-backups",
-			"database-cluster-restores":
+		// These resources are not namespaced.
+		case "namespaces",
+			"backup-storages",
+			"monitoring-instances":
```
recharte, line 196: Actually, we discussed that even though the `BackupStorages` and `MonitoringInstances` themselves aren't namespaced we would still enforce some sort of namespacing based on the `.spec.allowedNamespaces`. It doesn't need to be in this PR but we need to think about how to implement this...

recharte (approved):

oksana-grishchenko (approved):

pkg/rbac/rbac.go
mayankshah1607: Created a ticket: https://perconadev.atlassian.net/browse/EVEREST-1250

PR #501, opened 2024-07-12
EVEREST-1247 | Re-design of operator upgrades API

mayankshah1607:
Adds new APIs for operator upgrades such that all operators in a namespace are upgraded at once. This is done due to an OLM limitation that doesn't actually allow us to individual upgrade an operator in the namespace.

- Add `GET /namespaces/{namespace}/database-engines/upgrade-plan`
  - Returns a list of pending upgrades in the given namespace
  - Returns a list of pending action items in order to perform the upgrades
  - If there are no upgrades, the action items contains post-upgrade tasks
- Add `POST /namespaces/{namespace}/database-engines/upgrade-plan/approval`
  - Starts the operators upgrade process
- Earlier, the actual upgrade itself was handled by everest-operator for each individual dbengine. Now that this is no longer possible, we directly approve InstallPlans from the Everest API. (A corresponding PR has been created in everest-operator to remove this functionality) - https://github.com/percona/everest-operator/pull/452
- Older APIs are still kept for backward compatibility for now. They should be removed in future release. We must also refactor some code when we remove these deprecated APIs.


api/database_engine.go
```diff
@@ -67,7 +67,147 @@ func (e *EverestServer) UpdateDatabaseEngine(ctx echo.Context, namespace, name s
...
+		return err
+	}
+
+	// lock all engines that will be upgraded.
+	if err := e.setLockDBEnginesForUpgrade(ctx, namespace, up, true); err != nil {
+		return errors.Join(err, errors.New("failed to lock engines"))
+	}
+	// unlock if we return with error.
+	defer func() {
+		if err != nil {
```
recharte, line 120: I don't think this condition will ever be true...
This `err` that you're checking here is the `err` returned from `up, err := e.getUpgradePlan(ctx, namespace)` in line 109.
The `err` in line 137 `if err := e.startOperatorUpgradeWithRetry(ctx, "", namespace, ""); err != nil {` only exist in the scope of the conditional and won't get picked up by the deferred function.

In order to get it in the correct scope you'll need to remove the `:` from the `:=` operator

recharte: What about the return in line 131? Wouldn't we want to unlock there as well?

api/database_engine.go
```diff
@@ -67,7 +67,147 @@ func (e *EverestServer) UpdateDatabaseEngine(ctx echo.Context, namespace, name s
...
+			return nil, err
+		}
+		for _, db := range pointer.Get(pointer.Get(pf).Databases) {
+			*result.PendingActions = append(*result.PendingActions, db.toUpgradeTask())
+		}
+	}
+	return result, nil
+}
+
+// ToOperatorUpgradeDatabaseItem converts OperatorUpgradePreflightForDatabase to OperatorUpgradeDatabaseItem.
```
recharte, line 180: Incorrect comment

api/database_engine.go
```diff
@@ -67,7 +67,147 @@ func (e *EverestServer) UpdateDatabaseEngine(ctx echo.Context, namespace, name s
...
+// TODO: Remove this function when the deprecated API is removed.
```
recharte, line 181: Can this function be removed? It's part of the new solution...

deploy/quickstart-k8s.yaml
```diff
@@ -38,6 +38,9 @@ rules:
   - apiGroups: ["everest.percona.com"]
     resources: ["*"]
     verbs: ["*"]
+  - apiGroups: ["operators.coreos.com"]
+    resources: ["installplans"]
+    verbs: ["*"]
```
recharte, line 43: Can we be less broad and just have the verbs we actually need?

go.mod
```diff
@@ -27,15 +27,15 @@ require (
...
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/viper v1.18.2
 	github.com/stretchr/testify v1.9.0
 	go.uber.org/zap v1.27.0
-	golang.org/x/crypto v0.25.0
-	golang.org/x/mod v0.18.0
-	golang.org/x/net v0.27.0
+	golang.org/x/crypto v0.24.0
+	golang.org/x/mod v0.19.0
+	golang.org/x/net v0.26.0
```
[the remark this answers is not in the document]
mayankshah1607: No idea, I didn't notice.. I just ran go mod tidy.. Will reverse it

go.mod
```diff
@@ -162,9 +162,9 @@ require (
 	go.starlark.net v0.0.0-20230525235612-a134d8f9ddca // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/exp v0.0.0-20240213143201-ec583247a57a // indirect
-	golang.org/x/oauth2 v0.17.0 // indirect
-	golang.org/x/sys v0.22.0 // indirect
-	golang.org/x/term v0.22.0 // indirect
+	golang.org/x/oauth2 v0.18.0 // indirect
+	golang.org/x/sys v0.21.0 // indirect
+	golang.org/x/term v0.21.0 // indirect
```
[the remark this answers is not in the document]
mayankshah1607: No idea, I didn't notice.. I just ran go mod tidy.. Will reverse it

pkg/kubernetes/client/client.go
```diff
@@ -255,6 +255,7 @@ func NewInCluster() (*Client, error) {
 		clientset:    clientset,
 		olmClientset: olmClientset,
 		restConfig:   config,
+		rcLock:       &sync.Mutex{},
```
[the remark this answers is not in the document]
mayankshah1607: It is used here: https://github.com/percona/everest/blob/main/pkg/kubernetes/client/install_plan.go#L12

.. which is used here: https://github.com/percona/everest/blob/main/pkg/kubernetes/install_plan.go#L85

... which is finally used here for approving IPs: https://github.com/percona/everest/blob/188e87feeb92cf9a83862ed988ca34fb4edcc1e9/api/database_engine.go#L367

api/database_engine.go
mayankshah1607: > won't get picked up by the deferred function.

That's not true, it will get picked up by the defer function. Any value I return after the defer will be contained in `err`. You can test this with a simply program:

```go
package main

import "fmt"

func testDefer() (m int) {
	m = 0
	defer func() {
		if m == 3 {
			fmt.Println("value is 3")
		}
	}()
	return 3
}

func main() {
	testDefer()
}
```

pkg/kubernetes/client/client.go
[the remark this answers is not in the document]
recharte: ok nevermind

go.mod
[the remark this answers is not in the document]
mayankshah1607: oh, never mind. Its mostly because I've changed the everest-operator package to `EVEREST-1247`  to be able to use the `GetNextUpgrade` function. It will be reverted back once I merged the everest-operator PR and clean-up go mod finally..

go.mod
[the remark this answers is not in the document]
mayankshah1607: oh, never mind. Its mostly because I've changed the everest-operator package to EVEREST-1247 to be able to use the GetNextUpgrade function. It will be reverted back once I merged the everest-operator PR and clean-up go mod finally..

api/database_engine.go
recharte: You're completely correct, I also tested this with a simple program just like yours but I was blind to my own test results because I got the exact same result as you and my mind saw something else... :face_exhaling: fuck confirmation bias :facepalm:

api/database_engine.go
mayankshah1607: I think I forgot to remove the TODO there, but this was supposed to be a helper. I can still remove it and in-line it on the caller. Although, what's wrong in keeping this?

api/database_engine.go
recharte: Regardless of that...
> What about the return in line 131? Wouldn't we want to unlock there as well?

api/database_engine.go
recharte: Nothing wrong, I just thought it was an incorrect comment because it's actually being used in the new solution, if you plan to inline it that's fine.

api/database_engine.go
mayankshah1607: Yeah, we need to unlock on 131. For some reason, I thought returning a non-200 status in `c.JSON` results in an error being returned, but that doesn't make sense.. will update

mayankshah1607: Updated

I ended up removing the defer and named return, since I had to explicitly call unlock anyway on line 131. So the only place where the defer could have been triggered is towards the end of the function. Not worth the added complexity..

api/database_engine.go
mayankshah1607: Updated

deploy/quickstart-k8s.yaml
mayankshah1607: Done

docs/spec/openapi.yml
```diff
@@ -799,11 +799,14 @@ paths:
...
     x-everest-resource-name: database-engines
     get:
+      deprecated: true
       tags:
         - Operators
       summary: Upgrade preflight
       description: |
         This API gets the results of the preflight checks that must be performed before upgrading a database engine operator.
+
+        *This API is deprecated and will be removed from v1.2.0*
```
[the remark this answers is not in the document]
mayankshah1607: Done

api/database_engine.go
```diff
@@ -67,7 +67,152 @@ func (e *EverestServer) UpdateDatabaseEngine(ctx echo.Context, namespace, name s
...
+	}
+
+	// start upgrade process.
+	if err := e.startOperatorUpgradeWithRetry(ctx, "", namespace, ""); err != nil {
+		// Upgrade has failed, so we release the lock.
+		if err := e.setLockDBEnginesForUpgrade(ctx, namespace, up, false); err != nil {
+			e.l.Errorf("Cannot unlock engines: %w", err)
+			return errors.Join(err, errors.New("failed to release lock"))
+		}
+		e.l.Errorf("Failed to upgrade operators: %w", err)
```
recharte, line 143:
```suggestion
		e.l.Errorf("Failed to upgrade operators: %w", err)
		// Upgrade has failed, so we release the lock.
		if err := e.setLockDBEnginesForUpgrade(ctx, namespace, up, false); err != nil {
			e.l.Errorf("Cannot unlock engines: %w", err)
			return errors.Join(err, errors.New("failed to release lock"))
		}
```

recharte (commented):
Sorry for the pedantic review but these are user facing docs :see_no_evil:

oksana-grishchenko (approved):

fabio-silva:
@mayankshah1607 can we merge this?

PR #504, opened 2024-07-12
EVEREST-107 Fixes for automated catalog release

oksana-grishchenko:
- use the updated catalog generation tool with the simplified interface
- fixed some bugs in how the tool was used 
- use fixed opm version


.github/workflows/release.yml
```diff
@@ -387,7 +361,7 @@ jobs:
           git checkout $BRANCH_NAME
           
           # Update deploy manifest 
-          if [[ $env.IS_RC ]]; then 
+          if [[ $IS_RC == 1 ]]; then 
             sed -i "s/perconalab\/everest.*/perconalab\/everest:$VERSION/g" deploy/quickstart-k8s.yaml
             sed -i "s/percona\/percona-everest.*/perconalab\/everest:$VERSION/g" deploy/quickstart-k8s.yaml
           else
```
oksana-grishchenko, line 367: btw this might have been one of the reason of the mistakenly used `perconalab` vs `percona`

.github/workflows/release.yml
```diff
@@ -250,62 +250,36 @@ jobs:
...
-              --current-version "$CURRENT_STABLE_VERSION" \
-              \
-              >| veneer/everest-operator.yaml
-
-            go run ./tools/ \
-              --veneer-file veneer/everest-operator-original.yaml \
-              --version-type ${{ env.RELEASE_TYPE }} \
+              >| ../veneer/everest-operator.yaml
+          
+            cp ../veneer/everest-operator.yaml ../veneer/everest-operator-original.yaml
```
recharte, line 273: I'm a bit confused, in the IS_RC conditional you don't output it to the non original file and here you do and then you copy it to the original

.github/workflows/release.yml
oksana-grishchenko: what the `sed` command does still needs to be re-checked, but the fact is that the condition was wrong

.github/workflows/release.yml
oksana-grishchenko: Thanks for drawing attention to it. After testing I forgot to put back the writing of the command output to the original file. So we always should write to the original file

recharte: Ok now that part looks good but do we actually need 2 files here? Couldn't we always use `../veneer/everest-operator.yaml` instead of copying it back and forth to `../veneer/everest-operator-original.yaml`?

oksana-grishchenko: The temporary file was needed bc the command reads the file and writes the output of itself to the same file. The operational system doesn't handle it well and at some point the file become empty. 
Reworked the update tool so that we won't need a temporary file, ptal https://github.com/percona/everest-catalog/pull/33

recharte (approved):

PR #519, opened 2024-07-15
Remove not needed step for port forwarding for UI E2E tests

tplavcic:


tplavcic:
Closing as it seems it cannot be done this way.

PR #522, opened 2024-07-16
EVEREST-1092 | Fix psmdb on-demand backups allowed to run alongside scheduled backups

mayankshah1607:
To check if a scheduled backup is running, before the creation of an on-demand backup, the current logic is to list all dbbbackups for the cluster and see if they're running.

However, with PSMDB, due to a known limitation, everest-operator does not immediately create a dbbackups object (it waits until progressing beyond `waiting`). Due to this, there's a considerable delay for creating a dbbackup object, and hence Everest ends up allowing the creation of a on-demand backup even though the psmdb-backup object is created.

To work around this issue, we lookup the backup schedule and figure out if a scheduled backup is expected to start.


api/validation.go
```diff
@@ -1356,3 +1384,44 @@ func validateBucketName(s string) error {
...
+			b.DeletionTimestamp.IsZero()
+	}), nil
+}
+
+// isBackupScheduleRunning returns true if a backup schedule is running at the current time.
+func isBackupScheduleRunning(
+	currTime time.Time,
+	backupSchedules []everestv1alpha1.BackupSchedule,
+) (bool, error) {
+	currTime = currTime.Truncate(time.Minute)
```
oksana-grishchenko, line 1413: The `.Truncate()` function removes the seconds like `16:05:01 -> 16:05:00, 16:05:59 -> 16:05:00`, is that the intended behaviour? Shouldn't the `.Add(-1 * time.Minute)` be used here?

mayankshah1607: `Truncate()` is intended here. In this case, it truncates to the nearest minute, so it removes everything from the seconds. Reason for this is that `NextTickAfter` does not return seconds, so the `Equals` check will fail.

> Shouldn't the .Add(-1 * time.Minute) be used here?

That won't work here. If we do this, we are simply avoiding creation of backups 1minute before the actual schedule. But what happens when the schedule is triggered? This check will allow us to create an on-demand backup even though a `psmdb-backup` object has been created (note that dbbackup object creation will be delayed for a few seconds). This won't fix the issue..

recharte:
I still didn't think this through so assume this comment as me thinking out loud...

The only reason why we need this PR is because we have a strange behavior introduced by the everest operator to work around a PSMDBO bug.
The bug is that the [Operator can start two PBM backup operations if backup object is updated](https://perconadev.atlassian.net/browse/K8SPSMDB-1088) and before implementing the workaround we were triggering this bug due to the following events timeline:
1. PSMDBO creates a `psmdb-backup` CR
2. Event 1. triggers the everest operator to create a `dbbackup` CR
3. Everest operator reconciles the  `dbbackup` CR which updates the initial `psmdb-backup` CR
4. Updating the `psmdb-backup` CR while in one of the initial states triggers the bug

The way we worked around this bug was to stop point 2. from happening until the `psmdb-backup` CR is at a state that won't trigger the bug.

If instead of stopping 2. from happening we stop point 3. from happening until the `psmdb-backup` CR is at a state that won't trigger the bug we would avoid triggering the PSMDBO bug and we'd also avoid the behavior that this PR is trying to solve, right? By stopping at 3. we have the `dbbackup` CR created which is all we need to avoid this behavior.

Is this possible? Is there any drawback I'm not considering at this first look?

api/validation.go
oksana-grishchenko: Ok, I see. Thanks, the concern is resolved 👍

mayankshah1607:
@recharte I implemented your suggestion in https://github.com/percona/everest-operator/pull/459 and it seems to work as expected (and I don't see the issue with the duplicate S3 backups either).

I very vaguely remember that this was actually our first idea when we implemented that PSMDB workaround in everest-operator, but can't remember why we decided to do it the way it is now.

I'll close this PR for now, thanks for your idea 👍🏻

PR #544, opened 2024-07-24
EVEREST-1274: check k8s version during install/upgrade

michal-kralik:
EVEREST-1274

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -160,6 +160,10 @@ type (
 		// PXC stores if XtraDB Cluster shall be installed.
 		PXC bool `mapstructure:"xtradb-cluster"`
 	}
+
+	supportedVersion struct {
+		kubernetes goversion.Constraints
+	}
 )
 
 // NewInstall returns a new Install struct.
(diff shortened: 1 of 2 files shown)


pkg/install/install.go
```diff
@@ -244,6 +253,57 @@ func (o *Install) Run(ctx context.Context) error {
...
+
+		if !supVer.kubernetes.Check(k8sVersion) {
+			return fmt.Errorf(
+				"kubernetes version %q does not meet minimum requirements of %q",
+				k8sVersion.String(), supVer.kubernetes.String(),
+			)
+		}
+
+		o.l.Debugf("Finished requirements check for Kubernetes version")
+	}
```
recharte, line 302: Can we use the `checkK8sRequirements()` function defined in the upgrade module, or vice-versa?

recharte: Just trying to avoid duplicate code :slightly_smiling_face:

michal-kralik: we could refactor, but it required extracting upgrade/install structures into a common package so I decided it's maybe better not to touch this behavior so we don't break it and make the PR smaller.

recharte: It's not a big deal but I think it's still work considering this refactor but it can be done in a follow-up PR.

recharte (approved):
Thanks Michal, it looks good. Let's consider doing the refactoring of the supported versions functions in a follow-up PR.

PR #564, opened 2024-07-30
EVEREST-1288 - Add initial release test

tplavcic:
This is a draft PR to start a discussion if it makes sense to use something like this for sanity/release checks to reduce the amount of manual checks.

Notes:
1. Tests are tagged with `@release` and put into `release` subfolder
2. One scenario executes test for multiple configurations (see [forEach in init-deploy.e2e.ts](https://github.com/percona/everest/compare/EVEREST-1288-release-test?expand=1#diff-f5d9340547dea7f86a1072b7c2e8ca8b873565beebbcbdaacb2ec50c6523a2c3R28-R34)) - does that make sense or we want to execute only one (read which one from env variable) and handle multiple executions with the pipeline?
3. Some existing functions could be made more generic like I did here with `resourcesStepCheck`, but we need to see do we want to do it for all since these are currently used by FE UI tests. Here I created simple `populateBasicInformation`, but there is already `basicInformationStepCheck` which is not generic but mongodb specific and it checks some other stuff (does it make sense to use these `check` functions for our populating purposes or we should leave that and just create generic `populate` functions` which will not always check if all elements are populated correctly)
4. I think we should write functions to verify some status on k8s level (skip Everest objects and just check operator objects)
5. In this scenario we could add monitoring, external access check, db parameters, edit cluster, other completely new scenarios could be:
- demand full backup and restore
- demand backup and pitr restore
- scheduled backup and restore
6. Pipelines would need to be created to execute these tests on each supported platform


ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -0,0 +1,180 @@
...
+
+      test(`Cluster creation with ${db} and size ${size}`, async ({
+        page,
+        request,
+      }) => {
+        test.setTimeout(size * 120000);
+
+        expect(storageClasses.length).toBeGreaterThan(0);
+
+        await page.goto('/databases/new');
```
puneet0191, line 74: Any specific reason you would like to keep URL's and locators for pages directly in the test files, could they be part of page classes or some other helpers?

tplavcic: No specific reason, I was looking at page object pattern as well, but current setup is not like that so didn't want to change to much. But it can be done.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -0,0 +1,180 @@
...
+test.describe.configure({ retries: 0 });
+
+[
+  { db: 'psmdb', size: 1 },
+  { db: 'psmdb', size: 3 },
+  { db: 'pxc', size: 1 },
+  { db: 'pxc', size: 3 },
+  { db: 'postgresql', size: 1 },
+  { db: 'postgresql', size: 3 },
+].forEach(({ db, size }) => {
```
oksana-grishchenko, line 43: I like the idea of table tests, it's concise and observable. 
Let me to bring one thing to consideration. If the test fails, is there a way to quickly identify which one has failed? Maybe we could add definitive names (like `pxc1n`, `pxc3n`) which will be shown somehow during execution so when the test fails it won't be a problem to identify which case has failed.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -0,0 +1,180 @@
...
+        const response = await request.get(
+          `/v1/namespaces/${namespace}/database-clusters`,
+          {
+            headers: {
+              Authorization: `Bearer ${await getTokenFromLocalStorage()}`,
+            },
+          }
+        );
+
+        expect(response.ok()).toBeTruthy();
```
oksana-grishchenko, line 122: the fact that the response is not ok is not really informative and we would need to debug the test to understand what is the error. For the response checking consider to use something like this
https://github.com/percona/everest/blob/3f2baae7a2432c4b957210365690a22371bd357b/api-tests/tests/helpers.ts#L11-L16 
so if the response is not ok we would see the error message right away

.github/workflows/dev-fe-e2e.yaml
```diff
@@ -149,7 +149,7 @@ jobs:
           MONITORING_PASSWORD: "admin"
         run: |
           cd ui
-          pnpm --filter "@percona/everest" e2e
+          pnpm --filter "@percona/everest" e2e --grep-invert @release
```
oksana-grishchenko, line 152: Thinking about how to trigger the `@release` tests then. 
Triggering them by a new tag is not correct, bc the build does not exist at that moment. 
Looks like [the release trigger](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release) would be relevant:
```
on:
  release:
    types: [published]
```
It should work for both prereleases (RCs) and releases. And then we can use the appropriate build to set up Everest.

oksana-grishchenko (commented):
I absolutely love the idea of covering the manual post-release checks with automated tests! Thank you Tomislav for this initiative👍

> 4. I think we should write functions to verify some status on k8s level (skip Everest objects and just check operator objects)

FYI the Everest API objects(e.g. [DatabaseCluster](https://github.com/percona/everest/blob/3f2baae7a2432c4b957210365690a22371bd357b/docs/spec/openapi.yml#L2407)) are exactly the same as k8s objects (e.g. [DatabaseCluster](https://github.com/percona/everest-operator/blob/efd490357118cea7461b407f0e6874ccf3061275/deploy/bundle.yaml#L383)) since the API proxies requests to k8s. The Everest API spec for custom resources is copied from the operator spec, and then the API server is generated by a generation tool from that spec.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
tplavcic: This is a good idea and I will see to add some identifier, initially my idea was if the test will be parametrized that the parameters need to be visible in test name like `test(`Cluster creation with ${db} and size ${size}` so if it fails you will see which combination has failed.
But in general I don't think that the structure with this parameters will stay because if one combination fails to re-run just that combination you need to remove other combinations from the test parameters and since we will run this in github action we need to be able to select which one to re-run.
But for now it is convenient to run it like this locally and try all combinations.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
tplavcic: Thank you! I have applied this.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
tplavcic: Just for reference this is how it looks now:
```
Running 35 tests using 1 worker

  ✓  1 [auth] › ../auth.setup.ts:20:1 › Login (5.4s)
  ✓  2 [setup] › ../global.setup.ts:57:1 › Backup storages (1.8s)
  ✓  3 [setup] › ../global.setup.ts:120:1 › Close modal permanently (6.4s)
  ✓  4 [Chrome Stable] › ../release/init-deploy.e2e.ts:55:5 › Init deploy workflow › Cluster creation with psmdb and size 1 (1.2m)
  ✓  5 [Chrome Stable] › ../release/init-deploy.e2e.ts:121:5 › Init deploy workflow › Suspend cluster with psmdb and size 1 (9.9s)
  ✓  6 [Chrome Stable] › ../release/init-deploy.e2e.ts:131:5 › Init deploy workflow › Resume cluster with psmdb and size 1 (35.8s)
  ✓  7 [Chrome Stable] › ../release/init-deploy.e2e.ts:139:5 › Init deploy workflow › Restart cluster with psmdb and size 1 (31.0s)
  ✓  8 [Chrome Stable] › ../release/init-deploy.e2e.ts:150:5 › Init deploy workflow › Delete cluster with psmdb and size 1 (10.1s)
  ✓  9 [Chrome Stable] › ../release/init-deploy.e2e.ts:55:5 › Init deploy workflow › Cluster creation with psmdb and size 3 (1.9m)
  ✓  10 [Chrome Stable] › ../release/init-deploy.e2e.ts:121:5 › Init deploy workflow › Suspend cluster with psmdb and size 3 (25.1s)
  ✓  11 [Chrome Stable] › ../release/init-deploy.e2e.ts:131:5 › Init deploy workflow › Resume cluster with psmdb and size 3 (1.1m)
  ✓  12 [Chrome Stable] › ../release/init-deploy.e2e.ts:139:5 › Init deploy workflow › Restart cluster with psmdb and size 3 (1.5m)
  ✓  13 [Chrome Stable] › ../release/init-deploy.e2e.ts:150:5 › Init deploy workflow › Delete cluster with psmdb and size 3 (11.1s)
  ✓  14 [Chrome Stable] › ../release/init-deploy.e2e.ts:55:5 › Init deploy workflow › Cluster creation with pxc and size 1 (1.5m)
  ✓  15 [Chrome Stable] › ../release/init-deploy.e2e.ts:121:5 › Init deploy workflow › Suspend cluster with pxc and size 1 (25.1s)
  ✓  16 [Chrome Stable] › ../release/init-deploy.e2e.ts:131:5 › Init deploy workflow › Resume cluster with pxc and size 1 (56.2s)
  ✓  17 [Chrome Stable] › ../release/init-deploy.e2e.ts:139:5 › Init deploy workflow › Restart cluster with pxc and size 1 (1.1m)
  ✓  18 [Chrome Stable] › ../release/init-deploy.e2e.ts:150:5 › Init deploy workflow › Delete cluster with pxc and size 1 (10.5s)
  ✓  19 [Chrome Stable] › ../release/init-deploy.e2e.ts:55:5 › Init deploy workflow › Cluster creation with pxc and size 3 (3.9m)
  ✓  20 [Chrome Stable] › ../release/init-deploy.e2e.ts:121:5 › Init deploy workflow › Suspend cluster with pxc and size 3 (56.3s)
  ✓  21 [Chrome Stable] › ../release/init-deploy.e2e.ts:131:5 › Init deploy workflow › Resume cluster with pxc and size 3 (2.1m)
  ✓  22 [Chrome Stable] › ../release/init-deploy.e2e.ts:139:5 › Init deploy workflow › Restart cluster with pxc and size 3 (2.7m)
  ✓  23 [Chrome Stable] › ../release/init-deploy.e2e.ts:150:5 › Init deploy workflow › Delete cluster with pxc and size 3 (11.2s)
  ✓  24 [Chrome Stable] › ../release/init-deploy.e2e.ts:55:5 › Init deploy workflow › Cluster creation with postgresql and size 1 (1.0m)
  ✓  25 [Chrome Stable] › ../release/init-deploy.e2e.ts:121:5 › Init deploy workflow › Suspend cluster with postgresql and size 1 (29.7s)
  ✓  26 [Chrome Stable] › ../release/init-deploy.e2e.ts:131:5 › Init deploy workflow › Resume cluster with postgresql and size 1 (51.7s)
  ✓  27 [Chrome Stable] › ../release/init-deploy.e2e.ts:139:5 › Init deploy workflow › Restart cluster with postgresql and size 1 (31.0s)
  ✓  28 [Chrome Stable] › ../release/init-deploy.e2e.ts:150:5 › Init deploy workflow › Delete cluster with postgresql and size 1 (10.0s)
  ✓  29 [Chrome Stable] › ../release/init-deploy.e2e.ts:55:5 › Init deploy workflow › Cluster creation with postgresql and size 3 (1.3m)
  ✓  30 [Chrome Stable] › ../release/init-deploy.e2e.ts:121:5 › Init deploy workflow › Suspend cluster with postgresql and size 3 (10.0s)
  ✓  31 [Chrome Stable] › ../release/init-deploy.e2e.ts:131:5 › Init deploy workflow › Resume cluster with postgresql and size 3 (26.7s)
  ✓  32 [Chrome Stable] › ../release/init-deploy.e2e.ts:139:5 › Init deploy workflow › Restart cluster with postgresql and size 3 (1.3m)
  ✓  33 [Chrome Stable] › ../release/init-deploy.e2e.ts:150:5 › Init deploy workflow › Delete cluster with postgresql and size 3 (11.3s)
  ✓  34 [teardown] › ../global.teardown.ts:21:3 › Teardown › Delete backup storage (2.5s)
  ✓  35 [teardown] › ../global.teardown.ts:54:3 › Teardown › Logout (4.6s)
::warning title=Slow Test,file=[Chrome Stable] › ../release/init-deploy.e2e.ts::[Chrome Stable] › ../release/init-deploy.e2e.ts took 28.4m
::notice title=🎭 Playwright Run Summary::  35 passed (29.5m)
```

oksana-grishchenko: In that case I think the output looks clear enough, my main concern was about how to identify which test has failed. I've missed that it's going to print the messages

.github/workflows/dev-fe-e2e.yaml
tplavcic: I was thinking for these to just build the cli and use the current dev latest build and if the release branch exists to be able to point it to the release branch and use already pre-built cli and package.
But these will likely be executed as well on different platforms and will need some credentials for cloud providers so these pipelines will be added to private [k8s-testing-automation](https://github.com/percona/k8s-testing-automation) repo.

ui/apps/everest/.e2e/utils/generic.ts
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { expect } from '@playwright/test';
+
+export const checkError = async (response) => {
+  if (!response.ok()) {
+    console.log(`${response.url()}: `, await response.json());
+  }
+  expect(response.ok()).toBeTruthy();
```
fabio-silva, line 22: if we do `expect(response.ok()).toBeTruthy();` first, I think we can then just log the response, as so:

```
    expect(response.ok()).toBeTruthy();
    console.log(`${response.url()}: `, await response.json());
```

The way it is now, I think `expect(response.ok()).toBeTruthy()` is redundant, as you already checked if it was falsy in the first if clause.

ui/apps/everest/.e2e/utils/db-wizard.ts
```diff
@@ -54,3 +54,152 @@ export const goToLastAndSubmit = async (page: Page) => {
...
+  if (externalAccess != '') {
+    await page.getByLabel('Enable External Access').check();
+    await page
+      .getByTestId('text-input-source-ranges.0.source-range')
+      .fill(externalAccess);
+  }
+  if (engineParameters != '' || addDefaultEngineParameters) {
+    //await page.getByRole('checkbox', {name: "engineParametersEnabled"}).check({force: true})
+    await page.getByLabel('Database engine parameters').check();
+    if (engineParameters != '')
```
fabio-silva, line 148: missing brackets

fabio-silva (changes requested):

ui/apps/everest/.e2e/utils/db-wizard.ts
tplavcic: thank you! fixed.

ui/apps/everest/.e2e/utils/generic.ts
fabio-silva: Fair point, haven't thought of that!

oksana-grishchenko (approved):
LGTM👍 Approving the PR since my concerns were resolved, however let's have an approval also at least from @fabio-silva who requested some changes earlier

fabio-silva (approved):

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -0,0 +1,202 @@
...
+        storageClasses = storageClassNames;
+      });
+
+      test(`Cluster creation with ${db} and size ${size}`, async ({
+        page,
+        request,
+      }) => {
+        test.setTimeout(size * 120000);
+
+        expect(storageClasses.length).toBeGreaterThan(0);
```
edithturn, line 80: Hi Tomislav!  :)
The script assumes that there will always be at least one storage class available. If no storage classes are present, the test will fail early. I might be missing something here, but to clarify, are we ensuring that at least one storage class is available for the end-to-end test to run? Is that right?

Thanks in advance for your clarification!

edithturn: Perhaps adding an error message could provide clearer feedback about the cause of the failure 🙂 

`if (storageClasses.length === 0) {
  throw new Error("No storage classes available. Please ensure the environment is properly set up.");
}
`

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -0,0 +1,202 @@
...
+          request
+        );
+        storageClasses = storageClassNames;
+      });
+
+      test(`Cluster creation with ${db} and size ${size}`, async ({
+        page,
+        request,
+      }) => {
+        test.setTimeout(size * 120000);
```
edithturn, line 78: I was going through this part and noticed that you made the timeout dynamic by calculating it as size * 120,000 milliseconds. So, for a cluster of size 1, the timeout would be 120,000 milliseconds (or 2 minutes).

I'm curious...what would happen if the timeout ends up being too short in one of those test environments? Should we display an error or maybe a warning?

Also, adding a comment in the code explaining why "120000 milliseconds per size" was necessary could help to understand the reason for using `120000`. Something like:

```
// Timeout is scaled based on the cluster size to account for the additional resources needed.
test.setTimeout(size * 120000);
```

tplavcic: Thank you, I have added your comment. If the test fails you will get a message that the timeout was exceeded and in which place.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -0,0 +1,202 @@
...
+        expect(addedCluster?.spec.engine.storage.size.toString()).toBe('1G');
+        expect(addedCluster?.spec.proxy.expose.type).toBe('internal');
+        expect(addedCluster?.spec.proxy.replicas).toBe(size);
+      });
+
+      test(`Suspend cluster with ${db} and size ${size}`, async ({ page }) => {
+        test.setTimeout(size * 120000);
+
+        await suspendDbCluster(page, clusterName);
+        if (size != 1 && db != 'postgresql') {
```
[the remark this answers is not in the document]
tplavcic: Added a comment

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -0,0 +1,202 @@
...
+        );
+
+        expect(addedCluster).not.toBeUndefined();
+        expect(addedCluster?.spec.engine.type).toBe(db);
+        expect(addedCluster?.spec.engine.replicas).toBe(size);
+        expect(['600m', '0.6']).toContain(
+          addedCluster?.spec.engine.resources?.cpu.toString()
+        );
+        expect(addedCluster?.spec.engine.resources?.memory.toString()).toBe(
+          '1G'
```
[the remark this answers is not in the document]
tplavcic: That would be good if the function was generic and re-used somewhere, but these values are used only in two places in a function which will not be re-used by anything so I don't want to complicate logic.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
tplavcic: We are not ensuring that at least one storage class is available, but usually all k8s environments have at least one. But if we accidentally find ourselves in such environment the check will fail and it should be self explanatory so I don't think we need to add something here.

PR #583, opened 2024-08-12
EVEREST-1292 | Migrate backup-storages and monitoring-configs during 1.2.0 upgrade

mayankshah1607:


oksana-grishchenko (approved):
LGTM. I would've concerned about the code duplication but it's a one-time thing, I don't think we need to overcomplicate (it looks like making the duplicated functions reusable would require even more code).

pkg/upgrade/upgrade.go
```diff
@@ -215,12 +215,16 @@ func (u *Upgrade) Run(ctx context.Context) error {
 					return err
 				}
 			}
-			// RBAC is added in 1.2.x, so if we're upgrading to that version, we need to ensure
-			// that the RBAC configmap is present.
 			if common.CheckConstraint(upgradeEverestTo, "~> 1.2.0") {
```
recharte, line 218: What if we eventually release a `1.2.1`? Can we make this check a `1.2.x`?

mayankshah1607: Yes, that constraint does indeed check for `1.2.x`. The `~>` is called the [pessimistic constraint operator](https://developer.hashicorp.com/terraform/language/expressions/version-constraints#version-constraint-syntax)

recharte: Ohhh yeah, sorry for that I didn't properly check it :see_no_evil:

PR #604, opened 2024-08-21
[UI] EVEREST-1296 Integration with new backup-storages and monitoring-instances APIs

dianabirs:
[EVEREST-1296](https://perconadev.atlassian.net/browse/EVEREST-1296)

[EVEREST-1296]: https://perconadev.atlassian.net/browse/EVEREST-1296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/apps/everest/src/components/confirm-dialog-namespace/confirm-dialog-namespace.tsx
```diff
@@ -0,0 +1,49 @@
+import { Button, Dialog, DialogActions, DialogContent } from '@mui/material';
+import { DialogTitle } from '@percona/ui-lib';
+import { kebabize } from '@percona/utils';
+import { ConfirmDialogNamespaceProps } from './confirm-dialog-namespace.types';
+
+export const ConfirmDialogNamespace = ({
```
solovevayaroslavna, line 6: Why we can't use standart confirm dialog with customization via props?

ui/apps/everest/src/pages/settings/monitoring-endpoints/createEditModal/create-edit-modal.tsx
```diff
@@ -62,13 +61,14 @@ export const CreateEditEndpointModal = ({
               placeholder: Messages.fieldPlaceholders.name,
             }}
           />
-          <AutoCompleteSelectAll
-            name={EndpointFormFields.namespaces}
-            label={Messages.fieldLabels.namespaces}
+          <AutoCompleteInput
```
solovevayaroslavna, line 64: Do we have changes in mocups? Where can i check them? We couldn't select all namespaces now?

dianabirs: The logic changes and now we deprecated `allowedNamespaces` and can only have one namespaces per resource, because every monitoring instance and backup storage will be namespaced

ui/apps/everest/src/components/confirm-dialog-namespace/confirm-dialog-namespace.tsx
dianabirs: I duplicated this component because the way we handle confirm for these two resources has to include a namespace param (which we don't need for the others; the hooks need to be sent the namespace)
```
  const onClick = () => {
    handleConfirm(selectedId, selectedNamespace);
  };
```
But now that you mentioned prop customization, I think we can have two separate props `selectedNamespace` and `handleConfirmNamespace` and use those in these two cases, instead of duplicating the component.. What do you say about this approach?

.github/workflows/dev-fe-e2e.yaml
```diff
@@ -138,7 +138,7 @@ jobs:
       - name: Run integration tests
         shell: bash
         env:
-          EVEREST_BUCKETS_NAMESPACES_MAP: '[["bucket-1",["everest-ui"]],["bucket-2",["psmdb-only"]],["bucket-3",["pxc-only"]],["bucket-4",["pg-only"]]]'
+          EVEREST_BUCKETS_NAMESPACES_MAP: '[["bucket-1","my-special-place"],["bucket-2","psmdb-only"],["bucket-3","pxc-only"],["bucket-4","pg-only"]]'
```
oksana-grishchenko, line 141: The e2e `EVEREST_CI_NAMESPACES` still have the `everest-ui` namespace in it: https://github.com/percona/everest/blob/2ad899853af7dbe963e71dc9d9bdee15b7c594ba/ui/apps/everest/.e2e/constants.ts#L7-L12  
Are you sure that `my-special-place` wasn't commited by a mistake?

dianabirs: yes, it was, reverting it now. Thanks for noticing!

solovevayaroslavna (approved):

PR #643, opened 2024-09-09
EVEREST-1405-hiding-sharding: hide sharding from UI and API

solovevayaroslavna:
image
hiding sharding from preview/steps/db details and API requests (create/update)
image
image


fabio-silva (approved):

recharte:
@solovevayaroslavna since we already have the release branch can we target base branch `release-1.2` instead of `main`? This way this change only affects the release branch and main continues to have this feature.

solovevayaroslavna:
@recharte then i will cherrypick commit to release, because it allows to avoid extra BE conflicts

recharte:
> @recharte then i will cherrypick commit to release, because it allows to avoid extra BE conflicts

Why do we need this in `main`? The BE will also only revert the sharding API changes in the `release-1.2` branch. If there's something in the PR that's no just about hiding the sharding options then it should probably be separated onto another PR :shrug:

solovevayaroslavna:
cherry-picked directly to release-1.2 branch

PR #644, opened 2024-09-09
EVEREST-1428 | Filter ListDatabaseEngines API call based on RBAC

mayankshah1607:
Similar to `ListNamespaces` API call, we will allow users to list db-engines, however the response itself shall be filtered based on the RBAC policy.

### DIFF
--- a/pkg/rbac/rbac.go
+++ b/pkg/rbac/rbac.go
@@ -47,6 +47,7 @@ const (
 	ResourceNamespaces              = "namespaces"
 	ResourceDatabaseClusterBackups  = "database-cluster-backups"
 	ResourceDatabaseClusterRestores = "database-cluster-restores"
+	ResourceDatabaseEngines         = "database-engines"
 )
 
 // RBAC actions.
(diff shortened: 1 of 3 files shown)


pkg/rbac/rbac.go
```diff
@@ -211,21 +212,23 @@ func NewEnforceHandler(basePath string, enforcer *casbin.Enforcer) func(c echo.C
...
 			return false, errors.New("invalid method")
 		}
+		// Always allowing listing all namespaces.
+		// The result is filtered based on permission.
+		if resource == ResourceNamespaces {
+			return true, nil
+		}
+		// Always allow listing database engines.
+		// The result is filtered based on permission.
+		if resource == ResourceDatabaseEngines && object == "" && action == ActionRead {
```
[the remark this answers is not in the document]
mayankshah1607: Yes.. you're right. Fixed it, thanks!

pkg/rbac/rbac.go
```diff
@@ -211,21 +212,23 @@ func NewEnforceHandler(basePath string, enforcer *casbin.Enforcer) func(c echo.C
...
+		name := c.Param("name")
+		object = namespace + "/" + name
+		// Always allowing listing all namespaces.
+		// The result is filtered based on permission.
+		if resource == ResourceNamespaces {
+			return true, nil
+		}
+		// Always allow listing database engines.
+		// The result is filtered based on permission.
+		if resource == ResourceDatabaseEngines && name == "" && action == ActionRead {
```
recharte, line 229: Should this be the following?
```suggestion
		if resource == ResourceDatabaseEngines && name == "" && (action == ActionRead || action == "*") {
```

mayankshah1607: No, I don't think so..

The value of `action` is derived from the HTTP method, so it will always be one of `read`, `create`, `updated`... etc. Whereas `*` is a wildcard we set in the RBAC policy

Moreover, we want to bypass the handler only when Listing database-engines, i.e, `GET /v1/namespaces/database-engines`. For the other `database-engines` APIs, we still need this enforcer

oksana-grishchenko (approved):

pkg/rbac/rbac.go
recharte: You're right, casbin handles that internally

recharte (approved):

PR #648, opened 2024-09-10
EVEREST-1433 | [BE] Check restore permissions when creating a new DB from backup

mayankshah1607:


api/validation.go
```diff
@@ -626,6 +626,16 @@ func (e *EverestServer) validateDatabaseClusterOnCreate(
 			return errors.Join(errInsufficientPermissions, errors.New("missing permission to take backups"))
 		}
 	}
+	// To be able to create a cluster from a backup (restore backup to new cluster), the user needs to explicitly
+	// have permissions to take restores.
+	sourceBackup := pointer.Get(pointer.Get(databaseCluster.Spec.DataSource).DbClusterBackupName)
```
oksana-grishchenko, line 631: Since `databaseCluster.Spec` is `*struct` should we apply another `pointer.Get` to it?  
```suggestion
	sourceBackup := pointer.Get(pointer.Get(pointer.Get(databaseCluster.Spec).DataSource).DbClusterBackupName)
```

oksana-grishchenko (approved):

PR #655, opened 2024-09-11
EVEREST-1450 | [RBAC] policy is not honored for backups and restores

mayankshah1607:
### Problem

RBAC does not honor the policies for `database-cluster-backups` and `database-cluster-restores` when performing List operations

### Cause

The RBAC system in Everest uses the `x-everest-resource-name` tag defined in the swagger spec to identify the resource types per API call. The tag had incorrect values for the backups and restores list APIs. As a result, the permissions were not being checked correctly.

### Fix

- Update the `x-everest-resource-name` tags for these APIs
- For the same APIs, rename the `name` query parameter with `cluster-name`. This is done because the RBAC enforcer uses the value of `name` when check for db-cluster permissions, however, for these APIs, we don't need the db cluster name.

### DIFF
--- a/ui/apps/everest/src/utils/useGetPermissions.ts
+++ b/ui/apps/everest/src/utils/useGetPermissions.ts
@@ -52,7 +52,7 @@ export const useGetPermissions = ({
     canRead: false,
     canUpdate: false,
     canDelete: false,
-    canCreate: true,
+    canCreate: false,
   });
 
   const { authorize } = useContext(AuthContext);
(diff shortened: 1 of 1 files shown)


docs/spec/openapi.yml
```diff
@@ -481,7 +481,7 @@ paths:
           required: true
           schema:
             type: string
-        - name: name
+        - name: cluster-name
```
recharte, line 484: Doesn't this require some changes in the api package? :thinking:

recharte: I don't know if it does, I'm searching for it now...

mayankshah1607: I don't think it does for path parameters (only query parameters, I think..)

recharte: :thinking: 
Doesn't [this function](https://github.com/percona/everest/blob/f0b8dc8f40d3643c71ae1f71ae22587b62815fae/api/database_cluster_backup.go#L49) need to match [this interface](https://github.com/percona/everest/blob/957d3844fc54e8bdd52eec43cdc2c12a60b971a4/api/everest-server.gen.go#L1610)?

mayankshah1607: And it does, the variable names don't have to be the same for it to match 😅

recharte: Right, it's all good then

recharte (approved):

PR #684, opened 2024-09-17
EVEREST-1488 | [RBAC] Improved debuggability and code refactoring

mayankshah1607:
- Add log messages whenever permissions are denied, for improved debugging
- Get rid of all `canGetXXX`, `canUpateXXX`, etc. function calls, and instead use a single wrapper method for checking explicit permissions
- Use centralised error handling for 403 errors
- Remove duplicate code
- Remove duplicate proxy calls


oksana-grishchenko (approved):

api/everest.go
```diff
@@ -330,3 +335,37 @@ func (e *EverestServer) k8sToAPIErrorHandler(next echo.HTTPErrorHandler) echo.HT
...
+				Code:    http.StatusForbidden,
+				Message: errInsufficientPermissions.Error(),
+			}
+		}
+		next(err, c)
+	}
+}
+
+// enforceOrErr is a wrapper around casbin.Enforce that returns an error if the enforcement fails.
+func (e *EverestServer) enforceOrErr(subject, resource, action, object string) error {
```
recharte, line 352: I don't like the name `enforceOrErr` very much. Most functions in go return error and we don't call all of them `xOrErr` we call them just `x` :sweat_smile:

recharte (approved):
Logic looks good, just a small naming nitpick that's up to you

PR #685, opened 2024-09-17
EVEREST-1475 use type generation for API tests

yurkovychv:


recharte:
Our Makefile has a [target (`gen`) ](https://github.com/percona/everest/blob/8642f65b75111397fe16f1f7fcb572e6362baa6f/Makefile#L48-L50) that among other things generates go files based on the openapi schema and then we have a step in our CI that runs `make gen` and [checks that we didn't miss committing any auto-generated files](https://github.com/percona/everest/blob/8642f65b75111397fe16f1f7fcb572e6362baa6f/.github/workflows/dev-be-ci.yaml#L86-L94).
We should probably add this generation of JS files from the openapi schema to the `gen` makefile target so that we don't forget to update the API tests files every time that we change the schema.

recharte (approved):

oksana-grishchenko (approved):

api-tests/support/types/database-clusters.ts
```diff
@@ -0,0 +1,6 @@
+import { paths } from '@generatedTypes';
```
tplavcic, line 1: If I see correctly types from this file are currently not used?

tplavcic (changes requested):

api-tests/support/types/database-clusters.ts
tplavcic: ok, let's leave it

api-tests/support/types/database-engines.ts
```diff
@@ -0,0 +1,4 @@
+import { paths } from '@generatedTypes';
+
+export type DatabaseEnginesResponse = paths['/namespaces/{namespace}/database-engines']['get']['responses']['200']['content']['application/json'];
```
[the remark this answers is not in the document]
tplavcic: That was my initial thinking when I saw the naming yes, it would be good to have the consistent naming in all types.

tplavcic (approved):

PR #686, opened 2024-09-17
EVEREST-1477 | [RBAC] Enforce backup-storage and monitoring-instances permission for database-clusters

mayankshah1607:
Since database-clusters contain information about backup-storages and monitoring-instances, we need to check if the user has permission for those specified resources.

### DIFF
--- a/api/database_cluster.go
+++ b/api/database_cluster.go
@@ -88,17 +88,9 @@ func (e *EverestServer) CreateDatabaseCluster(ctx echo.Context, namespace string
 	return err
 }
 
-// ListDatabaseClusters lists the created database clusters on the specified kubernetes cluster.
-func (e *EverestServer) ListDatabaseClusters(ctx echo.Context, namespace string) error {
-	user, err := rbac.GetUser(ctx)
-	if err != nil {
-		return ctx.JSON(http.StatusInternalServerError, Error{
-			Message: pointer.ToString("Failed to get user from context" + err.Error()),
-		})
-	}
-
-	// Filter out all DatabaseClusters whose backup-storages are not accessible by the user.
-	rbacFilter := transformK8sList(func(l *unstructured.UnstructuredList) error {
+//nolint:gocognit
+func (e *EverestServer) enforceDatabaseClusterRBAC(user string) func(l *unstructured.UnstructuredList) error {
+	return func(l *unstructured.UnstructuredList) error {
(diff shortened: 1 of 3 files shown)


api/database_cluster.go
```diff
@@ -133,6 +192,50 @@ func (e *EverestServer) DeleteDatabaseCluster(
...
+			e.l.Error(errors.Join(err, errors.New("failed to check monitoring-config permissions")))
+			return err
+		} else if !ok {
+			return ctx.JSON(http.StatusForbidden, Error{
+				Message: pointer.ToString(errInsufficientPermissions.Error()),
+			})
+		}
+	}
+	// TODO: duplicate call to Kubernetes, need to figure out how to fix this.
 	return e.proxyKubernetes(ctx, namespace, databaseClusterKind, name)
```
recharte, line 239: Won't this 2nd call return the unfiltered list, thus completely bypassing the RBAC filtering we just did?

recharte: Okay I just found that this is the GET not the LIST but still could't we just remove this 2nd altogether? What's stopping us from doing that?

mayankshah1607: We don't return any list here, this is a `Get` call. If the RBAC fails, it returns a 403.

So it either returns the whole object or an error.

mayankshah1607: > Okay I just found that this is the GET not the LIST but still could't we just remove this 2nd altogether? What's stopping us from doing that?

TBH, I'm not sure. I see that the proxy does some response modification, and I assume we need to do the same for the K8s response - https://github.com/percona/everest/blob/main/api/proxy.go#L194

I'm yet to figure out what this does, so I left a TODO here to unblock the PR and RC-3 release..

mayankshah1607: BTW, I'm doing some code refactoring in a [separate PR](https://github.com/percona/everest/pull/684), and thought I'll address it there. I see we do the duplicate calls in other places too..

recharte: I see... I'm also not a big fan of the fact that we had to duplicate what's effectively the same filtering for the List and the Get. It would be great if we could find a way to use the same logic on both.

recharte (approved):

api/database_cluster.go
recharte: But that can be done on the separate refactoring PR

PR #689, opened 2024-09-17
EVEREST-1469 | [RBAC] Fix broken matcher function

mayankshah1607:
### Problem

We added a custom globMatch function in https://github.com/percona/everest/pull/665 to satisfy the use case of being able to define  objects in the format`*/<resource-1>`, however, it broke the namespaces permission.

### Solution

It looks like we do not really need a custom globMatch afterall, and our use-case can be handled well by the default `globMatch`. The reason for adding a custom function earlier was so that it could also satisfy a wildcard in the request (for example, in the `can` command), however, we never get a wildcard from the request side, except for the `can` command, so we can explicitly handle it there instead of adding any custom logic in the key matcher.

I'm also not really sure how the broken matcher got through, since it worked fine during the local testing, and also did not CI errors.

### DIFF
--- a/pkg/rbac/rbac.go
+++ b/pkg/rbac/rbac.go
@@ -145,8 +145,8 @@ func NewEnforcer(ctx context.Context, kubeClient *kubernetes.Kubernetes, l *zap.
 	return enforcer, refreshEnforcerInBackground(ctx, kubeClient, enforcer, l)
 }
 
-// We use a custom defined keyMatch function since we have use-cases that are not covered
-// by the default matchers in the casbin library.
+// We use a custom defined keyMatch function since we have use-cases that are not
+// properly handled by the key matchers in the casbin library.
 //
 // For example:
 // - keyMatch("ns-1/obj-1", "*/obj-1") => true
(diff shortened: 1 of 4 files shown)


pkg/rbac/validate_test.go
```diff
@@ -287,3 +289,21 @@ func TestCan(t *testing.T) {
...
+	assert.True(t, objectMatch("namespace-1/", "namespace-1/*"))
+	assert.True(t, objectMatch("namespace-1/object-1", "namespace-1/*"))
+	assert.False(t, objectMatch("namespace-2/", "namespace-1/*"))
+	assert.False(t, objectMatch("namespace-2/object-1", "namespace-1/*"))
+	assert.True(t, objectMatch("namespace-1/", "*/*"))
+	assert.True(t, objectMatch("namespace-1/object-1", "*/*"))
+	assert.False(t, objectMatch("namespace-1/object-1", "namespace-2/object-1"))
+	assert.False(t, objectMatch("namespace-2/object-2", "namespace-2/object-1"))
+	assert.False(t, objectMatch("namespace-2/object-2", "*/object-1"))
+	assert.True(t, objectMatch("namespace-2/object-1", "*/object-1"))
```
mayankshah1607, line 307: This is the use-case we needed, which was not covered originally by the `keyMatch` function (but covered in globMatch)

pkg/rbac/validate_test.go
```diff
@@ -287,3 +289,21 @@ func TestCan(t *testing.T) {
 		})
 	}
 }
+
+func TestObjectMatch(t *testing.T) {
+	t.Parallel()
+	assert.True(t, objectMatch("", "*"))
+	assert.True(t, objectMatch("namespace-1/object-1", "*"))
```
mayankshah1607, line 296: These are the cases which broke when we used `globMatch` instead. Using `keyMatch() || globMatch` also did not work, since it wrongly allowed certain cases..

pkg/rbac/rbac.go
```diff
@@ -142,30 +140,6 @@ func NewEnforcer(ctx context.Context, kubeClient *kubernetes.Kubernetes, l *zap.
 	return enforcer, refreshEnforcerInBackground(ctx, kubeClient, enforcer, l)
 }
 
-// To understand why we need this custom globMatch, consider the below policy:
-// ```
-// p, adminrole:role, database-clusters, read, */*
-// ```
-// Given a request `adminrole:role read database-clusters *`, the default globMatch returns false.
```
mayankshah1607, line 149: We never actually get '*' from the request side. I was testing all this while using the `can` command, where a `*` is a valid request parameter, so I just handled the case there internally to make it resemble a real enforce request (like at the API level)

fabio-silva (approved):

recharte (approved):

PR #691, opened 2024-09-18
EVEREST-1491 | Add support for toggling RBAC

mayankshah1607:
- RBAC can be toggled on/off by setting the `enabled` property of the `everest-rbac` configmap
- By default, RBAC is off
- If unspecified, RBAC is off
- No restart required for this toggle to take effect


pkg/rbac/rbac.go
```diff
@@ -137,6 +141,11 @@ func NewEnforcer(ctx context.Context, kubeClient *kubernetes.Kubernetes, l *zap.
 	if err != nil {
 		return nil, err
 	}
+	cm, err := adapter.ConfigMap(ctx)
+	if err != nil {
+		return nil, errors.Join(err, errors.New("failed to get RBAC ConfigMap"))
+	}
+	enforcer.EnableEnforce(IsEnabled(cm))
```
recharte, line 148: Just to be sure, this will affect all of our custom validations because we always use the `can...` functions that call the enforcer, correct? Is there any custom validation that doesn't rely on the enforcer under the hood?

pkg/rbac/configmap-adapter/adapter.go
```diff
@@ -50,6 +51,15 @@ func New(
...
 
+// ConfigMap returns the configmap used for RBAC.
+func (a *Adapter) ConfigMap(ctx context.Context) (*corev1.ConfigMap, error) {
+	return a.kubeClient.GetConfigMap(
+		ctx,
+		a.namespacedName.Namespace,
+		a.namespacedName.Name,
+	)
+}
+
```
recharte, line 62: You defined this function but I don't think you've used it anywhere...

pkg/rbac/rbac.go
mayankshah1607: There isn't.. Yes this will disable the enforcer, so even the custom validations will effectively be ignored

pkg/rbac/configmap-adapter/adapter.go
mayankshah1607: I did, it's in line 144 of the pkg/rbac/rbac.go file..

recharte: My bad, sorry for that :see_no_evil:

recharte (approved):
The CI is failing because you missed regenerating the mocks after updating the kubernetes client package but everything else is looking good so I'm approving the PR already

recharte:
@mayankshah1607 Ran autogen for you to unblock the PR :slightly_smiling_face:

recharte:
One last thing...
Should we explicitly set the
```enabled: false```
[here](https://github.com/percona/everest/blob/d2ec12096d3021716ce2434a1b14de6041d2d291/deploy/quickstart-k8s.yaml#L92-L93)?

mayankshah1607:
Good idea. If unspecified, it is false, but good to add for the sake of clarity..

mayankshah1607:
@recharte CI is super flaky in this PR. Please feel free to force merge, I see it has already passed once..

PR #696, opened 2024-09-18
EVEREST-1500 | Enforce database-cluster-credentials permission on restores

mayankshah1607:


api/validation.go
```diff
@@ -633,19 +633,48 @@ func (e *EverestServer) validateDatabaseClusterOnCreate(
...
+		return false, err
+	} else if !can {
+		return false, nil
+	}
+
+	// Get the name of the source database cluster.
+	bkp, err := e.kubeClient.GetDatabaseClusterBackup(ctx, namespace, sourceBackup)
+	if err != nil {
+		return false, errors.Join(err, errors.New("failed to get database cluster backup"))
+	}
```
recharte, line 666: I know it's more or less implied (it would be hard to know the backup name without it) but shouldn't we also validate if the user has read access to that backup?

recharte: I don't know if this is not overdoing stuff... One could argue that if you check for read permissions on the backup you should also have read permissions on the backup storage where that backup is stored. Am I jumping onto a rabbit hole?

mayankshah1607: Updated

recharte (approved):

PR #746, opened 2024-10-08
Everest 1306  sharding

solovevayaroslavna:
[sharding demo.webm](https://github.com/user-attachments/assets/1ac57e2d-e263-4a11-bf68-c2c7482b5063)

full changes scope is under the ticket => https://perconadev.atlassian.net/browse/EVEREST-1306


ui/apps/everest/src/components/cluster-form/resources/constants.ts
```diff
@@ -68,14 +68,16 @@ export const DEFAULT_SIZES = {
...
-  },
+export const DEFAULT_CONFIG_SERVERS = ['1', '3', '5', '7'];
+
+export const MIN_NUMBER_OF_SHARDS = '1';
+
+export const getDefaultNumberOfconfigServersByNumberOfNodes = (
+  numberOfNodes: number
+) => {
+  if (DEFAULT_CONFIG_SERVERS.includes(numberOfNodes.toString())) {
+    return numberOfNodes <= 5 ? numberOfNodes.toString() : '7';
```
fabio-silva, line 79: not sure I understand this.

`DEFAULT_CONFIG_SERVERS.includes(numberOfNodes.toString())` means `numberOfNodes` has to be 1, 3, 5 or 7.
If `numberOfNodes > 5`, we can still return `numberOfNodes.toString()`, no need for ternary operator

ui/apps/everest/src/components/cluster-form/resources/messages.tsx
```diff
@@ -4,6 +4,10 @@ export const Messages = {
     min: (val: number) => `The value cannot be less than ${val}`,
     max: (val: number) => `The value cannot be more than ${val}`,
     odd: 'The value cannot be even',
+    numberOfShards: '№ of shards',
```
fabio-silva, line 7: what character is this? Seems different from standard ones

ui/apps/everest/src/pages/database-form/database-form-body/steps/resources/resources-step.tsx
```diff
@@ -22,7 +22,7 @@ export const ResourcesStep = () => {
         pairProxiesWithNodes={mode !== 'edit'}
         disableDiskInput={mode === 'edit'}
         allowDiskInputUpdate={mode !== 'edit'}
-        showSharding={dbType === DbType.Mongo && mode !== 'edit'}
+        showSharding={dbType === DbType.Mongo}
```
fabio-silva, line 25: sharding should not be available when editing a cluster. Any reason why we are removing this condition?

ui/apps/everest/src/components/cluster-form/resources/constants.ts
```diff
@@ -68,14 +68,16 @@ export const DEFAULT_SIZES = {
...
-  [DbWizardFormFields.shardConfigServers]: {
-    min: '1',
-    max: '7',
-  },
-  [DbWizardFormFields.shardNr]: {
-    min: '1',
-  },
+export const DEFAULT_CONFIG_SERVERS = ['1', '3', '5', '7'];
+
+export const MIN_NUMBER_OF_SHARDS = '1';
```
fabio-silva, line 73: Setting default number as strings seem counter-intuitive and prone to error.
I think we should define numbers instead, and convert to string where necessary.,

ui/apps/everest/src/components/cluster-form/resources/resources.tsx
```diff
@@ -1,33 +1,46 @@
...
   NODES_DB_TYPE_MAP,
   ResourceSize,
-  SHARDING_DEFAULTS,
 } from './constants';
 import { DbWizardFormFields } from 'consts';
 import { DbType } from '@percona/types';
 import { getProxyUnitNamesFromDbType } from './utils';
 import { ResourcesTogglesProps, ResourceInputProps } from './resources.types';
 import { Messages } from './messages';
+import { Theme } from '@emotion/react';
```
fabio-silva, line 43: I think this is the wrong theme. Should be the one from MUI

fabio-silva (changes requested):

ui/apps/everest/src/components/cluster-form/resources/constants.ts
solovevayaroslavna: @fabio-silva we have only 4 values for config servers, that's why
image

ui/apps/everest/src/components/cluster-form/resources/messages.tsx
solovevayaroslavna: good catch, Fabio! I used one from my keyboard, but i see we use another one Nº, so will change it

ui/apps/everest/src/pages/database-form/database-form-body/steps/resources/resources-step.tsx
solovevayaroslavna: @fabio-silva yes, new requirenments (https://perconadev.atlassian.net/browse/EVEREST-1306)
image

we will not be able to enable/disable it, so toggle will be disabled but we should be able to change number of shards and number of config servers

solovevayaroslavna: also in restoreToNewDb mode

ui/apps/everest/src/components/cluster-form/resources/constants.ts
fabio-silva: Right, but here's the thing:
- Options are 1, 3, 5, 7
- `numberOfNodes <= 5 ? numberOfNodes.toString() : '7';` -> if number is greater than 5, it can only be 7 anyways, so we can just do `numberOfNodes.toString() : '7'` right away without this condition

fabio-silva (changes requested):
see last comment

ui/apps/everest/src/components/cluster-form/resources/constants.ts
solovevayaroslavna: fixed in https://github.com/percona/everest/pull/746/commits/a7a359f836db0456bfadd1842b1af6e6f877a4a8

ui/apps/everest/src/components/cluster-form/resources/messages.tsx
solovevayaroslavna: fixed in https://github.com/percona/everest/pull/746/commits/e10cef70b608f1c8b6169813e10e2b54dac6b435

ui/apps/everest/src/components/cluster-form/resources/resources.tsx
solovevayaroslavna: fixed in https://github.com/percona/everest/pull/746/commits/e10cef70b608f1c8b6169813e10e2b54dac6b435

fabio-silva (approved):

PR #776, opened 2024-10-21
EVEREST-1562 | Add --dry-run flag to upgrade cmd

mayankshah1607:
Setting the `--dry-run` flag to the upgrade command will only run the pre-requisite checks, and not perform the actual upgrade.


commands/upgrade.go
```diff
@@ -70,6 +70,7 @@ func newUpgradeCmd(l *zap.SugaredLogger) *cobra.Command {
 func initUpgradeFlags(cmd *cobra.Command) {
 	cmd.Flags().String("version-metadata-url", "https://check.percona.com", "URL to retrieve version metadata information from")
 	cmd.Flags().BoolP("logs", "l", false, "If set, logs are printed during the upgrade process")
+	cmd.Flags().Bool("dry-run", false, "If set, simulate the upgrade process")
```
oksana-grishchenko, line 73: It could be not obvious what does the "simulate" actually mean in this case. Should we be more specific and explain that only the upgrade prerequisites will be checked but the upgrade itself won't be performed?

oksana-grishchenko (approved):

PR #792, opened 2024-10-28
EVEREST-1590: support for installation in OpenShift

michal-kralik:
EVEREST-1590

OpenShift support
- Installation
- Upgrade
- Uninstallation

### DIFF
--- a/commands/install.go
+++ b/commands/install.go
@@ -17,6 +17,7 @@
 package commands
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
@@ -24,6 +25,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/percona/everest/pkg/install"
+	"github.com/percona/everest/pkg/kubernetes"
 	"github.com/percona/everest/pkg/output"
 )
 
@@ -78,6 +80,12 @@ func initInstallFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(install.FlagOperatorMongoDB, true, "Install MongoDB operator")
 	cmd.Flags().Bool(install.FlagOperatorPostgresql, true, "Install PostgreSQL operator")
 	cmd.Flags().Bool(install.FlagOperatorXtraDBCluster, true, "Install XtraDB Cluster operator")
+
+	cmd.Flags().Bool(install.FlagSkipEnvDetection, false, "Skip detecting Kubernetes environment where Everest is installed")
+	cmd.Flags().String(install.FlagCatalogNamespace, kubernetes.OLMNamespace,
+		fmt.Sprintf("Namespace where Everest OLM catalog is installed. Implies --%s", install.FlagSkipEnvDetection),
+	)
+	cmd.Flags().Bool(install.FlagSkipOLM, false, fmt.Sprintf("Skip OLM installation. Implies --%s", install.FlagSkipEnvDetection))
 }
 
 func initInstallViperFlags(cmd *cobra.Command) {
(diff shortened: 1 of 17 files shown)


pkg/uninstall/uninstall.go
```diff
@@ -172,14 +212,25 @@ func (u *Uninstall) Run(ctx context.Context) error { //nolint:funlen
...
+		uninstallSteps = append(uninstallSteps, common.Step{
+			Desc: "Delete Everest Catalog",
+			F: func(ctx context.Context) error {
+				u.l.Info("Deleting Everest catalog")
+				return u.kubeClient.DeleteCatalogSource(ctx, u.config.CatalogNamespace, catalogSource)
+			},
+		})
+	} else {
+		// There are no resources with finalizers in the monitoring namespace, so
+		// we can delete it directly
```
mayankshah1607, line 226: I know this comment was already there, but it doesn't really make any sense. Maybe it was placed there by mistake during a previous  refactor?

pkg/upgrade/upgrade.go
```diff
@@ -47,6 +47,13 @@ const (
 	contextTimeout    = 5 * time.Minute
 	backoffInterval   = 5 * time.Second
 	backoffMaxRetries = 5
+
+	// FlagCatalogNamespace is the name of the catalog namespace flag.
+	FlagCatalogNamespace = "catalog-namespace"
+	// FlagSkipEnvDetection is the name of the skip env detection flag.
+	FlagSkipEnvDetection = "skip-env-detection"
+	// FlagSkipOLM is the name of the skip OLM flag.
+	FlagSkipOLM = "skip-olm"
```
mayankshah1607, line 56: These flag constants are declared multiple times, would be nice if we can declare in one place and re-use in all 3 commands, wdyt?

mayankshah1607 (changes requested):

commands/install.go
```diff
@@ -78,6 +80,12 @@ func initInstallFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(install.FlagOperatorMongoDB, true, "Install MongoDB operator")
 	cmd.Flags().Bool(install.FlagOperatorPostgresql, true, "Install PostgreSQL operator")
 	cmd.Flags().Bool(install.FlagOperatorXtraDBCluster, true, "Install XtraDB Cluster operator")
+
+	cmd.Flags().Bool(install.FlagSkipEnvDetection, false, "Skip detecting Kubernetes environment where Everest is installed")
+	cmd.Flags().String(install.FlagCatalogNamespace, kubernetes.OLMNamespace,
+		fmt.Sprintf("Namespace where Everest OLM catalog is installed. Implies --%s", install.FlagSkipEnvDetection),
+	)
+	cmd.Flags().Bool(install.FlagSkipOLM, false, fmt.Sprintf("Skip OLM installation. Implies --%s", install.FlagSkipEnvDetection))
```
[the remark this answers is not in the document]
michal-kralik: Yes, this is great for environments where we cannot detect the type properly but we give users full control over the OLM. Environments similar to OpenShift would benefit from this flag.

pkg/uninstall/uninstall.go
```diff
@@ -92,14 +106,40 @@ func NewUninstall(c Config, l *zap.SugaredLogger) (*Uninstall, error) {
...
+
+	if !u.config.SkipEnvDetection {
+		u.l.Info("Detecting Kubernetes environment")
+		env, err := u.kubeClient.DetectEnvironment(ctx)
+		if err != nil {
+			return err
+		}
+		u.config.SkipOLM = env.SkipOLM
+		u.config.CatalogNamespace = env.CatalogNamespace
+	}
```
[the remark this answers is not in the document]
michal-kralik: Yes, good catch. Some leftover from refactoring.

pkg/uninstall/uninstall.go
michal-kralik: updated

pkg/upgrade/upgrade.go
michal-kralik: I considered this originally but then I thought even though they have the same string representation, they may have a slightly different meaning and I did not want to couple them together across commands.
If they were common constants, changing them seemed prone to errors - i.e. we may want to change one command but inadvertently change the other commands as well.

mayankshah1607 (approved):

PR #801, opened 2024-10-31
fix: wrong proxy expose type for psmdb

fabio-silva:
https://perconadev.atlassian.net/browse/EVEREST-1603

### DIFF
--- a/ui/apps/everest/src/pages/databases/expandedRow/ExpandedRow.tsx
+++ b/ui/apps/everest/src/pages/databases/expandedRow/ExpandedRow.tsx
@@ -52,6 +52,7 @@ export const ExpandedRow = ({
     port,
     raw,
   } = row.original;
+  console.log('row.original', row.original);
   const parsedDiskValues = memoryParser(storage.toString());
   const parsedMemoryValues = memoryParser(memory.toString());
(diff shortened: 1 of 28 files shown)


ui/apps/everest/src/pages/databases/expandedRow/ExpandedRow.tsx
```diff
@@ -52,6 +52,7 @@ export const ExpandedRow = ({
     port,
     raw,
   } = row.original;
+  console.log('row.original', row.original);
```
solovevayaroslavna, line 55: console.log ☝️

solovevayaroslavna, line 55: console.log ☝️

solovevayaroslavna (changes requested):

ui/apps/everest/src/pages/databases/expandedRow/ExpandedRow.tsx
fabio-silva: probably we don't have that rule, better to add it in a PR 🤔

solovevayaroslavna (approved):

PR #812, opened 2024-11-07
EVEREST-1518: validate db engine read access during db cluster read/update/delete

michal-kralik:
EVEREST-1518

### DIFF
--- a/api/database_cluster.go
+++ b/api/database_cluster.go
@@ -87,8 +87,9 @@ func (e *EverestServer) CreateDatabaseCluster(ctx echo.Context, namespace string
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


mayankshah1607 (approved):

recharte:
I haven't reviewed the PR yet, but decided to comment since it's already approved. Does this apply to create as well?

michal-kralik:
I don't think it does. I'll add it.

recharte:
> I don't think it does. I'll add it.

Thanks, this way you'll be able to fix [EVEREST-1604](https://perconadev.atlassian.net/browse/EVEREST-1604) as well :slightly_smiling_face:

PR #825, opened 2024-11-11
EVEREST-1514 | [CLI] use helm chart for install/upgrade/uninstall

mayankshah1607:


pkg/cli/install/install.go
```diff
@@ -0,0 +1,623 @@
...
+	kubeClient     *kubernetes.Kubernetes
+	versionService versionservice.Interface
+
+	// these are set only when Run is called.
+	clusterType    kubernetes.ClusterType
+	installVersion string
+	helmInstaller  *helm.Installer
+}
+
+const operatorInstallThreads = 1
```
recharte, line 113: To keep things tidy I suggest we move this to the consts block above.

pkg/cli/install/install.go
```diff
@@ -0,0 +1,623 @@
...
+
+	// everestDBNamespaceSubChartPath is the path to the everest-db-namespace subchart relative to the main chart.
+	dbNamespaceSubChartPath = "/charts/everest-db-namespace"
+
+	pollInterval    = 5 * time.Second
+	pollTimeout     = 10 * time.Minute
+	backoffInterval = 5 * time.Second
+)
+
+const postInstallMessage = "Everest has been successfully installed!"
```
recharte, line 96: To keep things tidy I suggest we move this to the consts block above.

pkg/cli/install/steps.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+
+			o.l.Infof("Installing DB namespace Helm chart in namespace ", namespace)
+			if err := installer.Install(ctx); err != nil {
+				return fmt.Errorf("could not install Helm chart: %w", err)
+			}
+
+			if err := o.installDBOperators(ctx, namespace); err != nil {
+				return err
+			}
```
recharte, line 187: I'm quite confused about this :see_no_evil: 
I thought that the DB namespace sub chart would handle all logic of installing the DB operators. Why are we managing the Subscriptions/InstallPlans manually here?

pkg/cli/install/steps.go
```diff
@@ -0,0 +1,191 @@
...
+			cs, err := helmutils.GetEverestCatalogSource(rel)
+			if err != nil {
+				return fmt.Errorf("could not get Everest CatalogSource from Helm chart: %w", err)
+			}
+			return wait.PollUntilContextTimeout(ctx, pollInterval, pollTimeout, false, func(ctx context.Context) (bool, error) {
+				cs, err = o.kubeClient.GetCatalogSource(ctx, cs.GetName(), cs.GetNamespace())
+				if err != nil {
+					return false, fmt.Errorf("catalog source not found")
+				}
+				return pointer.Get(cs.Status.GRPCConnectionState).LastObservedState == "READY", nil
```
recharte, line 106: Nice

pkg/cli/install/steps.go
```diff
@@ -0,0 +1,191 @@
...
+		Desc: "Ensuring OLM components are ready",
+		F: func(ctx context.Context) error {
+			depls, err := o.kubeClient.ListDeployments(ctx, kubernetes.OLMNamespace)
+			if err != nil {
+				return err
+			}
+			for _, depl := range depls.Items {
+				if err := o.kubeClient.WaitForRollout(ctx, depl.GetName(), depl.GetNamespace()); err != nil {
+					return err
+				}
```
recharte, line 73: Would the logging provided by the `waitForDeployment` wrapper be useful here or would it be too much?

pkg/cli/helm/installer.go
```diff
@@ -0,0 +1,351 @@
...
+		// Release does not exist, install it.
+		if errors.Is(err, driver.ErrReleaseNotFound) {
+			return i.install(ctx)
+		}
+	}
+	// If the release already exists, we will re-apply the manifests using upgrade.
+	// We're not actually upgrading to a new version, but using upgrade to re-apply manifests.
+	// This is how Helm expects us to re-apply manifests.
+	// To prevent accidental version upgrades, we will explicitly check that the resolved chart version matches the installed chart version.
+	if i.chart.Metadata.Version != rel.Chart.Metadata.Version {
```
recharte, line 132: Nice

pkg/cli/helm/installer.go
```diff
@@ -0,0 +1,351 @@
...
+
+func (i *Installer) install(ctx context.Context) error {
+	install := action.NewInstall(i.cfg)
+	install.ReleaseName = i.ReleaseName
+	install.Namespace = i.ReleaseNamespace
+	install.DryRun = i.DryRun
+	install.CreateNamespace = i.CreateReleaseNamespace
+	install.Wait = false
+	install.TakeOwnership = true
+	install.DisableHooks = true
```
recharte, line 169: :thinking: does this mean that we need to manage the IP approval process ourselves? Is this the reason for calling `installDBOperators` inside `provisionDBNamespace`? Why did you go with this approach?

pkg/install/install.go
```diff
@@ -1,940 +0,0 @@
...
-
-func (o *Install) checkRequirements(meta *versionpb.MetadataVersion) error {
-	supVer, err := common.NewSupportedVersion(meta)
-	if err != nil {
-		return err
-	}
-
-	if err := common.CheckK8sRequirements(supVer, o.l, o.kubeClient); err != nil {
-		return err
-	}
```
recharte, line 288: Just to confirm, now the k8s version requirement is baked into the chart itself so we no longer need to specify it in the version service, hence the removal of this check, correct?

pkg/common/constants.go
```diff
@@ -32,8 +32,10 @@ const (
 	MonitoringNamespace = "everest-monitoring"
 	// PerconaEverestDeploymentName stores the name of everest API Server deployment.
 	PerconaEverestDeploymentName = "percona-everest"
+	// PerconaEverestCatalogName is the name of the Everest catalog source.
+	PerconaEverestCatalogName = "everest-catalog"
 	// PerconaEverestOperatorDeploymentName stores the name of everest operator deployment.
-	PerconaEverestOperatorDeploymentName = "everest-operator-controller-manager"
+	PerconaEverestOperatorDeploymentName = "everest-operator"
```
recharte, line 38: I'm so glad that the helm chart brought this along :pray: I hated the name :joy:

pkg/common/constants.go
```diff
@@ -32,8 +32,10 @@ const (
 	MonitoringNamespace = "everest-monitoring"
 	// PerconaEverestDeploymentName stores the name of everest API Server deployment.
 	PerconaEverestDeploymentName = "percona-everest"
```
recharte, line 34: I know that I should have commented this in the PR of the helm chart but I'm only paying attention to it now that I commented about the operator name below. Can we change this name without breaking stuff? :sweat_smile:

pkg/cli/install/steps.go
mayankshah1607: Yeah, the sub-chart can actually handle this.. initially I didn't want to run any hooks at all in the CLI so that we could show the steps in detail to the user. For example, with the monitoring, I wanted it to be a separate step rather than being a part of the "Installing Helm chart". (now its a different story since with [this](https://github.com/percona/percona-helm-charts/pull/416) PR all hooks are gone from the core chart)

But now that I think about it, maybe we can allow the use of hooks in the sub-chart since the DB provisioning happens in 1 step.

Another reason is that if I enable hooks in the sub-chart, it will also run the clean-up hooks (remove DB, BackupStorage, etc.) and we won't be able to show those steps in the CLI. But later I figured a workaround to selectively disable hooks..

So I think we can get rid of this code afterall and let the sub-chart hooks handle it. Thanks for noticing this :)

pkg/cli/helm/installer.go
mayankshah1607: I answered here - https://github.com/percona/everest/pull/825#discussion_r1849470128

pkg/install/install.go
mayankshah1607: Yes, correct. Helm checks it for us

This is where it is specified - https://github.com/percona/percona-helm-charts/blob/main/charts/everest/Chart.yaml#L7

pkg/common/constants.go
mayankshah1607: At least within Everest, I don't think it breaks anything. The updated upgrade logic has some code to remove the EverestOperator Subscription first and then apply the chart so that the new Deployment can take over..

But if a user has this name hard-coded somewhere in their infrastructure (like CI, scripts, or code that depends on it).. then yes, it could break that.

Maybe we can document this in the upgrade guide. Optionally, we can go back to the original name (which I don't like either BTW 😂 ). Do you have any specific scenario you're particularly concerned about?

pkg/cli/install/steps.go
mayankshah1607: Hmm.. I think it could be useful. Updated

pkg/common/constants.go
recharte: I'm not concerned about any particular scenario I just don't like `percona-everest` because it's not descriptive. I'd rather have `everest-operator` and `everest-api-server` or `everest-server` :shrug:

pkg/cli/upgrade/steps.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// cleanupLegacyResources removes resources that were created as a part of the legacy installation method
+// and no longer exist as a part of the Helm chart.
+func (u *Upgrade) cleanupLegacyResources(ctx context.Context) error {
+	// Delete OLM PackageServer CSV.
+	if err := wait.PollUntilContextTimeout(ctx, pollInterval, pollTimeout, true, func(ctx context.Context) (bool, error) {
+		if err := u.kubeClient.DeleteClusterServiceVersion(ctx, types.NamespacedName{
+			Namespace: kubernetes.OLMNamespace,
+			Name:      "packageserver",
```
michal-kralik, line 145: Wouldn't this break some k8s clusters which have OLM preinstalled - i.e. openshift?

pkg/cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,432 @@
...
+		if err != nil {
+			return errors.Join(err, fmt.Errorf("invalid cli version %s", cliVersion.Version))
+		}
+
+		// if !supVer.Cli.Check(cli.Core()) {
+		// 	return fmt.Errorf(
+		// 		"cli version %q does not meet minimum requirements of %q",
+		// 		cli, supVer.Cli.String(),
+		// 	)
+		// }
```
michal-kralik, line 382: Do we need to run Everest 1.3.x before migrating to 1.4.0?

pkg/cli/upgrade/steps.go
mayankshah1607: This deletes the CRV only in `everest-olm` namespace, so I don't think it will break on Openshift since it has OLM in a different namespace

pkg/cli/upgrade/upgrade.go
mayankshah1607: I had this commented out only for some local testing. I forgot to uncomment it. Thanks for noticing :) 

> Do we need to run Everest 1.3.x before migrating to 1.4.0?

but to answer this, no they don't need to.. they can directly install 1.4.0

mayankshah1607: if you're talking about upgrades, I don't think we allow skipping versions. So you can upgrade to 1.4.0 only if you're on 1.3.x (or else you can directly install 1.4.0)

pkg/common/constants.go
mayankshah1607: sorry, I thought we were still talking about the operator name 🤦🏻 

but I don't think my answer changes much.. internally, there shouldn't be breaking anything. During upgrades, we can ensure that we delete the old deployment first, just like we do with the operator. If an existing user depends on the name in someway, it could break things (but we can document it ..)

I'm fine with changing the name. We would have to put in some extra effort while testing though..

michal-kralik:
@mayankshah1607 two questions:

1. Would the install/upgrade testing instructions change with this change?
2. Can you describe what actions we expect from users running e.g. 1.2.0 to upgrade to 1.4.0?

mayankshah1607:
> Would the install/upgrade testing instructions change with this change?

Not by much, I'm expecting it to mostly stay the same. One thing that I can think of is that we will no longer need to change the Everest Operator Subscription channel to `fast-v0` during RC upgrades because we got rid of the OLM way of installing it. I'm planning on reviewing that Notion doc (and also create a new one) once the changes in this PR are approved.

> Can you describe what actions we expect from users running e.g. 1.2.0 to upgrade to 1.4.0?

I don't think that upgrade path is possible, is it? Our CLI allows you to upgrade only 1 minor version at a time. So they need to upgrade to 1.3.x first, and then to 1.4.x. The user does not have to do anything special to upgrade from 1.3 -> 1.4, everything is taken care of automatically. The only new restriction we will be adding is that you may not use CLI 1.4 to perform older upgrades (like 1.1 to 1.2), as we discussed on slack..

.github/workflows/dev-be-ci.yaml
```diff
@@ -330,22 +334,10 @@ jobs:
...
-                "value": {
-                  "name": "PERCONA_VERSION_SERVICE_URL",
-                  "value": "https://check-dev.percona.com/versions/v1"
-                }
-            }]' --type=json
-
+            --namespaces everest \
+            --helm-set server.image=localhost:5000/perconalab/everest \
+            --helm-set server.apiRequestsRateLimit=200 \
+            --helm-set versionMetadataURL=https://check-dev.percona.com
```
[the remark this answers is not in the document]
mayankshah1607: Yes, I think you're right, I missed it..

I've updated the code now so that the value specified in --version-metadata-url flag is used in Helm chart

pkg/cli/uninstall/uninstall.go
```diff
@@ -146,87 +126,19 @@ func (u *Uninstall) Run(ctx context.Context) error { //nolint:funlen,cyclop
...
-		},
-	})
-
-	if u.config.SkipOLM {
-		u.l.Debug("Skipping OLM uninstallation")
-		uninstallSteps = append(uninstallSteps, common.Step{
-			Desc: "Delete Everest Catalog",
-			F: func(ctx context.Context) error {
-				u.l.Info("Deleting Everest catalog")
-				return u.kubeClient.DeleteCatalogSource(ctx, u.config.CatalogNamespace, catalogSource)
```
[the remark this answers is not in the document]
mayankshah1607: Removed

pkg/version/version_test.go
```diff
@@ -1,83 +1,39 @@
...
-			assert.Equal(t, tt.expectDev, dev)
-
-			if tt.version != "" {
-				dev := isDevVersion("v" + tt.version)
-				assert.Equal(t, tt.expectDev, dev)
-			}
-		})
+	for _, tc := range testCases {
+		actual := IsRC(tc.version)
+		assert.Equal(t, tc.expected, actual)
```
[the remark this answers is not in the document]
mayankshah1607: Added

pkg/cli/install/steps.go
mayankshah1607: Updated

cli-tests/tests/flow/all-operators.spec.ts
```diff
@@ -43,39 +43,21 @@ test.describe('Everest CLI install', async () => {
...
-      await out.outErrContainsNormalizedMany([
-        'percona-xtradb-cluster-operator operator has been installed',
-        'percona-server-mongodb-operator operator has been installed',
-        'percona-postgresql-operator operator has been installed',
-        'everest-operator operator has been installed',
-      ]);
-    });
-    await page.waitForTimeout(10_000);
-    await verifyClusterResources();
-
```
mayankshah1607, line 61: It seemed redundant to run the CLI tests twice for verbose logs and pretty logs. I think its just redundant and doesn't provide any real value, so removed them. Let me know if you think otherwise

cli-tests/tests/flow/pg-operator.spec.ts
```diff
@@ -67,45 +51,17 @@ test.describe('Everest CLI install', async () => {
...
-      const operator = await cli.exec(`kubectl -n everest-system get po | grep everest|awk {'print $1'}`);
-      await operator.assertSuccess();
-
-      const out = await cli.everestExecSkipWizard(
-        `install -v --operator.mongodb=false --operator.postgresql=true --operator.xtradb-cluster=true --namespaces=everest-operators`,
-      );
-      const restartedOperator = await cli.exec(`kubectl -n everest-system get po | grep everest|awk {'print $1'}`);
-      await restartedOperator.assertSuccess();
-
-      expect(operator.getStdOutLines()[0]).not.toEqual(restartedOperator.getStdOutLines()[0]);
```
mayankshah1607, line 96: I removed this test step because Everest operator does not need to restart anymore when the install command is re-run.

deploy/quickstart-k8s.yaml
```diff
@@ -1,164 +0,0 @@
-apiVersion: v1
```
oksana-grishchenko, line 1: Will the release pipeline be updated in the scope of this PR or in a separate one? It references the deleted quickstart-k8s

cli-tests/tests/flow/pg-operator.spec.ts
```diff
@@ -44,21 +43,6 @@ test.describe('Everest CLI install', async () => {
         ]);
       });
     };
-    const clusterName = `test-${faker.number.int()}`;
-
-    await test.step('run everest install command', async () => {
-      const out = await cli.everestExecSkipWizard(
-        `install -v --operator.mongodb=false --operator.postgresql=true --operator.xtradb-cluster=false --namespaces=everest-operators`,
-      );
```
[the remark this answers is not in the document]
mayankshah1607: No, the `-v` option is still there. I removed the test because I didn't think it was useful to test the same thing twice - once with pretty output and once with debug logs.. Especially now since Helm takes care of the main heavy lifting, there isn't much output in the verbose logs

deploy/quickstart-k8s.yaml
mayankshah1607: Yes, I have it in mind. I planned to do it in a separate PR, so its easy to review. This one is already pretty big 🙈

mayankshah1607: Okay nevermind, the change was small, so I included it in this PR.. Can you please have another look and let me know if that is all that's needed for the release?

recharte (approved):
I reached the point when looking closer at this PR's code won't help anymore, instead I'll dedicate some time to testing this manually

deploy/quickstart-k8s.yaml
oksana-grishchenko: Great, thanks Mayank, looks good!

oksana-grishchenko (approved):

PR #826, opened 2024-11-11
Everest-1645 | tech preview badge

fabio-silva:
https://perconadev.atlassian.net/browse/EVEREST-1645

Refactored LabeledContent and created a new "ActionableLabeledContent" that allows actions & tech preview chip.
Composition FTW.

Screenshot 2024-11-11 at 17 01 28


ui/packages/ui-lib/src/actionable-labeled-content/ActionableLabeledContent.tsx
```diff
@@ -0,0 +1,62 @@
+import { Button, ButtonProps, Chip } from '@mui/material';
+import AddIcon from '@mui/icons-material/Add';
+import LabeledContent, { LabeledContentProps } from '../labeled-content';
+
+type Props = LabeledContentProps & {
```
solovevayaroslavna, line 5: Are we going to have types separately in all files? Or there are some rules where we can keep it or not?

ui/packages/ui-lib/src/labeled-content/labeled-content.tsx
```diff
@@ -13,64 +13,60 @@
...
+          mt: 1.5,
+          ...textFieldRootSx,
         },
         '.MuiAutocomplete-root': {
           mt: 1.5,
+          ...autocompleteRootSx,
         },
         mt: 2,
-        ...sx,
+        ...verticalStackSx,
```
solovevayaroslavna, line 51: Hey! Can you help with understanding why we need verticalStack/horizontalStack spreading instead of common sx? Because we need both in one time and it's just a naming?

ui/packages/ui-lib/src/actionable-labeled-content/ActionableLabeledContent.tsx
fabio-silva: My feeling is that our file number is growing a lot nowadays, a lot because we always have a .types and .messages files. This is something we can discuss, but for simple use cases (e.g. one typing) I think not having a separate file is OK.

ui/packages/ui-lib/src/labeled-content/labeled-content.tsx
fabio-silva: Yes. At this point we have two stacks, so just one "sx" wouldn't be clarifying regarding which one we're targetting

PR #830, opened 2024-11-13
[UI] [Chore] Add error for unwanted console.logs in the UI

dianabirs:
- added rule to mark console logs as errors by linter
- added exceptions in the files that have necessary log statements
- left a `console.log` in `ui/apps/everest/src/components/nav-item/NavItem.tsx` for visibility (will remove before merging)


.github/workflows/dev-fe-gatekeeper.yaml
```diff
@@ -2,18 +2,8 @@
...
-      - ".github/workflows/dev-fe-e2e.yaml"
     branches:
-      - main
-  pull_request:
-    paths:
-      - "ui/**"
-      - ".github/workflows/dev-fe-gatekeeper.yaml"
-      - ".github/workflows/dev-fe-e2e.yaml"
-    types: [opened, synchronize]
+      - chore/add-eslint-warning-for-logs
```
oksana-grishchenko, line 6: looks like these changes should be reverted, we probably still want the workflow to be triggered by old triggers, not only by pushing to the current branch.

oksana-grishchenko (approved):

ui/apps/everest/src/pages/settings/default-configurations/default-configurations.tsx
```diff
@@ -56,6 +56,7 @@ export const DefaultConfigurations = () => {
   });
 
   const onSubmit: SubmitHandler<DefaultConfigurationsType> = (data) => {
+    // eslint-disable-next-line no-console
     console.log(data);
```
fabio-silva, line 60: do we need this log? I think we can just remove it

ui/apps/everest/src/components/nav-item/NavItem.tsx
```diff
@@ -19,6 +19,7 @@ export const NavItem = ({
   onClick,
   ...listItemProps
 }: NavItemProps) => {
+  console.log('🚀 ~ open:', open);
```
fabio-silva, line 22: should we then add the ignore comment here?

fabio-silva (changes requested):

ui/apps/everest/src/components/nav-item/NavItem.tsx
fabio-silva: never mind, it's gonna be removed, gotcha

dianabirs: I left this log intentionally so that the pipeline failure would be visible

fabio-silva (approved):

PR #860, opened 2024-11-25
[UI] EVEREST-1684 Remove blue background from resources step accordion

dianabirs:
[EVEREST-1684](https://perconadev.atlassian.net/browse/EVEREST-1684)

image

[EVEREST-1684]: https://perconadev.atlassian.net/browse/EVEREST-1684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/apps/everest/src/components/cluster-form/resources/resources.tsx
+++ b/ui/apps/everest/src/components/cluster-form/resources/resources.tsx
@@ -572,7 +572,6 @@ const ResourcesForm = ({
         onChange={handleAccordionChange('nodes')}
         sx={{
           px: 2,
-          backgroundColor: 'transparent',
         }}
       >
         <CustomAccordionSummary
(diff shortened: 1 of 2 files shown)


solovevayaroslavna:
Shouldn't we remove it globally in ui-library? It seems that we don't have any blue accordions in the application at all  🤔 What do you think?

fabio-silva:
> Shouldn't we remove it globally in ui-library? It seems that we don't have any blue accordions in the application at all 🤔 What do you think?

I think we should remove the background at app level to all accordions, not just this one.
But I wouldn't do it at the ui-library level, as this is an Everest-only change.
@dianabirs

fabio-silva:
> @fabio-silva we have a separate theme for everest, can we use it?

@solovevayaroslavna I think that's a good approach, yes

fabio-silva (approved):

PR #862, opened 2024-11-25
[UI] EVEREST-1688 Blur text input on scroll

dianabirs:
[EVEREST-1688](https://perconadev.atlassian.net/browse/EVEREST-1688)

[EVEREST-1688]: https://perconadev.atlassian.net/browse/EVEREST-1688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/packages/ui-lib/src/form/inputs/text/text.tsx
+++ b/ui/packages/ui-lib/src/form/inputs/text/text.tsx
@@ -31,6 +31,9 @@ const TextInput = ({
           }}
           inputProps={{
             'data-testid': `text-input-${kebabize(name)}`,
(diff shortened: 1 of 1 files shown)


dianabirs:
@fabio-silva @solovevayaroslavna I decided to blur all inputs for the `onWheel` event, since I see no benefit of editing the input while scrolling. If you think this is only necessary for number inputs, I will modify this, so lmk

fabio-silva (approved):

PR #867, opened 2024-11-26
EVEREST-1698 filter monitoring instances list based on RBAC permissions

recharte:
EVEREST-1698

The monitoring instance listing operation must be allowed if a given user has permissions to read at least one monitoring instance but the list should be filtered to include only the allowed monitoring instances.


mayankshah1607 (approved):
LGTM, but I think the base of this branch was `EVEREST-1699` and not `main`, as a result there are some conflicts now since `EVEREST-1699` was merged ...

recharte:
> LGTM, but I think the base of this branch was `EVEREST-1699` and not `main`, as a result there are some conflicts now since `EVEREST-1699` was merged ...

That was actually intentional to aid in the review and it actually avoids real conflicts. If you see, the last force push diff is empty, it just complained because the commit hash was different. All good now.

PR #869, opened 2024-11-27
EVEREST-1673 | New CLI command for managing DB namespaces

mayankshah1607:
This PR introduces the following changes:

#### 1. New command for adding namespaces
```sh
./bin/everestctl namespaces add [NAMESPACE] [FLAGS]

./bin/everestctl namespaces add everest-dev

# allows using a Kubernetes namespace that already exists
./bin/everestctl namespaces add everest-existing --take-ownership
```

#### 2. New command for removing namespaces
```sh
./bin/everestctl namespaces remove [NAMESPACE] [FLAGS]

./bin/everestctl namespaces remove everest

# does not delete the k8s namespace, but all the DB resources only
./bin/everestctl namespaces remove everest-dev --keep-namespace
```

#### 3. New command for adding operators to existing namespaces
```sh
./bin/everestctl namespaces update everest
```

#### Notes:
* The existing `everestctl install ...` command can still be used to create DB namespaces like before. However, this command may be run only once, i.e, when installing Everest for the first time. Running it a second time will throw an error. You are expected to use `everestctl namespaces add ...` command to add DB namespaces
* It is also possible to NOT specify a DB namespace when running `everestctl install ...`, unlike before, when it would default to `everest`
* The following commands will work only if installed Everest version is `>= 1.4.0`:
  * `everestctl uninstall ...`
  * `everestctl upgrade ...`
  * `everestctl namespaces [add|remove|update]`


pkg/cli/uninstall/steps.go
```diff
@@ -118,29 +72,6 @@ func (u *Uninstall) deleteEverestCRDs(ctx context.Context) error {
 	return nil
 }
 
-// Older versions of Everest (> 1.3.0) were not installed using Helm.
-// So we need to render the manifests using the Helm chart and delete them using the kube client.
-// Note that the uninstallation process takes care of most resources, but cluster-scoped resources like CRDs and ClusterRoles are
-// not deleted without this step.
-func (u *Uninstall) cleanupLeftovers(ctx context.Context) error {
```
oksana-grishchenko, line 125: VS [allows](https://github.com/Percona-Lab/percona-version-service/blob/main/sources/metadata/everest/1.3.0.yaml) the previous version (1.3.0) to be installed using cli 1.2.0, so without helm. Would it mean that the cleanup is still relevant?

mayankshah1607: Not sure what you mean.. I added this clean-up thing earlier to make it possible to uninstall Everest < 1.4.0 (i.e, without helm) with CLI 1.4.0.

But now I added a [check](https://github.com/percona/everest/pull/869/files#diff-a8915bc94071568fb5c4f87eba9318927746fe4316029eb2caa21aad7b824e1dR100) such that when using CLI 1.4.0, you cannot uninstall if Everest < 1.4.0 (i.e, without helm, so that we don't have to maintain this code..)

> VS [allows](https://github.com/Percona-Lab/percona-version-service/blob/main/sources/metadata/everest/1.3.0.yaml) the previous version (1.3.0) to be installed using cli 1.2.0,

I think this is fine, as long as you uninstall previous version (1.3.0) with CLI 1.3.0, which is what the above check ensures. Am I missing something?

pkg/cli/install/steps.go
```diff
@@ -127,40 +123,3 @@ func (o *Install) installEverestHelmChart(ctx context.Context) error {
...
-			if err := installer.Init(o.config.KubeconfigPath, helm.ChartOptions{
-				Directory: chartDir,
-				URL:       o.config.RepoURL,
-				Name:      helm.EverestDBNamespaceChartName,
-				Version:   ver,
-			}); err != nil {
-				return fmt.Errorf("could not initialize Helm installer: %w", err)
-			}
-
-			o.l.Infof("Installing DB namespace Helm chart in namespace ", namespace)
```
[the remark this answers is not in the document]
mayankshah1607: Thanks, will keep this in mind next time. This code is already removed so nothing to change here..

PR #877, opened 2024-11-28
EVEREST-1707: RBAC on DB restore

michal-kralik:
EVEREST-1707

Access control for `/namespaces/{namespace}/database-clusters/{cluster-name}/restores` and `/namespaces/{namespace}/database-cluster-restores` based on db cluster name


api/database_cluster_restore.go
```diff
@@ -107,6 +138,10 @@ func (e *EverestServer) enforceDBRestoreRBAC(user, namespace, srcBackupName, dbC
 	if err := e.enforce(user, rbac.ResourceDatabaseClusterBackups, rbac.ActionRead, rbac.ObjectName(namespace, srcBackupName)); err != nil {
 		return err
 	}
```
mayankshah1607, line 140: I think we also need to update this to use the DB cluster name

api/database_cluster_restore.go
```diff
@@ -34,6 +37,8 @@ const (
 )
 
 // ListDatabaseClusterRestores List of the created database cluster restores on the specified kubernetes cluster.
+//
+//nolint:dupl
```
mayankshah1607, line 41: Is it possible to tackle this linter?

api/database_cluster_restore.go
```diff
@@ -49,7 +54,33 @@ func (e *EverestServer) ListDatabaseClusterRestores(ctx echo.Context, namespace,
...
+	}
+	rbacFilter := transformK8sList(func(l *unstructured.UnstructuredList) error {
+		allowed := []unstructured.Unstructured{}
+		for _, obj := range l.Items {
+			restore := &everestv1alpha1.DatabaseClusterRestore{}
+			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, restore); err != nil {
+				e.l.Error(errors.Join(err, errors.New("failed to convert unstructured to DatabaseClusterRestore")))
+				return err
+			}
+			if err := e.enforceDBClusterListRestoreRBAC(user, restore); errors.Is(err, errInsufficientPermissions) {
```
mayankshah1607, line 72: just a nitpick, but is a separate method actually needed? I see it calls just one `e.enforce()` internally, so I'm not sure its actually useful

pkg/rbac/rbac.go
```diff
@@ -270,13 +270,19 @@ func NewEnforceHandler(l *zap.SugaredLogger, basePath string, enforcer *casbin.E
 		if resource == ResourceNamespaces {
 			return true, nil
 		}
+		// Allow creating a restore without a name.
+		// RBAC is enforced in the individual methods.
+		if resource == ResourceDatabaseClusterRestores && name == "" && action == ActionCreate {
+			return true, nil
+		}
```
mayankshah1607, line 277: Since we'll be doing the same for `ResourceDatabaseClusterBackups` as well, WDYT about moving the `create-able` objects to a list and checking that? Like we do below for `allowedObjectsForListing`?

mayankshah1607 (changes requested):
Few comments, also, shouldn't we handle `DeleteDatabaseRestore` too?

api/database_cluster_restore.go
michal-kralik: It's not necessary per se, I just wanted to be consistent with the rest of the codebase.
I think it hides unnecessary logic well in this case - it's easy to understand what the line does without extra error handling, logging, etc.

api/database_cluster_restore.go
michal-kralik: We could abstract the functionality, although I think it will be harder to read. The code was duplicated before this change as well.
I understand, since we use `nolint:dupl` in other places, this is a good tradeoff between abstraction and readability.

If you look at the two functions, there would be 4-5 parameters passed in, most probably we would pass around interfaces, would need type assertion, etc. Not sure if it's adding any value tbh.

api/database_cluster_restore.go
mayankshah1607: Because with [EVEREST-1706](https://perconadev.atlassian.net/browse/EVEREST-1706), we will be using DB cluster name for Backups as well, just like we do with Restores in this PR.

So this check will basically never pass, since this permissions will not be associated to a specific backup, but its DB.

mayankshah1607:
@michal-kralik don't we need to handle the permission for `DeleteDatabaseClusterRestore` API as well?

michal-kralik:
> don't we need to handle the permission for `DeleteDatabaseClusterRestore` API as well?

From the ticket it does not seem we do.
It is only about `GET` and `POST`. Maybe there's some other plan with `DELETE`. @recharte?

api/database_cluster_restore.go
michal-kralik: Right, but EVEREST-1706 is not part of this PR.
I planned to address backups in a follow up PR so it's not mixed together.

recharte:
> > > don't we need to handle the permission for `DeleteDatabaseClusterRestore` API as well?
> > 
> > 
> > From the ticket it does not seem we do. It is only about `GET` and `POST`. Maybe there's some other plan with `DELETE`. @recharte?
> 
> Yes, we do need to handle the DELETE as well.

Updated the ticket to reflect the GET and DELETE on `/namespaces/{namespace}/database-cluster-restores/{name}` as well

mayankshah1607 (approved):

PR #888, opened 2024-12-02
EVEREST-1180 | TLS support

mayankshah1607:


oksana-grishchenko (approved):

cmd/config/config.go
```diff
@@ -39,9 +39,9 @@ var (
 
 // EverestConfig stores the configuration for the application.
 type EverestConfig struct {
-	DSN      string `default:"postgres://admin:pwd@127.0.0.1:5432/postgres?sslmode=disable" envconfig:"DSN"`
-	HTTPPort int    `default:"8080" envconfig:"HTTP_PORT"`
-	Verbose  bool   `default:"false" envconfig:"VERBOSE"`
+	DSN        string `default:"postgres://admin:pwd@127.0.0.1:5432/postgres?sslmode=disable" envconfig:"DSN"`
+	ListenPort int    `default:"8080" envconfig:"PORT"`
```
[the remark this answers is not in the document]
mayankshah1607: good idea, updated..

pkg/certwatcher/watcher.go
maxkondr, line 1: Licensing and package description is missed

pkg/certwatcher/watcher.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return err
+	}
+	w.cert = &cert
+	return nil
+}
+
+// Start the certificate files watcher until the context is closed.
+func (w *certWatcher) Start(ctx context.Context) error {
+	watcher, err := fsnotify.NewWatcher()
```
maxkondr, line 54: are we really sure that you will get there filesystem notify signal in case certs mounted to Everest pod are changed?

We have re-check this with K8S because I remember earlier not everywhere it worked.

I remember NGINX recommended to implement it in the following way:
- use sidecar container
- subscribe in K8S API to event "secret changed"
- re-read and write new secret data to 
- send signal to the main container to re-read certs.

internal/server/everest.go
```diff
@@ -294,8 +302,34 @@ func newSkipperFunc() (echomiddleware.Skipper, error) {
...
+	if e.config.TLSCertsPath != "" {
+		return e.startHTTPS(ctx, addr)
+	}
+	return e.echo.Start(addr)
+}
+
+func (e *EverestServer) startHTTPS(ctx context.Context, addr string) error {
+	tlsKeyPath := path.Join(e.config.TLSCertsPath, "tls.key")
+	tlsCertPath := path.Join(e.config.TLSCertsPath, "tls.crt")
+
```
maxkondr, line 316:
```suggestion
	tlsKeyPath := path.Join(filepath.Clean(e.config.TLSCertsPath), "tls.key")
	tlsCertPath := path.Join(filepath.Clean(e.config.TLSCertsPath), "tls.crt")

```

maxkondr: or even better - apply cleaning during reading the value from config. In such a case you always refer to already cleaned path each time you need to get this value from config obj

pkg/certwatcher/watcher.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (w *certWatcher) loadCertificate() error {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	cert, err := tls.LoadX509KeyPair(w.certFile, w.keyFile)
+	if err != nil {
+		return err
+	}
```
maxkondr, line 47: it seems `w.mutex.Lock()` should be only there, no need to lock it before loading certs

pkg/certwatcher/watcher.go
```diff
@@ -0,0 +1,79 @@
...
+type certWatcher struct {
+	cert              *tls.Certificate
+	certFile, keyFile string
+	log               *zap.SugaredLogger
+
+	mutex sync.RWMutex
+}
+
+// GetCertificate returns the certificate.
+func (w *certWatcher) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
maxkondr, line 21: shouldn't it return a copy of certs? otherwise there is no much sense to protect a pointer by mutex.

pkg/certwatcher/watcher.go
mayankshah1607: > should be only there

should be where?

> no need to lock it before loading certs

Why not? The point is to ensure safety when the secret is changed, and someone tries to read at the same time. By removing `Lock()` from here, you're removing this safety and now readers can potentially read an outdated cert. IMO we need it here.

pkg/certwatcher/watcher.go
mayankshah1607: > otherwise there is no much sense to protect a pointer by mutex

Why not? You're just synchronising the timing here, which guarantees that even the pointer read will give you a consistent result. As long as the reader reads it at the "right" time, even the pointer gives a correct result. And the "right" timing is ensured by the lock, so I don't see a problem here

pkg/certwatcher/watcher.go
mayankshah1607: > are we really sure that you will get there filesystem notify signal in case certs mounted to Everest pod are changed?

Yes, please see: are we really sure that you will get there filesystem notify signal in case certs mounted to Everest pod are changed?

```
When a volume contains data from a Secret, and that Secret is updated, Kubernetes tracks this and updates the data in the volume, using an eventually-consistent approach.
```

> earlier not everywhere it worked.

Yes, earlier, I don't think it worked very well. I have also had to use some sidecar watcher. But Kubernetes has fixed it to make this work more reliably, I can't remember from which version, but its probably 1.18.

Now this fsnotify mechanism is also used almost everywhere in most popular Kubernetes projects, like [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/certwatcher/certwatcher.go), so I wouldn't worry

pkg/certwatcher/watcher.go
mayankshah1607: Hmm, sure, yeah.. the caller can change the pointer which bypasses the lock protection.

But I am returning a pointer here mainly to comply with the signature needed in the `TLSConfig`, which is `GetCertificate func(*ClientHelloInfo) (*Certificate, error)`.

Moreover, [this](// Once a Certificate is returned it should not be modified.) comment on `GetCertificate` says:
```
// Once a Certificate is returned it should not be modified.
```

and we are complying with it, so I don't think its a big deal in this case. But for the sake of complete correctness, I'm happy to change it

mayankshah1607: Updated to return a pointer to a copy of the certs

cmd/config/config.go
```diff
@@ -39,9 +40,11 @@ var (
 
 // EverestConfig stores the configuration for the application.
 type EverestConfig struct {
-	DSN      string `default:"postgres://admin:pwd@127.0.0.1:5432/postgres?sslmode=disable" envconfig:"DSN"`
-	HTTPPort int    `default:"8080" envconfig:"HTTP_PORT"`
-	Verbose  bool   `default:"false" envconfig:"VERBOSE"`
+	DSN string `default:"postgres://admin:pwd@127.0.0.1:5432/postgres?sslmode=disable" envconfig:"DSN"`
```
maxkondr, line 43: btw, do we still need DSN? it seems to be an artefact

maxkondr (approved):

recharte:
We also need to address the TODOs left in https://github.com/percona/everest/pull/1195.

mayankshah1607:
@recharte done, PTAL. But I don't think the SSL redirect actually works

internal/server/middlewares.go
```diff
@@ -129,23 +131,25 @@ func securityHeaders(oidcProvider *oidc.ProviderConfig) echo.MiddlewareFunc {
...
+		PermissionsPolicy:             PermissionsPolicy,
+		ReferrerPolicy:                "no-referrer",
 		XPermittedCrossDomainPolicies: "none",
-	})
+	}
+
+	if useTLS {
+		opts.SSLRedirect = true
+		opts.STSIncludeSubdomains = true
+		opts.STSSeconds = 31536000
```
recharte, line 149: You missed the TODO in the CSP headers above.

PR #894, opened 2024-12-04
EVEREST-1706: db backup RBAC based on db cluster name

michal-kralik:
EVEREST-1706

### DIFF
--- a/pkg/rbac/rbac.go
+++ b/pkg/rbac/rbac.go
@@ -270,11 +270,17 @@ func NewEnforceHandler(l *zap.SugaredLogger, basePath string, enforcer *casbin.E
 		if resource == ResourceNamespaces {
 			return true, nil
 		}
+
 		// Allow creating a restore without a name.
 		// RBAC is enforced in the individual methods.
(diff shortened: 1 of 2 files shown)


pkg/rbac/rbac.go
```diff
@@ -297,6 +285,34 @@ func NewEnforceHandler(l *zap.SugaredLogger, basePath string, enforcer *casbin.E
 	}
 }
 
+func isAllowedRBAC(name, action, resource string) bool {
+	// Allow creating a restore without a name.
```
mayankshah1607, line 289:
```suggestion
	// Allow creating a the following resources without a name.
```

mayankshah1607 (commented):
Overall LGTM, but shouldn't [this](https://github.com/percona/everest/pull/877/files#r1863145258) be addressed in this PR?

pkg/rbac/rbac.go
michal-kralik: updated

michal-kralik:
> Overall LGTM, but shouldn't [this](https://github.com/percona/everest/pull/877/files#r1863145258) be addressed in this PR?

Updated

mayankshah1607 (approved):

PR #899, opened 2024-12-05
EVEREST-1537: increase db removal timeout

michal-kralik:
EVEREST-1537

Increase timeout for DB removal in case there's too many DB clusters running.


pkg/kubernetes/database_cluster.go
```diff
@@ -39,7 +42,19 @@ func (k *Kubernetes) GetDatabaseCluster(ctx context.Context, namespace, name str
...
-	return wait.PollUntilContextTimeout(ctx, pollInterval, pollTimeout, true, func(ctx context.Context) (bool, error) {
+	timeout := pollTimeout
+	list, err := k.ListDatabaseClusters(ctx, namespace)
+	if err != nil {
+		k.l.Debugf("Could not list db clusters: %s", err)
+	}
+	if list != nil && len(list.Items) > 0 {
+		// We increase the timeout if there's too many DB clusters.
+		timeout = pollTimeout * time.Duration(len(list.Items)/dbsCountTimeoutMultiply)
+	}
```
oksana-grishchenko, line 53: The formula doesn't work as expected because of integer division (for example, the timeout is 0 if len = 1 or 2)
It will work correctly however if the `time.Duration()` is removed 

I would suggest to extract the multiplier logic to a separate function and cover it with unit tests to be sure it works as expected.

michal-kralik: nice catch, updated

oksana-grishchenko: Ok, good, it fixes the critical issue about multiplying by 0. 
But why don't you like the idea of moving the logic to a separate function and cover it with tests? It would be prominent then that the function still works weird bc of the integer division.

given the default pollTimeout 15 min, with the current implementation we'll have:

	1 dbs - 15 min
	2 dbs - 15 min
	3 dbs - 15 min
	4 dbs - 15 min
	5 dbs - 15 min
	6 dbs - 30 min
	7 dbs - 30 min
	8 dbs - 30 min
	9 dbs - 45 min

However it looks like the intention was to have something like this:

	1 dbs - 15 min
	2 dbs - 15 min
	3 dbs - 15 min
	4 dbs - 20 min
	5 dbs - 25 min
	6 dbs - 30 min
	7 dbs - 35 min
	8 dbs - 40 min
	9 dbs - 45 min


Also probably we could move this `dbsCountTimeoutMultiply` constant inside the new function because it's not used anywhere else, so wdyt about
```
func flexibleTimeout(basicTimeout time.Duration, dbsLen int) time.Duration {
	const dbsCountTimeoutMultiply = 3
	timeout := basicTimeout * time.Duration(dbsLen) / dbsCountTimeoutMultiply
	return time.Duration(math.Max(float64(basicTimeout), float64(timeout)))
}
``` 
?

michal-kralik: The first example was my intention. I know it looks odd if written one after each other but the logic is - if there's 5 or less DBs, let's wait 15 minutes (i.e. 1 timeout) and then for each 3, add 15 minutes extra (i.e. +1 timeout).

The reason I opted for this approach is that it's not clear if for deleting a single DB we need to add 5 more minutes. So instead of having 5 minutes hardcoded, I opted for doing it in batches of +1 timeout per 3 dbs which shall yield lower probability of hitting the timeout due to the number of DBs.

Re unit tests - the logic is 1-2 lines of math operations. It's not so complex that it would require unit tests, imo.
If you feel this would benefit from unit tests though, I can always add them, of course.

I moved the constant to a new location

PR #904, opened 2024-12-06
EVEREST-1709: backup rbac tests

ghost:
EVEREST-1709


api/database_cluster_backup_test.go
```diff
@@ -0,0 +1,226 @@
...
+						},
+						Spec: everestv1alpha1.DatabaseClusterBackupSpec{
+							DBClusterName: "db-cluster-name",
+						},
+					},
+				},
+			},
+
+			expectedEnforce: []enforceStrings{
+				{"alice", rbac.ResourceBackupStorages, rbac.ActionRead, "ns/name1"},
```
recharte, line 86: I think these tests just helped us find a bug :sweat_smile: 
The enforce for ResourceBackupStorages not be `ns/name1`. Instead it should be `ns/backup-storage-name` that is part of the Spec like so:

```suggestion
						Spec: everestv1alpha1.DatabaseClusterBackupSpec{
							DBClusterName: "db-cluster-name",
							BackupStorageName: "backup-storage-name",
						},
					},
				},
			},

			expectedEnforce: []enforceStrings{
				{"alice", rbac.ResourceBackupStorages, rbac.ActionRead, "ns/backup-storage-name"},
```

The bug is [here](https://github.com/percona/everest/blob/316f08d9c843d2b63f455a524238b6097e12e62f/api/database_cluster_backup.go#L100). It should be `dbbackup.Spec.BackupStorageName` instead of `dbbackup.GetName()`

recharte: The other functions have the same bug :see_no_evil:

recharte (commented):
Overall I like the unit test based approach, it was a good suggestion :clap:  However, I have 2 concerns with the tight coupling with the actual implementation.
1. We are checking for enforce calls but we are not asserting what was done with those enforcements. For example, if we had a bug [here](https://github.com/percona/everest/blob/316f08d9c843d2b63f455a524238b6097e12e62f/api/database_cluster_backup.go#L101) and called something other than `continue` the unit tests would pass because the enforce call was done but the end result would be wrong, the list wouldn't have been properly filtered.
2. The tight coupling with the implementation can make these tests brittle with refactors. I know that @mayankshah1607 had some plans to refactor the way we are handling the filtering functions and I'm concern that this testing approach will need many adjustments as part of that refactoring. Let's hear from @mayankshah1607

Issue #903, opened 2024-12-05
Kubernetes Cluster is Unavailable


kjordan2001:
I keep running into this message on my Talos cluster, but everything is up.  How can I debug what it thinks is unavailable since everything is up and operational?  It can create new database clusters, although Mongo and MySQL are always stuck in the Initializing state on version 1.3.0 even after everything is running.

recharte:
Hi @kjordan2001, the `Kubernetes cluster is unavailable` message is potentially a kubernetes RBAC issue. However, I find this strange because you were able to create and list the database clusters.
We need to figure out which requests are returning this error. Can you please share the Everest API server logs? You can get them by running the following command:
```
kubectl logs -n everest-system -l app.kubernetes.io/component=everest
```

kjordan2001:
[everest-log.txt](https://github.com/user-attachments/files/18027961/everest-log.txt)
Here is the log for after I refresh.  The message seems to come up when this is logged:
`{"level":"error","T":"2024-12-05T18:28:00Z","caller":"api/proxy.go:156","msg":"invalid character '\\x1f' looking for beginning of value\nfailed to transform response body"}`

recharte:
That's very strange :sweat_smile: 
Can you show me the output of:
```
kubectl get db/mongodb-li8 -n rocketchatdb -o yaml
```

kjordan2001:
[rocketchatdb-yaml.txt](https://github.com/user-attachments/files/18028261/rocketchatdb-yaml.txt)

recharte:
The error message came when trying to list the backups for that DB so it seems that there might be some backups done even though the DB was stuck on initializing. They most likely failed but that still doesn't justify the strange character.
Let's have a look at the backup objects themselves to see if we spot something
```
kubectl get dbb -n rocketchatdb -o yaml
```

recharte:
In order to understand why the DB is stuck initializing we'll also need to look into the percona-server-mongodb-operator logs.
```
kubectl logs -n rocketchatdb -l name=percona-server-mongodb-operator
```

kjordan2001:
Output for both.
[rocketchatdb-yaml.txt](https://github.com/user-attachments/files/18028736/rocketchatdb-yaml.txt)
[mongodb-operator.txt](https://github.com/user-attachments/files/18028737/mongodb-operator.txt)

kjordan2001:
Also for a mysql xtradb cluster that's stuck at initializing too.
[...]

PR #908, opened 2024-12-07
EVEREST-1733 gzip support in proxy response modifier

efimkin:
Hi! After creating the 25th cluster, I get an error message in browser Kubernetes Cluster is Unavailable and in operator log:
`{"level":"error","T":"2024-12-02T09:05:44Z","caller":"api/proxy.go:156","msg":"invalid character '\\x1f' looking for beginning of value\nfailed to transform response body"}`
Looks like same problem in https://github.com/percona/everest/issues/903

### DIFF
--- a/api/proxy.go
+++ b/api/proxy.go
@@ -18,6 +18,7 @@ package api
 
 import (
 	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


recharte (approved):
Hi @efimkin thank you for your contribution.

PR #916, opened 2024-12-10
EVEREST-1724-sharding-is-reset-to-default

solovevayaroslavna:
https://perconadev.atlassian.net/browse/EVEREST-1724 Sharding was reset to default when returning to the previous step of the form


fabio-silva (approved):

fabio-silva:
@solovevayaroslavna I think all of these "resetField" can be safely removed, as we only had that when db type was changed in the wizard. Since we now can't change it, we no longer need those

solovevayaroslavna:
@fabio-silva i think it's better to do this as part of the of https://perconadev.atlassian.net/browse/EVEREST-1234, because it would be good to double-check that this will not bring any errors +it will be better to track this as part of the same ticket 🤔

solovevayaroslavna:
@fabio-silva i will mark this idea under the ticket to not forget

PR #929, opened 2024-12-12
EVEREST-1617: api for affinity settings

michal-kralik:
EVEREST-1617

### DIFF
--- a/docs/spec/openapi.yml
+++ b/docs/spec/openapi.yml
@@ -2426,404 +2426,3049 @@ components:
     # please do not edit the definition below manually
     # -------------------------
     DatabaseCluster:
-      description: DatabaseCluster is the Schema for the databaseclusters API.
-      properties:
-        apiVersion:
-          description: |-
-            APIVersion defines the versioned schema of this representation of an object.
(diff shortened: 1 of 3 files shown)


docs/spec/openapi.yml
```diff
@@ -2426,404 +2426,3049 @@ components:
...
-                      type: integer
-                  required:
-                  - enabled
-                  type: object
-                schedules:
-                  description: Schedules is a list of backup schedules
-                  items:
-                    description: BackupSchedule is the backup schedule configuration.
+        description: DatabaseCluster is the Schema for the databaseclusters API.
+        properties:
```
oksana-grishchenko, line 2430: The schema is still valid, however the indent is bigger than it was before and as it is for other schemas, that's why the diff shows more changes than was actually done. Please update it

oksana-grishchenko (approved):

docs/spec/openapi.yml
michal-kralik: updated

recharte:
Waiting on https://github.com/percona/percona-helm-charts/pull/449 before merging

PR #946, opened 2024-12-18
EVEREST-1563 Add FB bash scripts

oksana-grishchenko:
EVEREST-1563

**Problem:**
Feature build installation requires to many manual steps, our QAs won't be happy about it. 

**Reason:**
For the FB to work, the VS needs to be deployed locally. Then we also need to figure out the internal IP of the VS to pass it to helm. In case of CLI we would also need the external access to the VS which is done by port-forwarding. 

**Solution:** 
Add bash scripts that will abstract away the complexity and only require the path to the helm chart (or the helm chart and the cli).
There are three scripts added:
1. `vs.sh` - deploys VS and returns its internal IP. This script is being run by other two scripts. 
2. `helm.sh` - runs `vs.sh` under the hood and then runs Everest helm installation. 
3. `cli.sh` -  runs `vs.sh` under the hood, exposes VS with port-forwarding and then runs Everest cli installation. 

The pre-built examples of how to use the scripts for this particular FB are printed to the FB action run in the build summary section: 
Screenshot 2024-12-18 at 19 07 13

See more details in [How to use FB](https://www.notion.so/percona/Everest-Feature-Builds-FBs-153674d091f38054baf9c4160b8e62dc?pvs=4#153674d091f381b88366e40b39665b73) section.

### DIFF
--- a/.github/workflows/feature-build.yaml
+++ b/.github/workflows/feature-build.yaml
@@ -1,0 +1,423 @@
+---
+name: Custom build
+on:
+  workflow_dispatch:
+    inputs:
+      everest_branch:
+        required: false
+        default: main
+        description: "Everest branch with the feature to checkout"
+      everest_operator_branch:
+        required: false
+        default: main
+        description: "Everest operator branch with the feature to checkout"
+      vs_branch:
+        required: false
+        default: main
+        description: "Version Service branch with the feature to checkout"
+      helm_branch:
+        required: false
+        default: main
+        description: "Helm charts branch with the feature to checkout"
+      upstream_operator:
+        type: choice
+        description: "The upstream operator to upgrade"
+        required: false
+        options:
+          - ""
+          - percona-xtradb-cluster-operator
+          - percona-server-mongodb-operator
+          - percona-postgresql-operator
+      upstream_operator_bundle_image:
+        required: false
+        description: "Full name of the upstream bundle image to test"
+      artifacts_retention_days:
+        required: false
+        description: "How many days the artifacts will be stored in GitHub"
+        default: 30
+
+permissions:
+  contents: read
+  packages: write
+  checks: write
+  pull-requests: write
+
+jobs:
+  build:
+    name: Build
+    timeout-minutes: 30
+
+    strategy:
+      fail-fast: false
+      matrix:
+        go-version: [ 1.23.x ]
+        may-fail: [ false ]
+
+    continue-on-error: ${{ matrix.may-fail }}
+    runs-on: ubuntu-20.04
+
+    env:
+      TOOLS_PATH: "/opt/tools/bin"
+      ARTIFACTS_RETENTION_DAYS: ${{ github.event.inputs.artifacts_retention_days }}
+      VS_BRANCH: ${{ github.event.inputs.vs_branch }}
+      HELM_BRANCH: ${{ github.event.inputs.helm_branch }}
+      EVEREST_BRANCH: ${{ github.event.inputs.everest_branch }}
+      EVEREST_OPERATOR_BRANCH: ${{ github.event.inputs.everest_operator_branch }}
+      UPSTREAM_OPERATOR_IMAGE: ${{ github.event.inputs.upstream_operator_bundle_image }}
+      UPSTREAM_OPERATOR: ${{ github.event.inputs.upstream_operator }}
+
+      VS_PORT: 8081
+      ARCH: ""
+      OS: ""
+      VERSION: ""
+      VERSION_TAG: ""
+      VS_URL: ""
+      VS_TAG: ""
+
+    steps:
+      - name: Input validation
+        run: |
+          if [[ ! "$ARTIFACTS_RETENTION_DAYS" =~ ^[0-9]+$ ]]; then
+            echo "Wrong artifacts_retention_days format. Use an integer number. 0 means default GitHub repo settings."
+            exit 1
+          fi
+
+      - name: Set environment variables
+        run: |
+          echo "ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')" >> $GITHUB_ENV
+          echo "OS=$(uname | awk '{print tolower($0)}')" >> $GITHUB_ENV
+          echo "VS_URL=http://localhost:$VS_PORT" >> $GITHUB_ENV
+          
+          TIMESTAMP=$(date +'%Y%m%d%H%M%S')
+          V="1.10000.0-rc$TIMESTAMP"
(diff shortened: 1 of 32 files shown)


.github/workflows/feature-build.yaml
```diff
@@ -13,7 +13,7 @@ on:
         description: "Everest operator branch with the feature to checkout"
       vs_branch:
         required: false
-        default: main
+        default: production
         description: "Version Service branch with the feature to checkout"
```
oksana-grishchenko, line 17: Using `production` instead of `main`, because `main` can contain RC builds which means we would always need to install the latest RC to be able to upgrade to FB using CLI. 
I think the most frequent use case for FB is to test upgrade from the latest released version to FB. If we need to include RCs we can always change the `vs_branch` to `main` when running the pipeline

dev/fb/vs.sh
```diff
@@ -0,0 +1,20 @@
+curl https://raw.githubusercontent.com/Percona-Lab/percona-version-service/main/deploy.yaml  > vs_deploy.yaml
```
recharte, line 1:
```suggestion
#!/usr/bin/env bash

curl https://raw.githubusercontent.com/Percona-Lab/percona-version-service/main/deploy.yaml  > vs_deploy.yaml
```

dev/fb/helm.sh
```diff
@@ -0,0 +1,24 @@
+if [[ -z "${HELM_PATH+set}" ]]; then
```
recharte, line 1:
```suggestion
#!/usr/bin/env bash

if [[ -z "${HELM_PATH+set}" ]]; then
```

dev/fb/cli.sh
```diff
@@ -0,0 +1,41 @@
+if [[ -z "${VS_IMAGE+set}" ]]; then
```
recharte, line 1:
```suggestion
#!/usr/bin/env bash

if [[ -z "${VS_IMAGE+set}" ]]; then
```

dev/fb/cli.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo "Error: VS_IMAGE environment variable is not set." >&2
+  exit 1
+elif [[ -z "$VS_IMAGE" ]]; then
+    echo "Error: VS_IMAGE environment variable is empty." >&2
+    exit 1
+fi
```
recharte, line 7: Do we really need to distinguish between an unset and an empty env var? We don't treat them differently so I would suggest to just reduce the scripts to half the size with just this:
```suggestion
if [[ -z "$VS_IMAGE" ]]; then
    echo "Error: VS_IMAGE environment variable is empty." >&2
    exit 1
fi
```

dev/fb/cli.sh
```diff
@@ -0,0 +1,41 @@
...
+  exit 1
+elif [[ -z "$EVEREST_CTL_PATH" ]]; then
+    echo "Error: EVEREST_CTL_PATH environment variable is empty. Please add the absolute path to your everestctl binary to the EVEREST_CTL_PATH environment variable." >&2
+    exit 1
+fi
+
+# deploy VS and get it's internal IP
+curl -O https://raw.githubusercontent.com/percona/everest/main/dev/fb/vs.sh vs.sh > /dev/null
+chmod +x vs.sh
+SERVICE_IP=$(bash ./vs.sh)
```
recharte, line 36:
```suggestion
curl -O https://raw.githubusercontent.com/percona/everest/main/dev/fb/vs.sh vs.sh > /dev/null
chmod +x vs.sh
SERVICE_IP=$(curl -sfL https://raw.githubusercontent.com/percona/everest/main/dev/fb/vs.sh | bash -s)
```

dev/fb/cli.sh
```diff
@@ -0,0 +1,41 @@
...
+
+kubectl port-forward svc/percona-version-service 8081:80 &
+
+# run everest installation with everest CLI
+$EVEREST_CTL_PATH install --chart-dir "$HELM_PATH/charts/everest" --version $VERSION --version-metadata-url http://localhost:8081  --operator.xtradb-cluster --operator.mongodb --operator.postgresql --skip-wizard --namespaces everest -v --helm.set "versionMetadataURL=http://$SERVICE_IP"
```
recharte, line 41:
```suggestion
"$EVEREST_CTL_PATH" install --chart-dir "$HELM_PATH/charts/everest" --version "$VERSION" --version-metadata-url http://localhost:8081  --operator.xtradb-cluster --operator.mongodb --operator.postgresql --skip-wizard --namespaces everest -v --helm.set "versionMetadataURL=http://$SERVICE_IP"
```

dev/fb/helm.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+# deploy VS and get it's internal IP
+curl -O https://raw.githubusercontent.com/percona/everest/main/dev/fb/vs.sh vs.sh > /dev/null
+chmod +x vs.sh
+
+SERVICE_IP=$(bash ./vs.sh)
+
+# run everest installation with helm
+helm install everest-core "$HELM_PATH/charts/everest" --namespace=everest-system --create-namespace --set versionMetadataURL=http://$SERVICE_IP --timeout=10m --devel
```
recharte, line 24:
```suggestion
helm install everest-core "$HELM_PATH/charts/everest" --namespace=everest-system --create-namespace --set versionMetadataURL="http://$SERVICE_IP" --timeout=10m --devel
```

dev/fb/helm.sh
```diff
@@ -0,0 +1,24 @@
...
+elif [[ -z "$VS_IMAGE" ]]; then
+    echo "Error: VS_IMAGE environment variable is empty." >&2
+    exit 1
+fi
+
+# deploy VS and get it's internal IP
+curl -O https://raw.githubusercontent.com/percona/everest/main/dev/fb/vs.sh vs.sh > /dev/null
+chmod +x vs.sh
+
+SERVICE_IP=$(bash ./vs.sh)
```
recharte, line 21:
```suggestion
SERVICE_IP=$(curl -sfL https://raw.githubusercontent.com/percona/everest/main/dev/fb/vs.sh | bash -s)
```

dev/fb/cli.sh
oksana-grishchenko: Updated in all files, thanks. 

First I thought to print the pre-built commands to the `build summary` with empty env values. So in case the user just copied the command and didn't provide the values I added this additional check. However, the final version prints the placeholders (`<path_to_everestctl>`, `<path_to_your_helmchart>`) so there is no point in keeping the checks.

recharte (approved):

PR #956, opened 2024-12-20
EVEREST-1542 - Extend upgrade tests

tplavcic:
The upgrade test was not finished so in this PR we added/fixed:
- PXC cluster is included in the upgrade.
- Created shell script (`percona-everest-ui.sh --pre-upgrade` in `everest-qa` repo) which prepares local environment for the upgrade so it is easy to run it locally.
- Added data before upgrade and do a check after upgrade.
- Added upgrade for databases so at the moment we do Everest upgrade, operators upgrade and database upgrade.
- Unified the environment so that it is completely identical to normal UI tests for easier debugging.
- Added database deletion and Everest uninstall check.
- Added util functions `getVersionServiceURL`, `getDbClusterAPI`, `updateDbClusterAPI`, `getDbAvailableUpgradeVersionK8S` for future use.
- Tweaked resources so the whole setup can actually run on the cloud instances which we use.

**NOTE:**
For the util functions the naming rule will be:
- Add suffix `API` if the function is getting/setting something only through api calls.
- Add suffix `K8S` if the function is getting/setting something only in kubernetes.
- If the function doesn't have suffix it means it does some actions through UI or it does something irrelevant like some calculation or data transformation.


fabio-silva:
@tplavcic I think we need an empty commit to trigger the checks, as there was a CI commit that won't trigger those (we'll need a token to fix this issue)

fabio-silva (approved):

PR #988, opened 2025-01-07
Add Helm Charts Guide for our Readme

edithturn:
I added the information for the HELM installation and updated the CLI installation section as well, but I would appreciate a second pair of eyes to review it, please.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,14 +32,14 @@ Helm is the recommended installation method for Percona Everest as it simplifies
 
 ### Steps to Install
 
-1. Add the Percona Helm repository:
+1. **Add the Percona Helm repository:**
 
 ```bash
 helm repo add percona https://percona.github.io/percona-helm-charts/
 helm repo update
 ```
 
-2. Install the Percona Everest Helm Chart:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,117 +14,175 @@ Here’s why you should try Percona Everest:
...
+## Install Percona Everest using CLI
 
-    ```sh
-    everestctl install
-    ```
+> **Note**:
+>
+> Google Container Registry (GCR) is scheduled to be deprecated and will officially shut down on March 18, 2025. All versions of Percona Everest prior to 1.4.0 depend on images hosted on GCR. After the shutdown date, downloading those images will fail.
+>
+> We strongly recommend upgrading to Percona Everest version 1.4.0 as soon as possible. For more details, refer to the [Container Registry Deprecation Documentation](https://cloud.google.com/artifact-registry/docs/transition/prepare-gcr-shutdown)
```
recharte, line 79: Why was this note added in this section? This is not specific to the installation using the CLI we should move this elsewhere.

recharte (approved):

PR #990, opened 2025-01-08
EVEREST-1709 | RBAC unit tests

mayankshah1607:


internal/server/handlers/rbac/database_cluster_restore_test.go
```diff
@@ -0,0 +1,405 @@
...
+				policy: newPolicy(
+					"p, role:test, database-cluster-restores, update, default/cluster1",
+					"g, bob, role:test",
+				),
+			},
+			{
+				desc:    "missing update permission for database-cluster-restores on 'cluster1'",
+				policy:  newPolicy(),
+				wantErr: ErrInsufficientPermissions,
+			},
```
recharte, line 310: Should the Update require similar permissions to Create? I mean these :point_down: 
```
					"p, role:test, database-cluster-credentials, read, default/cluster1",
					"p, role:test, database-cluster-backups, read, default/cluster1",
					"p, role:test, database-cluster-restores, read, default/cluster1",
```

internal/server/handlers/rbac/monitoring_instance_test.go
```diff
@@ -0,0 +1,743 @@
...
+				policy: newPolicy(
+					"p, role:test, monitoring-instances, update, default/*",
+					"g, bob, role:test",
+				),
+				wantErr: ErrInsufficientPermissions,
+			},
+			{
+				desc: "update only action for some monitoring-instances in default namespace",
+				policy: newPolicy(
+					"p, role:test, monitoring-instances, update, default/some",
```
recharte, line 352: Commenting here in the `update` but the same is true in the `read` and `delete`. Shouldn't this match the name of the resource (`monitoring-instance-1`) to make sure that `update` isn't a super-set of `create`? By using `some` we are not validating that `update default/monitoring-instance-1` doesn't lead to `create` as well which I think was the purpose of iterating through `read`, `update`, `delete`.

internal/server/handlers/rbac/monitoring_instance_test.go
```diff
@@ -0,0 +1,743 @@
...
+					"g, bob, role:test",
+				),
+			},
+			{
+				desc: "create only action for 'monitoring-instance-1' in default namespace",
+				policy: newPolicy(
+					"p, role:test, monitoring-instances, create, default/monitoring-instance-1",
+					"g, bob, role:test",
+				),
+			},
```
recharte, line 332: Should we have a negative case for the resource name enforcement?
```suggestion
			},			{
				desc: "create only action for 'some' in default namespace",
				policy: newPolicy(
					"p, role:test, monitoring-instances, create, default/some",
					"g, bob, role:test",
				),
				wantErr: ErrInsufficientPermissions,
			},
```

internal/server/handlers/rbac/database_engine_test.go
```diff
@@ -0,0 +1,581 @@
...
+					"p, role:test, database-engines, read, */percona-server-mongodb-operator",
+					"g, bob, role:test",
+				),
+				assert: func(list *everestv1alpha1.DatabaseEngineList) bool {
+					return len(list.Items) == 1 &&
+						slices.ContainsFunc(list.Items, func(engine everestv1alpha1.DatabaseEngine) bool {
+							return engine.GetName() == common.PSMDBOperatorName
+						})
+				},
+			},
```
recharte, line 180: I think we should add a negative case for a namespace, e.g.:
```suggestion
			},
			{
				desc: "read all in kube-system namespace",
				policy: newPolicy(
					"p, role:test, database-engines, read, kube-system/*",
					"g, bob, role:test",
				),
				assert: func(list *everestv1alpha1.DatabaseEngineList) bool {
					return len(list.Items) == 0
				},
			},
```

internal/server/handlers/rbac/database_cluster_restore_test.go
mayankshah1607: I'm actually not sure, but I did think about it. For now I just added the test cases based on the current RBAC implementation..

But I think it makes sense to have the same permissions and create..

I've noted that there are some minor RBAC permission changes required for DBB and DBR, and planned to do them in a separate PR. Shall I include this in that PR?

recharte: Sure, feel free to handle this in a follow-up PR.

internal/server/handlers/rbac/monitoring_instance_test.go
mayankshah1607: Right, yeah, makes sense.

I think I might have intended to do `create default/some` - to test that having `create` for some, does not lead to giving permissions for another

recharte (approved):

PR #992, opened 2025-01-09
EVEREST-1778 Ask user provide namespace name if it is missed in `everestctl namespaces add` command

maxkondr:


commands/namespaces/add.go
```diff
@@ -39,11 +40,29 @@ func NewAddCommand(l *zap.SugaredLogger) *cobra.Command {
...
+					output.PrintError(err, l, true)
+					os.Exit(1)
+				}
+
+				if resp == "" {
+					// user didn't provide namespace name
+					output.PrintError(namespaces.ErrNamespaceIsMissed, l, true)
+					os.Exit(1)
+				}
+				c.Namespaces = resp
```
mayankshah1607, line 64: If you see below, there is already a function call `c.Populate` that can run this namespace name prompt if the second arg is set to `true` ... I think we should re-use that

maxkondr: we decided to change help message and keep current logic as is.

mayankshah1607 (approved):

commands/namespaces/add.go
```diff
@@ -25,10 +25,11 @@ var (
 // NewAddCommand returns a new command to add a new namespace.
 func NewAddCommand(l *zap.SugaredLogger) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "add",
+		Use:     "add [flags] namespace",
 		Long:    "Add a new namespace",
 		Short:   "Add a new namespace",
-		Example: `everestctl namespaces add [NAMESPACE] [FLAGS]`,
+		Example: `everestctl namespaces add --operator.mongodb=true --skip-wizard ns-1,ns-2`,
```
oksana-grishchenko, line 31: The example might be confusing because it will install all 3 operators still. If the goal is to install only the psmdb then the command would be:
```
everestctl namespaces add --operator.postgresql=false --operator.xtradb-cluster=false --skip-wizard ns-1,ns-2
```

maxkondr: done

oksana-grishchenko (approved):

commands/namespaces/add.go
```diff
@@ -25,10 +25,11 @@ var (
 // NewAddCommand returns a new command to add a new namespace.
 func NewAddCommand(l *zap.SugaredLogger) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "add",
+		Use:     "add [flags] namespace",
```
maxkondr, line 28:
```suggestion
		Use:     "add [flags] NAMESPACES",
```

PR #1015, opened 2025-01-16
EVEREST-1819 [CLI] Get rid of spf13/viper lib for parsing CLI params

maxkondr:
- Use spf13/cobra lib only for parsing CLI params
- Parse global flags (--verbose, --json, --kubeconfig) on root level
- Extend examples
- Use approach for commands initialisation suggested by cobra lib

### DIFF
--- a/commands/accounts.go
+++ b/commands/accounts.go
@@ -18,24 +18,25 @@ package commands
 
 import (
 	"github.com/spf13/cobra"
-	"go.uber.org/zap"
 
 	"github.com/percona/everest/commands/accounts"
 )
 
-func newAccountsCmd(l *zap.SugaredLogger) *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "accounts",
-		Long:  "Manage Everest accounts",
-		Short: "Manage Everest accounts",
-	}
+var accountsCmd = &cobra.Command{
+	Use:   "accounts <command> [flags]",
+	Args:  cobra.ExactArgs(1),
+	Long:  "Manage Everest accounts",
+	Short: "Manage Everest accounts",
+	Run:   func(_ *cobra.Command, _ []string) {},
+}
 
-	cmd.AddCommand(accounts.NewCreateCmd(l))
-	cmd.AddCommand(accounts.NewListCmd(l))
-	cmd.AddCommand(accounts.NewDeleteCmd(l))
-	cmd.AddCommand(accounts.NewSetPwCommand(l))
-	cmd.AddCommand(accounts.NewResetJWTKeysCommand(l))
-	cmd.AddCommand(accounts.NewInitialAdminPasswdCommand(l))
+func init() {
+	rootCmd.AddCommand(accountsCmd)
 
-	return cmd
+	accountsCmd.AddCommand(accounts.GetCreateCmd())
+	accountsCmd.AddCommand(accounts.GetListCmd())
+	accountsCmd.AddCommand(accounts.GetDeleteCmd())
+	accountsCmd.AddCommand(accounts.GetSetPasswordCmd())
+	accountsCmd.AddCommand(accounts.GetResetJWTKeysCmd())
+	accountsCmd.AddCommand(accounts.GetInitAdminPasswordCmd())
 }
--- a/commands/install.go
+++ b/commands/install.go
@@ -20,106 +20,91 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
-	"go.uber.org/zap"
 
 	"github.com/percona/everest/pkg/cli"
 	"github.com/percona/everest/pkg/cli/helm"
 	"github.com/percona/everest/pkg/cli/install"
+	"github.com/percona/everest/pkg/logger"
 	"github.com/percona/everest/pkg/output"
 )
 
-func newInstallCmd(l *zap.SugaredLogger) *cobra.Command {
-	cmd := &cobra.Command{
-		Use: "install",
+var (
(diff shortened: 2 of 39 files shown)


commands/accounts.go
```diff
@@ -18,24 +18,25 @@ package commands
...
+}
 
-	cmd.AddCommand(accounts.NewCreateCmd(l))
-	cmd.AddCommand(accounts.NewListCmd(l))
-	cmd.AddCommand(accounts.NewDeleteCmd(l))
-	cmd.AddCommand(accounts.NewSetPwCommand(l))
-	cmd.AddCommand(accounts.NewResetJWTKeysCommand(l))
-	cmd.AddCommand(accounts.NewInitialAdminPasswdCommand(l))
+func init() {
+	rootCmd.AddCommand(accountsCmd)
```
mayankshah1607, line 34: TBH I have some mixed feelings about the multiple `init()` functions in a single package.. In the current implementation I don't see any problems, but I'm worried going forward they could end up being used for more than flag initialisation or binding sub-commands, which would introduce hard to debug side-effects 😅 

It might just be a personal preference, but I would rather we have a something like `NewAccountsCmd()` and do the flag initialisation and sub-command binding there. Here's what I mean:
```go
func newAccountsCmd() *cobra.Command {
        // setup flags..
        
        // setup sub-commands
	accountsCmd.AddCommand(accounts.GetCreateCmd())
	accountsCmd.AddCommand(accounts.GetListCmd())
	accountsCmd.AddCommand(accounts.GetDeleteCmd())
	accountsCmd.AddCommand(accounts.GetSetPasswordCmd())
	accountsCmd.AddCommand(accounts.GetResetJWTKeysCmd())
	accountsCmd.AddCommand(accounts.GetInitAdminPasswordCmd())

}
```

and then on the rootCmd side:
```go
func NewRootCmd() *cobra.Cobra {
     rootCmd.AddCommand(accounts.NewAccountsCmd())
     // ...
}
```

mayankshah1607 (commented):
Just some suggestions to consider :) 

Overall LGTM, great improvements. Thanks!

commands/install.go
```diff
@@ -20,106 +20,91 @@ import (
...
+	installCmd.Flags().StringVar(&installCfg.Namespaces, cli.FlagNamespaces, install.DefaultDBNamespaceName, "Comma-separated namespaces list Percona Everest can manage")
+	installCmd.Flags().BoolVar(&installCfg.SkipWizard, cli.FlagSkipWizard, false, "Skip installation wizard")
+	installCmd.Flags().StringVar(&installCfg.VersionMetadataURL, cli.FlagVersionMetadataURL, "https://check.percona.com", "URL to retrieve version metadata information from")
+	installCmd.Flags().StringVar(&installCfg.Version, cli.FlagVersion, "", "Everest version to install. By default the latest version is installed")
+	installCmd.Flags().BoolVar(&installCfg.DisableTelemetry, cli.FlagDisableTelemetry, false, "Disable telemetry")
+	_ = installCmd.Flags().MarkHidden(cli.FlagDisableTelemetry)
+	installCmd.Flags().BoolVar(&installCfg.SkipEnvDetection, cli.FlagSkipEnvDetection, false, "Skip detecting Kubernetes environment where Everest is installed")
+	installCmd.Flags().BoolVar(&installCfg.SkipDBNamespace, cli.FlagInstallSkipDBNamespace, false, "Skip creating a database namespace with install")
+
+	// --namespaces and --skip-db-namespace flags are mutually exclusive
```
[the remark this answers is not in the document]
maxkondr: yes, found in cobra guideline

commands/namespaces.go
```diff
@@ -18,19 +18,22 @@ package commands
...
-	cmd.AddCommand(namespaces.NewAddCommand(l))
-	cmd.AddCommand(namespaces.NewRemoveCommand(l))
-	cmd.AddCommand(namespaces.NewUpdateCommand(l))
-	return cmd
+var namespacesCmd = &cobra.Command{
+	Use:   "namespaces <command> [flags]",
+	Args:  cobra.ExactArgs(1),
+	Long:  "Managed Everest database namespaces",
+	Short: "Managed Everest database namespaces",
+	Run:   func(_ *cobra.Command, _ []string) {},
```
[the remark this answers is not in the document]
maxkondr: if Run function is not defined -> help message will not be showed in case you execute just `namespaces`, so that you will not be able to check out what sub-commands it supports. However if you pass `namespaces -h` - help message will be shown. I thought it is convenient to show help message each time when not all command params are passed

commands/accounts.go
maxkondr: First of all it is needed to explain why sub-commands initialisation here (and actually there are a couple of places more) is differ from the rest ones. Sub-commands for `accounts` are located in a separate package (not in the same where `rootCmd` is defined) so that they can reach `rootCmd` using module scope. Importing `everest/commands` module from `everest/commands/accounts` module is not possible as well because it creates cyclic imports.

Go provides with possibility to use multiple `init()` functions in one more exactly for such purposes:
- make initialisations of structs as close as possible to their declarations
- avoid using one huge `init()` that gathers logic from different files in module.

Of course this feature shall be used with caution, at least do not rely on `init()` calls order.
Personally I would like to avoid using `New*Cmd()` because this is more error prone way - it assumes that fund may be called several times and produce several Cmd instances. In this case with (CLI) this is not allowed. I think this is the main reason why cobra lib suggests init commands in this way.

mayankshah1607: Okay, makes sense 👍🏻

mayankshah1607 (approved):

PR #1019, opened 2025-01-17
EVEREST-107 Simplify Feature Builds

recharte:
By creating the VS deployment manifest during the build process and by embedding the helm chart and CLI binaries into the same FB artifact we can simplify the users experience of installing a feature build. Now, after downloading and unzipping 1 artifact, users only need to run `./helm.sh` or `./cli.sh` to install the FB.

### DIFF
--- a/.github/workflows/feature-build.yaml
+++ b/.github/workflows/feature-build.yaml
@@ -99,6 +99,9 @@ jobs:
           echo "VERSION_TAG=v$V" >> $GITHUB_ENV
           echo "VS_TAG=everest-test$TIMESTAMP" >> $GITHUB_ENV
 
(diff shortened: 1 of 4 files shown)


oksana-grishchenko (approved):
Thank you Diogo for implementing the changes, it reduces the amount additional manual steps so smoothly! 

One suggestion which is not directly related to the PR, but since we are changing the workflow, let's increase the timeout in vs.sh
```
kubectl wait --for=jsonpath='{.status.readyReplicas}'=3 deployment/percona-version-service > /dev/null
``` 
default is 30sec, but today Yusaf faced 
`error: readyReplicas is not found`, so i think it worth to increase the waiting time to 60s at least

PR #1021, opened 2025-01-20
EVEREST-1729 | check CLI constraints from VS during install

mayankshah1607:


pkg/cli/install/install.go
```diff
@@ -242,6 +242,21 @@ func (o *Install) setVersionInfo(ctx context.Context) error {
...
 	o.l.Debugf("Everest version information %#v", latestMeta)
 	o.installVersion = latest.String()
+
+	supVer, err := common.NewSupportedVersion(latestMeta)
+	if err != nil {
+		return err
+	}
+	if err := o.checkRequirements(supVer); err != nil {
+		return err
+	}
```
maxkondr, line 252: I would place these lines before line 242, because line 242 prints debug logs that "it looks OK" but later we return an error that "version is not compatible".
An in addition there is line 244 `o.installVersion = latest.String()` that changes the state of Install obj before the verification is passed - it is error-prone way.

mayankshah1607: Sure, updated

maxkondr (approved):

PR #1033, opened 2025-01-21
EVEREST-107 Trigger FB automatically

recharte:
This PR adds the ability to automatically create feature builds for every PR labeled with `feature-build`. A new feature is built on every change to the PR.

To achieve this, the existing feature build workflow was changed to:
* Receive a catalog branch instead of an OLM bundle of an individual operator to upgrade
* Fallback to the default branches (main or production in the case of the VS repo) if the specified branch couldn't be found in the corresponding repo.

This new workflow assumes that the branches in the other repos have the same name of the branch of the PR.
E.g. if the branch of the PR is called `EVEREST-107-trigger-fb-automatically`, the branches for the everest-operator, everest-catalog, percona-helm-charts and percona-version-service must also be called `EVEREST-107-trigger-fb-automatically`. If these repos don't have such a branch, main (or production in the case of VS) will be used instead.
If custom branches need to be used, the feature build must be created manually using the existing action.

### DIFF
--- a/.github/workflows/feature-build.yaml
+++ b/.github/workflows/feature-build.yaml
@@ -101,6 +101,7 @@ jobs:
     steps:
       - name: Input validation
         run: |
+          echo "ARTIFACTS_RETENTION_DAYS=$ARTIFACTS_RETENTION_DAYS"
           if [[ ! "$ARTIFACTS_RETENTION_DAYS" =~ ^[0-9]+$ ]]; then
             echo "Wrong artifacts_retention_days format. Use an integer number. 0 means default GitHub repo settings."
             exit 1
--- a/.github/workflows/feature-build.yaml
+++ b/.github/workflows/feature-build.yaml
@@ -109,22 +109,22 @@ jobs:
           
       - name: Set environment variables
         run: |
-          echo "ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')" >> $GITHUB_ENV
-          echo "OS=$(uname | awk '{print tolower($0)}')" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


.github/workflows/feature-build.yaml
```diff
@@ -77,27 +101,30 @@ jobs:
     steps:
       - name: Input validation
         run: |
+          echo "ARTIFACTS_RETENTION_DAYS=$ARTIFACTS_RETENTION_DAYS"
```
recharte, line 104: Debugging leftover
```suggestion
```

oksana-grishchenko (approved):
🚀

.github/workflows/feature-build.yaml
```diff
@@ -266,43 +292,10 @@ jobs:
...
-          cd ..
-          curl -Lo /tmp/opm https://github.com/operator-framework/operator-registry/releases/download/v1.48.0/${OS}-${ARCH}-opm
-          chmod +x /tmp/opm
-          /tmp/opm alpha render-template basic --skip-tls -o yaml < veneer/everest-operator.yaml > catalog/everest-operator/catalog.yaml
-
-          # Check if catalog has the new version listed
-          if ! grep -q "$VERSION$" catalog/everest-operator/catalog.yaml; then
-            echo "catalog/everest-operator/catalog.yaml does not include the version $VERSION"
-            exit 1
-          fi
```
oksana-grishchenko, line 289: Sorry, I missed it when was reviewing yesterday. We still need to add the FB version to the catalog image.

oksana-grishchenko: https://github.com/percona/everest/pull/1038 PR that puts it back

PR #1035, opened 2025-01-22
[UI] EVEREST-1781 Update message for empty state when user has no create permissions

dianabirs:
[EVEREST-1781](https://perconadev.atlassian.net/browse/EVEREST-1781)

https://github.com/user-attachments/assets/96e258a5-2b8a-4d0e-aff4-2afef5adbe10

[EVEREST-1781]: https://perconadev.atlassian.net/browse/EVEREST-1781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/apps/everest/src/pages/common/empty-state/databases/index.tsx
+++ b/ui/apps/everest/src/pages/common/empty-state/databases/index.tsx
@@ -1,4 +1,4 @@
-import { Box, Divider, Typography } from '@mui/material';
+import { Box, Divider, Link, Typography } from '@mui/material';
 import { EmptyStateIcon } from '@percona/ui-lib';
 import { Messages } from './messages';
 import CreateDbButton from 'pages/databases/create-db-button/create-db-button';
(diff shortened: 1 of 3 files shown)


solovevayaroslavna (approved):

ui/apps/everest/src/pages/common/empty-state/databases/index.tsx
```diff
@@ -22,7 +22,24 @@ export const EmptyStateDatabases = ({
         <EmptyStateIcon w="60px" h="60px" />
         <Box sx={centeredContainerStyle}>
           <Typography>{Messages.noDbClusters}</Typography>
-          <Typography> {Messages.createToStart}</Typography>
+          {showCreationButton ? (
+            <Typography> {Messages.createToStart} </Typography>
+          ) : (
+            <>
+              <Typography>{Messages.noPermissions}</Typography>
```
fabio-silva, line 29: I think we should not use `!showCreationButton` to assume there's no permissions. That prop should have only that purpose itself, which is to show/hide the button, as the name implies.
Here we are showing the "no permission" message when "showCreationButton" is false, but that's because we know the business logic from the outside. The component, however, should not be aware of it.
If we want to control the display of the "lack of permissions" message, we better have another prop for that purpose, even though both props will have the same expression to control them.

fabio-silva (changes requested):

fabio-silva (approved):

PR #1047, opened 2025-01-23
EVEREST-107 remove dead code

recharte:
[EVEREST-107](https://jira.percona.com/browse/EVEREST-107) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-107]: https://perconadev.atlassian.net/browse/EVEREST-107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/kubernetes/informer/informer.go
```diff
@@ -99,16 +99,6 @@ func (i *Informer) OnUpdate(cb func(oldObj, newObj interface{})) {
...
-// OnAdd is triggered when an object is added.
-func (i *Informer) OnAdd(cb func(obj interface{})) {
-	i.eventHandlers.AddFunc = cb
-}
-
-// OnDelete is triggered when an object is deleted.
-func (i *Informer) OnDelete(cb func(obj interface{})) {
-	i.eventHandlers.DeleteFunc = cb
-}
-
```
mayankshah1607, line 111: I would actually not remove this, they are a part of the `Informer` utility. We are currently using only `OnUpdate` for reloading the RBAC configMap, but it is possible to use `OnAdd` and `onDelete` too for other use-cases.

What I mean is this is a generic helper library to watch Kubernetes events and perform actions, so without these 2 methods, the library is kind of incomplete even though we aren't using it anywhere 😅

pkg/accounts/cli/accounts.go
```diff
@@ -75,11 +75,6 @@ func NewAccounts(c Config, l *zap.SugaredLogger) (*Accounts, error) {
 	return cli, nil
 }
 
-// WithAccountManager sets the account manager for the Accounts.
-func (c *Accounts) WithAccountManager(m accounts.Interface) {
-	c.accountManager = m
-}
-
```
mayankshah1607, line 82: Are you sure this is unused? 🤔

mayankshah1607: nevermind, it is unused. There's another `WithAccountManager` in the session package that I got confused with - https://github.com/percona/everest/blob/main/pkg/session/manager.go#L64

pkg/kubernetes/client/database/database.go
```diff
@@ -47,40 +46,6 @@ type DBClusterClient struct {
...
-}
-
-// DBClusters returns a db cluster client.
-func (c *DBClusterClient) DBClusters(namespace string) DBClusterInterface { //nolint:ireturn
-	return &dbClusterClient{
-		restClient: c.restClient,
-		namespace:  namespace,
-	}
-}
-
```
oksana-grishchenko, line 83: Is there a point in keeping anything else in this file? By removing this method the `dbClusterClient` struct also becomes unused. The `addToScheme` gets unused by deleting `NewForConfig`.  The interfaces `DBClusterClientInterface` `DBClusterInterface` also seems to be not used anywhere.

oksana-grishchenko (approved):

PR #1063, opened 2025-01-28
EVEREST-1799 Configure OIDC scope

recharte:
[EVEREST-1799](https://jira.percona.com/browse/EVEREST-1799) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

In #980 we added the ability to fetch user groups from the external IdP. However, we can't always request the groups scope during the OIDC flow. This is because the IdP might not support it, or the user might not have configured it.
In order to keep backwards compatibility, we need to add a new field to the settings API that returns the user configured OIDC scope. Like the other OIDC settings this can only be configured through the CLI. By default, the scope should be `openid profile email`.

Related PRs:
* https://github.com/percona/percona-helm-charts/pull/478

[EVEREST-1799]: https://perconadev.atlassian.net/browse/EVEREST-1799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commands/settings/oidc/configure.go
```diff
@@ -42,8 +43,9 @@ var (
 
 func init() {
 	// local command flags
-	settingsOIDCConfigureCmd.Flags().StringVar(&settingsOIDCConfigureCfg.IssuerURL, cli.FlagOIDCIssueURL, "", "OIDC issuer url")
-	settingsOIDCConfigureCmd.Flags().StringVar(&settingsOIDCConfigureCfg.ClientID, cli.FlagOIDCIssueClientID, "", "OIDC application client ID")
+	settingsOIDCConfigureCmd.Flags().StringVar(&settingsOIDCConfigureCfg.IssuerURL, cli.FlagOIDCIssuerURL, "", "OIDC issuer url")
+	settingsOIDCConfigureCmd.Flags().StringVar(&settingsOIDCConfigureCfg.ClientID, cli.FlagOIDCClientID, "", "OIDC application client ID")
+	settingsOIDCConfigureCmd.Flags().StringVar(&settingsOIDCConfigureCfg.Scope, cli.FlagOIDCScope, common.DefaultOIDCScope, "Space-separated list of requested scopes")
```
oksana-grishchenko, line 48: Although I see the reason why here we use space-separated values here, I'm still not sure we should define the cli interface differently from what we have for namespaces (comma-separated). Maybe we should have an additional step to convert the values to space-separated format under the hood so we keep a unified cli interface and don't expose implementation details.

oksana-grishchenko: although looks like we'll have different format for helm and cli then...

oksana-grishchenko (approved):

commands/settings/oidc/configure.go
oksana-grishchenko: then let's have it as it is, sorry for the noise 😅

commands/settings/oidc/configure.go
```diff
@@ -33,7 +34,7 @@ var (
 		Args:    cobra.NoArgs,
 		Long:    "Configure OIDC settings",
 		Short:   "Configure OIDC settings",
-		Example: `everestctl settings oidc configure --issuer-url https://example.com --client-id 123456`,
+		Example: `everestctl settings oidc configure --issuer-url https://example.com --client-id 123456 --scope "openid profile email groups"`,
```
maxkondr, line 37: I would make format the same as for --namespaces flag, i.e
```suggestion
		Example: `everestctl settings oidc configure --issuer-url https://example.com --client-id 123456 --scopes openid,profile,email,groups`,
```

And in addition this is less error-prone way to pass values

commands/settings/oidc/configure.go
maxkondr: I would insist on having comma-separated.

pkg/cli/flags.go
```diff
@@ -76,10 +76,12 @@ const (
...
-	// FlagOIDCIssueURL is the name of the issuer-url flag.
-	FlagOIDCIssueURL = "issuer-url"
-	// FlagOIDCIssueClientID is the name of the client-id flag.
-	FlagOIDCIssueClientID = "client-id"
+	// FlagOIDCIssuerURL is the name of the issuer-url flag.
+	FlagOIDCIssuerURL = "issuer-url"
+	// FlagOIDCClientID is the name of the client-id flag.
+	FlagOIDCClientID = "client-id"
+	// FlagOIDCScope is the name of the scope flag.
+	FlagOIDCScope = "scope"
```
maxkondr, line 84:
```suggestion
	FlagOIDCScopes = "scopes"
```

pkg/oidc/configure.go
```diff
@@ -45,6 +47,8 @@ type Config struct {
 	IssuerURL string
 	// ClientID ID of the client OIDC app.
 	ClientID string
+	// Scope requested scopes.
+	Scope string
```
maxkondr, line 51:
```suggestion
	Scopes []string
```

pkg/oidc/configure.go
```diff
@@ -93,6 +98,16 @@ func (u *OIDC) Run(ctx context.Context) error {
 		return errors.New("clientID and/or issuerURL are not provided")
 	}
 
+	// Even though in other commands we use comma separated lists, we use a
+	// space separated list here because that's how the OIDC spec defines
+	// scopes. See Section 3.3 of OAuth 2.0
```
maxkondr, line 103:
```
// Even though in other commands we use comma separated lists, we use a
// space separated list here because that's how the OIDC spec defines
// scopes.
```

OIDC spec doesn't force us to have the same format in our CLI, this is our task to convert from one format to another

maxkondr (changes requested):

commands/settings/oidc/configure.go
PeterSzcz: We should follow known patterns for OIDC configuration, this operation is not "day-to-day" stuff you do but a set-up/configuration which happens once in blue moon and can be performed by a completely different "person" (which should be familiar with OIDC formats IMHO)

recharte: Well, I feel very split on this topic.
My initial implementation was using the comma-separated list but then I thought that people that are used to configuring OIDC will be slightly confused by this unusual format, so I changed to the space-separated list. Even the naming `scope` instead of `scopes` was intentional because that's how OIDC spec defines it.
Unfortunately, we need to compromise somewhere, we either keep the consistency with our other CLI commands or we keep consistency with the standard.
Let me share how both options would look like from CLI, API and Helm chart. I believe we need to put a bigger emphasis on the Helm part because that's now our default installation method and will quickly become the most popular.

## Comma-separated/array approach
### CLI
```
everestctl settings oidc configure \
    --issuer-url=https://example.com \
    --client-id=123456789 \
    --scopes="openid,profile,email,groups"
```
### API (`/settings`)
```
{
    "oidcConfig": {
        "clientId": "123456789",
        "issuerURL": "https://example.com",
        "scopes": [
                "openid",
                "profile",
                "email",
                "groups"
        }
    }
}
```
### Helm (`values.yaml`)
```
  oidc:
    issuerUrl: "https://example.com"
    clientId: "123456789"
    scopes:
        - openid
        - profile
        - email
        - groups
```
## Space-separated/string approach
### CLI
```
everestctl settings oidc configure \
    --issuer-url=https://example.com \
    --client-id=123456789 \
    --scope="openid profile email groups"
```
### API (`/settings`)
```
{
    "oidcConfig": {
        "clientId": "123456789",
        "issuerURL": "https://example.com",
        "scope": "openid profile email groups"
    }
}
```
### Helm (`values.yaml`)
```
  oidc:
    issuerUrl: "https://example.com"
    clientId: "123456789"
    scope: "openid profile email groups"
```

Again, I'm having a hard time coming up with a good reason for going with one over the other, they both end up being very similar, with the nuances I shared above...

oksana-grishchenko: the consumer of this setting is UI eventually. As I understand, UI needs the setting [space separated](https://github.com/percona/everest/blob/7f097f855befe5bcbeaaf672474b51d472d1ad84/ui/apps/everest/src/App.tsx#L75). 

I also thought about a combined approach. What if we'll have 
### CLI
```
everestctl settings oidc configure \
    --issuer-url=https://example.com \
    --client-id=123456789 \
    --scopes="openid,profile,email,groups"
```
which is aligned with the existing CLI interface. 
However under the hood we convert that input to: 
### API (`/settings`)
```
{
    "oidcConfig": {
        "clientId": "123456789",
        "issuerURL": "https://example.com",
        "scope": "openid profile email groups"
    }
}
```
### Helm (`values.yaml`)
```
  oidc:
    issuerUrl: "https://example.com"
    clientId: "123456789"
    scope: "openid profile email groups"
```
so in the API we'll have what UI needs and in helm we'll have something that is familiar to people who usually configure such stuff. 

However it would mean that when we configure OIDC and running the actual commands, we would need to provide the values space-separated for helm but comma-separated for CLI, which sucks. 

So any of the options sucks in some way. I don't insist on any of the approaches, let's just pick one of them. However I think the idea of returning space-separated values at least for UI does make sense.

recharte: > However it would mean that when we configure OIDC and running the actual commands, we would need to provide the values space-separated for helm but comma-separated for CLI, which sucks.
Being inconsistent between CLI and API/Helm doesn't sound all that great, I'd rather go with one of the other approaches...

oksana-grishchenko: Maybe then we could go with the space-separated approach but add some validation about the format, so even though the format is not aligned with other cli commands, we could make sure that user inputs the values in a correct format.

fabio-silva (approved):

oksana-grishchenko (approved):

maxkondr (approved):

PR #1088, opened 2025-02-06
Everest-1862 | topology diagram view

fabio-silva:
[EVEREST-1862](https://jira.percona.com/browse/EVEREST-1862) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-1862]: https://perconadev.atlassian.net/browse/EVEREST-1862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot 2025-02-04 at 18 36 09


dianabirs:
The zoom icons are not visible in dark mode
image

dianabirs:
@fabio-silva I have a suggestion here: 

Because only the grab cursor is displayed inside the topology diagram, it makes it impossible for a user to copy a component's name. Could we change this that the cursor is changed to arrow when hovering the name of the component? I think a user should be able to copy that name for debugging purposes

fabio-silva:
> @fabio-silva I have a suggestion here:
> 
> Because only the grab cursor is displayed inside the topology diagram, it makes it impossible for a user to copy a component's name. Could we change this that the cursor is changed to arrow when hovering the name of the component? I think a user should be able to copy that name for debugging purposes

That's a valid point. However, it seems like it's not plain simple as changing the cursor type. The text remains unselectable even with the cursor. I imagine react-flow manages events in the diagram for dragging purposes, which makes this more difficult without probably having to mess with CSS and events.
TBH, I wouldn't change this for now. I added the "cursor" pointer to component nodes so that users know they are selectable anyways.

dianabirs (approved):

dianabirs (approved):

PR #1124, opened 2025-02-18
Fix afterAll hook in release tests

tplavcic:
At some point Playwright decided that they want to execute only afterAll hook (beforeAll hook is skipped) even if the whole test group is skipped so we need to add a condition in afterAll hook to not execute the content if the group is skipped.

### DIFF
--- a/ui/apps/everest/.e2e/release/psmdb-sharding.e2e.ts
+++ b/ui/apps/everest/.e2e/release/psmdb-sharding.e2e.ts
@@ -95,20 +95,23 @@ test.describe(
     });
 
     test.afterAll(async ({ request }) => {
-      // we try to delete all monitoring instances because cluster creation expects that none exist
-      // (monitoring instance is added in the form where the warning that none exist is visible)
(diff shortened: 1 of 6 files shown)


ui/apps/everest/.e2e/release/psmdb-sharding.e2e.ts
```diff
@@ -95,20 +95,26 @@ test.describe(
...
-      const monitoringInstances = await listMonitoringInstances(
-        request,
-        namespace,
-        token
-      );
-      for (const instance of monitoringInstances) {
-        await deleteMonitoringInstance(
+      // Playwright decided to execute only afterAll hook even if the group is skipped so we need a condition here
+      if (
+        (SELECT_DB ? SELECT_DB === db : true) &&
```
fabio-silva, line 100: can we maybe refactor these conditions into a function, given we use it multiple times?

fabio-silva (changes requested):

fabio-silva (approved):

PR #1132, opened 2025-02-22
EVEREST-1654 Use controller-runtime client instead of Rest

maxkondr:
[EVEREST-1654](https://jira.percona.com/browse/EVEREST-1654) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

This is a PoC.

The idea is the following:
- in Everest server our own rest client (`pkg/kubernetes/client/` and `pkg/kubernetes/client/customresources`) can be removed. Instead, the logic can be placed in `pkg/kubernetes.Kubernetes`.
- `controller-runtime.Client` with cache is used in `pkg/kubernetes.Kubernetes` in the same way it is used in Everest-operator.
- cache allows to use FieldIndexer to index fields in our CRs. It allows later to lookup our CRs by such fields.
-- this allows us to remove the logic in `everest-operator` that copies values from `.spec.` fields to `metadata.labels` for our CRs.

[EVEREST-1654]: https://perconadev.atlassian.net/browse/EVEREST-1654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/kubernetes/backup_storage.go
```diff
@@ -82,34 +83,49 @@ func (k *Kubernetes) IsBackupStorageUsed(ctx context.Context, namespace, name st
 	}
 
 	// Check if it is in use by clusters?
-	clusters, err := k.client.ListDatabaseClusters(ctx, namespace, metav1.ListOptions{})
-	if err != nil {
+	listOpts := &ctrlclient.ListOptions{
```
maxkondr, line 86: this is an example of FieldSelector usage over `.spec.backup.schedules.backupStorageName` in DatabaseCluster CR

pkg/kubernetes/client/customresources/databaseclusters.go
maxkondr, line 1: this file is not needed any more since all logic is placed in `pkg/kubernetes/database_cluster.go`

pkg/kubernetes/client/database_cluster.go
maxkondr, line 1: this file is not needed any more since all logic is placed in `pkg/kubernetes/database_cluster.go`

pkg/kubernetes/client/kubeclient_interface.go
maxkondr, line 1: Once all logic is moved to `pkg/kubernetes/` - this file can be removed as well

pkg/kubernetes/database_cluster.go
```diff
@@ -29,27 +30,39 @@ import (
 
 // ListDatabaseClusters returns list of managed database clusters.
 func (k *Kubernetes) ListDatabaseClusters(ctx context.Context, namespace string) (*everestv1alpha1.DatabaseClusterList, error) {
-	return k.client.ListDatabaseClusters(ctx, namespace, metav1.ListOptions{})
+	result := &everestv1alpha1.DatabaseClusterList{}
```
maxkondr, line 33: this is an example of the logic that replaces one from `pkg/kubernetes/client/` + `pkg/kubernetes/client/customresources` layers

pkg/kubernetes/database_cluster.go
```diff
@@ -29,27 +30,39 @@ import (
...
+	err := k.k8sClient.List(ctx, result, &ctrlclient.ListOptions{Namespace: namespace})
+	return result, err
 }
 
 // GetDatabaseCluster returns database clusters by provided name.
 func (k *Kubernetes) GetDatabaseCluster(ctx context.Context, namespace, name string) (*everestv1alpha1.DatabaseCluster, error) {
-	return k.client.GetDatabaseCluster(ctx, namespace, name)
+	result := &everestv1alpha1.DatabaseCluster{}
+	err := k.k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, result)
```
maxkondr, line 41: `k.k8sClient.Get` and `k.k8sClient.List` operations use cache that is being updated in background and in case resource in K8S is changed

pkg/kubernetes/kubernetes.go
```diff
@@ -90,7 +93,10 @@ var ErrEmptyVersionTag = errors.New("got an empty version tag from Github")
 
 // Kubernetes is a client for Kubernetes.
 type Kubernetes struct {
-	client     client.KubeClientConnector
+	// client client.KubeClientConnector
+	client     *client.Client
```
maxkondr, line 97: this `client` will be removed once all logic from `pkg/kubernetes/client/` + `pkg/kubernetes/client/customresources` layers is moved to `pkg/kubernetes/` layer

pkg/kubernetes/kubernetes.go
```diff
@@ -146,13 +152,92 @@ func NewInCluster(l *zap.SugaredLogger) (*Kubernetes, error) {
...
+			k.l.Error("failed to start cache", zap.Error(err))
+		}
+	}()
+	return nil
+}
+
+// initIndexers initializes the field indexers for the DatabaseCluster.
+func (k *Kubernetes) initIndexers(ctx context.Context) error {
+	if err := k.k8sCache.IndexField(
+		ctx, &everestv1alpha1.DatabaseCluster{}, ".spec.backup.schedules.backupStorageName",
```
maxkondr, line 211: this is an example of indexing field in `.spec` for our DatabaseCluster CR. This index is later used for DBs lookup by field

recharte:
Hi @maxkondr in general I like this idea but I also have a few questions/concerns.

First of all, besides the `FieldIndexer` benefit you already identified in this PR's description, I see the following pros:
1. Gets rid of the kubernetes/client abstraction. This sounds good because, currently, we are not taking advantage of the `KubeClientConnector` interface. Unless I'm missing something, it just leads to boilerplate code.
2. (this can also be a con see point 2. below) Using the controller-runtime's client forces us to use a single interface for managing all k8s resources regardless of their API groups and versions. This will allow us to get rid of [this mess](https://github.com/percona/everest/blob/0c1b217018006541c96e2f37cca3fd60583712d5/pkg/kubernetes/client/client.go#L109-L113) :see_no_evil: 

Here are the concerns that came through my mind that might merit some discussion:
1. Does the controller-runtime's client and cache have any limitations when running outside the k8s cluster? This will be important in the future when we add support for connecting multiple k8s clusters to a single everest instance.
From my initial assessment, I couldn't find any relevant limitations. In the end, it seems like both clients communicate similarly with the k8s API server, they just provide different interfaces for users.
2. (this can also be a pro see point 2. above) Using the controller-runtime's client forces us to use a single interface for managing all k8s resources regardless of their API groups and versions. This means that we'll lose access to the (arguably) useful abstractions the respective client sets introduce. Being used to the generic interface provided by the controller-runtime's client, I personally don't feel like these client set interfaces are that useful but some might disagree.
3. Using the controller-runtime's client forces us to maintain a runtime Scheme. The DB operators can be installed/uninstalled during the lifetime of our API server, meaning that the k8s API's registered CRDs can change. This means that if we ever need to interface with such resources we'll need to mutate the scheme accordingly. However, since our API server doesn't interface directly with the DB operators' CRs we should probably be fine and can avoid this complexity.
4. Introducing caching will definitely increase the memory footprint. The question is: by how much? Our API server queries a multitude of resources, not just our CRs, e.g. we fetch pods and secrets. These resources can be plentiful in a large cluster. I believe that this can potentially have a big impact the memory consumption.

[Here's an interesting read](https://medium.com/@timebertt/kubernetes-controllers-at-scale-clients-caches-conflicts-patches-explained-aa0f7a8b4332) that, among other things, discusses the differences from the client-go and controller-runtime clients and cache mechanisms. Even though, the author approaches it from the point of view of a controller/operator we can still use this info to make a more informed decision about what to use in our API server.

recharte:
Thanks @maxkondr for the follow-up on this discussion. I agree with you and I think it's probably worth going ahead with implementing your proposal but before you jump to a full implementation let's hear from @mayankshah1607.

mayankshah1607:
In general I also like the idea of using the controller-runtime client mainly because we can get rid of the `KubeClientConnector`, etc. mess and use the simpler interface it provides (Get, Update, Create, etc.). 

But I have concerns about the caching:
1. Cache updates cannot always be instant. The delays are more evident in larger clusters. We perform certain checks on the state of objects before performing any action (like see if BackupStorages are in use before deleting it). If the cache is inconsistent, these checks can result in false results.
2. As pointed out above, memory footprint is also another issue. I'm not sure its enough to just skip ConfigMaps and Secrets (we have certain OLM objects as well that are huge). We need to be very thoughtful about which resources we are requesting for otherwise we can overwhelm the cache
3. When using a cached client, we will start observing problems when reading an object immediately after writing it. This is related to problem (1) -- the cache is used only for reading, writes are performed on the Kube API directly and later reconciled using the informers. This means that reading immediately after writing an object will likely give us an outdated result. I don't know for sure if we have such a scenario in the code already, but even if not, we need to be careful in the future, otherwise this can be hard to troubleshoot.

tl;dr - controller-runtime.Client is good, but I am not convinced we need caching. Yes, it can give some performance boost, but maybe it is premature optimisation considering the added complexity with the cache. I would at least consider providing an option to disable caching (us an uncached client)

maxkondr:
@mayankshah1607 thanks for your comments.

In such a case, I would suggest the following:
- we switch from REST to controller-runtime.Client -> get rid of `KubeClientConnector` mess
- we disable caching at all for now -> achieve the same results (i.e. we call K8S API intensively and get consistent state) but with less code and mess.
- later we will come back to caching topic again and move step by step

@mayankshah1607 , @recharte  - comments/objections?

mayankshah1607:
Yes, that sounds good. 👍🏻

PR #1145, opened 2025-02-26
Add possibility to Tilt to create BackupStorages during namespace provisioning

maxkondr:
this PR extends the Tilt functionality with the following:

1. It introduces the possibility to define all mandatory/optional params for running Tilt (and Everest + Everest operator) in local `.env` file that is placed in the same `dev` directory - see `dev/.env.example` file.
2. It introduces the possibility to define in `config.yaml` config file the list of backup storages that shall be created in the particular namespace during namespace provisioning - see `dev/config.yaml.example` file
2.1 Tilt watches for changes in `config.yaml` and creates new BackupStorages once they appear in config.
2.2. Tilt removes the existing BackupStorages in case they disappear from config.


dev/Tiltfile
```diff
@@ -569,3 +580,54 @@ k8s_resource(
...
+            request_body=getBackupStorageRequestBody(storage_name)
+            local_resource(
+                'bs:{0}:{1}'.format(ns_name, storage_name),
+                '{0} && {1}'.format(api_token_cmd, backup_storage_cmd_tmpl.format(request_body, ns_name)),
+                resource_deps = [
+                    'minIO-deploy',
+                    'everest-server-deploy',
+                ],
+                labels=["backup-storage"]
+            )
```
recharte, line 633: Can't we do this the cloud-native way and create the BackupStorage CRs and respectively Secrets instead of using the API?
Relying on the API doesn't feel right. If you are testing some changes to the API this can fail. Also, this is not really idempotent, subsequent calls to this will have the curl command fail. Handling the k8s resources directly sounds better to me.

dev/.env.example
```diff
@@ -0,0 +1,43 @@
...
+# Optional, default: ../ from Tiltfile
+#EVEREST_BACKEND_DIR=
+
+# Path to Everest server frontend source code directory.
+# Optional, default: ../ui from Tiltfile
+#EVEREST_FRONTEND_DIR=
+
+# Path to Everest CLI source code directory.
+# Optional, default: ../ from Tiltfile
+#EVEREST_CLI_DIR=
```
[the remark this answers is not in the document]
maxkondr: done

dev/README.md
```diff
@@ -50,20 +50,33 @@ gcloud auth configure-docker <REGISTRY_REGION>-docker.pkg.dev
...
+
+Copy file dev/config.yaml.example to dev/config.yaml and set the needed DB namespaces that will be created automatically.
+
+3. (Optional) If you want to test a specific version of a given DB operator you can set the following environment variables in .env file or in the terminal:
 ```sh
-export PXC_OPERATOR_VERSION=1.13.0
-export PSMDB_OPERATOR_VERSION=1.15.0
-export PG_OPERATOR_VERSION=2.3.1
+export PXC_OPERATOR_VERSION=1.16.1
+export PSMDB_OPERATOR_VERSION=1.19.0
```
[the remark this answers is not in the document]
maxkondr: done

dev/Tiltfile
recharte: Hmmm 🤔 I see 
But in this case do we really need to set the owner ref in the secret? That is only useful for garbage collection, so that when a BS is deleted the accompanying secret is also deleted. In this case since tilt would be handling both, it would also delete both when tearing down. So maybe we can avoid setting the owner ref in the secret 🤷

maxkondr: Done

dev/resources/backupstorage.yaml
```diff
@@ -0,0 +1,17 @@
...
+data:
+  AWS_ACCESS_KEY_ID: bWluaW9hZG1pbg==
+  AWS_SECRET_ACCESS_KEY: bWluaW9hZG1pbg==
+kind: Secret
+type: Opaque
+---
+apiVersion: everest.percona.com/v1alpha1
+kind: BackupStorage
+metadata:
+  generation: 1
```
recharte, line 11: I don't think this should be included, this is automatically managed by k8s.
```suggestion
```

oksana-grishchenko (approved):

PR #1171, opened 2025-03-06
EVEREST-1280 | Moving Storage class from first step to Advanced Configurations step

Tikdev00:
[EVEREST-1280](https://jira.percona.com/browse/EVEREST-1280) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-1280]: https://perconadev.atlassian.net/browse/EVEREST-1280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot from 2025-03-06 19-41-17


dianabirs:
I think the container needs `padding: 0` in order to look like in figma 
image

ui/apps/everest/src/components/advanced-card/advanced-card.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+const AdvancedCard: React.FC<AdvancedCardProps> = ({
+  title,
+  description,
+  controlComponent,
+  gapFromText = 48,
+}) => {
+  return (
+    <Card variant="outlined" sx={{ marginBottom: 2, padding: 2 }}>
```
dianabirs, line 18:
```suggestion
    <Card variant="outlined" sx={{ marginBottom: 2 }}>
```

dianabirs (approved):

Tikdev00:
> @Tikdev00 I think you mentioned changing the other fields to be consistent with this one, right? In terms of UI, with that new border around. Do you think we can pull that off?

@fabio-silva Yes I think we can do it with the help of advanced card component I've created, we can put other fields inside it, maybe it will need a little bit more configurations but overall it's doable, just as we discussed we have missing things like Affinity and as task was mainly about Storage Class I didn't touched other existing fields which as you can see are changed, if it will not harm the scope of this task, I can refactor them as well. I am about external access and DB engine parameters

PR #1174, opened 2025-03-07
Everest-1901 backup bucket trimming issue

Tikdev00:
[EVEREST-1901](https://jira.percona.com/browse/EVEREST-1901) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-1901]: https://perconadev.atlassian.net/browse/EVEREST-1901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/packages/ui-lib/src/form/inputs/text/text.tsx
```diff
@@ -10,35 +10,44 @@ const TextInput = ({
...
-          }}
-          helperText={error ? error.message : textFieldProps?.helperText}
-        />
-      )}
+      render={({ field: { onChange, ...field }, fieldState: { error } }) => {
+        const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+          onChange(
+            (onValueChange && onValueChange(event)) ?? event.target.value
+          );
+        };
```
dianabirs, line 25: is there a way to pass the `onChange` handler to the `textFieldProps`, instead of adding a prop to the component for this case?
Or maybe we want this trim at the end of inputs to be something that is done for every input, I can't think of a case in which we want an input value with a space at the end of the string.

Tikdev00: If we don't have any input that needs end spaces we can adjust this logic for all of them, but as far as in ticket it was mentioned only for this one apply trim logic I decided to not expand it for all of them.
About other inputs I am not sure if they need it or no so would like to hear your opinions, if yes then we can get rid of extra props and handle it inside Text component

dianabirs: @fabio-silva do you know of any inputs that we want to leave end spaces for?

dianabirs (approved):

ui/packages/ui-lib/src/form/inputs/text/text.tsx
Tikdev00: @dianabirs I've changed and onChange is now being passed through textFieldProps, please take a look again

PR #1195, opened 2025-03-11
EVEREST-1772 add security headers

recharte:
[EVEREST-1772](https://jira.percona.com/browse/EVEREST-1772) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

Adds the headers recommend by [OWASP's secure headers project](https://owasp.org/www-project-secure-headers/), with two exceptions:
* HSTS: will be added when we have native support for TLS ([EVEREST-1180](https://perconadev.atlassian.net/browse/EVEREST-1180))
* Clear-site-data: we don't set the recommended `storage` because we require local storage to be persisted.

[EVEREST-1772]: https://perconadev.atlassian.net/browse/EVEREST-1772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EVEREST-1180]: https://perconadev.atlassian.net/browse/EVEREST-1180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/server/middlewares.go
```diff
@@ -84,3 +93,56 @@ func (e *EverestServer) checkOperatorUpgradeState(next echo.HandlerFunc) echo.Ha
...
+		PermissionsPolicy:         PermissionsPolicy,
+		ReferrerPolicy:            "no-referrer",
+		// TODO (EVEREST-1180): Once we have native support for TLS, we should
+		// probably redirect all HTTP traffic to HTTPS.
+		// SSLRedirect:                   true,
+		// And we should set the HSTS header.
+		// STSIncludeSubdomains:          true,
+		// STSSeconds:                    31536000,
+		XPermittedCrossDomainPolicies: "none",
+	})
```
oksana-grishchenko, line 145: In [the list](https://owasp.org/www-project-secure-headers/ci/headers_add.json) of the recommended headers there are also those two:
```
   {
      "name": "Cache-Control",
      "value": "no-store, max-age=0"
    },
    {
      "name": "Clear-Site-Data",
      "value": "\"cache\",\"cookies\",\"storage\""
    }
```
Are they not relevant in our case?

oksana-grishchenko (approved):

internal/server/middlewares.go
recharte: To be honest I'm not 100% sure. I thought that users could benefit from the performance benefits of storing some content in a proxy server. Everest being a self-hosted solution maybe this isn't very relevant and we could just disable it.
@fabio-silva what do you think?

recharte: After a deeper discussion with Fábio, we decided to [add them](https://github.com/percona/everest/pull/1195/commits/62b895c5462f3ebdc2937d0a51f30f377ddbe897). However, the `Clear-Site-Data` will only have `cache` and `cookies` and it won't have the `storage` because that would screw up our local storage where we store stuff like the token.

PR #1203, opened 2025-03-13
UI tests - Fail global setup if storage or monitoring endpoints not created

tplavcic:
Currently if `minio` or `PMM` are not running the global setup just marks the setup as passed, but the tests will just fail and it doesn't make sense to run them.

### DIFF
--- a/ui/apps/everest/.e2e/setup/global.setup.ts
+++ b/ui/apps/everest/.e2e/setup/global.setup.ts
@@ -46,6 +46,10 @@ const doBackupCall = async (fn: () => Promise<APIResponse>, retries = 3) => {
       return Promise.resolve();
     }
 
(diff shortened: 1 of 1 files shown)


dianabirs (approved):

ui/apps/everest/.e2e/setup/global.setup.ts
```diff
@@ -46,6 +46,10 @@ const doBackupCall = async (fn: () => Promise<APIResponse>, retries = 3) => {
       return Promise.resolve();
     }
 
+    if (response.status() !== 201 && response.status() !== 200) {
+      Promise.reject();
```
fabio-silva, line 50:
```suggestion
      return Promise.reject();
```

fabio-silva (changes requested):

fabio-silva (approved):

PR #1204, opened 2025-03-13
EVEREST-1918 | check for dupicate IP and netmask entries for external access

Tikdev00:
[EVEREST-1918](https://jira.percona.com/browse/EVEREST-1918) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-1918]: https://perconadev.atlassian.net/browse/EVEREST-1918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tikdev00:
@fabio-silva should I point target branch as 1909 which is base for this, or keep main?

fabio-silva:
> @fabio-silva should I point target branch as 1909 which is base for this, or keep main?

@Tikdev00 1909, please

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration.tsx
```diff
@@ -68,10 +69,32 @@ export const AdvancedConfigurationForm = ({
...
     }
   }, [clusterInfo]);
+
   const handleBlur = (value: string, fieldName: string, hasError: boolean) => {
     if (!hasError && !value.includes('/') && value !== '') {
       setValue(fieldName, `${value}/32`);
     }
+
+    // Check for duplicate entries
+    const existingEntries: Array<{ sourceRange?: string | undefined }> =
```
fabio-silva, line 79: @Tikdev00 let's try to do this at the schema level, using Zod.
The file is at `ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration-schema.ts`

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration.tsx
```diff
@@ -68,10 +69,32 @@ export const AdvancedConfigurationForm = ({
...
+      getValues(AdvancedConfigurationFields.sourceRanges) || [];
+    const isDuplicate = existingEntries.filter(
+      (item) => item.sourceRange === value
+    );
+    // Get unique entries to check if all are unique to clear the error
+    const uniqueEntries = Array.from(
```
fabio-silva, line 85: When using Zod, if we set an error whenever the user types a duplicate, it's guaranteed all entries will be unique, so we can skip all this section ahead

fabio-silva (changes requested):

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration-schema.ts
```diff
@@ -55,6 +55,24 @@ export const advancedConfigurationsSchema = () =>
           });
         }
       });
+      const seen = new Set<string>();
+      sourceRanges.forEach(({ sourceRange }, index) => {
```
fabio-silva, line 59: There's already a `forEach` loop right above here, we can rework for our purposes, `if (sourceRange)` should be our external condition, and the other follow internally.
Also, if `seen.has(sourceRange)`, why are we doing `seen.add(sourceRange)` again?
Something to look at: we automatically add the `/32` netmask when the users blurs the output, if they don't do it themselves. I think this piece of code doesn't have that in account.

fabio-silva (changes requested):
@Tikdev00  Please check the comment. 
Also, let's add a unit test to make sure we get an error when we type duplicate entries, with and without netmask

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration-schema.ts
```diff
@@ -55,6 +55,24 @@ export const advancedConfigurationsSchema = () =>
...
+        if (sourceRange && seen.has(sourceRange)) {
```
dianabirs, line 60: I would use this approach, instead of having a set for seen 
```suggestion
         const uniqueRanges = new Set(
        Object.values(sourceRanges).map((item) => item.sourceRange)
      );
      if (uniqueRanges.size !== sourceRanges.length) { ...
```

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration.tsx
```diff
@@ -40,7 +40,8 @@ export const AdvancedConfigurationForm = ({
   dbType,
   loadingDefaultsForEdition,
 }: AdvancedConfigurationFormProps) => {
-  const { watch, setValue, getFieldState } = useFormContext();
+  const { watch, setValue, getFieldState, getValues, setError, clearErrors } =
```
dianabirs, line 43: These are not used
```suggestion
  const { watch, setValue, getFieldState} =
```

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration-schema.ts
dianabirs: @fabio-silva for blur we are setting field value to append `/32` so the `sourceRanges` list should have all the correct values in it, with the netmask as well

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration.tsx
```diff
@@ -68,6 +68,7 @@ export const AdvancedConfigurationForm = ({
       );
     }
   }, [clusterInfo]);
+
   const handleBlur = (value: string, fieldName: string, hasError: boolean) => {
     if (!hasError && !value.includes('/') && value !== '') {
       setValue(fieldName, `${value}/32`);
```
fabio-silva, line 74: @Tikdev00 please add `{ shouldValidate: true }` as third argument here, or IP addresses without netmask won't fail the validation (e.g. "192.168.1.1" when there's already "192.168.1.1/32"

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration-schema.ts
```diff
@@ -42,36 +42,38 @@ export const advancedConfigurationsSchema = () =>
...
             message: Messages.errors.sourceRange.invalid,
           });
         }
-      });
-      const seen = new Set<string>();
-      sourceRanges.forEach(({ sourceRange }, index) => {
-        if (sourceRange && seen.has(sourceRange)) {
-          seen.add(sourceRange);
+
+        const nonEmptyRanges = sourceRanges
```
fabio-silva: this should not be done in the loop, as it does not depend on the current element.
Same for `uniqueRanges`

fabio-silva (changes requested):

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration-schema.ts
Tikdev00: @fabio-silva yes you are right, it's already fixed

fabio-silva (approved):

ui/apps/everest/src/components/cluster-form/advanced-configuration/advanced-configuration-schema.ts
```diff
@@ -41,18 +41,48 @@ export const advancedConfigurationsSchema = () =>
...
+              path: [
+                AdvancedConfigurationFields.sourceRanges,
+                index,
+                'sourceRange',
+              ],
+              message: Messages.errors.sourceRange.invalid,
+            });
+          }
+
+          const isCurrentOneOfDuplicates = nonEmptyRanges.filter(
```
dianabirs, line 67: I think that by its name, this variable should reflect if current IP is one of duplicates. Right now, it reflects the length of the filter. Can we add `> 1` condition here, so it's more readable?

Tikdev00: @dianabirs done

PR #1221, opened 2025-03-17
Add possibility to set Go build ENV vars to Tiltfile

maxkondr:
in addition the following improvements are introduced:
- New ENV var EVEREST_OPERATOR_BUILD_ENV for setting Go build env vars for Everest operator
- New ENV var EVEREST_BACKEND_BUILD_ENV for setting Go build env vars for Everest server backend
- New ENV var K8S_CONTEXT for setting K8S context to use during deployment - used in case of remote K8S cluster
- New ENV var DOCKER_REGISTRY_URL for setting docker registry URL where new docker images shall be pushed - used in case of remote K8S cluster
- optimize source code file watchers (tuned patterns for watching and ignores) - now rebuild is executed for appropriate part (operator, FE, BE).
- rename generated file (created by `go::generate` directive) - now they have suffix `.gen.go`


dev/Tiltfile
```diff
@@ -19,47 +19,64 @@ if os.path.exists(local_env_file):
...
+
+# Check for Everest server backend Go build options
+everest_backend_build_env = os.getenv('EVEREST_BACKEND_BUILD_ENV', go_default_build_env)
+print('Using Everest server backend build ENV vars: {0}'.format(everest_backend_build_env))
+
+# External resources set up
+k8s_context = os.getenv('K8S_CONTEXT', '')
+if k8s_context:
+    print('Using Kubernetes context: {0}'.format(k8s_context))
+    allow_k8s_contexts(k8s_context)
```
oksana-grishchenko, line 73: Can we please allow a list of contexts? I often have two and modify the Tiltfile to keep them both like
```
    allow_k8s_contexts("context1")
    allow_k8s_contexts("context2")
```
it would be great if it could be set in environment and no edits to the Tiltfile were needed anymore🙏

maxkondr: this is my idea - define everything we need in `.env` or `config.yaml` and do not edit Tiltfile.

Please provide more details about your use-case when >1 contexts are required. Anyway the deployment will be performed into one K8S cluster. What is the reason for defining a list of contexts?

I see it as the following:
- you deploy into one cluster
- once you need to deploy into another one - you just change (or uncomment another ENV var) `.env` with appropriate context and deploy into another K8S cluster. Or I missed smth?

oksana-grishchenko: You're right, the problem is too specific, it would be better to handle it by commenting/uncommenting lines in .env, feel free to resolve the comment.

Here is my use case:
A local k8s (kind, k3d, minikube) can't handle many-nodes clusters, such clusters just don't come up because of lack of the resources. So I use GKE which means I need to set the context. Then sometimes tilt deployment got screwed (e.g., one of the components starts to fail after many redeployments). I can't cleanup the k8s cluster from a Everest-tilt installation because the version is something like tilt-b07eed973aa620e2, so usually I just delete the old cluster and create another one. This another k8s cluster should have a different name while the old one is being deleted (I interchangeable use `oxana-everest` and `oxana-everest-2`). That's why I keep two contexts in the Tiltfile😅

oksana-grishchenko (approved):
🚀

One note: there were no changes regarding victoriamertics in this PR, however just as a heads-up: I was able to run Tilt only after removing the lines
```
        "victoriametrics\\:admin:clusterrole",
        "victoriametrics\\:view:clusterrole",
```
  - the resources weren't found:
  ```
No object identified by the fragment "victoriametrics\\:admin:clusterrole" could be found. Possible objects are: "everest-server:Deployment:everest-system",...
```

dev/Tiltfile
maxkondr: in fact Tilt has command `tilt down` that cleans up resources in cluster. But it will remove the resources it tracks only (i.e. defined in Tiltfile only), it will not delete DB clusters, backups, etc that were created from UI/API...

So the only way I see right now - recreate K8S cluster and switch context :(

PR #1275, opened 2025-04-03
EVEREST-1975 | [bugfix] DBengine min version preflight checks

mayankshah1607:
[EVEREST-1975](https://jira.percona.com/browse/EVEREST-1975) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

Due to a bug in how we use the version library, PG DB versions are returned in an incorrect form. For example, `13.16` is parsed and returned as `13.16.0`, which is a non-existent version. This causes issues when trying to upgrade PG version during operator upgrades.

[EVEREST-1975]: https://perconadev.atlassian.net/browse/EVEREST-1975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/server/handlers/k8s/database_engine.go
+++ b/internal/server/handlers/k8s/database_engine.go
@@ -308,7 +308,7 @@ func preflightCheckDBEngineVersion(
 	if err != nil {
 		return false, "", err
 	}
-	return currentVersion.GreaterThanOrEqual(minVersion), minVersion.String(), nil
+	return currentVersion.GreaterThanOrEqual(minVersion), minVersion.Original(), nil
 }
 
 func (h *k8sHandler) getDBPostUpgradeTasks(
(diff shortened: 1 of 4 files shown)


internal/server/handlers/k8s/database_engine.go
```diff
@@ -308,7 +308,7 @@ func preflightCheckDBEngineVersion(
 	if err != nil {
 		return false, "", err
 	}
-	return currentVersion.GreaterThanOrEqual(minVersion), minVersion.String(), nil
+	return currentVersion.GreaterThanOrEqual(minVersion), minVersion.Original(), nil
```
mayankshah1607, line 311: `minVersion` itself is parsed correctly for comparison and internal use, but we need to return back the original form, not the internal representation used by the libaray

oksana-grishchenko (approved):

maxkondr (approved):

PR #1293, opened 2025-04-08
EVEREST-1925 Blocklisting mechanism 

oksana-grishchenko:
[EVEREST-1925](https://jira.percona.com/browse/EVEREST-1925) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-1925]: https://perconadev.atlassian.net/browse/EVEREST-1925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**Problem:** 
When user logs out, their token remains valid until its natural expiration time specified in the token itself. For the built-in Everest authentication the token lifetime is 24h which introduces security concerns even though the built-in authentication is not recommended for the production use. 

**Solution:**
Implement the blocklisting mechanism according to the [proposal](https://www.notion.so/percona/Research-Session-management-1ab674d091f38010989df39028d1b110?pvs=4#1b2674d091f3809296bdf495196afa9f) 

This PR does:
- Add the new `DELETE /session` endpoint, which adds the current token information to the blacklist
- On each request, check the blacklist and reject the request if the token is blacklisted. The check uses cashed data, so no overhead is added in terms of additional k8s requests. 
- Add mechanism to cleanup the blacklist - each time a new value needs to be blocklisted we cleanup outdated data from the list
- Add mechanism to cache the blacklist. Using a separate controller-runtime client which can cache the secret for the blocklist. 
- Add rate limit for the `DELETE /session` endpoint. The rate limiting rule which was applied to the `POST /session endpoint` is now applied to the `DELETE /session` as well, for more details see https://github.com/percona/everest/pull/530


docs/spec/openapi.yml
```diff
@@ -59,6 +59,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+        '429':
+          description: Too many attempts
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
oksana-grishchenko, line 67: 429 status was there before, but it wasn't reflected in the API spec

pkg/session/content_processor.go
```diff
@@ -0,0 +1,102 @@
...
+func newContentProcessor(secret *corev1.Secret) ContentProcessor {
+	return &contentProcessor{
+		cachedSecret: secret,
+	}
+}
+
+// Block adds the token to the blacklist, cleans up the outdated data from the secret.
+// The method returns the updated secret and a bool that indicates if the attempt was successful.
+func (d *contentProcessor) Block(l *zap.SugaredLogger, secret *corev1.Secret, shortenedToken string) (*corev1.Secret, bool) {
+	if d.mutex.TryLock() {
```
mayankshah1607, line 43: Why not just use `d.mutex.Lock()` and let the caller block?

pkg/session/content_processor.go
```diff
@@ -0,0 +1,102 @@
...
+		defer d.mutex.Unlock()
+		updatedSecret := addDataToSecret(l, secret, shortenedToken, time.Now().UTC())
+		return updatedSecret, true
+	}
+	return secret, false
+}
+
+// IsBlocked returns true if the given shortened token is in the blocklist.
+func (d *contentProcessor) IsBlocked(shortenedToken string) bool {
```
mayankshah1607, line 52: Should we use a RLock here?

mayankshah1607:
Overall, I think this approach is great. I do have a couple of suggestions:


1. Right now, `Blocklist` is a stateful object. What if we made it stateless and pushed state management behind an interface? That way, we could easily swap out the underlying storage mechanism in the future without touching the blocklist logic.

Here’s what I mean:

```go
type TokenStore interface {
	Add(token string) error
	Delete(token string) error
	Exists(token string) (bool, error)
}

type blocklist struct {
	l          *zap.SugaredLogger
	tokenStore TokenStore
}
```

And then the logic becomes simpler and more focused:

```go
// Block invalidates the token from the context by adding it to the blocklist.
func (b *blocklist) Block(ctx context.Context) error {
	token, err := extractToken(ctx)
	if err != nil {
		return err
	}
	shortenedToken, err := shortenToken(token)
	if err != nil {
		return err
	}

	for attempts := 0; attempts < maxRetries; attempts++ {
		if err := b.tokenStore.Add(shortenedToken); err != nil {
			b.l.Errorf("failed to add token to store: %v", err)
			return err
		}
		return nil
	}
	return fmt.Errorf("failed to block token after %d attempts", maxRetries)
}
```

With this setup, the `blocklist` no longer deals with the underlying storage (or secrets) directly. It just delegates to `TokenStore`. This opens up some nice benefits:

- Easy to plug in different storage backends later (e.g., Redis, etcd, etc.) by just implementing `TokenStore`
- Easier testing with an in-memory or mock implementation
- Clear separation of concerns


2. It looks like `ContentProcessor` is manually caching `Secret` objects. But, if we use a `controller-runtime` client, it comes with built-in caching out of the box. You can likely remove the custom caching logic and rely on that instead — less code to maintain.

What do you think? :)

pkg/session/content_processor.go
oksana-grishchenko: A good point, thanks, changed the Mutex to RWMutex so the locks could be separated.

oksana-grishchenko:
@mayankshah1607, thanks for the suggestions! 

1. Considering that there is no need for caching anymore, such interface is much cleaner indeed, I love it! I reworked your suggestion a bit but kept the essentials. Please review. 

2. Since the PR with the controller-runtime client is merged let's use the controller-runtime cache. However the caching was not enabled, so I added a new way of creating an in-cluster k8s client (the one with cache) and using it for Everest API server. To be sure it works as expected, I compared the k8s audit logs for a version with cache and without it - there is less requests indeed.

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
...
+
+type tokenStore struct {
+	kubeClient kubernetes.KubernetesConnector
+	l          *zap.SugaredLogger
+	mutex      sync.RWMutex
+}
+
+func (ts *tokenStore) Add(ctx context.Context, shortenedToken string) error {
+	ts.mutex.Lock()
+	defer ts.mutex.Unlock()
```
mayankshah1607, line 53: Do we actually need locking here (since we're not manually maintaining a cache anymore)?

The Kube API provides atomic updates to resources using resource versioning, so we already have protection from concurrent updates.. And we also already have retries on the caller side

pkg/session/token_store_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestCleanupOld(t *testing.T) {
```
mayankshah1607, line 15: In addition to these, I think it would be nice to have tests for our token store implementation as well. We can create a Kube connector using a fake controller-runtime client. Example here - https://github.com/percona/everest/blob/main/pkg/kubernetes/accounts_test.go#L46

pkg/kubernetes/kubernetes.go
```diff
@@ -148,6 +147,42 @@ func NewInCluster(l *zap.SugaredLogger) (KubernetesConnector, error) {
 	}, nil
 }
 
+// NewInClusterWithCache creates a new kubernetes client using incluster authentication with cache enabled
+func NewInClusterWithCache(l *zap.SugaredLogger, ctx context.Context) (KubernetesConnector, error) {
+	restConfig := inClusterRestConfig()
+	cacheOptions := cache.Options{
+		Scheme: CreateScheme(),
```
maxkondr, line 154: em, I assume this is not definitely what you wanted to achieve.  The cache will be used for ALL resources that will be touched by kubernetesConnector at least once. More over it will call such resources from ALL K8S cluster that may leads to huge memory consumption and OOM problems.

This is exactly what we were discussing in https://github.com/percona/everest/pull/1132#issuecomment-2689884499

maxkondr (changes requested):

pkg/session/blocklist.go
```diff
@@ -0,0 +1,108 @@
...
+	// IsBlocked checks if the token from the context is blocked.
+	IsBlocked(ctx context.Context) (bool, error)
+}
+
+type blocklist struct {
+	tokenStore TokenStore
+	l          *zap.SugaredLogger
+}
+
+type TokenStore interface {
```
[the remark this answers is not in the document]
mayankshah1607: The idea was to have de-coupling between the blocklist logic and the storage layer, see this comment - https://github.com/percona/everest/pull/1293#issuecomment-2815241150

pkg/session/token_store.go
oksana-grishchenko: right, seems like the mutex not needed anymore. The resouce version protects from concurent updates and in case the update fails we have this retry mechanism. removed, thanks!

pkg/session/content_processor.go
oksana-grishchenko: Resolving the thread, mutex is not used anymore, see https://github.com/percona/everest/pull/1293#discussion_r2058204411

docs/spec/openapi.yml
```diff
@@ -72,6 +78,28 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/UserCredentials'
+    delete:
+      tags:
+        - Authentication & Authorization
+      summary: Everest UI Logout
```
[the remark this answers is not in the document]
oksana-grishchenko: a good Idea, I'll create an overhead ticket for this

pkg/kubernetes/kubernetes.go
```diff
@@ -170,6 +205,16 @@ func getKubernetesClientOptions() ctrlclient.Options {
 	}
 }
 
+func getKubernetesClientOptionsWithCache(k8sCache cache.Cache) ctrlclient.Options {
```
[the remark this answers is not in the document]
oksana-grishchenko: Not relevant anymore since the controller-runtime cache is not used anymore

pkg/kubernetes/kubernetes.go
```diff
@@ -148,6 +147,42 @@ func NewInCluster(l *zap.SugaredLogger) (KubernetesConnector, error) {
 	}, nil
 }
 
+// NewInClusterWithCache creates a new kubernetes client using incluster authentication with cache enabled
+func NewInClusterWithCache(l *zap.SugaredLogger, ctx context.Context) (KubernetesConnector, error) {
```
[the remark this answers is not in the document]
oksana-grishchenko: Not relevant anymore since the controller-runtime cache is not used anymore

pkg/session/blocklist.go
```diff
@@ -0,0 +1,108 @@
+package session
```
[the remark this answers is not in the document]
oksana-grishchenko: done, thanks

pkg/session/blocklist.go
```diff
@@ -0,0 +1,108 @@
...
+
+const (
+	dataKey    = "list"
+	maxRetries = 10
+)
+
+// Blocklist represents interface to block JWT tokens and check if a token is blocked.
+type Blocklist interface {
+	// Block invalidates the token from the context by adding it to blocklist.
+	Block(ctx context.Context) error
```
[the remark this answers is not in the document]
oksana-grishchenko: I see. There are two points which made me choose to pass the context:
- we still need the context inside because of the k8s API calls
- since we still need the whole context, there is no point in extracting the token outside, so we can hide the extractToken call behind the interface "

pkg/session/blocklist.go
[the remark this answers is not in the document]
oksana-grishchenko: I think we do. As Mayank mentioned, it's beneficial in terms of abstract away where and how do we store the data

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
+package session
```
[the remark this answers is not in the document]
oksana-grishchenko: done, thanks

.github/workflows/dev-be-ci.yaml
```diff
@@ -349,8 +349,10 @@ jobs:
         run: |
           kubectl port-forward --namespace everest-system deployment/everest-server 8080:8080 &
 
-      - name: Create Everest test user
+      - name: Create Everest test users
         run: |
+          ./bin/everestctl accounts create -u test -p password
+          echo "API_TOKEN_TEST=$(curl --location -s 'localhost:8080/v1/session' --header 'Content-Type: application/json' --data '{"username": "test","password": "password"}' | jq -r .token)" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
oksana-grishchenko: a good Idea, I'll create an overhead ticket for this

docs/spec/openapi.yml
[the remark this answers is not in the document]
oksana-grishchenko: right, changed the summary for the /session endpoints from Everest UI -> Everest API because it describes the actual job more accurately. Wdyt?

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
...
+	store := &tokenStore{
+		l:          logger,
+		kubeClient: kubeClient,
+	}
+	_, err := kubeClient.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestBlocklistSecretName})
+	if err == nil {
+		return store, nil
+	}
+	if !k8serrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get secret: %w", err)
```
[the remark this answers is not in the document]
oksana-grishchenko: Updated, thanks

pkg/session/blocklist.go
```diff
@@ -0,0 +1,108 @@
...
+
+func extractToken(ctx context.Context) (*jwt.Token, error) {
+	token, ok := ctx.Value(common.UserCtxKey).(*jwt.Token)
+	if !ok {
+		return nil, fmt.Errorf("failed to get token from context")
+	}
+	return token, nil
+}
+
+func blockListSecretTemplate(stringData string) *corev1.Secret {
```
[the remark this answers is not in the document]
oksana-grishchenko: I see, I meant this method to be like a getter - according to Go best practices https://go.dev/doc/effective_go#Getters there is no need to put the get verb into the method name.  But I agree this name is confusing because 
- the name already starts with a verb ("block") 
- the name does not end with what the method returns so we could assume a getter 

Renamed to `getBlockListSecretTemplate`. Wdyt?

pkg/session/blocklist.go
```diff
@@ -0,0 +1,108 @@
...
+	return &corev1.Secret{
```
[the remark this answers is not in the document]
oksana-grishchenko: good catch, thanks, it got messed up during refactoring. Updated.

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
...
+	mutex      sync.RWMutex
+}
+
+func (ts *tokenStore) Add(ctx context.Context, shortenedToken string) error {
+	ts.mutex.Lock()
+	defer ts.mutex.Unlock()
+
+	secret, err := ts.kubeClient.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestBlocklistSecretName})
+	if err != nil {
+		ts.l.Errorf("failed to get %s secret: %v", common.EverestBlocklistSecretName, err)
```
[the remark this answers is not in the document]
oksana-grishchenko: Updated

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
...
+		return err
+	}
+
+	addDataToSecret(ts.l, secret, shortenedToken, time.Now().UTC())
+	_, updateErr := ts.kubeClient.UpdateSecret(ctx, secret)
+	if updateErr != nil {
+		ts.l.Errorf("failed to update %s secret, retrying: %v", common.EverestBlocklistSecretName, updateErr)
```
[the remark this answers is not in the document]
oksana-grishchenko: Updated

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
...
+func cleanupOld(l *zap.SugaredLogger, list string, now time.Time) []string {
+	tokens := strings.Split(list, sep)
+	newList := make([]string, 0, len(tokens))
+	for _, shortenedToken := range tokens {
+		length := len(shortenedToken)
+		if length < timestampLen {
+			l.Info("blocklist contains irregular data format")
+			continue
+		}
+		ts := shortenedToken[length-10 : length]
```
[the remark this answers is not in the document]
oksana-grishchenko: true, but the json format is verbose while we have this 1Mb size limitation for secrets, so we store only the necessary symbols

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
...
+		return secret
+	}
+	list := append(cleanupOld(l, string(existingList), now), shortenedToken)
+	secret.StringData = map[string]string{
+		dataKey: strings.Join(list, sep),
+	}
+	return secret
+}
+
+func cleanupOld(l *zap.SugaredLogger, list string, now time.Time) []string {
```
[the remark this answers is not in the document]
oksana-grishchenko: I would keep passing it as a parameter, it's like a dependency injection - the `time.Now()` adds the non-deterministic behaviour which makes it difficult to cover the function with unit-tests

pkg/kubernetes/kubernetes.go
oksana-grishchenko: Thanks Maxim for sharing the link. As I see there were some doubts about caching general and we specifically didn't want to cache secrets because it could lead to overwhelmed memory in the big clusters. So I decided not to use the the controller-runtime cache but move back to my initial idea about the locally cached secret. 

@mayankshah1607 @maxkondr wdyt?

.github/workflows/dev-be-ci.yaml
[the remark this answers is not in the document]
oksana-grishchenko: here is the ticket https://perconadev.atlassian.net/browse/EVEREST-2016

pkg/kubernetes/kubernetes.go
mayankshah1607: I believe we wanted to disable caching for the client used by the API methods because of the large memory foot-print.

But controller-runtime allows you to configure the cache so that it will watch only specified resource kinds (you can further restrict it by providing namespaces, label filters, etc.). Can't we use that? Then we can also get rid of the `KubeConnector` here and directly use the controller-runtime client methods. This will actually result in lesser code too

mayankshah1607: Yes, but that is in the context of using it for the API requests (unless I'm missing something) ...

It is possible to create a cache that will watch just this one Secret and nothing else. In this case we will not have any memory problems, it is the same as maintaining our own cache logic (performance wise)

For example, here is how you can do it:
```go
	// cache that will watch ONLY the blacklist secret.
	secretCache, err := cache.New(nil, cache.Options{
		ByObject: map[ctrlclient.Object]cache.ByObject{
			&corev1.Secret{}: {
				Namespaces: map[string]cache.Config{
					"everest-system": {},
				},

				Label: labels.SelectorFromSet(map[string]string{
					"everest-blacklist-secret-key": "",
				}),
			},
		},
	})
	if err != nil {
		// ... handle error
	}

	// create client using this cache.
	client, err := ctrlclient.New(nil, ctrlclient.Options{
		Cache: &ctrlclient.CacheOptions{
			Reader: secretCache,
		},
	})
	if err != nil {
		// ... handle error
	}
```

Also, by using the client directly, we will not have to go via the `KubeConnector` route, and the token store can directly use the controller-runtime client methods.. Let me know if I'm missing any point 😅

mayankshah1607: That informer actually serves a different use-case. We are using it in RBAC implementation to live-reload the casbin object. So the idea is that the informer watches the ConfigMap, and calls internal code to reload casbin. See the usage here: https://github.com/percona/everest/blob/main/pkg/rbac/rbac.go#L92

I don't think we can actually replace this logic with a controller-runtime cache -- it simply acts as a cache, whereas here we need something to tell us about changes in-real time and execute callbacks.

But I still don't understand why we cannot use a separate cache (like showed in the above example) for the blocklist use-case. As far as I understand, Max's original comment was about enabling a cluster-wide cache. But we can overcome it by setting up a restricted cache. 😅

pkg/session/token_store.go
```diff
@@ -0,0 +1,152 @@
...
+	if ts.secret == nil {
+		secret, err := ts.kubeClient.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestBlocklistSecretName})
+		if err != nil {
+			ts.l.Errorf("failed to get %s secret in the %s namespace: %v", common.EverestBlocklistSecretName, common.SystemNamespace, err)
+			return false, err
+		}
+		ts.secret = secret
+	}
+	list, ok := ts.secret.Data[dataKey]
+	return ok && strings.Contains(string(list), shortenedToken), nil
```
maxkondr, line 96: do not like this place...
the issue here - this method will be called on each single incoming request and search over string is not performance optimal at all. 

We may consider using bloom filter instead.

pkg/session/blocklist.go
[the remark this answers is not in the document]
maxkondr: ok

pkg/session/token_store.go
```diff
@@ -0,0 +1,152 @@
...
+	secret, err := ts.kubeClient.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestBlocklistSecretName})
+	if err != nil {
+		ts.l.Errorf("failed to get %s secret in the %s namespace: %v", common.EverestBlocklistSecretName, common.SystemNamespace, err)
+		return err
+	}
+
+	secret = addDataToSecret(ts.l, secret, shortenedToken, time.Now().UTC())
+	updatedSecret, updateErr := ts.kubeClient.UpdateSecret(ctx, secret)
+	if updateErr != nil {
+		ts.l.Errorf("failed to update %s secret in the %s namespace, retrying: %v", secret.Name, secret.Namespace, updateErr)
```
maxkondr, line 78: still there is no info about the token that has problems with adding into secret

pkg/kubernetes/kubernetes.go
maxkondr: > I believe we wanted to disable caching for the client used by the API methods because of the large memory foot-print.


I thought we decided to disable it because we didn't know for sure how to use this new feature and for us it would be an overkill to replace REST-based client with controller-runtime + enable caching at the same time.

> As far as I understand, Max's original comment was about enabling a cluster-wide cache.

Mayank is right here. 

In general I support the idea to enable caching for Everest resources only. The only one thing that concerns me here - Mayank suggests using a separate k8S client with enabled cache instead of `kubernetesConnector` - it will create a confusion "why are there different ways used for interacting with Kubernetes?".  

And I do think that we need to enable caching for blockList due to this https://github.com/percona/everest/pull/1293#discussion_r2058973567 - we need to minimise the interaction with KubeAPI, request rate is getting higher and higher, one day we will hit K8S API rate limits. The `blockList.Exist` may increase it significantly.

So, instead of introducing a separate kubeClient I would suggest the following:

- enable caching in `kubernetesConnector` for resources in all namespaces that have our cache label (for example `everest-cache-enabled: true`).
- in this PR set this label on blockList Secret only.
As result only one blockList Secret will be cached.

Later in separate tickets/PRs:
- add this cache label to other resource types that we want to cache one by one.

P.S.
as soon as we already have controller-runtime, I would recheck our RBAC informer implementation - whether it is possible to add our own hooks on Informer events that is used under the hood of `controller-runtime.cache`. So that we could just enable caching of this Secret with RBAC, use common `kubernetesConnector` and get rid of our Informer implementation. but this is a completely different story...

oksana-grishchenko: As discussed in Slack, the `ByObject` option fails the requests which does not fit the filter, so it was decided to use a separate client for the blocklist and enable cache in it. Another benefit of using the client in compare with the in-memory cache is HA compatibility

pkg/session/token_store.go
```diff
@@ -0,0 +1,120 @@
...
+	return secret
+}
+
+func cleanupOld(l *zap.SugaredLogger, list string, now time.Time) []string {
+	tokens := strings.Split(list, sep)
+	newList := make([]string, 0, len(tokens))
+	for _, shortenedToken := range tokens {
+		length := len(shortenedToken)
+		if length < timestampLen {
+			l.Info("blocklist contains irregular data format")
```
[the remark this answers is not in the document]
oksana-grishchenko: A fair point, added a comment to the `shortenToken` function, thanks. The initial request "please write somewhere" was a bit too broad, sorry, I did not get the idea to have this information in the codebase

pkg/session/blocklist.go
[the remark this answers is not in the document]
oksana-grishchenko: Ok, I agree that assuming additional content in context complicates things. Updated.

internal/server/everest.go
```diff
@@ -110,13 +112,19 @@ func NewEverestServer(ctx context.Context, c *config.EverestConfig, l *zap.Sugar
...
+	}
+
 	e := &EverestServer{
 		config:        c,
 		l:             l,
 		echo:          echoServer,
 		kubeConnector: kubeConnector,
 		sessionMgr:    sessMgr,
 		attemptsStore: store,
+		blocklist:     blockList,
```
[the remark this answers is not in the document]
oksana-grishchenko: Agree, it can be a part of the session manager. Updated, thanks

pkg/session/token_store.go
```diff
@@ -0,0 +1,152 @@
...
+	store := &tokenStore{
+		l:          logger,
+		kubeClient: kubeClient,
+	}
+	_, err := kubeClient.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestBlocklistSecretName})
+	if err == nil {
+		return store, nil
+	}
+	if !k8serrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get %s secret in the %s namespace: %w", common.EverestBlocklistSecretName, common.SystemNamespace, err)
```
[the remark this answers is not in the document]
oksana-grishchenko: updated, thanks

internal/server/session.go
```diff
@@ -66,6 +66,20 @@ func (e *EverestServer) CreateSession(ctx echo.Context) error {
...
 }
 
+// DeleteSession invalidates the user token by adding it to the blocklist
+func (e *EverestServer) DeleteSession(ctx echo.Context) error {
+	e.attemptsStore.IncreaseTimeout(ctx.RealIP())
+	err := e.blocklist.Block(ctx.Request().Context())
+	if err != nil {
+		e.l.Errorf("blocklist error: %v", err)
+		return ctx.JSON(http.StatusInternalServerError, api.Error{
+			Message: pointer.To("Failed to logout user"),
```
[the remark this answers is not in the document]
oksana-grishchenko: If an error appeared t's either we couldn't get the secret or we couldn't update the secret, both mean the token is still valid and not in the blocklist, so user can retry. So I would keep the error

pkg/session/token_store.go
oksana-grishchenko: Created a benchmark that measures how long does it take in average to search a non-present substring in a 1mb size string, its `453438 ns/op` which means `0.453438 ms`. Do you think we need to optimize it?  

Included the benchmarks `BenchmarkStringsContains`, `BenchmarkStringsContainsNotFound` to the storage tests file

pkg/session/token_store.go
oksana-grishchenko: updated, thanks

pkg/cli/upgrade/upgrade.go
```diff
@@ -102,7 +102,7 @@ func NewUpgrade(cfg *Config, l *zap.SugaredLogger) (*Upgrade, error) {
 
 	var kubeClient kubernetes.KubernetesConnector
 	if cfg.InCluster {
-		k, err := kubernetes.NewInCluster(cli.l)
+		k, err := kubernetes.NewInCluster(cli.l, context.Background(), nil)
```
maxkondr, line 105: this is an issue - because when user terminates the process via `Esc` or `Ctrl+C` the CLI context is being canceled as well, but cache will continue running.

pkg/session/blocklist.go
```diff
@@ -0,0 +1,163 @@
...
+	// Add adds the shortened token to the blocklist
+	Add(ctx context.Context, shortenedToken string) error
+	// Exists checks if the shortened token is in the blocklist
+	Exists(ctx context.Context, shortenedToken string) (bool, error)
+}
+
+// BlocklistClient supports only the k8s API methods that are needed for blocklist management.
+// A separate client is needed to apply the controller-runtime cache only to the related objects.
+// Using the controller-runtime client is also beneficial because it supports HA mode.
+type BlocklistClient interface {
```
maxkondr, line 71: Secrets are part of TokenStore and are required in it's context only, blockList knows nothing about Secrets, it operates with TokenStore only.
it seems this is more `TokenStoreSecretClient` rathe then `BlocklistClient`

pkg/session/blocklist.go
```diff
@@ -0,0 +1,163 @@
...
+	// GetSecret returns a secret that matches the criteria.
+	GetSecret(ctx context.Context, key client.ObjectKey) (*corev1.Secret, error)
+	// CreateSecret creates a secret.
+	CreateSecret(ctx context.Context, secret *corev1.Secret) (*corev1.Secret, error)
+	// UpdateSecret updates a secret.
+	UpdateSecret(ctx context.Context, secret *corev1.Secret) (*corev1.Secret, error)
+}
+
+// NewBlocklist creates a new block list
+func NewBlocklist(ctx context.Context, bc BlocklistClient, logger *zap.SugaredLogger) (Blocklist, error) {
```
mayankshah1607, line 81: I feel that there's a bit of abstraction leak here ..

The idea behind having a `TokenStore` interface was to have loose coupling between blocklist logic and the storage code. But by having to specify a `BlocklistClient` (which is an implementation detail of the `TokenStore`) while initialising `Blocklist`, we're introducing close-coupling again. 

Currently we're initialising the `blocklistClient` object in the call to `NewEverestServer`, I think that we should move that into `newTokenStore` and remove the `BlockListClient` arg from here

pkg/session/blocklist.go
```diff
@@ -0,0 +1,163 @@
...
+// A separate client is needed to apply the controller-runtime cache only to the related objects.
+// Using the controller-runtime client is also beneficial because it supports HA mode.
+type BlocklistClient interface {
+	// GetSecret returns a secret that matches the criteria.
+	GetSecret(ctx context.Context, key client.ObjectKey) (*corev1.Secret, error)
+	// CreateSecret creates a secret.
+	CreateSecret(ctx context.Context, secret *corev1.Secret) (*corev1.Secret, error)
+	// UpdateSecret updates a secret.
+	UpdateSecret(ctx context.Context, secret *corev1.Secret) (*corev1.Secret, error)
+}
```
mayankshah1607, line 78: We should move this to `token_store.go` as well since this is an implementation detail of the token store, not the blocklist. 

Also, should we name it `SecretClient` instead?

pkg/session/token_store.go
```diff
@@ -0,0 +1,159 @@
...
+	secret, err := ts.client.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestBlocklistSecretName})
+	if err != nil {
+		ts.l.Errorf("failed to get %s secret in the %s namespace: %v", common.EverestBlocklistSecretName, common.SystemNamespace, err)
+		return err
+	}
+
+	secret = addDataToSecret(ts.l, secret, shortenedToken, time.Now().UTC())
+	_, updateErr := ts.client.UpdateSecret(ctx, secret)
+	if updateErr != nil {
+		ts.l.Errorf("failed to update %s secret in the %s namespace withe the %s shortened token, retrying: %v", secret.Name, secret.Namespace, shortenedToken, updateErr)
```
maxkondr, line 88:
```suggestion
		ts.l.Errorf("failed to update %s secret in the %s namespace with the %s shortened token, retrying: %v", secret.Name, secret.Namespace, shortenedToken, updateErr)
```

pkg/session/token_store.go
```diff
@@ -0,0 +1,159 @@
...
+	return secret
+}
+
+func cleanupOld(l *zap.SugaredLogger, list string, now time.Time) []string {
+	tokens := strings.Split(list, sep)
+	newList := make([]string, 0, len(tokens))
+	for _, shortenedToken := range tokens {
+		length := len(shortenedToken)
+		if length < timestampLen {
+			l.Warn("blocklist contains irregular data format")
```
maxkondr, line 127:
```suggestion
			l.Warnf("blocklist token='%s' contains irregular data format", shortenedToken)
```

pkg/session/blocklist.go
oksana-grishchenko: My first idea was also to hide the client creation inside the token store, however I don't think it's a right thing to do anymore because:
1. the k8s calls are an external dependency. External dependencies should be injected to be able to cover the functionality with unit-tests which mock that external interface, otherwise we tightly couple the token store implementation with the context - it would mean we assume there is the k8s API available. 
2. the BlocklistClient interface is just a subset of methods from kubeConnector. we can pass the whole kubeConnector or another client to it, it's interchangeable, we only define that whatever client you use, the blocklist needs it to define this particular subset of methods to communicate with the external dependency - the k8s API.

pkg/session/blocklist.go
oksana-grishchenko: > We should move this to token_store.go
I disagree, see https://github.com/percona/everest/pull/1293#discussion_r2063187037 

> Also, should we name it SecretClient instead?
what if for another implementation of token storage we would need another subset of methods to communicate with the k8s API? The `BlocklistClient` defines the list of whatever external methods we need to make the current implementation of the Blocklist work. Currently in includes only secrets, but we may need something else to be included and then the name `SecretClient` will get stale.

pkg/session/blocklist.go
mayankshah1607: But both arguments then assume that every token store is a k8s dependency which defeats the purpose of defining a `TokenStore` interface in the first place.

1. For the `Blocklist`, the `TokenStore` should be the dependency, not `k8s`. And for the `TokenStore`, k8s should be the dependency. The purpose of `TokenStore` was to make `Blocklist` completely unaware of k8s. Also, since `Blocklist` is itself an interface, we can mock it (or create a stub implementation using an in-memory may). So the `Blocklist` should test the `TokenStore`, and the `TokenStore` should test k8s. Currently it seems we are mixing everything in `Blocklist`.
2. Yes, I like the idea of having a `BlocklistInterface` method, but again, it should not expose itself in the `Blocklist`. The `Blocklist` talks to the `TokenStore`, and the `TokenStore` talks to `BlocklistClient` (or k8s).. That's how I see the separation of concerns in this architecture. In the future if I need to add non-k8s storage for the tokens, how would I do it through the current implementation?

pkg/session/blocklist.go
mayankshah1607: > what if for another implementation of token storage we would need another subset of methods to communicate with the k8s API?

Then it needs to be coupled with the implementation of `TokenStore`, not `Blocklist`.. the current implementation assumes that a `TokenStore` will always be a `k8s` dependency, which defeats the purpose of introducing a `TokenStore` interface ...

pkg/session/blocklist.go
oksana-grishchenko: The `BlocklistClient` defines the list of whatever external methods we need to make the current implementation of the Blocklist work. Otherwise we would assume that our global context allows us to make those external requests (that there is the k8s API available). So having the `BlocklistClient` is a dependency injection that allows the blocklist operate independently from the environment

oksana-grishchenko: The list of the methods which are needed to make the Blocklist work may change with time. That's why I would prefer not specifying "Secrets" in the name

pkg/cli/upgrade/upgrade.go
oksana-grishchenko: there is no cache if no cacheOptions are specified

internal/server/everest.go
```diff
@@ -416,3 +444,31 @@ func everestErrorHandler(next echo.HTTPErrorHandler) echo.HTTPErrorHandler {
 		next(err, c)
 	}
 }
+
+func (e *EverestServer) blocklistMiddleWare() (echo.MiddlewareFunc, error) {
```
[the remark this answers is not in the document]
oksana-grishchenko: I understand your concern, but I would keep it here. The method uses this `newSkipperFunc` which is defined in the everest.go (and it makes sense since it defines the options for the echo server). The method also uses logger which is not defined for the session manager. Both can be fixed with a refactoring which will touch the parts that wasn't touched before which will cause another round of review, which I would avoid for the sake of moving things forward

pkg/session/blocklist.go
oksana-grishchenko: > For the Blocklist, the TokenStore should be the dependency, not k8s. And for the TokenStore, k8s should be the dependency.

But if the `tokenStore` is a part of the `Blocklist` and it depend on k8s, then the whole Blocklist depends on k8s. This is how dependencies work, and this is why I want to abstract this dependency away. What do I miss here?

pkg/session/blocklist.go
oksana-grishchenko: Renaming the interface to TokenStoreClient would make sense, wdyt?

pkg/session/blocklist.go
oksana-grishchenko: I think renaming the `BlockListClient` to `TokenStoreClient` would eliminate the confusion. And this is actually what the interface represents - the set of external methods to make the TokenStore work.

mayankshah1607: Imagine this as an example scenario: what if I want to add a token store that does not use k8s? For example, vault or redis. There is no k8s used here with regards to storage of tokens. Or what if I want to use `ConfigMaps` instead of `Secrets`? Will I create a new `BlocklistClient` that updates configmaps under the `UpdateSecret` method?

How would this work in the current architecture? What does the `BlocklistClient` mean in this context? (probably nothing because Redis/Vault/ConfigMaps has nothing to do with Secrets).

We introduced `TokenStore` interface so we do not have to provide k8s or Secrets knowledge to `Blocklist` -- anything that can perform `Add` and `Exists` can store tokens.

But now, `Blocklist` depends on a `BlocklistClient` -- something that directly deals with `Secrets`. So we're essentially exposing `Blocklist` to the internal storage details again, so why did we have a `TokenStore` abstraction in the first place?

What I'm suggesting is that we introduce the `BlocklistClient` coupling in the `TokenStore` implementation. Right now, it seems to me that having a `TokenStore` and `BlocklistClient` both as dependencies for `Blocklist` are contradicting each other's usefullness.

oksana-grishchenko: Ok, I got the idea. Let me rework the solution

pkg/session/blocklist.go
oksana-grishchenko: This thread https://github.com/percona/everest/pull/1293#discussion_r2063029462 changed my mind, moved the dependencies inside the session manager and made it coupled with k8s API.

pkg/cli/upgrade/upgrade.go
oksana-grishchenko: to make it less confusing, now passing nil instead of an empty context. wdyt?

internal/server/everest.go
```diff
@@ -98,7 +102,25 @@ func NewEverestServer(ctx context.Context, c *config.EverestConfig, l *zap.Sugar
...
+	}
+	blocklistClient, err := kubernetes.NewInCluster(l, ctx, options)
+	if err != nil {
+		return nil, err
+	}
+
+	blockList, err := session.NewBlocklist(ctx, blocklistClient, l)
+	if err != nil {
+		return nil, errors.Join(err, errors.New("failed to configure tokens blocklist"))
+	}
```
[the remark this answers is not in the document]
oksana-grishchenko: This thread https://github.com/percona/everest/pull/1293#discussion_r2063029462 changed my mind about injecting dependencies, moved the dependencies inside the session manager and made it coupled with k8s API.

internal/server/everest.go
```diff
@@ -98,7 +102,25 @@ func NewEverestServer(ctx context.Context, c *config.EverestConfig, l *zap.Sugar
...
 	echoServer.Use(middleware)
+
+	options := &cache.Options{
+		ByObject: map[client.Object]cache.ByObject{
+			&corev1.Secret{}: {
+				Field: fields.SelectorFromSet(fields.Set{"metadata.name": common.EverestBlocklistSecretName}),
+			},
+		},
+	}
+	blocklistClient, err := kubernetes.NewInCluster(l, ctx, options)
```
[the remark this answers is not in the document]
oksana-grishchenko: there was such a comment over the BlocklistClient interface, however I agree it would make more sense to have it where the client is actually getting create, moved the comment there, thanks

pkg/session/blocklist.go
oksana-grishchenko: This thread https://github.com/percona/everest/pull/1293#discussion_r2063029462 changed my mind about how we handle dependencies, so moved the interface into the token store. Kept the common name TokenStoreClient without mentioning the secrets, because over the time we could not use Secrets anymore.

mayankshah1607 (approved):
LGTM! 🚀 
Thanks for making all those changes!

internal/server/everest.go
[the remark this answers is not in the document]
oksana-grishchenko: since with the latest changes the logger got inside the session manager anyway, I changed my mind and moved the `blocklistMiddleWare()` function into the session manager as requested

maxkondr (approved):

PR #1313, opened 2025-04-15
POC: BDD-style integration tests using Ginkgo+Gomega

mayankshah1607:
The following environment variables need be set first:
- `INTEGRATION_TEST_API_URL`
- `INTEGRATION_TEST_API_TOKEN`
- `INTEGRATION_TEST_NAMESPACE` (defaults to using `everest`)

Run the tests using `make integration`

### DIFF
--- a/integration/backup_storage_test.go
+++ b/integration/backup_storage_test.go
@@ -1,0 +1,136 @@
+package integration_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+
+	"github.com/AlekSi/pointer"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"oras.land/oras-go/pkg/context"
+
+	"github.com/percona/everest/client"
+)
+
+var _ = Describe("Backup Storages", func() {
+	Context("of type s3", func() {
+		createdBackupStorageName := ""
+		It("can be created", func() {
+			payload := client.CreateBackupStorageJSONRequestBody{
+				Type:        "s3",
+				Name:        "backup-storage-" + newTestSuffix(),
+				Url:         pointer.To("http://custom-url"),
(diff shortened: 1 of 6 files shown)


integration/backup_storage_test.go
```diff
@@ -0,0 +1,136 @@
...
+			resp, err := everestClient.CreateBackupStorage(context.Background(), testNs, payload)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.StatusCode).To(Equal(http.StatusCreated))
+
+			body, err := io.ReadAll(resp.Body)
+			Expect(err).NotTo(HaveOccurred())
+			defer resp.Body.Close()
+
+			var parsed *client.BackupStorage
+			Expect(json.Unmarshal(body, &parsed)).To(Succeed())
```
maxkondr, line 41: the generated client already provides method `CreateBackupStorageWithResponse` that returns parsed response and you can access it via `resp.JSON200`

mayankshah1607: ah, missed that. Yeah that's more convenient to use..

maxkondr:
checked PR and looked though Ginkgo docs.  It has good documentation.
But there are the following points that I would treat as drawbacks (that I caught from fast docs reading):
- it adds an additional layer of complexity, so that developer needs to study this framework first before starting writing tests. But to be honest - this point will be applicable to any another framework that we may choose (maybe exception is standard Testing  library that is widely used).
- it has own internal logic and rules for writing test-cases/suits/... and later for running them, that we need to understand during troubleshooting of failures.

At the same time it provides a very interesting and  useful functionality - [Patterns for Asynchronous Testing](https://onsi.github.io/ginkgo/#patterns-for-asynchronous-testing). I think we will definitely find the cases where to apply it and (what is more important) - where it is really required from testing framework.
And there are many other features that this framework provides but the requires an additional learning...

PR #1314, opened 2025-04-16
POC: Integration tests using gunit

mayankshah1607:
The following environment variables need be set first:
- `INTEGRATION_TEST_API_URL`
- `INTEGRATION_TEST_API_TOKEN`
- `INTEGRATION_TEST_NAMESPACE` (defaults to using `everest`)

Run the tests using `make integration-tests`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,3 +100,7 @@ CHART_BRANCH ?= main
 update-dev-chart:
 	GOPROXY=direct go get -u -v github.com/percona/percona-helm-charts/charts/everest@$(CHART_BRANCH)
 	go mod tidy
+
+
+integration-tests:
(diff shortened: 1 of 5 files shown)


maxkondr:
checked the framework description I see only only one benefit that will be useful for our project - using structure with `Setup/Teardown` methods where we can run only one instance of Everest server and then run all tests in parallel using this instance. But I am not sure that this a problem for as, we see that the same results can be achieved without `gunit` framework (like in https://github.com/percona/everest/pull/1319 with `envtest`).

One more point - I do not see how/where kubernetes is mocked in this case. Or it uses some real K8S and reads kubeconfig from user's ENV.

PR #1319, opened 2025-04-17
poc: envtest for integration testing

mayankshah1607:
Run the integration tests using `make integration-tests` (you don't have to setup anything before hand, not even a k8s cluster or an Everest installation)


maxkondr:
From the provided POCs only this one mocks K8S and runs Everest Server by itself - the rest require Everest server to be running before starting tests. 

Overall - `envtest` looks a little bit complicated. as I see the same results can be achieved with `sigs.k8s.io/controller-runtime/pkg/client/fake` FakeClient.

maxkondr:
>There are many benefits to using envtest over fake client -- for one, envtest actually runs a real KubeAPI server with etcd.

not exactly, please see https://book.kubebuilder.io/reference/envtest.html#testing-considerations
but it still may be enough for our needs.

Ok, fake client is better for unit tests - agree with this.


But I am still not sure that we need to mock K8S API - need to agreed on it.

PR #1339, opened 2025-04-29
EVEREST-1106 Check built-in user existence on every request

oksana-grishchenko:
[EVEREST-1106](https://jira.percona.com/browse/EVEREST-1106) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

EVEREST-1106

For the deleted built-in users, their access token was still valid until its natural expiration time.
To fix this, Everest server from now on checks if the user exists for every authenticated request for a built-in user. To avoid overloading k8s API with requests, Everest server uses a separate controller-runtime client with cache enabled for the needed resources.

[EVEREST-1106]: https://perconadev.atlassian.net/browse/EVEREST-1106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/session/manager.go
+++ b/pkg/session/manager.go
@@ -48,7 +48,10 @@ const (
 	SessionManagerClaimsIssuer = "everest"
 )
 
-var errExtractSub = errors.New("could not extract sub")
+var (
+	errExtractSub = errors.New("could not extract sub")
+	errExtractIss = errors.New("could not extract iss")
+)
 
 // Manager provides functionality for creating and managing JWT tokens.
 type Manager struct {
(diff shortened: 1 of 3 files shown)


pkg/session/manager.go
```diff
@@ -179,3 +204,37 @@ func (mgr *Manager) BlocklistMiddleWare(skipperFunc func() (echomiddleware.Skipp
...
+	content, err := extractContent(token)
+	if err != nil {
+		return "", false, err
+	}
+	sub, ok := content.Payload["sub"].(string)
+	if !ok {
+		return "", false, errExtractSub
+	}
+	username := strings.TrimSuffix(sub, ":login")
+	return username, len(sub) > len(username), nil
```
mayankshah1607, line 222: Can't we instead check if the issuer of the token is `everest` (the `SessionManagerClaimsIssuer` const). Won't that be more reliable?

mayankshah1607 (approved):
LGTM, just one question/suggestion

pkg/session/manager.go
oksana-grishchenko: A good point, I agree it's more reliable than checking the length. Updated, thanks!

mayankshah1607 (approved):

PR #1344, opened 2025-04-30
EVEREST-2025 fix(server): remove trailing slash from OIDC issuer in CSP connect-src

### LINKED ISSUE #1343: bug: `Content-Security-Policy` header includes invalid `connect-src` value when OIDC issuer ends with slash
Hi team,

There's a bug in the `connect-src` directive in the CSP header generated by the Everest server when OIDC authentication is enabled. Specifically, if the `Issuer` URL ends with a trailing slash, and the `.well-known` configuration path is appended (which starts with a slash), the resulting URL includes a double slash (`//`), which is not valid in CSP source expressions.

### How to reproduce

1. Configure OIDC authentication with an issuer URL that ends with a slash, for example:
   ```
   https://auth.example.com/
   ```
2. Start the Everest server.
3. Inspect the `Content-Security-Policy` response header.

### Actual result

The CSP header contains a directive like:
```
connect-src 'self' https://auth.example.com//.well-known/openid-configuration;
```

### Expected result

The issuer URL and `.well-known` path should be safely joined to avoid a double slash:
```
connect-src 'self' https://auth.example.com/.well-known/openid-configuration;
```

Thanks for your work on Everest! This fix would help ensure CSP compliance and prevent runtime issues.

aurimasniekis:
[EVEREST-2025](https://jira.percona.com/browse/EVEREST-2025) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

Ensure the issuer URL used in the Content-Security-Policy connect-src directive does not include a trailing slash, to avoid invalid URLs with double slashes  (e.g. https://auth.example.com//.well-known/openid-configuration).

Fixes #1343


recharte (approved):
Hi @aurimasniekis, thank you for your contribution.

PR #1354, opened 2025-05-08
EVEREST-1981 - Add test for specifying external DB access

tplavcic:
[EVEREST-1981](https://jira.percona.com/browse/EVEREST-1981) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-1981]: https://perconadev.atlassian.net/browse/EVEREST-1981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/apps/everest/.e2e/release/db-upgrade.e2e.ts
+++ b/ui/apps/everest/.e2e/release/db-upgrade.e2e.ts
@@ -178,7 +178,7 @@ test.describe.configure({ retries: 0 });
         });
 
         await test.step('Populate advanced db config', async () => {
-          await populateAdvancedConfig(page, db, '', true, '');
+          await populateAdvancedConfig(page, db, false, '', true, '');
           await moveForward(page);
         });
 
--- a/ui/apps/everest/.e2e/release/demand-backup.e2e.ts
+++ b/ui/apps/everest/.e2e/release/demand-backup.e2e.ts
@@ -113,7 +113,7 @@ test.describe.configure({ retries: 0 });
         });
 
         await test.step('Populate advanced db config', async () => {
-          await populateAdvancedConfig(page, db, '', true, '');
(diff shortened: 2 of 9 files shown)


mchawla16 (commented):
Hi Tomislav, Few points that came to my mind
1. Are we checking that the db is only accessible from the source IP range and not from other IP ranges? Should we also check that the service has load balancer set with an External IP?
2. Should we also check the case when the External access is open for all IP ranges? This is unsecure but for testing maybe we can do it.
3. Another suggestion would be to check for duplicate netmask entries. This will be fixed by https://perconadev.atlassian.net/browse/EVEREST-1918, currently it is in QA.

tplavcic:
> Hi Tomislav, Few points that came to my mind
> 
> 1. Are we checking that the db is only accessible from the source IP range and not from other IP ranges? Should we also check that the service has load balancer set with an External IP?
> 2. Should we also check the case when the External access is open for all IP ranges? This is unsecure but for testing maybe we can do it.
> 3. Another suggestion would be to check for duplicate netmask entries. This will be fixed by https://perconadev.atlassian.net/browse/EVEREST-1918, currently it is in QA.

@mchawla16 
1.  I have added a check for how the service is setup, we don't check from which IP it is accessible since it might be complicated to set it up, but if proper options are passed to K8s it would mean that there's some bug in Kubernetes - I doubt it would make sense to complicate the test this much.
2. We start without limiting source ranges and then later editing the setup to only allow some source ranges so this case is covered.
3. We can maybe add this check when the fix will be merged.

mchawla16 (approved):
Thanks, looks good.

PR #1368, opened 2025-05-13
EVEREST-2003 | expandable section in DB overview

Tikdev00:


fabio-silva (changes requested):
Just realised we have this "LabelValue" component, which is being imported by the StorageLocations component. Let's take this chance to move this component out of the "expandedRow" folder and move it to that part of the code and rename it to something more meaningful

fabio-silva (changes requested):
@Tikdev00 probably I explained myself wrongly.
LabelValue is only used by storage-locations.
Therefore:
1 - Move it next to storage-locations;
2 - Rename it to something more meaningful, as "LabelValue" is no longer fit

fabio-silva (approved):

PR #1371, opened 2025-05-14
Everest-2005 add view details in actions menu

Tikdev00:
[EVEREST-2005](https://jira.percona.com/browse/EVEREST-2005) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-2005]: https://perconadev.atlassian.net/browse/EVEREST-2005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fabio-silva (changes requested):
- We are showing "View DB details" on the DB details page itself;
- let's change `isStatusDetailView` and `isDbDetailsView` to something more generic, as the DbActions component only need to know whether to show or not those parts of the UI, it doesn't care if it's being shown in details page or other;
- we call a function `handleNavigateToDbDetails`, but the only thing it does is a navigate to a given URL. Either use a prop to handle the click or change this function to go straight away to the DB details page

ui/apps/everest/src/components/db-actions/db-actions.tsx
```diff
@@ -163,6 +169,18 @@ export const DbActions = ({
             'aria-labelledby': 'row-actions-button',
           }}
         >
+          {showDetailsAction && pathname !== redirectURL && (
```
fabio-silva, line 172: why all of this ?
We have the "showDetailsAction" flag, no need for further checks in the component.
The DbActions is rendered in two places:
- DbClusterView, aka cluster list;
- Cluster details page.

We need to show the icon on the **cluster list**, not on the details page itself.
So, on that page, just pass the flag as true and remove this pathname check.

fabio-silva (changes requested):

fabio-silva (approved):

fabio-silva (approved):

PR #1380, opened 2025-05-19
EVEREST-2028 | Tag test cases inside E2E tests

edithturn:
This PR adds Zephyr test case IDs to the E2E test coverage.


tplavcic (changes requested):
As discussed over slack, I don't think this will work good enough. Since we have a for loop for different databases we need to tag each test case in the runtime for each appropriate database and cluster size.
So I think we will need to create a map with test case IDs and db types and sizes and then in the runtime just print whichever case was executed at that time.
Otherwise it will not make sense since if we run for 1 database we will mark all 6-9 test cases as executed, but these are for different databases so it will be wrong.
I have some POC but I need to check if it really works or not.

PR #1413, opened 2025-06-02
EVEREST-2072 | Implement data importer APIs

mayankshah1607:
[EVEREST-2072](https://jira.percona.com/browse/EVEREST-2072) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-2072]: https://perconadev.atlassian.net/browse/EVEREST-2072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Adds the following API endpoints to Everest:
#### 1. `GET /v1/data-importers`
Lists all available data importers in the system.

#### 2. `POST /v1/namespaces/{namespace}/database-clusters/{dbName}/secret`
Allows the creation of a database-cluster Secret.

#### 3. `GET /v1/namespaces/{namespace}/database-clusters/{dbName}/data-import-jobs`
Lists all running data import jobs for a given DB.


go.mod
```diff
@@ -6,10 +6,12 @@ toolchain go1.24.1
 
 replace (
 	github.com/containerd/containerd v1.7.12 => github.com/containerd/containerd v1.7.13
-	k8s.io/api => k8s.io/api v0.32.0
+	github.com/percona/everest-operator => github.com/percona/everest-operator v0.6.0-dev1.0.20250610190300-1f40df0b2856
+	github.com/percona/percona-helm-charts/charts/everest => github.com/percona/percona-helm-charts/charts/everest v0.0.0-20250611064315-301cf3c025ed
```
mayankshah1607, line 10: Note to self - remove this before merging

docs/spec/openapi.yml
```diff
@@ -700,6 +700,107 @@ paths:
...
+  '/namespaces/{namespace}/database-clusters/{dbName}/secret':
+    post:
+      tags:
+        - Database Cluster
+      summary: Create a secret for the given database cluster
+      description: |
+        This API creates a secret for the database cluster specified by the `name` and `namespace`
+      operationId: createDatabaseClusterSecret
+      parameters:
+        - name: dbName
```
maxkondr, line 749: Do I understand properly that this secret with credentials shall be created before we create DBCluster? if yes - how could we know `dbName` beforehand?

docs/spec/openapi.yml
```diff
@@ -700,6 +700,107 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: secretName
+          in: query
+          description: Optional name of the secret to be created. If not provided, a random name will be generated.
+          required: false
+          schema:
+            type: string
+        - name: dbEngineType
```
maxkondr: what is the relation between the secret and DBEngine? it has reference to dbName

docs/spec/openapi.yml
```diff
@@ -1858,7 +1959,44 @@ paths:
...
+    get:
+      tags:
+        - Data Importer
+      summary: List data importers
+      description: |
+        This API lists all data importers in the kubernetes cluster.
+      operationId: listDataImporters
+      parameters:
+      - in: query
+        name: supportedEngine
```
maxkondr: if this is an array
```suggestion
        name: supportedEngines
```

docs/spec/openapi.yml
```diff
@@ -1858,7 +1959,44 @@ paths:
...
+        description: Filter data importers by supported database engine type.
```
maxkondr: comma-separated?

internal/server/handlers/k8s/data_importer.go
```diff
@@ -0,0 +1,37 @@
...
+		return list, nil
+	}
+
+	// filter the list based on the engines.
+	result := &everestv1alpha1.DataImporterList{}
+	for _, wantEngine := range supportedEngines {
+		for _, importer := range list.Items {
+			if slices.Contains(importer.Spec.SupportedEngines, everestv1alpha1.EngineType(wantEngine)) &&
+				// ensure we don't add duplicates
+				!slices.ContainsFunc(result.Items, func(i everestv1alpha1.DataImporter) bool {
```
maxkondr: you may use `slices.Compact` once just before returning the result but not on each loop iteration

internal/server/handlers/k8s/database_cluster.go
```diff
@@ -406,3 +406,33 @@ func valueOrDefault(value *int, defaultValue int) int {
...
+	return h.kubeConnector.CreateSecret(ctx, secretCpy)
+}
+
+func setDatabaseClusterSecretLabels(
+	namespace, dbName string,
+	engineType everestv1alpha1.EngineType,
+	secret *corev1.Secret,
+) {
+	labels := map[string]string{}
+	labels[common.DatabaseClusterNameLabel] = dbName
```
maxkondr: btw, shall we set `ownerReference` on this secret in operator?

api/everest-server.gen.go
```diff
@@ -111,11 +111,18 @@ const (
 	UpgradeEngine UpgradeTaskPendingTask = "upgradeEngine"
 )
 
+// Defines values for CreateDatabaseClusterSecretParamsDbEngineType.
+const (
+	CreateDatabaseClusterSecretParamsDbEngineTypePostgresql CreateDatabaseClusterSecretParamsDbEngineType = "postgresql"
```
maxkondr: can we create one common list of engine types (or import them from everest-operator) and avoid copy-pasting them each time we need to use them in API endpoint query params?

internal/server/handlers/rbac/data_import_job.go
```diff
@@ -0,0 +1,30 @@
...
+	filtered := make([]everestv1alpha1.DataImportJob, 0, len(list.Items))
+	for _, job := range list.Items {
+		if err := h.enforce(ctx, rbac.ResourceDataImportJobs,
+			rbac.ActionRead, rbac.ObjectName(namespace, job.Spec.TargetClusterName),
+		); errors.Is(err, ErrInsufficientPermissions) {
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+		filtered = append(filtered, job)
```
maxkondr: are we going to limit access to particular DataImportJob via RBAC? If yes - what benefit does it bring to user?
I thought user shall have access to all of them for particular DBCluster or nothing.

mayankshah1607: > are we going to limit access to particular DataImportJob via RBAC?

No, we are limiting access based on DBCluster name, same as DatabaseClusterBackups and DatabaseClusterRestores

> If yes - what benefit does it bring to user?

None actually, that's why we're not doing that

> I thought user shall have access to all of them for particular DBCluster or nothing.

Isn't this what we're doing? Please see the above line 20 and the unit test cases for it

internal/server/handlers/k8s/database_cluster.go
mayankshah1607: > btw, shall we set ownerReference on this secret in operator?

I don't think that's needed, the DB operator generally ends up owning it and cleaning it up upon deletion. The only thing we're doing differently now is instead of letting the everest-operator create it by default, we create it

docs/spec/openapi.yml
mayankshah1607: > Do I understand properly that this secret with credentials shall be created before we create DBCluster?

Not really, only the wizard step appears before the DB creation step. The API is called at the end of the wizard once the DBName is known

docs/spec/openapi.yml
mayankshah1607: > what is the relation between the secret and DBEngine?

Based on which DBEngine you are creating a Secret for, special labels need to be added (for now only for PG) on the Secret (please see the TODO comment in the k8s handler for this method)

> it has reference to dbName

As you pointed out in the previous comment already, the DB does not exist at the time of creation of the Secret, so it cannot know the DBEngine type. The UI needs to explicitly state it

docs/spec/openapi.yml
mayankshah1607: yeah, the `style: form` does that. Does it matter? Not sure what the question / concern was

docs/spec/openapi.yml
mayankshah1607: updated

internal/server/handlers/k8s/data_importer.go
mayankshah1607: updated (used `slices.CompactFunc` instead)

api/everest-server.gen.go
mayankshah1607: This is auto-generated based on the swagger spec, not copy-pasted manually. If you look at the code implementation, we are using the consts from everest-operator API. I do not know how to tell the code-gen not to generate these each time ...

docs/spec/openapi.yml
maxkondr: setting engineTypes in secret really confuses. And what if engineTypes in secret do not intersect with DBCluster's engine type?
we can assign labels on secret within the scope of DBCluster creation logic.

docs/spec/openapi.yml
maxkondr: if we accept comma-separated values - it shall be stated in API docs. It helps to understand API for user

internal/server/handlers/k8s/database_cluster.go
maxkondr: So, this secret will be tied to upstream DB cluster but not to Everest's DBCluster, right?

mayankshah1607: Yes

internal/server/handlers/rbac/data_import_job.go
maxkondr: in such a case you don't need to check RBAC for each DataimportJob - just check `read` permission to DBCluster - if OK -> pass the whole list to result, if no - pass empty list

docs/spec/openapi.yml
mayankshah1607: TBH - the reason we are setting `engineType` in Secret is so that it can work with PG. But as it turns out, we don't require the user to create Secret for the PG data-import. We can get rid of this param for now to avoid confusion ...

internal/server/handlers/rbac/data_import_job.go
mayankshah1607: Ah, okay, now I see what you mean - yes, it makes sense, the result is already a list of jobs for that databasecluster, so we don't have to check that in the loop. Just once is enough, so we return either the whole list or nothing. Thanks 👍🏻

docs/spec/openapi.yml
mayankshah1607: Done

docs/spec/openapi.yml
mayankshah1607: removed

maxkondr (approved):

PR #1414, opened 2025-06-02
EVEREST-1654 Get rid of labels usage in Everest objects

maxkondr:
Now Everest server doesn't check the dependencies (children objects that use parent one) during the validating `Delete` requests for `BackupStorage, MonitoringConfig, PodSchedulingPolicy`. All these objects now have `everest.percona.com/in-use-protection` finalizer in `.metadata.finalizers` in case they are in use.

Everest Operator PR: https://github.com/percona/everest-operator/pull/762
Helm PR: https://github.com/percona/percona-helm-charts/pull/561


internal/server/handlers/validation/backup_storage.go
```diff
@@ -80,7 +104,7 @@ func validateCreateBackupStorageRequest(
 		if storage.Spec.Region == params.Region &&
 			storage.Spec.EndpointURL == pointer.GetString(params.Url) &&
 			storage.Spec.Bucket == params.BucketName {
-			return errDuplicatedBackupStorage
+			return errDuplicatedBackupStorage(existingStorages.Items[0].GetNamespace())
```
oksana-grishchenko: is there any reason why `existingStorages.Items[0]` instead of `storage`?

go.mod
```diff
@@ -4,15 +4,6 @@ go 1.24.0
...
 
-replace (
-	github.com/containerd/containerd v1.7.12 => github.com/containerd/containerd v1.7.13
-	k8s.io/api => k8s.io/api v0.32.0
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.32.0
-	k8s.io/apimachinery => k8s.io/apimachinery v0.32.0
-	k8s.io/client-go => k8s.io/client-go v0.32.0
-	k8s.io/component-base => k8s.io/component-base v0.32.0
-)
-
```
oksana-grishchenko, line 15: For other reviewers who may also got concerned as I regarding the removal - this block was introduced in https://github.com/percona/everest/pull/901 but since the operator has updated the VM version and does not have the replaces anymore, the everest repo doesn't need the replacement as well

oksana-grishchenko (approved):
I love the idea of having that unified interface and use finalizers🙌

internal/server/handlers/validation/backup_storage.go
maxkondr: because I wanted to use `slices` lib here but missed for some reason...
Fixed

PR #1417, opened 2025-06-03
[UI] EVEREST-2073 - Data Importers

dianabirs:
[EVEREST-2073](https://jira.percona.com/browse/EVEREST-2073) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-2073]: https://perconadev.atlassian.net/browse/EVEREST-2073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/apps/everest/src/components/form-card/form-card.tsx
```diff
@@ -0,0 +1,133 @@
...
+const FormCard: React.FC<FormCardProps> = ({
+  title,
+  description = '',
+  controlComponent,
+}) => {
+  return (
+    <RoundedBox
+      title={<Header title={title} controlComponent={controlComponent} />}
+    >
+      {typeof description === 'string' ? (
```
Tikdev00: maybe we can avoid rendering Typography when string is empty, as in this way when it's empty string Typography will be rendered

dianabirs: description can also be `React.ReactNode` element, this is why we have this logic. This is actually an old component that I renamed to be a bit more reusable

  ```<FormCard
        title={Messages.cards.policies.title}
        description={
          <Box
            display="flex"
            justifyContent="space-between"
            alignItems="top"
            minHeight={'50px'}
          >...

Tikdev00: my point was about being empty string as there is typeof description === 'string' this check, my idea was to check if description exists then make type checking or render Typography

dianabirs: something like this?
```  tsx
 {description &&
        (typeof description === 'string' ? (
          <Typography variant="caption">{description}</Typography>
        ) : (
          description
        ))}

Tikdev00: yes!

dianabirs: ok, changing it then, thanks for the suggestion

PR #1434, opened 2025-06-11
EVEREST-2028 - Tag test cases from Zephyr inside E2E tests

tplavcic:
This should tag significant number of test cases from Zephyr inside E2E tests.
Second step will be automatically marking tests as passed/failed in zephyr based on automatic test runs.

### DIFF
--- a/ui/apps/everest/.e2e/release/init-deploy.e2e.ts
+++ b/ui/apps/everest/.e2e/release/init-deploy.e2e.ts
@@ -47,6 +47,42 @@ let token: string;
 
 test.describe.configure({ retries: 0 });
 
+const zephyrMap: Record<string, string> = {
+  'create-psmdb-1': 'T66',
+  'create-psmdb-3': 'T67',
+  'create-psmdb-5': 'T68',
+  'create-pxc-1': 'T72',
+  'create-pxc-3': 'T73',
+  'create-pxc-5': 'T74',
+  'create-postgresql-1': 'T77',
+  'create-postgresql-2': 'T78',
+  'create-postgresql-3': 'T79',
+  'edit-pxc': 'T85',
+  'edit-psmdb': 'T86',
+  'edit-postgresql': 'T87',
+  'external-pxc': 'T70',
+  'external-psmdb': 'T76',
+  'external-postgresql': 'T82',
+  'custom-pxc': 'T20',
+  'custom-psmdb': 'T46',
+  'custom-postgresql': 'T47',
+  'monitoring-pxc': 'T69',
+  'monitoring-psmdb': 'T75',
+  'monitoring-postgresql': 'T81',
+  'suspend-pxc': 'T90',
+  'suspend-psmdb': 'T91',
+  'suspend-postgresql': 'T92',
+  'resume-pxc': 'T93',
+  'resume-psmdb': 'T94',
+  'resume-postgresql': 'T95',
+  'restart-pxc': 'T87',
+  'restart-psmdb': 'T88',
+  'restart-postgresql': 'T89',
+  'delete-pxc': 'T96',
+  'delete-psmdb': 'T97',
+  'delete-postgresql': 'T98',
+};
+
 [
   { db: 'psmdb', size: 1 },
   { db: 'psmdb', size: 3 },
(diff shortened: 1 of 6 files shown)


ui/apps/everest/.e2e/release/init-deploy.e2e.ts
```diff
@@ -186,7 +224,8 @@ test.describe.configure({ retries: 0 });
         });
       });
 
-      test(`Check service type is LoadBalancer [${db} size ${size}]`, async () => {
+      zephyrId = zephyrMap[`external-${db}`];
+      test(`${zephyrId} - Check service type is LoadBalancer [${db} size ${size}]`, async () => {
         test.skip(size !== 3);
```
edithturn: This is not part of your change, but it will be nice to add some comments describing why tests are skipped to inform future maintainers.

ui/apps/everest/.e2e/release/pitr.e2e.ts
```diff
@@ -106,6 +118,7 @@ test.describe.configure({ retries: 0 });
       test.describe.configure({ timeout: 1_200_000 }); // 20 minutes
 
       const clusterName = `${db}-${size}-pitr`;
+      let zephyrId;
```
edithturn: The variable zephyrId is declared but not typed or initialised in this file.

ui/apps/everest/.e2e/release/pmm-k8s-metrics.e2e.ts
```diff
@@ -16,7 +16,7 @@
 import { test } from '@playwright/test';
 import { checkK8sMetrics } from '@e2e/utils/monitoring-instance';
 
-test(`Check PMM K8s metrics`, async () => {
+test(`T31 - Check PMM K8s metrics`, async () => {
```
edithturn, line 19: This test uses a hardcoded Zephyr `T31`  instead of fetching from a centralised zephyrMap object. I think we should keep consistency with the other files as well: ` test(`${zephyrId} - `

ui/apps/everest/.e2e/release/restore-new-cluster.e2e.ts
```diff
@@ -350,53 +358,51 @@ function getNextScheduleMinute(incrementMinutes: number): string {
...
+        await test.step(`Delete primary database cluster`, async () => {
+          await deleteDbCluster(page, clusterName);
+          await waitForStatus(page, clusterName, 'Deleting', 15000);
+          await waitForDelete(page, clusterName, 240000);
+        });
 
-      test(`Verify and Delete Restore History for the restored database [${db} size ${size}]`, async ({
-        page,
-      }) => {
-        // TODO: Remove the if statement after fix for https://perconadev.atlassian.net/browse/EVEREST-1012
```
edithturn, line 385: Glad to have these comments! :)

ui/apps/everest/.e2e/release/scheduled-backup.e2e.ts
```diff
@@ -47,6 +47,15 @@ let token: string;
 
 test.describe.configure({ retries: 0 });
 
+const zephyrMap: Record<string, string> = {
```
edithturn, line 50: I think we should add to the Readme a summary of how to approach Test Cases  Labels. Happy to help oni t!

ui/apps/everest/.e2e/release/pmm-k8s-metrics.e2e.ts
tplavcic: It would be best if all could be hardcoded (because it would be simpler), but since we have a `for` loop for each database type in most of the tests in that case the labels need to be dynamically decided.
So we will have a mix of hardcoded and dynamically decided (I don't think it's worth to add more code just for consistency - at least in this case).

ui/apps/everest/.e2e/release/pitr.e2e.ts
tplavcic: Thank you, it's fixed now.

ui/apps/everest/.e2e/release/init-deploy.e2e.ts
tplavcic: I have added some comments when skipping tests.

ui/apps/everest/.e2e/release/scheduled-backup.e2e.ts
tplavcic: I'm not sure what and where to add, but we will have some changes around this later since we need to automatically mark tests as passed/failed in zephyr so maybe we can add this later.

edithturn (approved):

PR #1459, opened 2025-06-25
Update helper scripts for Feature Branch

maxkondr:


oksana-grishchenko (approved):
Great improvements, thanks for implementing it! 
I think we should let @tplavcic know about the changes so that he could update the `everest-qa` repo accordingly

dev/fb/install-cli.sh
```diff
@@ -0,0 +1,52 @@
...
+  --chart-dir "./helm-chart" \
+  --version "$(cat version.txt)" \
+  --version-metadata-url "http://localhost:${vs_local_port}"  \
+  --operator.mysql=true \
+  --operator.mongodb=true \
+  --operator.postgresql=true \
+  --skip-wizard \
+  --namespaces everest \
+  --helm.set "versionMetadataURL=http://percona-version-service.default.svc.cluster.local" \
+  "$@"
```
mayankshah1607, line 45: Just a suggestion - same like our Tilt env, can we set the initial password to 'admin'? So that we don't have to do an extra step while testing?

mayankshah1607 (approved):
Thanks for the improvements, just one suggestion

dev/fb/install-cli.sh
maxkondr: I thought about this but decided to keep as is because QA may need to use the functionality with default password.
Anyway you still can achieve the same result with 1 command only:
```
install-cli.sh --helm.set server.initialAdminPassword=admin
```

or 
```
install-helm.sh --set server.initialAdminPassword=admin
```

PR #1490, opened 2025-07-07
EVEREST-1961 Update monitoring validation

oksana-grishchenko:
[EVEREST-1961](https://jira.percona.com/browse/EVEREST-1961) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

EVEREST-1961

Related PRs:
- https://github.com/percona/everest-operator/pull/816
- https://github.com/percona/percona-helm-charts/pull/587

In the operator PR, added webhook to check PMM access when creating or updating monitoring-instance.

In this PR: 
- Remove mentions of webhooks from the validation errors text to make the error observable and hide the implementation details
- Change the monitoring API tests. They now to use real PMM instances and cover the new cases described in the ticket comments:

    **Case 1.**
       Change the URL of an existing monitoring instance to a non-existing URL. Error example: `Get "<http://10.102.88.191sdas/v1/version>": dial tcp: lookup 10.102.88.191sdas on 10.96.0.10:53: no such host`

    **Case 2.**
       Change the URL of an existing monitoring instance to the URL of another (existing) PMM, no Login and Password provided. In that case Everest tries to use the existing apiKey but now since we are trying to connect to another instance, such apiKey does not exist in it. Everest returns the "authorization failed, please provide the correct credentials" error. To fix that the user needs to provide login and password for the new PMM - then the monitoring config is updated successfully and a new api key is added to the new PMM instance.

    **Case 3.**
        Change the URL of an existing monitoring instance to the URL that references the same PMM instance, no Login and Password provided. In that case Everest tries to use the existing apiKey and such apiKey exists for the instance, so no error returned, the monitoring instance is updated successfully. (example of alternative URLs that point to the same PMM instance: http://10.99.53.153 and http://monitoring-service.default.svc.cluster.local)

[EVEREST-1961]: https://perconadev.atlassian.net/browse/EVEREST-1961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/apps/everest/.e2e/release/rbac/create-db-clusters.e2e.ts
+++ b/ui/apps/everest/.e2e/release/rbac/create-db-clusters.e2e.ts
@@ -119,6 +119,8 @@ test.describe(
         ['database-clusters', 'create', `${namespace2}/*`],
       ]);
 
+      await page.reload({ waitUntil: 'networkidle' });
+
       await expect(page.getByText(pxcDb)).not.toBeVisible();
       await expect(page.getByText(psmdbDb)).toBeVisible();
 


pkg/pmm/apikey.go
```diff
@@ -100,3 +100,45 @@ func CreatePMMApiKey(
 
 	return key, nil
 }
+
+func CheckAccess(url, apiKey string, insecure bool, ctx context.Context) error {
```
mayankshah1607: WDYT about moving this validation to the Everest Operator ValidationWebhook? We already implemented some for the DatabaseCluster and DataImportJobs - https://github.com/percona/everest-operator/blob/main/internal/webhooks/databasecluster.go

We created the task [EVEREST-1300](https://perconadev.atlassian.net/browse/EVEREST-1300) to research on this topic but I think with the work done for data importers work we now have a feasible way to actually implement it.

mayankshah1607: One thing missing is that we don't run the webhooks in the E2E/Integration tests, but its something I'm working on

oksana-grishchenko: Thanks for the suggestion, I'll check it. 

> One thing missing is that we don't run the webhooks in the E2E/Integration tests, but its something I'm working on

Note: if we use validation webhooks in the E2E/Integration tests and add the PMM access validation to webhooks then we would need to update the tests so they would deploy and use a real PMM instance. So this should be done either in the scope of this ticket or in the scope of enabling webhooks for tests, depending on what gets implemented first.

oksana-grishchenko: The validation is moved to a new webhook, PTAL. 
The validation error from the WH appears as 500 and with the `admission webhook "vmonitoringconfig-v1alpha1.everest.percona.com" denied the request:` prefix. In order to hide the internal details and make the error observable, we trim this prefix.  


Regarding the failing Integration tests. The error `'Internal error occurred: failed calling webhook "vmonitoringconfig-v1alpha1.everest.percona.com": failed to call webhook: the server could not find the requested resource'` appears when creating any monitoring instance since dev-latest everest-operator which is used in tests, does not yet contain the webhook. Once the operator PR is merged, the tests should get fixed. Tested the cases locally.

internal/server/handlers/k8s/monitoring_instance.go
```diff
@@ -169,12 +172,23 @@ func (h *k8sHandler) createMonitoringK8sResources(
...
+	if err == nil {
+		return err
+	}
+	webhookPrefix := `admission webhook "vmonitoringconfig-v1alpha1.everest.percona.com" denied the request: `
+	if strings.Contains(err.Error(), webhookPrefix) {
+		return errors.New(strings.TrimPrefix(err.Error(), webhookPrefix))
+	}
+	return err
+}
+
```
mayankshah1607: Can we move this to the central error handler here? https://github.com/percona/everest/blob/main/internal/server/everest.go#L430

I think we can check for admission webhook errors using `k8serrors.IsInvalid` or `k8serrors.IsForbidden`
If you cast the `err` object to `k8serrors.APIStatus`, then you will get access to the underlying error object as well, so we may not have to perform such string operations either

oksana-grishchenko: Thanks Mayank, that's a great catch. Processing `k8serrors.IsForbidden` allows to figure out the http status code correctly instead of returning 500. We still need to trim the error message though because the underlying error contains that prefix in the message text as well, but it's not a big problem I believe.

mayankshah1607 (approved):

PR #1590, opened 2025-08-21
EVEREST-2177 Remove initContainer from Tilt

oksana-grishchenko:
EVEREST-2177 Remove initContainer from Tilt

initContainer uses the same `everest-operator` image as the main operator container but with the different entrypoint which is not supported by `docker_build_with_restart` which assumes a single entrypoint. So tilt was failing to run the everest-operator pod and now we patch the operator deployment to delete initContainer.

Related PRs:
- https://github.com/percona/everest-operator/pull/861

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -344,8 +344,17 @@ def patch_everest_operator_debug(in_yaml):
           c['resources'] = None
   return encode_yaml_stream(objects)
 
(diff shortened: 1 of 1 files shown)


mayankshah1607 (approved):

oksana-grishchenko:
CI failure comments: 
CI installs operator 0.0.0 which does not yet contain the migrator binary yet, so Everest installation fails in CI

PR #1597, opened 2025-08-26
Remove everest-operator catalog update from release workflow

recharte:
Since we moved to installing the everest operator directly through helm instead of through OLM, we no longer need it in the catalog, so https://github.com/percona/everest-catalog/pull/65 removed it. Now there's nothing to update in the catalog during a release so the release workflow can be simplified.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -219,45 +219,6 @@ jobs:
             exit 1
           fi
 
-      - name: Catalog - update veneer file
-        run: |
-          cd everest-catalog/tools
-
-          # configure userdata for commits
-          git config --global user.email "everest-ci@percona.com"
-          git config --global user.name "Everest RC CI triggered by ${{ github.actor }}"
-
-          if [[ $IS_RC == 1 ]]; then
-            go run . \
-              --veneer-file ../veneer/everest-operator.yaml \
-              --channel fast-v0  \
-              --new-version ${{ env.VERSION }} 
-          else
-            go run . \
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -219,45 +219,6 @@ jobs:
...
           fi
 
-      - name: Catalog - update veneer file
-        run: |
-          cd everest-catalog/tools
-
-          # configure userdata for commits
-          git config --global user.email "everest-ci@percona.com"
-          git config --global user.name "Everest RC CI triggered by ${{ github.actor }}"
-
```
oksana-grishchenko: we should keep the step itself, the `cd` and the git config in order to have the `git tag $GH_TAG` which is still called

oksana-grishchenko: we would need to rename the step though

recharte: Regarding the `cd` I think we should be good because I merged this into the previous step and it already has the `cd everest-catalog` [here](https://github.com/percona/everest/blob/3faef882bba46f9a77f0710b35472d5af6c5ec96/.github/workflows/release.yml#L205).

Regarding the `git tag` requiring the `git config` I'm not 100% sure if it's required but it also doesn't hurt so I just added it back in 5b00a13a96ffe85ec1b7c3b3594b1f44523d34ca.

oksana-grishchenko: I see, I didn't realize that it's intended that the remains of the step is merged to the previous step

oksana-grishchenko (approved):

PR #1622, opened 2025-09-03
EVEREST-2223 - Make PITR test to use AWS S3 instead of MinIO if credentials available

tplavcic:


dianabirs (approved):

ui/apps/everest/.e2e/pr/db-cluster/db-wizard/db-wizard-utils.ts
```diff
@@ -143,7 +158,7 @@ export const fillScheduleModalForm = async (
   page: Page,
   timeOptions: ScheduleTimeOptions = defaultTimeOptions,
   scheduleName: string,
-  testStorageOption: boolean = true,
+  backupStorage: string,
```
tplavcic: We change `testStorageOption: boolean = true` to `backupStorage: string` and it has 3 modes:
- undefined - we don't change or test storage option (same as `testStorageOption=false` before)
- 'testFirst' - tests the storage option, whichever is first in the combobox (same as `testStorageOption=true` before)
- 'any_other_value' - sets and tests the storage option to desired storage location (new behaviour)

PR #1658, opened 2025-09-16
EVEREST-2304 | Add image for running Helm chart hooks

mayankshah1607:


Dockerfile.helmtools
```diff
@@ -0,0 +1,11 @@
+# This image is used to run Everest Helm chart hooks.
+FROM alpine:3.20
+ARG TARGETOS
+ARG TARGETARCH
+
+# Install kubectl
+RUN apk add --no-cache curl \
+    && curl -LO "https://dl.k8s.io/release/v1.33.5/bin/${TARGETOS}/${TARGETARCH}/kubectl" \
```
recharte: Do you think it would be a good idea for the Dockerfile to receive the kubectl version as an env var so that we can control it from the workflow?

mayankshah1607: Updated

PR #1698, opened 2025-10-16
EVEREST-2239 Implement Split-Horizon DNS feature for PSMDB

maxkondr:
[EVEREST-2239](https://jira.percona.com/browse/EVEREST-2239) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

[EVEREST-2239]: https://perconadev.atlassian.net/browse/EVEREST-2239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/spec/openapi.yml
```diff
@@ -10932,3 +11168,88 @@ components:
...
+              properties:
+                certificate:
+                  description: Certificate is the TLS certificate and key for the
+                    split-horizon DNS configuration.
+                  properties:
+                    caCertFile:
+                      description: CaCertFile is based64 encoded CA certificate
+                        file content.
+                      type: string
+                    certFile:
```
oksana-grishchenko: The spec is not identical to what is generated now in the `everest-operator` repo, please keep it in sync before merging to avoid having unrelated updates in the spec in other PRs

docs/spec/openapi.yml
```diff
@@ -2156,6 +2156,219 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        '500':
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+    patch:
```
oksana-grishchenko, line 2280: PATCH assumes partial updates. In other APIs that use PATCH (update backup storages, update monitoring configs) in order to restrict what could be changed we use specific `params` types (`MonitoringInstanceUpdateParams` `UpdateBackupStorageParams`) in the request body but not the whole object. 

So in general, in the current API we have either: 
- `PUT + the whole CR`
- `PATCH + the 'params' object`

But now we want to use `PATCH + the whole CR` and I'm wondering if there is a real need for this inconsistency.

pkg/rbac/rbac.go
```diff
@@ -58,6 +58,10 @@ const (
 	ResourcePodSchedulingPolicies      = "pod-scheduling-policies"
 	ResourceDataImporters              = "data-importers"
 	ResourceDataImportJobs             = "data-import-jobs"
+
+	// Engine Features resources
+
+	ResourceEngineFeatures_SplitHorizonDNSConfigs = "engine-features/split-horizon-dns-configs"
```
oksana-grishchenko: The new constant so dramatically outstands from other constants (which have the same idea - the name of the resource), so just wondering - did some linter cause this difference? Not a big deal, just curious

internal/server/handlers/k8s/split_horizon_dns_config.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return h.kubeConnector.DeleteSplitHorizonDNSConfig(ctx, delObj)
+}
+
+func (h *k8sHandler) GetSplitHorizonDNSConfig(ctx context.Context, namespace, name string) (*enginefeaturesv1alpha1.SplitHorizonDNSConfig, error) {
+	return h.kubeConnector.GetSplitHorizonDNSConfig(ctx, types.NamespacedName{
+		Namespace: namespace,
+		Name:      name,
+	})
+}
```
oksana-grishchenko, line 72: Should we RBAC-restrict the DBs which use the splitDNS which the user does not have access to? We have such a restriction if a DB uses dbengines, backup storages, monitoring configs and lbc which the user does not have access to.

pkg/rbac/rbac.go
maxkondr: No, this is my intention to outline that this is Engine Feature take into account that in future we will have many of them:
ResourceEngineFeatures_feature_1
ResourceEngineFeatures_feature_2
ResourceEngineFeatures_feature_N

internal/server/handlers/k8s/split_horizon_dns_config.go
maxkondr: Adjusted RBAC validation for DBCluster resource

docs/spec/openapi.yml
maxkondr: Implemented PATCH + UpdateParams structure

oksana-grishchenko (approved):

PR #1723, opened 2025-11-10
Update GHA

maxkondr:
Add Go ENV vars to avoid fetching go modules from global cache


recharte (approved):

.github/workflows/dev-be-ci.yaml
```diff
@@ -30,14 +35,6 @@ jobs:
         with:
           go-version-file: 'go.mod'
           cache: true
-          
-      # The pxc-operator v1.18.0 go module gives checksum mismatch errors due to it being re-tagged.
-      # To workaround this problem, we need to clear the module cache and run `go mod tidy` again.
```
recharte, line 35: @maxkondr can we keep a comment like this so that we know why we have these env vars?

maxkondr: well, I wouldn't tie this comment to this particular problem. From my POV those env vars shall be set from the beginning (at least I always set them on my local machine)

PR #1750, opened 2026-01-02
Add schema definition of creating pluggable data store

chilagrow:
Refs solanicaio/engineering#15.

This PR adds an endpoint to get definition of providers' CR schema and UI components. Minimal changes were made in code so that generation of openapi doesn't fail.

### DIFF
--- a/docs/spec/openapi.yml
+++ b/docs/spec/openapi.yml
@@ -465,7 +465,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions':
+  '/clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extensionType}':
     post:
       tags:
         - Data Store Extension
@@ -500,6 +500,12 @@ paths:
         - name: dataStore
           in: path
           description: Name of the data store
+          required: true
+          schema:
+            type: string
+        - name: extensionType
+          in: path
+          description: Type of the extension
           required: true
           schema:
             type: string
@@ -546,7 +552,7 @@ paths:
         - Data Store Extension
       summary: List data store extensions
       description: |
-        This API lists all extensions configured for a specific data store.
+        This API lists extensions configured for a specific data store.
       operationId: listDataStoreExtensions
       parameters:
         - name: cluster
@@ -564,6 +570,12 @@ paths:
         - name: dataStore
           in: path
           description: Name of the data store
+          required: true
+          schema:
+            type: string
+        - name: extensionType
+          in: path
+          description: Type of the extension
           required: true
           schema:
             type: string
@@ -591,7 +603,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extension}':
+  '/clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extensionType}/{extension}':
     get:
       tags:
         - Data Store Extension
@@ -615,6 +627,12 @@ paths:
         - name: dataStore
           in: path
           description: Name of the data store
+          required: true
+          schema:
+            type: string
+        - name: extensionType
+          in: path
+          description: Type of the extension
           required: true
           schema:
             type: string
@@ -666,6 +684,12 @@ paths:
         - name: dataStore
           in: path
           description: Name of the data store
+          required: true
+          schema:
+            type: string
+        - name: extensionType
+          in: path
+          description: Type of the extension
           required: true
           schema:
             type: string
@@ -730,6 +754,12 @@ paths:
         - name: dataStore
           in: path
           description: Name of the data store
+          required: true
+          schema:
+            type: string
+        - name: extensionType
+          in: path
+          description: Type of the extension
           required: true
           schema:
             type: string
--- a/api/everest-server.gen.go
+++ b/api/everest-server.gen.go
@@ -5867,20 +5867,20 @@ type ServerInterface interface {
 	// (GET /cluster-info)
 	GetKubernetesClusterInfo(ctx echo.Context) error
 	// List data store extensions
-	// (GET /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions)
-	ListDataStoreExtensions(ctx echo.Context, cluster string, namespace string, dataStore string) error
+	// (GET /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extensionType})
+	ListDataStoreExtensions(ctx echo.Context, cluster string, namespace string, dataStore string, extensionType string) error
 	// Create data store extension
-	// (POST /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions)
-	CreateDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string) error
+	// (POST /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extensionType})
+	CreateDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string, extensionType string) error
 	// Delete data store extension
-	// (DELETE /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extension})
-	DeleteDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string, extension string) error
+	// (DELETE /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extensionType}/{extension})
+	DeleteDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string, extensionType string, extension string) error
 	// Get data store extension
-	// (GET /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extension})
-	GetDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string, extension string) error
+	// (GET /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extensionType}/{extension})
+	GetDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string, extensionType string, extension string) error
 	// Update data store extension
-	// (PUT /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extension})
-	UpdateDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string, extension string) error
+	// (PUT /clusters/{cluster}/namespaces/{namespace}/data-stores/{dataStore}/extensions/{extensionType}/{extension})
+	UpdateDataStoreExtension(ctx echo.Context, cluster string, namespace string, dataStore string, extensionType string, extension string) error
 	// List provider extensions
 	// (GET /clusters/{cluster}/namespaces/{namespace}/providers/{provider}/extensions)
 	ListProviderExtensions(ctx echo.Context, cluster string, namespace string, provider string) error
@@ -6109,10 +6109,18 @@ func (w *ServerInterfaceWrapper) ListDataStoreExtensions(ctx echo.Context) error
 		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter dataStore: %s", err))
 	}
 
+	// ------------- Path parameter "extensionType" -------------
+	var extensionType string
+
+	err = runtime.BindStyledParameterWithOptions("simple", "extensionType", ctx.Param("extensionType"), &extensionType, runtime.BindStyledParameterOptions{ParamLocation: runtime.ParamLocationPath, Explode: false, Required: true})
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter extensionType: %s", err))
+	}
+
 	ctx.Set(BearerAuthScopes, []string{})
 
 	// Invoke the callback with all the unmarshaled arguments
-	err = w.Handler.ListDataStoreExtensions(ctx, cluster, namespace, dataStore)
+	err = w.Handler.ListDataStoreExtensions(ctx, cluster, namespace, dataStore, extensionType)
 	return err
 }
 
@@ -6143,10 +6151,18 @@ func (w *ServerInterfaceWrapper) CreateDataStoreExtension(ctx echo.Context) erro
 		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter dataStore: %s", err))
 	}
 
+	// ------------- Path parameter "extensionType" -------------
+	var extensionType string
+
+	err = runtime.BindStyledParameterWithOptions("simple", "extensionType", ctx.Param("extensionType"), &extensionType, runtime.BindStyledParameterOptions{ParamLocation: runtime.ParamLocationPath, Explode: false, Required: true})
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter extensionType: %s", err))
+	}
+
 	ctx.Set(BearerAuthScopes, []string{})
 
 	// Invoke the callback with all the unmarshaled arguments
-	err = w.Handler.CreateDataStoreExtension(ctx, cluster, namespace, dataStore)
+	err = w.Handler.CreateDataStoreExtension(ctx, cluster, namespace, dataStore, extensionType)
 	return err
 }
 
@@ -6177,6 +6193,14 @@ func (w *ServerInterfaceWrapper) DeleteDataStoreExtension(ctx echo.Context) erro
 		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter dataStore: %s", err))
 	}
 
+	// ------------- Path parameter "extensionType" -------------
+	var extensionType string
+
+	err = runtime.BindStyledParameterWithOptions("simple", "extensionType", ctx.Param("extensionType"), &extensionType, runtime.BindStyledParameterOptions{ParamLocation: runtime.ParamLocationPath, Explode: false, Required: true})
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter extensionType: %s", err))
+	}
+
 	// ------------- Path parameter "extension" -------------
 	var extension string
 
@@ -6188,7 +6212,7 @@ func (w *ServerInterfaceWrapper) DeleteDataStoreExtension(ctx echo.Context) erro
 	ctx.Set(BearerAuthScopes, []string{})
 
 	// Invoke the callback with all the unmarshaled arguments
-	err = w.Handler.DeleteDataStoreExtension(ctx, cluster, namespace, dataStore, extension)
+	err = w.Handler.DeleteDataStoreExtension(ctx, cluster, namespace, dataStore, extensionType, extension)
 	return err
 }
 
@@ -6219,6 +6243,14 @@ func (w *ServerInterfaceWrapper) GetDataStoreExtension(ctx echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter dataStore: %s", err))
 	}
 
+	// ------------- Path parameter "extensionType" -------------
+	var extensionType string
+
+	err = runtime.BindStyledParameterWithOptions("simple", "extensionType", ctx.Param("extensionType"), &extensionType, runtime.BindStyledParameterOptions{ParamLocation: runtime.ParamLocationPath, Explode: false, Required: true})
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter extensionType: %s", err))
+	}
+
 	// ------------- Path parameter "extension" -------------
 	var extension string
 
@@ -6230,7 +6262,7 @@ func (w *ServerInterfaceWrapper) GetDataStoreExtension(ctx echo.Context) error {
 	ctx.Set(BearerAuthScopes, []string{})
 
 	// Invoke the callback with all the unmarshaled arguments
-	err = w.Handler.GetDataStoreExtension(ctx, cluster, namespace, dataStore, extension)
+	err = w.Handler.GetDataStoreExtension(ctx, cluster, namespace, dataStore, extensionType, extension)
 	return err
 }
 
@@ -6261,6 +6293,14 @@ func (w *ServerInterfaceWrapper) UpdateDataStoreExtension(ctx echo.Context) erro
 		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter dataStore: %s", err))
 	}
 
+	// ------------- Path parameter "extensionType" -------------
+	var extensionType string
+
+	err = runtime.BindStyledParameterWithOptions("simple", "extensionType", ctx.Param("extensionType"), &extensionType, runtime.BindStyledParameterOptions{ParamLocation: runtime.ParamLocationPath, Explode: false, Required: true})
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter extensionType: %s", err))
+	}
+
 	// ------------- Path parameter "extension" -------------
 	var extension string
 
@@ -6272,7 +6312,7 @@ func (w *ServerInterfaceWrapper) UpdateDataStoreExtension(ctx echo.Context) erro
 	ctx.Set(BearerAuthScopes, []string{})
 
 	// Invoke the callback with all the unmarshaled arguments
-	err = w.Handler.UpdateDataStoreExtension(ctx, cluster, namespace, dataStore, extension)
+	err = w.Handler.UpdateDataStoreExtension(ctx, cluster, namespace, dataStore, extensionType, extension)
 	return err
 }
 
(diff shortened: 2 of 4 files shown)


docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,233 @@ components:
...
+            - name: namespace
+              uiType: Toggle
+              description: Namespace
+            - name: name
+              uiType: TextBox
+              description: Display name
+            - name: dataStoreVersion
+              uiType: Toggle
+              description: Data store version
+            - name: nodes
```
chilagrow: What else is missing to display the UI below?

image

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,319 @@ components:
...
+                type: string
+              namespace:
+                description: Namespace to deploy the data store
+                type: string
+          spec:
+            description: Specification of the data store custom resource
+            type: object
+            required:
+              - provider
+              - topology
```
recharte: I'm not sure topology should be mandatory.

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,319 @@ components:
...
+                  monitoring:
+                    description: Monitoring component configuration
+                    type: object
+                    properties:
+                      type:
+                        description: Type of the monitoring component
+                        type: string
+                      version:
+                        description: Version of the monitoring component
+                        type: string
```
recharte: We can't have the components defined in this API schema because each provider will have a different set of components. But to be honest I'm not sure how to actually present the schema for this 🤔

recharte (commented):
Sorry @chilagrow I forgot about this PR 🙈

docs/spec/openapi.yml
chilagrow: Yeah, `DataStoreCRSchema` is just a blob, it has no schema, all above is under `example` of what a pxc provider may present. I am trying to show case how it may work with UI. I thought OpenAPI like schema definition might be one way. Let me think if there is other way of presenting what each CR exposes. 🤔 

```
    DataStoreCRSchema:
      description: DataStoreCRSchema defines the the custom resource of the data store
      type: object
      example:
```

docs/spec/openapi.yml
chilagrow: Perhaps not 😓

docs/spec/openapi.yml
chilagrow: Actually I would prefer an approach where CR schema can be generated from each provider.

Also do you think UI schema can be generated or do we see it as a manual task of adding a plugin?

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,319 @@ components:
...
+              description: Data store version
+              mapping: spec.version
+              selector:
+                - label: "v1.0"
+                  value: "1.0.0"
+                - label: "v2.0"
+                  value: "2.0.0"
+            - name: nodes
+              uiType: Group
+              description: Nodes configuration
```
solovevayaroslavna: The `description`, `name`, and `label` keys have different meanings in the UI, and their styles depend on where they are used.

In the case of node configuration, `description` means that this text will be rendered in the UI. Based on the screenshot, I don’t see any description with the text “Nodes configuration”, which is why I’m highlighting this. If we follow the spec as-is, an extra description text will appear in the UI.

For field descriptions we use the `body2` style, which is different. The same logic applies to `label`: when the UI encounters a `label` key, it knows that a label element should be rendered (example of label - below). 

Image

So for next several lines we will see the different result from what we have on your picture. There will be plain text (not bold) with "Number of nodes"

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,319 @@ components:
...
+              subParameters:
+                - name: replicas
+                  uiType: Number
+                  uiDescription: Number of nodes
+                  mapping: spec.engine.replicas
+                  params:
+                    badge: node
```
solovevayaroslavna: 1. Badge, what is it? Basing in MUI we use label for fields. For different type of fields it will looks different way, but we haven't any badge prop, so can you please explain what is it, and why we need both?

Image

solovevayaroslavna: Also! One important mention. If we want to see the same view as on the picture, we shouldn't have `uiType: Number` because number field looks like this one:

Image
I'm not sure is this expected or not, but if we want to see this as a number of buttons, we probably need to have a separate uiType like: "number of nodes" or "numberInoutButtonGroup" or something like that to allow user to use it the same way as on the screenshot

docs/spec/openapi.yml
chilagrow: Thank for identifying issues 🤗. The screenshot is the current UI on main branch and I would like to describe in UI schema, hence the question. Yeah current example doesn't work. From the spec https://github.com/openeverest/specs/blob/main/specs/001-plugins-architecture.md#ui-form-generator-yaml-schema-examples, what does `uiType: 'Group'` mean?

docs/spec/openapi.yml
```diff
@@ -899,7 +899,55 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/Secret'
-      
+
+  '/cluster/{cluster}/namespaces/{namespace}/data-store/{dataStore}/schema':
```
recharte: I don't think this is the correct path.
The schema is unique per provider and its purpose is to expose the information that clients (mostly the FE) need to understand to know how to build a DataStore CR for that given provider. Knowing the schema definition is a prerequisite to creating a DataStore. As such, it doesn't make sense to have this under `/data-store/{dataStore}`. 

The way I envision the flow from a client POV is:

1. Which providers are installed in a given cluster? GET `/clusters/{cluster}/providers`
2. How can I create a DataStore CR for a given provider? GET `/clusters/{cluster]/providers/{provider}/schema`
3. Now that I know that `ProviderA is installed and I understand the schema for that provider I can create a DataStore CR for it. POST `/clusters/{cluster}/namespaces/{namespace}/data-store`

Does this make sense?


```suggestion
  '/clusters/{cluster}/providers/{provider}/schema':
```

recharte: If we want to be less generic and since the schema actually refers to the dataStore maybe the following would also work:

```suggestion
  '/clusters/{cluster}/providers/{provider}/data-store-schema':
```

This is probably better because we might need to add some more schemas in the future and this way we are more future proof.
WDYT?

docs/spec/openapi.yml
recharte: >Also do you think UI schema can be generated or do we see it as a manual task of adding a plugin?

I would heavily push for auto-generation because it's a huge pain to do this manually, not to mention error-prone. But we need a mechanism that allows for extensive customization.
Something inspired by [kubebuilder tags](https://book.kubebuilder.io/reference/generating-crd) (e.g. `// +kubebuilder:validation:MaxLength=15`) might be the way to go.

docs/spec/openapi.yml
recharte: The group was meant to identify sections within the wizard. These sections could be collapsible for example.

docs/spec/openapi.yml
recharte: Looking at the example in the resources I think the badge refers to the units. For example the memory is in `GiB` but the replicas doesn't have a unit per se.
image

docs/spec/openapi.yml
chilagrow: Thanks, I am trying to identify what part of spec is Material UI and must follow the spec. There must be mapping between UI component and the CR. Any thoughts how best it can be done @solovevayaroslavna? 🤔

docs/spec/openapi.yml
chilagrow: Yeah agree `uiType: Number` doesn't look like the fit.

docs/spec/openapi.yml
chilagrow: Thanks, yes that's right about the path! 

As I understand, backup and restore will also be specific to provider CR. For example I would suppose it possible for a data store to not implement PITR, could that be possible?

docs/spec/openapi.yml
chilagrow: I agree, perhaps this is something to look into for me. It would be great amount of work to add even a few componets.

docs/spec/openapi.yml
solovevayaroslavna: @recharte correct me if i'm wrong, but i thought that group means group of anything. Like: group of fields, group of groups, etc. Because section, probably not the single case where we can need to group something. So... we can start with this approach, but during our work we can faced with the need to add some additional field like: group type to specify what it is - a section, a separate page, a dropdown with fields or something else. Maybe i'm wrong but it seems to me that just "group" is not enought to cover all ui situations

solovevayaroslavna: @chilagrow To be honest, not yet, I think the best way is to prepare UI examples along with the specification so that the reader of the specification can understand what a section is, what a label is, a group, etc. Perhaps a storybook with our components will help us with this. It would be ideal if we could integrate storybook or links to it into the documentation.

docs/spec/openapi.yml
solovevayaroslavna: @recharte yeah! you are absolutely right! We use badge for these purposes. 
This brings up another question. If a plugin developer uses a badge, should we also expose this string in the API? At the moment, the UI shows a “GB” badge, but the API request contains only “G”. But for CPU we have other logic

image

solovevayaroslavna: I see several possible approaches to address this.

1) We could add a configuration option for plugin developers that determines whether the badge value should be included in the API. In this case, the UI and API values would be identical.

2) Alternatively, we could provide predefined components for CPU, memory, and similar metrics to simplify adoption. However, I would still prefer the first option, as it offers greater flexibility in configuration.

docs/spec/openapi.yml
solovevayaroslavna: @chilagrow @recharte — could you help me understand this?

I see the `customResource` field, and its description suggests that this may not represent the entire schema, but rather only the part related to resources. However, in the `uiComponents` array I see UI components such as `name`, `namespace`, and `dataStoreVersion`. This implies that the schema is not currently split into logical parts and then mapped to steps in the UI, and I am not sure where the correct fix should be applied.

@chilagrow, to display steps in a wizard, we need to split the schema into sections corresponding to those steps—for example, resources, backups, and monitoring. If all components are listed sequentially in a single array, they will all be rendered on one page.

Should we remove `provider`, `name`, `namespace`, and `dataStoreVersion` from this array and place them elsewhere at a higher level, or should we introduce an additional level of hierarchy under `uiComponents`?

docs/spec/openapi.yml
chilagrow: Just to chime in from https://github.com/percona/everest/pull/1750#discussion_r2668002755. For plugin developers, being able to generate basic UI schema would ease adding new data store. So UI example or spec would also be valuable for identifying what should be generated.

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,203 @@ components:
...
+                  uiType: Number
+                  uiDescription: Resource size per proxy
+                  mapping: spec.proxy.resources.memory
+                  disable-edit: ['restore']
+                  params:
+                    badge: Gi
+                    label: Memory
+      type: object
+    DataStoreUISchema:
+      description: DataStoreUISchema defines a single property in the data store schema
```
solovevayaroslavna: Don't understand while "single" property.Do you mean one of many?

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,203 @@ components:
...
+          type: object
+          $ref: '#/components/schemas/DataStoreCRSchema'
+        uiComponents:
+          description: uiComponents is the schema definition for UI components supported by the data store
+          type: array # preserve the order of properties
+          items:
+            $ref: '#/components/schemas/DataStoreUISchema'
+          example:
+            - name: provider
+              uiType: Toggle
```
solovevayaroslavna:
```suggestion
              uiType: Select
```

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,203 @@ components:
...
+              description: Data store provider
+              mapping: spec.engine.type
+              disable-edit: ['all']
```
solovevayaroslavna: For me, it’s not obvious that “Edit” only works after submission, so i suppose this one, because it means i can edit during working with this part of form (as example). Maybe i'm wrong
```suggestion
              disable: ["update'', "restore"]
```

docs/spec/openapi.yml
```diff
@@ -3027,6 +3075,203 @@ components:
...
+              selector:
+                - label: "MySQL"
+                  value: "pxc"
+            - name: namespace
+              uiType: Toggle
+              description: Namespace
+              mapping: metadata.namespace
+              disable-edit: ['all']
```
solovevayaroslavna:
```suggestion
              disable: ["update'', "restore"]
```

docs/spec/openapi.yml
solovevayaroslavna: Should we separate the choosing of dbEngine(provider)? To avoid of problems during creation with default and already filled in parameters? In that case should we have it in schema at all? Or in some separate "global" section as it was done during [this](https://github.com/percona/everest/compare/main...openapi-ui-poc) POC?

solovevayaroslavna: and answering your question, I think you missed  small/medium/large options. It's just an addition to the UI that allows to set pre-prepared values for cpu/memory/disk size. But... it means user can add some "function" to schema, that tells how other fields should work with default values... we need to think how we can describe this on the UI from the plugin developer's side, and do we really need it

recharte: > Should we remove provider, name, namespace, and dataStoreVersion from this array and place them elsewhere at a higher level, or should we introduce an additional level of hierarchy under uiComponents?

I'm not sure we need any of these in the UI schema at all because these are fields that will always exist as part of any provider. I don't think that the UI schema needs to have absolutely everything that's in a DataStore CR, there are some fields that are mandatory and can never change, thus their component types can be hardcoded in the FE code. 

>Should we separate the choosing of dbEngine(provider)? To avoid of problems during creation with default and already filled in parameters? In that case should we have it in schema at all? Or in some separate "global" section as it was done during [this](https://github.com/percona/everest/compare/main...openapi-ui-poc) POC?

Yeah, I think we should separate it. And yes, I don't think it should be in the schema.

docs/spec/openapi.yml
recharte: > @recharte correct me if i'm wrong, but i thought that group means group of anything. Like: group of fields, group of groups, etc. Because section, probably not the single case where we can need to group something. So... we can start with this approach, but during our work we can faced with the need to add some additional field like: group type to specify what it is - a section, a separate page, a dropdown with fields or something else. Maybe i'm wrong but it seems to me that just "group" is not enought to cover all ui situations

Yeah, I agree that the groups construct can be for more than sections. If we can have that flexibility of choosing the type of group, that's even better.

docs/spec/openapi.yml
recharte: > As I understand, backup and restore will also be specific to provider CR.

Well, yes, but a given backup or restores object will always be coupled to a given data store. The we currently have in v1 is that you create backups by POST `/namespaces/{namespace}/database-cluster-backups` but then you can list the backups of a given database by GET `/namespaces/{namespace}/database-clusters/{cluster-name}/backups`. So it's kind of a hybrid, we can rethink this API design if needed.

internal/server/database_cluster.go
```diff
@@ -99,6 +99,12 @@ func (e *EverestServer) GetDatabaseCluster(c echo.Context, namespace, name strin
 	return c.JSON(http.StatusOK, result)
 }
 
+// GetDataStoreSchema implements [api.ServerInterface].
+// It retrieves the schema of CR and UI components for the given data store.
+func (e *EverestServer) GetDataStoreSchema(ctx echo.Context, cluster string, namespace string, dataStore string) error {
+	panic("unimplemented")
+}
+
```
chilagrow: Split horizon CR
```
Name:         split-horizon-1
Namespace:    my-special-place
Labels:       <none>
Annotations:  <none>
API Version:  enginefeatures.everest.percona.com/v1alpha1
Kind:         SplitHorizonDNSConfig
Metadata:
  Creation Timestamp:  2026-01-19T03:17:35Z
  Finalizers:
    everest.percona.com/cleanup-secrets
    everest.percona.com/in-use-protection
  Generation:        1
  Resource Version:  6214
  UID:               ace787f4-0ab4-4f90-8430-ebfc63a1bde7
Spec:
  Base Domain Name Suffix:  mydb.everest.local
  Tls:
    Secret Name:  splithorizon
Status:
  In Use:                    true
  Last Observed Generation:  1
Events:                      <none>

```
DatabaseCluster
```
Name:         mongodb-zc2
Namespace:    my-special-place
Labels:       <none>
Annotations:  <none>
API Version:  everest.percona.com/v1alpha1
Kind:         DatabaseCluster
Metadata:
  Creation Timestamp:  2026-01-19T03:18:22Z
  Finalizers:
    everest.percona.com/upstream-cluster-cleanup
    foregroundDeletion
  Generation:        2
  Resource Version:  7013
  UID:               f201e452-341e-4f69-bd8d-e2659451dbf1
Spec:
  Backup:
    Pitr:
      Enabled:  false
  Engine:
    Replicas:  3
    Resources:
      Cpu:     1
      Memory:  4G
    Storage:
      Class:            local-path
      Size:             2Gi
    Type:               psmdb
    User Secrets Name:  everest-secrets-mongodb-zc2
    Version:            8.0.12-4
  Engine Features:
    Psmdb:
      Split Horizon Dns Config Name:  split-horizon-1
  Monitoring:
    Resources:
  Pod Scheduling Policy Name:  everest-default-mongodb
  Proxy:
    Expose:
      Type:  internal
    Resources:
      Cpu:     0
      Memory:  0
  Sharding:
    Config Server:
      Replicas:  3
    Enabled:     false
    Shards:      2
Status:
  Cr Version:  1.21.1
  Details:     state: ready
mongoversion: 8.0.12-4
mongoimage: percona/percona-server-mongodb:8.0.12-4
message: ""
conditions:
    - status: "False"
      type: sharding
      lasttransitiontime: "2026-01-19T03:18:22Z"
      reason: ""
      message: ""
    - status: "True"
      type: initializing
      lasttransitiontime: "2026-01-19T03:18:23Z"
      reason: ""
      message: ""
    - status: "True"
      type: ready
      lasttransitiontime: "2026-01-19T03:20:21Z"
      reason: ""
      message: ""
replsets:
    rs0:
        members:
            mongodb-zc2-rs0-0:
                name: mongodb-zc2-rs0-0.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
                state: 1
                statestr: PRIMARY
            mongodb-zc2-rs0-1:
                name: mongodb-zc2-rs0-1.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
                state: 2
                statestr: SECONDARY
            mongodb-zc2-rs0-2:
                name: mongodb-zc2-rs0-2.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
                state: 2
                statestr: SECONDARY
        clusterrole: ""
        initialized: true
        addedasshard: null
        size: 3
        ready: 3
        status: ready
        message: ""
mongos: null
observedgeneration: 2
backupversion: 2.11.0
backupimage: percona/percona-backup-mongodb:2.11.0
backupconfighash: ""
pmmstatus: ""
pmmversion: ""
host: mongodb-zc2-rs0-0.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017,mongodb-zc2-rs0-1.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017,mongodb-zc2-rs0-2.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
size: 3
ready: 3

  Engine Features:
    Psmdb:
      Split Horizon:
        Domains:
          Domain:       mongodb-zc2-rs0-0-my-special-place.mydb.everest.local
          Private IP:   10.43.166.91
          Domain:       mongodb-zc2-rs0-1-my-special-place.mydb.everest.local
          Private IP:   10.43.65.49
          Domain:       mongodb-zc2-rs0-2-my-special-place.mydb.everest.local
          Private IP:   10.43.210.224
        Host:           mongodb-zc2-rs0-0-my-special-place.mydb.everest.local:27017,mongodb-zc2-rs0-1-my-special-place.mydb.everest.local:27017,mongodb-zc2-rs0-2-my-special-place.mydb.everest.local:27017
  Hostname:             mongodb-zc2-rs0-0.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017,mongodb-zc2-rs0-1.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017,mongodb-zc2-rs0-2.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
  Observed Generation:  2
  Port:                 27017
  Ready:                3
  Size:                 3
  Status:               ready
Events:                 <none>
```
PerconaServerMongoDB
```
Name:         mongodb-zc2
Namespace:    my-special-place
Labels:       <none>
Annotations:  <none>
API Version:  psmdb.percona.com/v1
Kind:         PerconaServerMongoDB
Metadata:
  Creation Timestamp:  2026-01-19T03:18:22Z
  Finalizers:
    percona.com/delete-psmdb-pods-in-order
    percona.com/delete-psmdb-pvc
  Generation:        2
  Resource Version:  7012
  UID:               54ab2a58-50d5-4000-806f-c2fb2bfff99a
Spec:
  Backup:
    Configuration:
      Backup Options:
        Oplog Span Min:  0
        Timeouts:
          Starting Status:  120
    Enabled:                true
    Image:                  percona/percona-backup-mongodb:2.11.0
    Pitr:
    Resources:
      Limits:
        Cpu:                300m
        Memory:             1G
  Cr Version:               1.21.1
  Enable Volume Expansion:  true
  Image:                    percona/percona-server-mongodb:8.0.12-4
  Image Pull Policy:        IfNotPresent
  Multi Cluster:
    Enabled:  false
  Pmm:
    Image:  
    Resources:
  Replsets:
    Affinity:
      Advanced:
        Pod Anti Affinity:
          Preferred During Scheduling Ignored During Execution:
            Pod Affinity Term:
              Topology Key:  kubernetes.io/hostname
            Weight:          1
    Arbiter:
      Enabled:  false
      Resources:
      Size:         0
    Configuration:  
      operationProfiling:
        mode: slowOp

    Expose:
      Enabled:  true
      Type:     ClusterIP
    Hidden:
      Enabled:  false
      Resources:
      Size:  0
    Name:    rs0
    Nonvoting:
      Enabled:  false
      Resources:
      Size:  0
    Pod Disruption Budget:
      Max Unavailable:  1
    Resources:
      Limits:
        Cpu:     1
        Memory:  4G
    Size:        3
    Split Horizons:
      mongodb-zc2-rs0-0:
        External:  mongodb-zc2-rs0-0-my-special-place.mydb.everest.local
      mongodb-zc2-rs0-1:
        External:  mongodb-zc2-rs0-1-my-special-place.mydb.everest.local
      mongodb-zc2-rs0-2:
        External:  mongodb-zc2-rs0-2-my-special-place.mydb.everest.local
    Volume Spec:
      Persistent Volume Claim:
        Resources:
          Requests:
            Storage:         2Gi
        Storage Class Name:  local-path
  Secrets:
    Encryption Key:  mongodb-zc2-mongodb-encryption-key
    Ssl:             mongodb-zc2-sh-cert
    Ssl Internal:    mongodb-zc2-ssl-internal
    Users:           everest-secrets-mongodb-zc2
  Sharding:
    Balancer:
    Enabled:  false
  Unsafe Flags:
  Update Strategy:  SmartUpdate
  Upgrade Options:
    Apply:     disabled
    Schedule:  0 4 * * *
    Set FCV:   true
Status:
  Backup Image:    percona/percona-backup-mongodb:2.11.0
  Backup Version:  2.11.0
  Conditions:
    Last Transition Time:  2026-01-19T03:18:22Z
    Status:                False
    Type:                  sharding
    Last Transition Time:  2026-01-19T03:18:23Z
    Status:                True
    Type:                  initializing
    Last Transition Time:  2026-01-19T03:20:21Z
    Status:                True
    Type:                  ready
  Host:                    mongodb-zc2-rs0-0.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017,mongodb-zc2-rs0-1.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017,mongodb-zc2-rs0-2.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
  Mongo Image:             percona/percona-server-mongodb:8.0.12-4
  Mongo Version:           8.0.12-4
  Observed Generation:     2
  Ready:                   3
  Replsets:
    rs0:
      Initialized:  true
      Members:
        mongodb-zc2-rs0-0:
          Name:       mongodb-zc2-rs0-0.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
          State:      1
          State Str:  PRIMARY
        mongodb-zc2-rs0-1:
          Name:       mongodb-zc2-rs0-1.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
          State:      2
          State Str:  SECONDARY
        mongodb-zc2-rs0-2:
          Name:       mongodb-zc2-rs0-2.mongodb-zc2-rs0.my-special-place.svc.cluster.local:27017
          State:      2
          State Str:  SECONDARY
      Ready:          3
      Size:           3
      Status:         ready
  Size:               3
  State:              ready
Events:               <none>

```

docs/spec/openapi.yml
```diff
@@ -277,6 +277,489 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        '500':
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  '/clusters/{cluster}/namespaces/{namespace}/providers/{provider}/extensions':
```
chilagrow: BackupStorage

```
Name:         bs-msp-1
Namespace:    my-special-place
Labels:       app.kubernetes.io/managed-by=tilt
Annotations:  <none>
API Version:  everest.percona.com/v1alpha1
Kind:         BackupStorage
Metadata:
  Creation Timestamp:  2026-01-20T05:09:25Z
  Generation:          1
  Resource Version:    94752
  UID:                 fc8a8149-cb29-4513-96bb-1d6eaca47053
Spec:
  Bucket:                   bs-msp-1
  Credentials Secret Name:  bs-msp-1
  Description:              bs-msp-1
  Endpoint URL:             https://minio.minio.svc.cluster.local
  Force Path Style:         true
  Region:                   us-east-1
  Type:                     s3
  Verify TLS:               false
Status:
  In Use:                    false
  Last Observed Generation:  1
Events:                      <none>
```

PR #1760, opened 2026-01-14
Rename container images

recharte:


dev/Tiltfile
```diff
@@ -658,7 +658,7 @@ local_resource(
     labels=["everest"]
 )
 
-docker_build_with_restart('perconalab/everest',
+docker_build_with_restart('ghcr.io/openeverest/openeverest',
```
recharte: I intentionally changed from the dev image to the production one because of the changes introduced here: https://github.com/percona/percona-helm-charts/pull/769/changes#r2691231999
Otherwise devs would have to run `make release-dev` in the helm charts repo before running `tilt up`.

dev/Tiltfile
```diff
@@ -472,7 +472,7 @@ local_resource(
 
 # Live update the Everest operator manager without generating a new pod
 load('ext://restart_process', 'docker_build_with_restart')
-docker_build_with_restart('perconalab/everest-operator',
+docker_build_with_restart('ghcr.io/openeverest/openeverest-operator',
```
recharte: I intentionally changed from the dev image to the production one because of the changes introduced here: https://github.com/percona/percona-helm-charts/pull/769/changes#r2691227712
Otherwise devs would have to run `make release-dev` in the helm charts repo before running `tilt up`.

.github/workflows/release.yml
```diff
@@ -209,7 +207,7 @@ jobs:
       - name: Catalog - checkout
         uses: actions/checkout@v5
         with:
-          repository: percona/everest-catalog
+          repository: openeverest/everest-catalog
```
chilagrow:
```suggestion
          repository: openeverest/openeverest-catalog
```

GitHub redirect would work, it's sensible to use our new repo name.

.github/workflows/release.yml
```diff
@@ -291,7 +289,7 @@ jobs:
           git checkout $BRANCH_NAME
 
           # Update the operator go module to reference the version tag
-          go get github.com/percona/everest-operator@$GH_TAG
+          go get github.com/openeverest/everest-operator@$GH_TAG
```
chilagrow, line 294:
```suggestion
          go get github.com/openeverest/openeverest-operator@$GH_TAG
```

dev/Tiltfile
recharte: Had to revert those changes in the helm chart so I'm also changing this back to the dev images.

dev/Tiltfile
recharte: Had to revert those changes in the helm chart so I'm also changing this back to the dev images.

.github/workflows/release.yml
recharte: Actually I reverted your suggestion because I want to handle everything go mod related at the same time.

PR #1761, opened 2026-01-15
Issue types after repo move

spron-in:
Changing the issue templates.
1. From Percona to OpenEverest
2. Remove JIRA links
3. Add security vulnerability private reporting


chilagrow:
> > Thank you 😄
> > For issues that do not result in user visible features, we probably a new category. It would include code maintenance, test improvements, refactor and others which are only for OpenEverest maintainers interest. What do you thing?
> 
> @chilagrow agree. But I would keep it simple and have just one new category, not a few. Sounds ok?

@spron-in One new category for all is totally fine. While I was creating new issues for maintainer tasks, I found it inflexible to limit within bug, feature and discussion.

PR #1783, opened 2026-01-16
feat: component logs

dianabirs:


recharte:
Hi @dianabirs overall the feature looks great :raised_hands: 

I noticed one thing we can probably improve. If the pod is in the `Pending` state and we try to access the logs we get an error message. This is because the pod isn't running yet and the API returns a 502 error. We should probably not allow users to try to access the logs of `Pending` pods.

@oksana-grishchenko do you think we should simply filter this from the FE side only or should we change the backend in some way?

ui/apps/everest/src/hooks/api/db-cluster/useDbClusterComponentLogsStream.ts
```diff
@@ -0,0 +1,116 @@
...
+  const [isConnecting, setIsConnecting] = useState(false);
+  const [error, setError] = useState<Error | null>(null);
+
+  const { canRead } = useRBACPermissions(
+    'database-clusters',
+    `${namespace}/${dbClusterName}`
+  );
+
+  const enabled =
+    (options?.enabled ?? true) &&
```
solovevayaroslavna, line 40: is this should be true by default?

solovevayaroslavna:
image

Maybe I'm wrong, but in my opinion it's strange to go to an empty log page via tab if there's anything there. It seems to me that we need to choose something by default anyway, and the user will choose whether to change the component or not. What if the log is unique for a single component? In this case, it will be strange to choose a single log. I would like to create an issue for improvement, what do you think?

ui/apps/everest/src/pages/db-cluster-details/components/diagram-view/component-node.tsx
```diff
@@ -1,19 +1,45 @@
...
 const ComponentNode = ({
   data: {
     selected,
     componentData: { name, status, ready, type, restarts, started = '' },
   },
 }: NodeProps<CustomNode<DBClusterComponent>>) => {
+  const { dbClusterName = '', namespace = '' } = useParams();
+  const navigate = useNavigate();
+  const { dbCluster } = useContext(DbClusterContext);
+  const isPending = dbCluster?.status?.status === DbClusterStatus.initializing;
```
solovevayaroslavna: Could individual component be pending even if cluster is not initializing? If so, probably we should also check componentData.status? Is this make sense?

ui/apps/everest/src/hooks/api/db-cluster/useDbClusterComponentLogsStream.ts
```diff
@@ -0,0 +1,116 @@
...
+        if (!reader) throw new Error('No response body');
+
+        const decoder = new TextDecoder();
+
+        // eslint-disable-next-line no-constant-condition
+        while (true) {
+          const { value, done } = await reader.read();
+          if (done || abortController.signal.aborted) break;
+          if (value) {
+            const chunk = decoder.decode(value, { stream: true });
```
solovevayaroslavna, line 97: How long can streaming take? My concern is that with long-term streaming of logs, the logs string can grow to huge sizes, which will cause memory problems. So i can suppose to think about:
- adding a limit on the maximum size of logs (for example, the last 10,000 rows)
- or to add pagination/virtualization for rendering

ui/apps/everest/src/hooks/api/db-cluster/useDbClusterComponentLogsStream.ts
dianabirs: we want to prevent streaming in case there is no selected component or reading permissions

ui/apps/everest/src/hooks/api/db-cluster/useDbClusterComponentLogsStream.ts
dianabirs: Added a helper to fetch full logs only for download/copy functionality

ui/apps/everest/src/pages/db-cluster-details/components/table-view/components-table-view.tsx
```diff
@@ -118,17 +118,22 @@ const ComponentsTableView = ({
       enableRowActions={!isPending}
       renderRowActions={({ row }) => {
         const component = row.original;
+        const isComponentPending =
+          isPending || component.status === COMPONENT_STATUS.PENDING;
```
recharte: This shouldn't depend on the DB status. We want to show the logs even if the DB is initializing, we just don't want to show on a given component if that component is in the pending state.

recharte (approved):
Everything looks good from my end.
Let's just wait for @solovevayaroslavna to do a final review before merging.
Thanks @dianabirs 🤗

PR #1788, opened 2026-01-20
Replace logo assets

spron-in:
Changing logos, ensuring vendor neutral support banners, fixing links to documentation

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -3,7 +3,3 @@ contact_links:
   - name: 🙋🏾 Generic question
     url: https://github.com/openeverest/openeverest/discussions/new
     about: If you need help debugging something or have a general question about OpenEverest
-
-  - name: 🔒 Report a Security Vulnerability
-    url: https://github.com/openeverest/openeverest/security/advisories/new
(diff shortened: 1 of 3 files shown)


ui/apps/everest/src/components/welcome-dialog/welcome-dialog.messages.ts
```diff
@@ -1,8 +1,8 @@
 export const Messages = {
   allSet: 'You´re all set!',
-  header: 'Welcome to Everest Beta!',
+  header: 'Welcome to OpenEverest Beta!',
```
recharte: AFAIK this header isn't being displayed anywhere but let's take this opportunity and fix it nonetheless.

```suggestion
  header: 'Welcome to OpenEverest!',
```

public/dist/index.html
chilagrow, line 1: I noticed `tilt up` produces this file for me, it's a build output. I think it's better suited in `.gitignore`.

recharte: Yes this should not be committed. @spron-in please keep this as an empty file.

@chilagrow regarding adding it to the .gitignore I don't recall the exact details but I believe it's not as simple as that, I think removing this file requires some pipeline changes. I started that work a long time ago and never finished it 🙈 https://github.com/openeverest/openeverest/pull/209

PR #1789, opened 2026-01-21
Renaming percona everest into openeverest in md files

spron-in:
Fix naming in .md files that are quite visible.

Not touching CONTRIBUTING.md file for now, but it clearly needs more love.

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -1,6 +1,6 @@
 # Adopters
 
-Below is a list of organizations and users who are using Percona Everest.
+Below is a list of organizations and users who are using OpenEverest.
 
 This list helps inspire others to join our community and demonstrates the real-world impact of the project. Adding your organization takes just a few minutes but means the world to us!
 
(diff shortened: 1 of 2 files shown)


ADOPTERS.md
```diff
@@ -1,26 +1,26 @@
...
 
-## Percona Everest Adopters
+## OpenEverest Adopters
 
 This list is sorted chronologically by submission date.
 
 | Organization | Contact | Date | Description of Use |
 | ------------ | ------- | ---- | ------------------ |
-| [Percona](https://percona.com) | @recharte | 2024-02-29 | Percona provides support and services for Percona Everest |
+| [Percona](https://percona.com) | @recharte | 2024-02-29 | Percona provides support and services for OpenEverest |
```
recharte, line 24: We definitely need to change the contact here since I no longer work at Percona.

recharte: I know this was not the scope of the PR but we might as well do it now if you don't mind :see_no_evil:

spron-in: Well, changing a contact is not mandatory. The contact is mostly for history if someone would want to check what was happening there, not the current state.

recharte (approved):

PR #1792, opened 2026-01-21
Add PXC 1.19.0

oksana-grishchenko:
Related PRs:
- https://github.com/openeverest/openeverest-operator/pull/921
- https://github.com/openeverest/openeverest-catalog/pull/75

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/oapi-codegen/echo-middleware v1.0.2
 	github.com/oapi-codegen/runtime v1.1.1
 	github.com/operator-framework/api v0.33.0
-	github.com/percona/everest-operator v0.6.0-dev1.0.20260125124540-ff8df087e04c
+	github.com/percona/everest-operator v0.6.0-dev1.0.20260128124631-ff82d11d17da
 	github.com/percona/percona-helm-charts/charts/everest v0.0.0-20260119185700-561f0f50e592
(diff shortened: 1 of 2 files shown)


ui/apps/everest/src/shared-types/restores.types.ts
```diff
@@ -63,6 +63,7 @@ export enum PXC_STATUS {
   STOPPING = 'Stopping Cluster',
   RESTORING = 'Restoring',
   STARTING_CLUSTER = 'Starting Cluster',
+  PREPARING_CLUSTER = 'Preparing Cluster',
```
oksana-grishchenko, line 66: Hi @solovevayaroslavna, PTAL - there is this new restore state for pxc clusters, so I added it to the FE files as well although I'm not sure why do we duplicate the statuses if the BE already maps them in the DatabaseClusterRestore status and we could rely on what's returned from the API

solovevayaroslavna: @oksana-grishchenko Thanks a lot for hilighting this issue! ❤️

I will keep this unreading, because probably you right and we don't need this more
We duplicate the types of statuses and constants in order to see the typing while writing code
And perhaps we can either: 
- painlessly get rid of unnecessary mapping
or
- reconsider our approach to describing types and generate them using the api
but both theories need to be tested, I'll leave it on my TODO list

PR #1834, opened 2026-02-03
ui-shema-generator

solovevayaroslavna:
image

- ui-generator documentation was added
- the ui-generator component has been added, which is responsible for generating the form using json schema
(Note: some parts of the ui-generator are still under development and may change during its implementation in db-flow and the introduction of new fields)
- a new page  /ui-generator-builde has been added. Plugin developer can check the form display depending on the schema (Note: ui-generator-builder is a separate issue with certain requirements, that haven't been discussed in detail, its implementation as part of this PR should provide an easier development and testing process, as well as recording demo videos, but its presence doesn't mean that it will be presented to the end user as is)
- number fields was implemented into ui-generator (https://github.com/issues/assigned?issue=openeverest%7Copeneverest%7C1848)
- the customAccordionSummary component was migrated to the ui library as part of the refactoring
- db-form-page was added as a part of https://github.com/issues/assigned?issue=openeverest%7Copeneverest%7C1839 (Note: at the moment, this page is hidden and inaccessible by the user through the UI, work on the implementation of the ui-generator in db-flow is a separate issue and is being conducted in a separate branch and will be allocated to a separate large PR)

### DIFF
--- a/ui/apps/everest/src/pages/db-form/db-form.tsx
+++ b/ui/apps/everest/src/pages/db-form/db-form.tsx
@@ -20,6 +20,7 @@ export const DatabasePageGenerated = () => {
 
   const methods = useForm({
     mode: 'onChange',
+    // TODO uncomment during move to the dbWizard
     // resolver: async (data, context, options) => {
     // //   const customResolver = zodResolver(schema);
     //   const result = await customResolver(data, context, options);
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -378,12 +378,3 @@ jobs:
           draft: true
           files: |
             dist/*
-
-      - name: Trigger Homebrew tap update
-        if: env.IS_RC == 0
-        run: |
-          curl -X POST \
-            -H "Accept: application/vnd.github+json" \
-            -H "Authorization: Bearer ${{ secrets.TAP_REPO_TOKEN }}" \
-            https://api.github.com/repos/openeverest/homebrew-tap/dispatches \
-            -d "{\"event_type\":\"new-release\",\"client_payload\":{\"version\":\"$VERSION\"}}"


ui/apps/everest/src/pages/ui-generator-builder/ui-generator-schema.json
chilagrow, line 1: This file is going to be a relevant one for the API 😃

ui/apps/everest/src/pages/db-form/db-form.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const [selectedTopology, setSelectedTopology] = useState<string>('');
+  const sections = getSteps(selectedTopology, topologyUiSchemas);
+  const stepLabels = ['Choose topology', ...Object.keys(sections)];
+
+  // const selectedTopology = 'replica';
+  const topologies = Object.keys(topologyUiSchemas);
+
+  const methods = useForm({
+    mode: 'onChange',
+    // resolver: async (data, context, options) => {
```
chilagrow, line 24: There are a few places with commented out code, perhaps you want to remove them?

ui/apps/everest/src/pages/ui-generator-builder/json-editor-panel/json-editor-panel.tsx
```diff
@@ -0,0 +1,84 @@
...
+        height: '100%',
+        display: 'flex',
+        flexDirection: 'column',
+        backgroundColor: '#f6f8fa',
+      }}
+    >
+      <Typography variant="h6" sx={{ mb: 2 }}>
+        JSON Editor
+      </Typography>
+      <Editor
```
chilagrow, line 51: I'm not sure what I'm doing wrong but editor is in `Loading...` and doesn't complete loading. I see this error in browser console below though. 🤔 

```
ui-generator-builder:1 Uncaught (in promise) 
Event {isTrusted: true, type: 'error', target: script, currentTarget: null, eventPhase: 0, …}
```

image

ui/apps/everest/src/components/drawer/Drawer.constants.ts
```diff
@@ -30,4 +31,10 @@ export const ROUTES: EverestRoute[] = [
     icon: SettingsIcon,
     text: 'Settings',
   },
+  //TODO ui-generator builder should be hided from ui and delivered separately with tests coverage
```
chilagrow, line 34: Yeah this comment makes sense, while this is helpful for plugin development, I'm not so sure if this is relevant for OpenEverest users.

ui/apps/everest/src/pages/db-form/db-form.tsx
solovevayaroslavna: To be honest, not really, since I have several branches related to this one and I would say that the work is still "in progress" because, for example, the db-form file was created essentially to integrate the ui-generator into the db-creation flow. But it's a big separate ticket, that already was started and live in separate branch. The reason why  I didn't make changes to the database-form, but started creating a separate db-form page, is that I would like to separate v1 and v2 and then quickly switch them between each other, keeping the previous version working until the v2 will be fully implemented. so there will have to be 5 more methods or more in this place.
I don't know if this is acceptable in our situation and if I can leave it like this until the https://github.com/openeverest/openeverest/issues/1839 will be implemented

solovevayaroslavna: if I put a //TODO next to each of them for easier ui-folder search at the end to make sure I haven't missed anything will it be okey for such a big issue for now? Or not?

chilagrow: Ok!

solovevayaroslavna: @chilagrow  I updated comments with extra TODO

ui/apps/everest/src/pages/ui-generator-builder/json-editor-panel/json-editor-panel.tsx
solovevayaroslavna: @chilagrow I've researched this issue and I want to come back with an answer about it. I used [monaco-react](https://github.com/suren-atoyan/monaco-react) as it was. But it's just a react wrapper on [microsoft/monaco-editor](https://github.com/microsoft/monaco-editor), that uses CDN under the hood. This is the exact problem why in your case you got an error with `tilt up` because tilt up use production build, that doesn't allow using CDN, that's expected behavior.
And, despite the fact that monaco-react library has [a way to change this behavior](https://github.com/suren-atoyan/monaco-react?tab=readme-ov-file#loader-config), microsoft/monaco-editor still have an [issue](https://github.com/microsoft/monaco-editor/issues/271) that needs from us to change our SPC policy and add 'unsafe-inline' to it. Without this setting, some of the styles in production don't work, which makes it impossible to work with the editor.

Since we are not working on an editor for plugin developer (as a new feature) at the moment and use it only for our internal needs, I wouldn't like to change anything in the csp policy and worsen the security of the project until a solution is found. We need more time to explore the options, and since this is not a priority right now, I plan to leave everything as it is. 
So the monaco editor will be available only in development mode.
You need to set `enableFrontendBuild: false` in `openeverest/dev/config.yaml`. And then after `tilt up` run `pnpm install` and `pnpm --filter @percona/everest dev` to see the working monaco editor.

solovevayaroslavna: I already investigate such editors as ACE Editor and CodeMirrorEditor and faced with the same problem of inline styles. [react-simple-code-editor](https://github.com/react-simple-code-editor/react-simple-code-editor) looks poor for our purposes, so i hope i can find something better. 
According to the documentation [this one](https://uiwjs.github.io/react-json-view/) has no problems with CSP, so I plan to try it next as soon as the priority of this issue is raised.

PR #1870, opened 2026-02-12
Add Providers and Instances to API

recharte:
Added a tool (gen-crds-openapi) to generate the OpenAPI schema components directly from the CRD manifests.
Added Providers and Instances API endpoints.


pkg/apis/v1alpha1/instance_types.go
chilagrow, line 1: I see, providers and instances are v1alpha1, they are new. While openeverest will be v2alpha.

recharte: yeah, we shouldn't start a new resource with v2. The resource's version is v1 even though the application will be v2

PR #1872, opened 2026-02-13
Add clusters API

recharte:


internal/server/handlers/k8s/cluster.go
```diff
@@ -0,0 +1,54 @@
+// everest
+// Copyright (C) 2023 Percona LLC
```
chilagrow, line 2: Could we use `Copyright (C) 2024-2026 Solanica Inc.` for newly added files please?

chilagrow (approved):
Thank you!

internal/server/handlers/k8s/cluster.go
recharte: I didn't worry too much about this because we need to fix license headers across the board and we should integrate with a tool that does this automatically for us.

recharte: I'll make sure to do this is a follow-up PR.

PR #1873, opened 2026-02-13
Add RBAC resource generation

recharte:
Add a tool to generate the RBAC resources from the OpenAPI spec.
The OpenAPI spec is static so there was no point to parse it at runtime to extract the RBAC resources. By generating the RBAC resources ahead of time we optimize runtime execution.
As a nice side-effect the unit tests (`make test`) now take only 14s to run instead of the >400s it used to take :sweat_smile: :see_no_evil:


chilagrow (approved):
Just wonderful improvement!

pkg/rbac/resources.gen.go
chilagrow, line 1: Could you run `make format` please, CI is complaining about formatting.

PR #1875, opened 2026-02-16
Add Backup and BackupClass APIs

recharte:


pkg/apis/v1alpha1/backup_types.go
```diff
@@ -0,0 +1,147 @@
...
+	// +kubebuilder:validation:Required
+	InstanceName    string `json:"instanceName,omitempty"`
+	*BackupTemplate `json:",inline"`
+}
+
+// BackupTemplate defines a shared template for the backup job.
+type BackupTemplate struct {
+	// BackupClassName is the backup tool to use for the backup.
+	// +kubebuilder:validation:Required
+	BackupClassName string `json:"backupToolName,omitempty"`
```
chilagrow: `BackupClassName` is Go struct field name and `backupToolName` is json key name. Which one is our preference? 🤗

pkg/apis/v1alpha1/backup_types.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// BackupDestination defines the destination for the backup data.
+type BackupDestination struct {
+	// BackupStorageName is the name of the BackupStorage to use for the backup.
+	// +optional
+	BackupStorageName *string `json:"backupStorageName,omitempty"`
+	// S3 contains the S3 information for the backup destination.
+	// +optional
+	S3 *BackupS3Destination `json:"s3,omitempty"`
```
chilagrow, line 56: Would we consider beyond S3 compatible storage?

chilagrow (commented):
Great, thanks for this! Perhaps the storage is beyond the scope of backup but I would like to ask you about current tie to s3 compatible storage.

pkg/apis/v1alpha1/backup_types.go
recharte: That's a leftover :see_no_evil: changed to `backupClassName`.

pkg/apis/v1alpha1/backup_types.go
recharte: Yes, in the future we can add support for other types of storages. That's why both `BackupStorageName` and `S3` are pointers, only one can be set at a given time and in the future there might be more options.

PR #1878, opened 2026-02-17
Add watch configuration and field indexing support to provider interface

recharte:
This pull request introduces a major refactor and extension of the controller SDK to support more flexible and advanced resource watching and indexing. The changes replace the old `OwnedTypes` mechanism with a new, configurable watch system, add support for field indexing, and provide new extension points for provider implementations. There are also improvements to metrics configuration and context access.

**Key changes:**

### Advanced Watch and Index Configuration

* Introduced the `WatchConfig` struct and `WatchProvider` interface, allowing providers to declaratively configure which Kubernetes resources to watch, whether they are owned or external, how events are handled, and which predicates filter events. Helper functions (`WatchOwned`, `WatchExternal`, `WatchWithHandler`) and commonly used predicates are provided for convenience. This replaces the old `OwnedTypes` approach.
* Added the `FieldIndex` struct and `FieldIndexProvider` interface, allowing providers to declare field indexes for efficient querying of related resources, which is especially useful when watching external resources. The reconciler now sets up these indexes automatically if implemented.

### Reconciler and Metrics Enhancements

* Added a `WithMetrics` option for configuring the metrics server bind address, including the ability to disable metrics entirely. The reconciler now passes this configuration to the controller manager.

### Context Utility Methods

* Added utility methods to the `Context` struct for accessing the underlying `context.Context` and Kubernetes client, improving ergonomics for provider implementations.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/oapi-codegen/runtime v1.1.1
 	github.com/operator-framework/api v0.33.0
 	github.com/percona/everest-operator v0.6.0-dev1.0.20260203073758-bcbe85312503
-	github.com/percona/percona-helm-charts/charts/everest v0.0.0-20260216180230-33983cd2bef3
+	github.com/percona/percona-helm-charts/charts/everest v0.0.0-20260217144450-06d30a0668c2
 	github.com/rodaine/table v1.3.0
 	github.com/spf13/cobra v1.10.1
 	github.com/stretchr/testify v1.11.1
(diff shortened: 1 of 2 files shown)


pkg/sdk/reconciler/provider.go
```diff
@@ -114,11 +139,33 @@ func newReconciler(p providerAdapter, opts ...ReconcilerOption) (*ProviderReconc
...
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), mgrOpts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create manager: %w", err)
 	}
 
+	// Setup field indexes if provider implements FieldIndexProvider
+	if fip, ok := p.(controller.FieldIndexProvider); ok {
+		for _, fi := range fip.FieldIndexes() {
+			if err := mgr.GetFieldIndexer().IndexField(
+				context.Background(),
```
chilagrow: Do you think we should pass this context from earlier code? Creating index feels possibly long operation and handling cancellation signal would be nice.

recharte: Yeah, I agree, thanks for the suggestion.

PR #1885, opened 2026-02-20
add copyright automation

spron-in:
We are going with a custom script to add copyright info.

1. `addlicense` - will skip the files that already have headers (like some have with Percona LLC)
2. `licensei` - outdated, we should not use it.
3. `skyuwalking-eyes` - it is going to strip percona headers and replace with our. Which is not right. 

We are using "Copyright (C) {YEAR} The OpenEverest Contributors" as the best practice for OSS projects.

When we are in CNCF, we will need to change it again, but it is ok.

### DIFF
--- a/.github/workflows/copyright.yml
+++ b/.github/workflows/copyright.yml
@@ -1,0 +1,73 @@
+---
+name: Copyright Headers
+
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


recharte (approved):
Thank you 🤗

PR #1895, opened 2026-02-27
feat(api): add OpenAPI types generation and provider API integration

solovevayaroslavna:
auto-generated index file for OpenAPI types in `api/openapi/generated/index.ts`

### DIFF
--- a/ui/api/README.md
+++ b/ui/api/README.md
@@ -19,8 +19,8 @@ make gen-openapi-ts-types
 
 This produces:
 
-- `ui/api/crds.gen.types.ts` — from `crds.gen.yaml`
-- `ui/api/http-api.types.ts` — from `http-api.yaml`
+- `ui/api/crds.gen.types.ts` — from `api/openapi/crds.gen.yaml`
+- `ui/api/http-api.types.ts` — from `api/openapi/http-api.yaml`
 - `ui/api/index.ts` — barrel re-exporting all types under named namespaces
 
 ### Regenerate types for a specific file


api/openapi/Makefile
chilagrow, line 1: I think the code in this Makefile can be part of ui/Makefile. `ui/Makefile` already has `pnpm install` that's why it feels like it should be there 😉

ui/api/crds.gen.types.ts
chilagrow, line 1: After some thoughts, I think the generated files can be `ui/api`, `api/ui` or `api-ui`. I don't think you necessary need `generated` directory, files define themselves they were generated 🤗. Also I thought it belongs outside openapi directory, `api/openapi` contains yaml files which can be consumed both by backend and frontend.

solovevayaroslavna: Thanks! I fixed the structure and moved code of makefile to the ui/makefile in [this commit](https://github.com/openeverest/openeverest/pull/1895/commits/a6b711d651fea70628b3630c67ab7b6f5f4c5ea9), so now it should looks better, but maybe not perfect

Makefile
```diff
@@ -33,6 +33,10 @@ CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 gen-crds-openapi: ## Extract OpenAPI schemas from CRD manifests.
 	go run hack/gen-crds-openapi/main.go
 
+.PHONY: gen-openapi-ts-types
+gen-openapi-ts-types: ## Generate TypeScript types from all OpenAPI YAML files in api/openapi/.
+	$(MAKE) -C ui generate-openapi-types
+
 .PHONY: gen
 gen: gen-crds-deepcopy gen-crds-manifests gen-crds-openapi ## Generate code.
```
recharte: I think `gen` should also call `gen-openapi-ts-types`

```suggestion
gen: gen-crds-deepcopy gen-crds-manifests gen-crds-openapi gen-openapi-ts-types ## Generate code.
```

ui/api/README.md
```diff
@@ -0,0 +1,32 @@
+# OpenAPI Schemas
+
+This directory contains OpenAPI schema YAML files consumed by both the backend and frontend.
```
recharte: This is no longer correct

PR #1896, opened 2026-02-27
manager can be set in controller runtime

chilagrow:


provider-runtime/reconciler/provider.go
```diff
@@ -124,47 +125,57 @@ func WithMetrics(bindAddress string) ReconcilerOption {
 	}
 }
 
+func WithManager(mgr ctrl.Manager) ReconcilerOption {
```
chilagrow, line 128: // TODO: it might be wiser to have it mandatory parameter rather than optional

PR #1922, opened 2026-03-12
Add connection details secret and API

recharte:
This pull request improves how instance connection details are managed and exposed in the OpenEverest API and CRDs. The changes replace the previous `ConnectionURL` and `CredentialSecretRef` fields with a new, standardized `ConnectionSecretRef` field that follows the Service Binding specification. Additionally, the API now exposes a dedicated endpoint for retrieving connection details, and supporting client and OpenAPI schema updates have been made.

Connection details management and specification:

* Replaced `ConnectionURL` and `CredentialSecretRef` in `InstanceStatus` with `ConnectionSecretRef`, which references a Secret containing well-known keys (type, provider, host, port, username, password, uri) following the Service Binding spec. This Secret is auto-generated and owned by the instance for easier integration and cleanup. 
* Added a new `InstanceConnectionDetails` schema to the OpenAPI and client, supporting both standard and provider-specific fields, with custom JSON marshaling/unmarshaling logic.

API enhancements:

* Introduced a new HTTP API endpoint (`/clusters/{cluster}/namespaces/{namespace}/instances/{instance}/connection`) to fetch instance connection details, improving discoverability and usability for clients.

Status and conditions:

* Added a new `ConditionConnectionDetailsReady` condition type to indicate when the connection details Secret has been populated by the provider.

OpenAPI and CRD generation improvements:

* Enhanced the CRD OpenAPI generation tool to extract Go type schemas from specified source files, allowing for easier extension and alignment between Go types and OpenAPI schemas.

These changes standardize and simplify how connection information is managed and accessed, making integration with external tools and clients much easier.

### DIFF
--- a/hack/gen-crds-openapi/main.go
+++ b/hack/gen-crds-openapi/main.go
@@ -136,7 +136,7 @@ func extractGoTypeSchemas(goFile string, schemas map[string]*openapi3.SchemaRef)
 				}
 				exportName = extractExportMarker(ts.Doc)
 				if exportName != "" {
-					if schema, err := buildSchemaFromStruct(ts, exportName); err == nil {
+					if schema, err := buildSchemaFromStruct(ts, exportName, ts.Doc); err == nil {
 						schemas[exportName] = openapi3.NewSchemaRef("", schema)
 						fmt.Printf("  Extracted Go type schema for %s from %s\n", exportName, filepath.Base(goFile))
 					} else {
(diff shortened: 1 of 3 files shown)


hack/gen-crds-openapi/main.go
```diff
@@ -73,6 +111,238 @@ func run(crdDir, outputFile string) error {
 	return nil
 }
 
+// extractGoTypeSchemas parses a Go source file and adds OpenAPI schemas for any
+// struct types annotated with the marker comment "// +openapi:export=<SchemaName>".
```
chilagrow, line 115: A simple test with testdata containing `openapi:export` annotation would be great!

PR #1924, opened 2026-03-12
Ensure index.html exists before building Everest API server to prevent build failures

solovevayaroslavna:
backend rebuild loop:

- tiltfile watches backend/public for backend-build
- make build always did touch ./public/dist/index.html
- that file timestamp changed every build, which retriggered backend-build, which restarted everest-server repeatedly
- during that loop, /healthz and /v1/settings intermittently failed with connection refused /empty reply, so login broke

Makefile change should help to check that placeholder file is created only if missing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,8 @@ build-server-helper: GOARCH = amd64
 build-server-helper: $(LOCALBIN)
 # We need to ensure that /public/dist/index.html exists before building Everest
 # API server because it's embedded into the binary and missing file will cause
-# build failure.
+# build failure. We avoid touching the file if it already exists to prevent
+# unnecessary rebuilds when only the timestamp of the file changes.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -73,7 +73,7 @@ build-server-helper: $(LOCALBIN)
 # We need to ensure that /public/dist/index.html exists before building Everest
 # API server because it's embedded into the binary and missing file will cause
 # build failure.
-	mkdir -p ./public/dist && touch ./public/dist/index.html
+	mkdir -p ./public/dist && [ -f ./public/dist/index.html ] || touch ./public/dist/index.html
```
chilagrow, line 77: Could you add comments about why we need conditional check before `touch` please 🙏

recharte: [5932418](https://github.com/openeverest/openeverest/pull/1924/commits/5932418e840dc47e8e4b9e9f12fb81020982e71d)

Issue #1799, opened 2026-01-26
PSMDB Provider: PMM monitoring


recharte:
### Task Description

Define and reconcile components/custom fields for monitoring using PMM.

### Reasoning / Benefit

Move to extensible provider framework

PR #1977, opened 2026-03-30
Add monitoring config api-tests and update server to pass tests

chilagrow:
This PR adds monitoring config API tests

* Add monitoring config API tests to run in CI
* v2 branch uses v2 branch of helm-charts in CI
* Fix monitoring config impl in server so tests pass
* PMM is optional field in MonitoringConfig CR

Refs #1799.


api/monitoring/v1alpha1/monitoringconfig_types.go
```diff
@@ -39,7 +39,7 @@ type MonitoringConfigSpec struct {
 	// It contains `apiKey` key with the API key value.
 	CredentialsSecretName string `json:"credentialsSecretName"`
 	// PMM is configuration specific for monitoring using PMM tool.
-	PMM PMMConfig `json:"pmm,omitempty"`
+	PMM *PMMConfig `json:"pmm,omitempty"`
```
chilagrow: I think `MonitoringConfigSpec` is not quite right. I think `CredentialsSecretName` and `VerifyTLS` should both be under PMM.

recharte: Why do you think they should be under PMM? Maybe I'm missing the big picture but I think that credentials and TLS verification would be fields that would exist for other monitoring solutions as well. The schema of the secret's data might be different but it would most likely need to exist for other technologies as well. What is your reasoning for moving these to the PMMConfig?

chilagrow: Thank you! Yeah you are right about them possibly required for other technologies. My thought was everything that are needed in PMM are under PMM, as we don't yet have use case for others yet. FYI prometheus, victoria metrics, datadog, coroot won't need MonitoringConfig. Having things under PMM has disadvantage of having bloated config, while giving the indication what are the supported config for PMM.

Yeah let's not have them under PMM. I copied URL under PMM but that's not quite right, it should be on the same level as TLS verification.

recharte (approved):

PR #2000, opened 2026-04-09
Version bundles

recharte:
This pull request enhances the provider and instance specification models by adding support for "version bundles." Version bundles are curated sets of component versions that are known to be compatible, making it easier for users to select and manage compatible software versions for their database instances. The changes also ensure that upgrades are safe and predictable by tracking the effective bundle in the instance status and preventing unintended upgrades on provider changes.

The most important changes are:

**Version Bundle Support in API and CRDs:**

* Added a `Version` field to `InstanceSpec` and `InstanceStatus` to allow users to select a version bundle and to record the effective bundle applied, respectively. This ensures safe upgrades and prevents unexpected changes when providers are updated.
* Added a `Versions` field to `ProviderSpec` to define named version bundles, each specifying component versions and a default bundle. Also introduced the `VersionBundle` struct. 

**Provider Runtime Helpers:**

* Added helper functions in `provider-runtime/controller/metadata.go` to resolve version bundles, get the default bundle, and fetch images for specific component versions, making it easier to work with version bundles in the controller logic.

**Validation Enhancements:**

* Extended provider spec validation to ensure that all component versions referenced in version bundles exist in the component catalog, preventing configuration errors.


chilagrow (approved):
Great!

recharte:
@chilagrow I had to change the versions to an array instead of a map in order to unify it with the existing component types versions and to integrate better with the existing UI schema that has `optionsPath` and `optionsPathConfig` that expect an array instead of a map.

PR #2001, opened 2026-04-10
Add kubernetes metrics to PMM

chilagrow:
This PR send kubernetes metrics to PMM server.

* Adding a MonitoringConfig adds remote writes configuration in `VMAgent`
* It copies secret to monitoring namespace used by `VMAgent`
* Add finalizer to clean up secret
* Tests were run locally, CI is added in #2011. 

helm-charts openeverest/helm-charts#30, failing integration test is due to not having correct cluster role, merging helm-charts PR first will fix it.


config/manager/manager.yaml
```diff
@@ -74,6 +74,7 @@ spec:
         args:
           - --leader-elect
           - --health-probe-bind-address=:8081
+          - --monitoring-namespace=everest-monitoring
```
recharte, line 77: We don't need to do it now but we might want to review the namespaces names.

recharte (approved):
Well done 🚀

PR #2003, opened 2026-04-13
Update monitoringconfig CR to extract generic field

chilagrow:
`MonitoringConfig` CR should have generic fields not under PMM. This PR moves `url` from PMM, https://github.com/openeverest/openeverest/pull/1977#discussion_r3061470208.

This PR only changes CR, because I learnt that CR change needs to be applied to `helm-charts` then api-tests in CI can use updated `helm-charts`.

provider openeverest/provider-percona-server-mongodb#24.
helm-charts openeverest/helm-charts#28.

Refs #1799.


api/monitoring/v1alpha1/monitoringconfig_types.go
recharte: Unrelated to this PR's changes but just a heads-up. Since we are changing this CRD in a backward incompatible way we'll eventually need to change its version or change the CRD name.

recharte (approved):

api/monitoring/v1alpha1/monitoringconfig_types.go
chilagrow: Thanks, yeah that's right. Do we want to start changing version for these breaking changes before we release v2? CRD versions and v2 are separate thing but I've been thinking of keeping the same version until we release, but that's technically not correct reading about CRD versioning. So make sense to change the version to v1alpha2 here.

Issue #2002, opened 2026-04-10
Add Controller tests for v2


chilagrow:
### Task Description

We are introducing controllers in v2 branch and adding CR tests would be beneficial. While we have api tests for end to end test cases, more fine grained functionality is suited for CR tests. We use kuttl in openeverest-operator and something similar would be useful.

### Reasoning / Benefit

Add more test coverage for v2 controllers.

PR #2011, opened 2026-04-14
Add CI for controller tests

chilagrow:
This PR adds CI workflow for controller tests.

It also builds separate docker images for just controller.

Although this PR does not add kuttl or chainsaw tests, it sets up kuttl test environment suitable for `MonitoringConfig` controller test. It checks controller deployment is available, so we get good idea of how long it takes to setup tests. It's around 3-4min.

Refs #2002.


build/package/server/Dockerfile
```diff
@@ -17,3 +17,12 @@ COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
...
 ENTRYPOINT ["/everest-api"]
+
+FROM scratch AS controller
+
+WORKDIR /
+
+COPY ./bin/manager  /everest-controller
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+ENTRYPOINT ["/everest-controller"]
```
chilagrow, line 28: It now creates a separate image for controller only, this is used for testing. The openeverest image contains both server and API as was before.

.github/actions/install-kubernetes-tools/action.yaml
```diff
@@ -21,6 +22,14 @@ inputs:
     description: 'K3D version to install'
     required: false
     default: '5.8.3'
+  install-kuttl:
```
chilagrow, line 25: Do we prefer to install chainsaw instead?

recharte: It's fine for now.

recharte (approved):

PR #2030, opened 2026-04-20
Hybrid backup architecture: dual execution modes, Restore CRD, and provider-runtime `BackupProvider` interface

recharte:
### Summary

This PR delivers a complete redesign of the OpenEverest backup system. The central change is the introduction of two mutually exclusive execution modes on `BackupClass`: `Job` (the existing external-tool approach) and `ProviderManaged` (engine-native agents coordinated by the provider's own reconciler). The PR covers all three layers — CRD API, provider-runtime SDK, and runtime reconciler wiring — and keeps both modes fully operational without overlap.

The changes are organized in three layers:
1. **CRD API surface** — `BackupClass`, `BackupStorage` (new), `Backup`, `Restore` (new), and `Instance`.
2. **Provider-runtime SDK** — the `BackupProvider` interface, `BackupConfigError`, and new `Context` helper methods.
3. **Runtime reconciler wiring** — ancillary reconcilers for `ProviderManaged` Backup/Restore CRs, and guarded Job-mode controllers.

---

### Motivation: why not just use Jobs?

The original architecture ran every backup as a Kubernetes Job that connected to the database from outside (e.g., `mongodump`, `pg_dump`, `mysqldump`). This approach has hard limits:

| Concern | Job-based | ProviderManaged |
|---|---|---|
| **PITR** | Not possible — requires continuous oplog/WAL streaming | Fully supported via engine agents (PBM, pgBackRest) |
| **Scheduling** | Platform-owned CronJobs, divorced from the engine scheduler | Engine manages its own schedule; platform mirrors results as Backup CRs |
| **Storage registration** | Every Backup CR re-describes its own storage, causing agent reconfiguration on every backup | Storages are registered once on the Instance; provider syncs them at reconcile time |
| **Snapshot consistency** | External tools can miss in-flight transactions | Engine agents coordinate internally for application-consistent snapshots |
| **Restore fidelity** | Full-dump restores only | Point-in-time recovery to sub-second granularity |
| **Operator overlap** | Reimplements logic the engine operator already owns | Delegates to the operator as the authoritative controller |

Job mode is preserved for engines without native backup agents and for external/offline workflows. `BackupClass.spec.executionMode` makes the choice explicit and operator-configurable without any provider code changes.

---

#### `BackupClass` (cluster-scoped)

- Added `spec.executionMode: ProviderManaged | Job` — the central dispatch field. Both controllers and the provider-runtime check this before touching a Backup or Restore CR.
- Added `spec.providerManaged.supportsPITR: bool` — advertises PITR capability to the Restore admission webhook.
- Added `spec.job` / `spec.restoreJob` (`JobExecution`) — scoped to `executionMode: Job`. Must be unset for `ProviderManaged`.
- Added `spec.instanceConstraints` — optional compatibility requirements an Instance must satisfy to use the class.
- Added `spec.config` / `spec.restoreConfig` (OpenAPI v3 schema) — `Backup.spec.config` and `Restore.spec.config` are validated against these at admission time.

#### `BackupStorage` (namespaced, new)

A reusable, namespaced reference to an S3-compatible object store plus its credentials. Decoupling storage from individual Backup CRs is what makes `ProviderManaged` practical: the provider registers a fixed set of storages on the engine at Instance reconcile time instead of recomputing them from a dynamic list of Backup CRs.

Key fields: `spec.type` (`s3`), `spec.s3.{bucket, region, endpointURL, verifyTLS, forcePathStyle, credentialsSecretName}`. The `accessKeyId` / `secretAccessKey` convenience fields are write-only — a webhook stores them in the referenced Secret and immediately clears the fields so credentials are never persisted on the CR.

#### `Backup` (namespaced)

- Added `spec.backupClassName` — explicit BackupClass reference; `executionMode` is resolved from it.
- Added `spec.storageName` — references a `BackupStorage` in the same namespace.
- Added `spec.scheduleName` — set by the provider's mirroring loop for engine-produced scheduled backups; empty for on-demand Backups.
- Added `status.executionMode` — recorded at start time for observability.
- Added `status.engineBackupRef` — typed object reference to the engine-native resource (e.g., `PerconaServerMongoDBBackup`) for `ProviderManaged` classes.

#### `Restore` (namespaced, new)

A new CRD representing a restore operation. Symmetrically structured to `Backup`.

- `spec.dataSource.backupName` — references a `Backup` CR; BackupClass and storage are resolved transitively.
- `spec.dataSource.external` — describes a backup with no `Backup` CR (`backupClassName` + `storageName`).
- `spec.dataSource.pitr` — point-in-time recovery target (`type: date | latest`, `date: RFC3339`). Validated at admission to require `spec.providerManaged.supportsPITR: true` on the resolved class.
- `status.engineRestoreRef` — mirrors `engineBackupRef` for restores.
- `status.state` — `Pending | Running | Succeeded | Failed | Error`.

#### `Instance` (namespaced)

- Added `spec.backup`:
  - `enabled` — master toggle.
  - `classRef.name` — the `BackupClass` to use.
  - `storages[]` — `{name, storageRef, main}` list registering `BackupStorage` resources on the engine.
  - `schedules[]` — `{name, enabled, cron, storageName, retentionCopies}` list. Translated by the provider into the engine's native scheduler. The runtime does **not** spawn CronJobs for `ProviderManaged` classes.
  - `pitr` — `{enabled, storageName, config}` for continuous log streaming.
- Added `spec.dataSource` — bootstraps a fresh Instance from an existing backup, mirroring `RestoreDataSource` without importing the backup package.

---

#### `BackupProvider` interface (`provider-runtime/controller/interface.go`)

An optional interface providers implement to handle `ProviderManaged` Backup and Restore CRs. When detected on the provider struct, the runtime automatically:

- Registers `backup.openeverest.io/v1alpha1` in the manager scheme.
- Registers field indexes on `Backup.spec.instanceName` and `Restore.spec.instanceName`.
- Starts ancillary `Backup` and `Restore` reconcilers that filter on `executionMode: ProviderManaged` and dispatch to the provider's methods.

```go
type BackupProvider interface {
    SyncBackup(c *Context, backup *backupv1alpha1.Backup) (BackupExecutionStatus, error)
    SyncRestore(c *Context, restore *backupv1alpha1.Restore) (RestoreExecutionStatus, error)
    CleanupBackup(c *Context, backup *backupv1alpha1.Backup) (done bool, err error)
    CleanupRestore(c *Context, restore *backupv1alpha1.Restore) (done bool, err error)
}
```

**Note: `ConfigureBackup` was removed from this interface.** The initial design included a post-Sync hook for translating `Instance.spec.backup` into engine configuration, but it is structurally broken for any provider whose backup config lives inside the engine CR — which is virtually all of them (PSMDB, PXC, CloudNativePG, etc.). Splitting engine CR writes across two hooks (`Sync` → writes CR without backup, `ConfigureBackup` → writes CR with backup) creates an infinite reconcile loop: each write bumps the generation, which fires the watch, which reruns `Sync`, which resets the fields. Backup wiring belongs inside `Sync`, which produces a single complete desired state for the engine CR in one write.

#### `BackupConfigError` (`provider-runtime/controller/common.go`)

A typed error providers return from `Sync` when backup configuration fails (missing storage, bad credentials, invalid schedule) while the engine itself is healthy. The runtime surfaces it on the `BackupConfigured` condition without marking the Instance `Failed` or blocking the main reconcile loop.

```go
// Inside Sync:
return &controller.BackupConfigError{
    Reason:  "StorageNotFound",
    Message: fmt.Sprintf("BackupStorage %q not found", name),
}
```

#### New `Context` helper methods

```go
func (c *Context) BackupClass(name string) (*backupv1alpha1.BackupClass, error)
func (c *Context) BackupStorage(name string) (*backupv1alpha1.BackupStorage, error)
func (c *Context) BackupStorageCredentials(bs *backupv1alpha1.BackupStorage) (accessKeyID, secretKey string, err error)
func (c *Context) BackupsForInstance() ([]backupv1alpha1.Backup, error)
func (c *Context) RestoresForInstance() ([]backupv1alpha1.Restore, error)
```

---

#### `BackupReconciler` (updated)

The existing Job-mode backup controller is updated to check `executionMode` **before** resetting status or setting up the deferred `Status().Update()`. For `ProviderManaged` Backups it returns immediately without any writes, ceding full status ownership to the provider-runtime. Previously the deferred status write would overwrite whatever the provider had written, causing a conflict loop.

#### `RestoreReconciler` (new)

A new Job-mode restore controller mirroring `BackupReconciler`. Applies the same early-exit guard: resolves the `BackupClass` and checks `executionMode` before touching status. Manages the full restore lifecycle — RBAC, payload Secret, Job creation, and state observation — for `executionMode: Job` restores.

---

### What is not in this PR

- The PSMDB provider `BackupProvider` implementation (`SyncBackup`, `SyncRestore`, `CleanupBackup`, `CleanupRestore`, schedule mirroring) — follow-up provider PR.
- UI changes, API server endpoints, or CLI commands surfacing backup/restore status.
- Kuttl integration tests for the new reconciler paths.


chilagrow (approved):
Looks good!

recharte:
After implementing `ConfigureBackup` for PSMDB we discovered it is structurally broken for any provider whose backup config lives inside the engine CR (PSMDB, PXC, CloudNativePG, MongoDB Community — effectively all of them). The problem: `Sync` calls `Apply(engineCR)` with backup disabled, then `ConfigureBackup` writes the same CR with backup enabled. Each write bumps the generation, which fires the provider's watch, which reruns `Sync`, which resets the backup fields, which triggers `ConfigureBackup` again — an infinite reconcile loop. This was visible in practice: the PSMDB CR reached generation 15,273 within minutes.

We considered keeping the hook with a merge-patch instead of a full Update, but that only suppresses the conflict error, not the write itself. The real fix is to produce a single, complete desired state for the engine CR inside `Sync`. `ConfigureBackup` is therefore removed from the `BackupProvider` interface. We also introduced a separate `BackupConfigured` condition for providers that detect a backup-specific failure (missing storage, bad schedule expression, etc.) inside `Sync` which can now return a `*controller.BackupConfigError`; the runtime surfaces it on `BackupConfigured=False` without failing the Instance or blocking the main reconcile loop.

recharte:
@chilagrow besides removing the `ConfigureBackup` method I added a few new methods to the provide-runtime's `BackupProvider` interface, as well as a Restore CRD and respective controller that follows the same pattern of the Job based Backups we already had before.
The PR description was updated with a summary of all of these.

internal/controller/backup/restore_controller.go
```diff
@@ -0,0 +1,831 @@
...
+				Namespace: restore.GetNamespace(),
+			},
+		},
+	}
+	allGone := true
+
+	for _, res := range resources {
+		err := r.Client.Delete(ctx, res)
+		if err == nil {
+			allGone = false
```
chilagrow, line 651: If delete function returns no error (deleted) or not found for all resources, I would consider resources are allGone. What do you think? 🤔 

At the moment, allGone condition is based on if all resources return not found. I think it means it needs extra reconciliation cycle for deleting resources.

internal/controller/backup/restore_controller.go
```diff
@@ -0,0 +1,831 @@
...
+
+	// Create RBAC resources.
+	requiresRbac := len(bc.Spec.RestoreJob.Permissions) > 0 || len(bc.Spec.RestoreJob.ClusterPermissions) > 0
+	if requiresRbac { //nolint:nestif
+		if err := r.ensureServiceAccount(ctx, restore); err != nil {
+			restore.Status.State = backupv1alpha1.RestoreStateError
+			restore.Status.Message = fmt.Errorf("failed to ensure service account: %w", err).Error()
+			return ctrl.Result{}, err
+		}
+		if err := r.ensureRBACResources(ctx, restore, bc.Spec.RestoreJob.Permissions, bc.Spec.RestoreJob.ClusterPermissions); err != nil {
```
chilagrow, line 201: What do you think of adding `restoreRBACCleanupFinalizer` before `ensureRBACResources`? In `ensureRBACResources` function, it creates roles, roleBindings and a few other thing. It may mean some resources may created while one resource fails to create. 🤔

api/backup/v1alpha1/backupclass_types.go
```diff
@@ -28,39 +28,101 @@ import (
...
+	// configuration accepted by this class. Restore.spec.config is validated
+	// against this schema.
 	// +optional
-	JobSpec *BackupJobSpec `json:"jobSpec,omitempty"`
-	// CleanupJobSpec is the specification of the cleanup job.
+	RestoreConfig BackupClassConfig `json:"restoreConfig,omitempty"`
+	// InstanceConstraints defines compatibility requirements that must be
+	// satisfied by an Instance before this backup class can be used with it.
 	// +optional
-	CleanupJobSpec *BackupJobSpec `json:"cleanupJobSpec,omitempty"`
```
chilagrow, line 45: Do we want to bump CRD version for these changes in BackupClass and Backup CR?

chilagrow (commented):
Thank you for these!

chilagrow:
By the way, failing integration test for monitoringconfigv2 indicates that roles needs to be added to helm-charts. It's the only set of tests for v2 webhook/controller deployment, and they fail for missing roles and stuff.

```
  ✘  31 [monitoring-config-v2] › tests/monitoring-config-v2.spec.ts:32:3 › Monitoring configs tests › create/update/delete monitoring config (218ms)
```

api/backup/v1alpha1/backupclass_types.go
recharte: No, these are new CRDs that are only in development we didn't release any of them yet.

internal/controller/backup/restore_controller.go
recharte: In Kubernetes, a successful `Delete` call does not guarantee the resource is gone — it only guarantees that the API server has accepted the deletion request. If the resource has finalizers (added by this controller, another controller, or an admission webhook), the API server sets `deletionTimestamp` and the resource enters a terminating state. It persists until all finalizers are removed. Treating a successful `Delete` as "resource is gone" would cause the `Restore` finalizer to be removed prematurely, while RBAC resources are still terminating in the cluster.

The only authoritative signal that a resource is truly absent is a `NotFound` error from a subsequent `Get` or `Delete`. The current pattern — setting `allGone = false` on a successful delete and requeuing — is the standard controller-runtime idiom for waiting on dependent resource cleanup. The extra reconcile cycle is intentional: it re-enters `deleteRBAC`, hits `NotFound` for each resource, leaves `allGone = true`, and only then proceeds to remove the finalizer from the `Restore`. This guarantees the cleanup invariant holds regardless of whether the RBAC resources have finalizers now or acquire them in the future.

internal/controller/backup/restore_controller.go
recharte: That's a great point, I'll implement that change, thanks.

internal/controller/backup/restore_controller.go
chilagrow: Thank you 🤗 I missed about successful delete call potentially leaving resources hanging around with finalizer.

PR #2034, opened 2026-04-22
Add backup storage API

chilagrow:
This PR adds backup storage API and API tests.

Merge after #2030.


internal/server/handlers/k8s/backup_storage_v1.go
chilagrow, line 1: Any files with `*_v1.go` suffix is not a new file, but a renamed file. New backup storage v2 logic lives in `backup_storage.go` file.

PR #2040, opened 2026-04-23
Add API for listing all backups of an instance

chilagrow:
This PR adds an endpoint to list backups for a specific instance.

It creates a new instance name label on `Backup` resources for filtering via `MatchingLabels`, avoiding fetching all backups in the namespace.

Manually tested, API test will be added in a separate PR after #2030.

Merge after #2030.


internal/server/handlers/k8s/backup.go
```diff
@@ -30,6 +32,12 @@ func (h *k8sHandler) GetBackup(ctx context.Context, namespace, name string) (*ba
 
 // CreateBackup creates a backup.
 func (h *k8sHandler) CreateBackup(ctx context.Context, backup *backupv1alpha1.Backup) (*backupv1alpha1.Backup, error) {
+	labels := backup.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[common.InstanceNameLabel] = backup.Spec.InstanceName
+	backup.SetLabels(labels)
```
recharte, line 40: It's fine to keep this here but we'll also need to have this logic in the controller otherwise Backups created directly through CRs might not show up in the UI.

recharte (approved):

internal/server/handlers/k8s/backup.go
chilagrow: True, this logic is suitable more in the controller 👍

PR #2041, opened 2026-04-23
[Security] Pin container base images by digest

### LINKED ISSUE #2018: Security: Pin container base images by digest
## Problem
Dockerfiles use mutable tags (e.g., `golang:1.25-alpine`, `alpine:3.22`) which can change between builds. A compromised base image registry could inject malicious code into our published images without changing the Dockerfile.

**Files affected:**
- Dockerfile
- Dockerfile.everestctl  
- Dockerfile.helmtools
- dev/everest-operator.Dockerfile
- dev/everest.Dockerfile

## Solution
Pin to immutable digests with version comments:

```dockerfile
# Before
FROM golang:1.25-alpine AS builder

# After
FROM golang:1.25-alpine@sha256:7a00384194cf2cb68924bbb918d675f1517357433c8541bac0ab2f929b9d5447 AS builder
```

## Getting digests
```bash
docker pull golang:1.25-alpine
docker inspect golang:1.25-alpine --format='{{index .RepoDigests 0}}'
```

## Automation
Renovate/Dependabot can update pinned digests automatically, so we don't lose the convenience of automated updates.

## Impact
- Prevents base image substitution attacks
- Reproducible builds (same digest = same bytes)
- Required for SLSA compliance

Ankit152:
Pin container base images by digest

Fixes #2018

### DIFF
--- a/dev/everest-operator.Dockerfile
+++ b/dev/everest-operator.Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine AS dev
+FROM alpine:3.22@sha256:310c62b5e7ca5b08167e4384c68db0fd2905dd9c7493756d356e893909057601 AS dev
 WORKDIR /home/everest
 RUN adduser -D everest
 USER 1000:1000
@@ -6,7 +6,7 @@ COPY ./bin/manager  ./manager
 ENTRYPOINT ["./manager"]
 
 # Build the Delve debuger
-FROM golang:1.26-alpine AS delve
+FROM golang:1.26-alpine@sha256:f85330846cde1e57ca9ec309382da3b8e6ae3ab943d2739500e08c86393a21b1 AS delve
 RUN go install github.com/go-delve/delve/cmd/dlv@v1.25.2
 RUN chmod +x /go/bin/dlv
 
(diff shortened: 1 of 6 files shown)


dev/everest-operator.Dockerfile
Ankit152: Don't we need to pin the digest here as well?
cc @spron-in

dev/everest.Dockerfile
Ankit152: Same as above ^
cc @spron-in

dev/everest-operator.Dockerfile
spron-in: Even StepSecurity is not showing anything - we need to pin this one to follow best practices. Now it is `:latest`. Can you just pin to the latest stable version? 3.22.4 ?

dev/everest.Dockerfile
spron-in: same as above :

spron-in (changes requested):

dev/everest-operator.Dockerfile
Ankit152: Something like this?

```
FROM alpine:3.22@sha256:310c62b5e7ca5b08167e4384c68db0fd2905dd9c7493756d356e893909057601
```

spron-in: Yeah

Ankit152: Done.

dev/everest.Dockerfile
Ankit152: Done.

spron-in (approved):
lgtm

PR #2047, opened 2026-04-23
Test SSO with keycloak

spron-in:
This PR adds the test for SSO with keycloak.
This test is optional and it only runs when `KEYCLOAK_ENABLED=true` is set. In `playwright.config.ts` the keycloak setup/teardown and sso projects are inside a conditional spread.

How to use:
```
make deploy-keycloak
make test-sso
```

Test results:
```
Running 13 tests using 5 workers

  ✓   1 …ycloak:setup] › tests/setup/keycloak/keycloak.setup.ts:38:3 › Keycloak SSO setup › Wait for Keycloak readiness (482ms)
  ✓   2 …› tests/setup/keycloak/keycloak.setup.ts:42:3 › Keycloak SSO setup › Configure Keycloak realm, client, and user (3.0s)
  ✓   3 …sts/setup/keycloak/keycloak.setup.ts:52:3 › Keycloak SSO setup › Configure Everest OIDC settings via ConfigMap (12.4s)
  ✓   4 …] › tests/setup/keycloak/keycloak.setup.ts:77:3 › Keycloak SSO setup › Login via Keycloak and obtain SSO token (665ms)
  ✓   5 …oak:setup] › tests/setup/keycloak/keycloak.setup.ts:84:3 › Keycloak SSO setup › Grant SSO user admin RBAC role (201ms)
  ✓   6 [sso] › tests/sso/keycloak.spec.ts:21:3 › SSO with Keycloak › OIDC settings are configured (364ms)
  ✓   7 [sso] › tests/sso/keycloak.spec.ts:43:3 › SSO with Keycloak › SSO token can access authenticated endpoints (753ms)
  ✓   8 [sso] › tests/sso/keycloak.spec.ts:55:3 › SSO with Keycloak › logout invalidates SSO session (959ms)
  ✓   9 [sso] › tests/sso/keycloak.spec.ts:72:3 › SSO with Keycloak › invalid SSO token is rejected (329ms)
  ✓  10 [sso] › tests/sso/keycloak.spec.ts:31:3 › SSO with Keycloak › SSO access token is accepted as Bearer token (710ms)
  ✓  11 [sso] › tests/sso/keycloak.spec.ts:81:3 › SSO with Keycloak › expired or tampered JWT is rejected (77ms)
  ✓  12 …own] › tests/teardown/keycloak/keycloak.teardown.ts:22:3 › Keycloak SSO teardown › Reset Everest OIDC settings (10.1s)
  ✓  13 …down] › tests/teardown/keycloak/keycloak.teardown.ts:29:3 › Keycloak SSO teardown › Delete Keycloak deployment (227ms)
::notice title=🎭 Playwright Run Summary::  13 passed (30.4s)
```

### DIFF
--- a/api-tests/Makefile
+++ b/api-tests/Makefile
@@ -66,6 +66,22 @@ deploy-minio:
 		exit 1; \
     fi
 
+KEYCLOAK_RESOURCE=$(TEST_ROOT)/manifests/keycloak.yaml
+.PHONY: deploy-keycloak
+deploy-keycloak: ## Deploy Keycloak to K3D test cluster for SSO tests.
+	$(info Deploying Keycloak to K3D cluster)
+	kubectl apply -f $(KEYCLOAK_RESOURCE)
+	if ! kubectl wait deploy/keycloak --for=condition=Available -n everest-system --timeout=120s; then \
+		echo "❌ Keycloak deployment did not become ready in time"; \
+		echo "🔎 Pods:"; \
+		kubectl get pods -n everest-system -l app=keycloak -o wide || true; \
+		echo "🔎 Events:"; \
+		kubectl get events -n everest-system --sort-by=.lastTimestamp || true; \
+		exit 1; \
+	fi
+	$(info Port-forwarding Keycloak to localhost:8180)
+	kubectl port-forward -n everest-system svc/keycloak 8180:8080 &
+
 .PHONY: k3d-upload-pg-images
 k3d-upload-pg-images: ## Pre-upload Postgresql images to K3D test  cluster to speedup tests execution.
 	$(info Uploading PG images to K3D test cluster)
@@ -151,4 +167,15 @@ test:               ## Run all tests.
 	  cmd_proj="$${cmd_proj} --project=$${project}" ;\
 	done ;\
 	npx playwright test --config=./playwright.config.ts $${cmd_proj} $${FLAGS};\
+	}
+
+.PHONY: test-sso
+test-sso:           ## Run SSO tests with Keycloak (requires deploy-keycloak first).
+	@{ \
+	set -xe ;\
+	export KEYCLOAK_ENABLED="true" ;\
+	export KEYCLOAK_URL="http://localhost:8180" ;\
+	export CI_USER="everest_ci" ;\
+	export CI_PASSWORD="password" ;\
+	npx playwright test --config=./playwright.config.ts --project=sso $${FLAGS};\
 	}
--- a/api-tests/Makefile
+++ b/api-tests/Makefile
@@ -178,4 +178,5 @@ test-sso:           ## Run SSO tests with Keycloak (requires deploy-keycloak fir
 	export CI_USER="everest_ci" ;\
 	export CI_PASSWORD="password" ;\
 	npx playwright test --config=./playwright.config.ts --project=sso $${FLAGS};\
+	$(EVERESTCTL) settings oidc configure --issuer-url="" --client-id="" ;\
 	}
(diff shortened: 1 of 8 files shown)


api-tests/tests/teardown/keycloak/keycloak.teardown.ts
```diff
@@ -0,0 +1,33 @@
...
+import {test as teardown} from '@playwright/test';
+import {CliHelper} from '@helpers/cliHelper';
+
+teardown.describe.serial('Keycloak SSO teardown', () => {
+  const cli = new CliHelper();
+
+  teardown('Reset Everest OIDC settings', async () => {
+    // Reset OIDC configuration by setting empty values
+    await cli.execute(
+      'everestctl settings oidc configure --issuer-url="" --client-id=""',
```
chilagrow: Could we do teardown in `Makefile` using `$(EVERESTCTL)` instead of using `everestctl` dependency within API test? `$(EVERESTCTL)` is the binary built within openeverest project, we can guarantee its existence and correct version.

api-tests/tests/setup/keycloak/keycloak.setup.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // Patch the ConfigMap directly — no network validation, works without port-forward.
+    const oidcYAML = buildOIDCConfigYAML(clientId!);
+    const patch = JSON.stringify({data: {'oidc.config': oidcYAML}});
+    const patchResult = await cli.execute(
+      `kubectl patch configmap everest-settings -n everest-system --type merge -p '${patch}'`,
+    );
+    expect(patchResult.code, `ConfigMap patch failed: ${patchResult.stderr}`).toEqual(0);
+
+    // Restart Everest so it picks up the new OIDC config.
```
chilagrow, line 63: Let me think, this feels like it should be set up before API test starts. Similarly to how `create-test-users` is done, so if restarting everest is needed, it can be done before the test starts. Could SSO user be created in `Makefile` before the test instead of within the test setup stage? 😃

That way, if something fails, we can poke `kubectl` and figure out what's going on.

api-tests/Makefile
```diff
@@ -152,3 +168,14 @@ test:               ## Run all tests.
 	done ;\
 	npx playwright test --config=./playwright.config.ts $${cmd_proj} $${FLAGS};\
 	}
+
+.PHONY: test-sso
```
chilagrow, line 172: We don't want to run this in CI right, any reason why? 🤗  I often notice, things break silently without CI.

chilagrow (commented):
Thank you for this, I have some comments about re-arranging code 🙏

api-tests/tests/teardown/keycloak/keycloak.teardown.ts
spron-in: I did that, thanks @chilagrow

api-tests/tests/setup/keycloak/keycloak.setup.ts
spron-in: Well, I thought about it. It is just that complexity to setup keycloak without playwright (I tried with `curl`) - is almost impossible. It never works for me :) 

So I decided to keep it here.

api-tests/Makefile
spron-in: Yeah, for now let's not go that route. I'm not 100% confident in these tests and we may want to run them manually for some time and then add them into CI.

chilagrow (approved):

chilagrow (commented):
Could you check the copyright header please?

PR #2051, opened 2026-04-24
feat: add SSO token exchange endpoint for opaque OIDC tokens

spron-in:
### Problem

Everest's JWT middleware (`echojwt`) only accepts tokens it signed itself. When using an IdP like Authentik that may issue opaque access tokens, the browser receives a token that Everest cannot parse or verify — causing all authenticated API calls to fail with 400/401.

### Solution

Added `POST /v1/session/sso` — a public endpoint (no JWT middleware) that:

- Accepts any OIDC access token (JWT or opaque)
- Validates it by calling the IdP's UserInfo endpoint
- Issues a short-lived (1h) Everest-signed JWT with an oidc_issuer claim
- All subsequent API calls use the Everest JWT, fully compatible with existing middleware and session blocklist

I created keycloak tests - #2047 - and these tests work with this change and without it.
I also created Authentik tests in `authentic-test` branch. Authentik passes only with the new changes.


ui/apps/everest/src/contexts/auth/auth.provider.tsx
```diff
@@ -137,7 +137,11 @@ const AuthProvider = ({ children, isSsoEnabled }: AuthProviderProps) => {
     try {
       const newLoggedUser = await userManager.signinSilent();
       if (newLoggedUser && newLoggedUser.access_token) {
-        localStorage.setItem('everestToken', newLoggedUser.access_token);
+        // Exchange the refreshed OIDC access token for an Everest JWT.
+        const response = await api.post('/session/sso', {
```
chilagrow, line 141: Do we want to call /session/sso only if sso is enabled? 🤔 

The server returns `BadRequest` error if oidc provider is not set

chilagrow (commented):
The change looks sensible, could you push `authentic-test` branch mentioned in the PR description? That will show me working OIDC token with test 🤗

Lappihuan:
Whats the blocker here?
SSO is not really usable with most 3rd party providers as is

spron-in:
@Lappihuan we need proper testing for this one. Which 3rd party provider are you using? Can you test this PR with it?

Lappihuan:
we are using dex and keycloak, i can try

spron-in:
@Lappihuan that would be awesome. I will resolve the conflicts meanwhile as @recharte also did some changes to sso.

Lappihuan:
@spron-in we only get empty responses from everest-server when building the image and overwriting it in our deployment. we did use the makefile to build the image.

spron-in:
@Lappihuan do you mean in the UI? Where do you see empty responses?

Lappihuan:
@spron-in the whole everest-server only responds with empty http responses.
we must have done something wrong in the image build i assume, since we only overlayed the image path in the deployment, everything else was as is.

PR #2063, opened 2026-04-27
feat: truncate long text database summary preview

### LINKED ISSUE #2035: Truncate long text values in Database Summary preview

When a text field with multiline (textarea) contains a long value, the Database Summary sidebar renders the entire text inline without any truncation. This creates a poor visual experience — the preview panel overflows with a wall of text that obscures other sections

**Current behavior:**
The "Engine configuration" value is rendered as-is inside a `<Typography variant="caption">`, producing an unreadable block that dominates the entire preview panel.

**Expected behavior:**
Long values should be truncated to ~2 lines with an ellipsis. A "Show more" link can expand the full text inline.

CSS line-clamp + JS "more" toggle link

Creation od TruncatedPreviewText component that
- Renders label + value in a <Typography variant="caption">
- Applies CSS line-clamp (2 lines) by default
- Shows a "Show more" / "Show less" <Link> toggle
- Uses React.useState for expanded/collapsed state

"-" — don't truncate the dash placeholder, use PreviewContentText as-is.

_No response_

_No response_

### Additional Context

ui/apps/everest/src/pages/database-form/database-preview/preview-section.tsx renders every field value identically with no truncation
ui/apps/everest/src/pages/database-form/database-preview/dynamic-section-preview/utils.tsx converts the value to a string via String(value) and passes the full text to PreviewContentText. There is no awareness of whether the source component is multiline.

AdityaPimpalkar:
Fixes #2035

https://github.com/user-attachments/assets/d19992af-0d45-4ac8-ada3-6e342f841078

Let me know if this is acceptable. Feedback is appreciated! :)


solovevayaroslavna:
Hey @AdityaPimpalkar! First of all, thank you for your interest in the UI part of the project, it is very valuable! 

Unfortunately, you have done PR for a task that has not been fully described. If we had waited for assign first, we would have avoided this unnecessary work. 

The fact is that this problem applies not only to preview. This is a universal problem that relates to different parts of the database creation and editing flow, and since in v2 we follow the concept: "give the rights to determine exactly what the UI should be to the plugin developer" by writing yaml. We must strictly monitor the universatility and reuse of components. 

That is, the problem of this issue is much deeper:
It is necessary to investigate in which other places we need to crop large text, what to do if the text is too large, say 1000 lines? I will click on more and what happens next? They won't fit in this interface. Should I show a modal window with scrolling in this case? What if I'm on the same page and I already have this textarea component with the same text displayed? Should I highlight it instead of re-rendering it as a text in a modal window? 

All these questions need to be answered before starting to work on this issue. Аnd I'm sure as a result, this task will turn into an epic with a lot of todos on board.

So I'm sorry, but we won't be able to accept this PR because it doesn't follow our real goals. But thank you so much for your participation, we've been receiving a lot of PR in recent days, and today [we've improved the process](https://github.com/openeverest/openeverest/pull/2077) by making the assign issue process more explicit.

AdityaPimpalkar:
Thank you for the explanation @solovevayaroslavna! Apologies for jumping on it quickly without confirming the full context, I will follow the contributing guidelines closely next time. 👍

PR #2069, opened 2026-04-27
Provider runtime interfaces to support scheduled backups

recharte:


provider-runtime/reconciler/mirror.go
```diff
@@ -0,0 +1,118 @@
...
+		out, err := bm.Mirror(ctx, c, src)
+		if err != nil {
+			return nil, err
+		}
+		if out == nil {
+			return nil, nil
+		}
+		return out, nil
+	}
+	return setupWatchAndCreate(mgr, providerName+"-backup-mirror", bm.OperatorBackupType(), translate)
```
chilagrow, line 117: Would it simplify if we pass `Mirror` interface directly to `setupWatchAndCreate(` func, then `translate` function is not necessary? 🤔

recharte: Agree, I simplified the Mirror reconciler in #2030.

PR #2072, opened 2026-04-28
Replace kuttl tests with chainsaw

chilagrow:
This PR migrates kuttl tests to chainsaw tests, and run them in CI.

Migrated kuttl config and tests using `chainsaw migrate kuttl config` and `chainsaw migrate kuttl tests`. Then removed kuttl specific CR fields such as `TestStep` from the tests.


test/integration/monitoring/monitoringconfig/chainsaw-test.yaml
```diff
@@ -118,6 +118,8 @@ spec:
               metadata:
                 name: mc-test
                 finalizers:
+                  - openeverest.io/vmagent
+                  - openeverest.io/cleanup-secrets
```
chilagrow, line 122: These chainsaw tests were not run in CI before, since then finalizers were added.

PR #2075, opened 2026-04-28
chore: [security] scope workflow permissions to least privilege (#2015)

### LINKED ISSUE #2015: Security: Add least-privilege permissions to all workflows
## Problem
Multiple workflows lack explicit `permissions` declarations, defaulting to wide access. Some workflows grant `write` permissions at the top level unnecessarily.

**Current violations:**
- api-docs.yml, cli-tests.yml, dev-*.yml: no topLevel permission defined
- dev-be-ci.yaml: unnecessary `packages: write` and `checks: write` at top level
- release.yml: `contents: write` at top level

## Solution
Apply principle of least privilege:

1. Set minimal top-level permissions:
```yaml
permissions:
  contents: read  # Minimum required for checkout
```

2. Grant specific permissions only at job level where needed:
```yaml
jobs:
  release:
    permissions:
      contents: write  # Only this job can write
      packages: write
```

## Reference
OpenSSF Scorecard check: https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

vyagh:
Went through all workflows and added missing top level permissions blocks, moved write permissions down to the specific jobs that need them. 
one more thing, dev-build.yaml was pushing to GHCR without packages: write explicitly set, fixed that too
fixes #2015

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -3,6 +3,9 @@ name: API Docs
 on:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   rdme-openapi:
     runs-on: ubuntu-latest
--- a/.github/workflows/auto-update-base.yaml
+++ b/.github/workflows/auto-update-base.yaml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   auto-update-base:
     runs-on: ubuntu-latest
--- a/.github/workflows/cli-tests.yml
+++ b/.github/workflows/cli-tests.yml
@@ -8,6 +8,9 @@ on:
         type: string
         required: true
 
+permissions:
+  contents: read
+
 env:
   GONOPROXY: 'github.com/percona'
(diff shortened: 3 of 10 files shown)


vyagh:
@spron-in

spron-in:
Thanks @vyagh . 
I'm not sure if it is okay to remove `read` permissions from release.yml. They were in write for some reason there. Maybe it is safer to set them all to `read`?

This one looks good, but would require deeper validation before merging.

vyagh:
hey, I might be a bit confused, but the write permissions aren't removed, i just moved them to job level. 

build job still has `contents: write` and `packages: write` (for action-gh-release and GHCR pushes
`checks: write` and `pull-requests: write` were removed because no step in release.yml uses themin the current workflow, nothing that would write checks or pr comments.

should i still keep the writes at top level for release.yml specifically?

vyagh:
or should i backtrack to the older commits to understand the situation?

spron-in:
@vyagh I did some fixes. Now it should be ok.

vyagh:
thanks a lot @spron-in

spron-in (approved):

PR #2084, opened 2026-04-28
test: add unit tests for pkg/accounts/cli (100% function coverage)

### LINKED ISSUE #2083: test: add unit tests for pkg/accounts/cli to improve code coverage

The file `pkg/accounts/cli/accounts.go` has zero test coverage for its 6 exported methods: `Create`, `Delete`, `SetPassword`, `List`, `GetInitAdminPassword`, and `CreateRSAKeyPair`.

The only existing test file (`pkg/accounts/cli/accounts_test.go`) covers only `ValidateUsername` and `ValidatePassword` from `utils.go`, and also contains a silent bug — valid-case assertions use `require.ErrorIs(t, err, nil)` instead of `require.NoError(t, err)`, so passing cases are never actually verified.

Add unit tests in pkg/accounts/cli/accounts_test.go  to cover all 6 untested methods using a mock for `accounts.Interface` (already defined in `pkg/accounts/types.go`).

_No response_

_No response_

### Additional Context

_No response_

adity1raut:
Improve the Test Coverage From 20%  to 76%

- Fixed #2083 

image

### DIFF
--- a/pkg/accounts/cli/accounts.go
+++ b/pkg/accounts/cli/accounts.go
@@ -29,9 +29,16 @@ import (
 	"github.com/percona/everest/pkg/accounts"
 	cliutils "github.com/percona/everest/pkg/cli/utils"
 	"github.com/percona/everest/pkg/common"
-	"github.com/percona/everest/pkg/kubernetes"
 	"github.com/percona/everest/pkg/output"
 )
+
+// osExit is a variable so it can be replaced in tests to avoid calling os.Exit.
+var osExit = os.Exit
+
+// rsaKeyCreator is the minimal interface needed by Accounts for RSA key pair operations.
+type rsaKeyCreator interface {
+	CreateRSAKeyPair(ctx context.Context) error
+}
 
 type (
 	// Config holds the configuration for the accounts subcommands.
@@ -47,7 +54,7 @@ type (
 		accountManager accounts.Interface
 		l              *zap.SugaredLogger
 		config         Config
-		kubeClient     kubernetes.KubernetesConnector
+		kubeClient     rsaKeyCreator
 	}
 )
 
@@ -240,7 +247,7 @@ func (c *Accounts) CreateRSAKeyPair(ctx context.Context) error {
 	c.l.Info("Creating/Updating JWT keys and restarting Everest.")
 	if err := c.kubeClient.CreateRSAKeyPair(ctx); err != nil {
 		c.l.Error(err)
-		os.Exit(1)
+		osExit(1)
 	}
 
 	c.l.Info("JWT keys have been created/updated successfully")
--- a/pkg/accounts/cli/accounts_test.go
+++ b/pkg/accounts/cli/accounts_test.go
@@ -1,13 +1,100 @@
+// everest
+// Copyright (C) 2023 Percona LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package cli
 
 import (
+	"context"
+	"errors"
+	"os"
 	"testing"
 
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/percona/everest/pkg/accounts"
+	"github.com/percona/everest/pkg/common"
 )
+
+// ---------------------------------------------------------------------------
+// Mocks
+// ---------------------------------------------------------------------------
+
+type mockAccountManager struct {
+	mock.Mock
+}
+
+func (m *mockAccountManager) Create(ctx context.Context, username, password string) error {
+	return m.Called(ctx, username, password).Error(0)
+}
+
+func (m *mockAccountManager) Get(ctx context.Context, username string) (*accounts.Account, error) {
+	args := m.Called(ctx, username)
+	acc, _ := args.Get(0).(*accounts.Account)
+	return acc, args.Error(1)
+}
+
+func (m *mockAccountManager) List(ctx context.Context) (map[string]*accounts.Account, error) {
+	args := m.Called(ctx)
+	list, _ := args.Get(0).(map[string]*accounts.Account)
+	return list, args.Error(1)
+}
+
+func (m *mockAccountManager) Delete(ctx context.Context, username string) error {
+	return m.Called(ctx, username).Error(0)
+}
+
+func (m *mockAccountManager) SetPassword(ctx context.Context, username, newPassword string, secure bool) error {
+	return m.Called(ctx, username, newPassword, secure).Error(0)
+}
+
+func (m *mockAccountManager) Verify(ctx context.Context, username, password string) error {
+	return m.Called(ctx, username, password).Error(0)
+}
+
+func (m *mockAccountManager) IsSecure(ctx context.Context, username string) (bool, error) {
+	args := m.Called(ctx, username)
+	return args.Bool(0), args.Error(1)
+}
+
+type mockRSACreator struct {
+	err error
+}
+
+func (m *mockRSACreator) CreateRSAKeyPair(_ context.Context) error {
+	return m.err
+}
+
+// newTestAccounts bypasses NewAccounts (which needs a real k8s cluster).
+func newTestAccounts(mgr accounts.Interface, kube rsaKeyCreator, pretty bool) *Accounts {
+	return &Accounts{
+		accountManager: mgr,
+		kubeClient:     kube,
+		l:              zap.NewNop().Sugar(),
+		config:         Config{Pretty: pretty},
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Existing validation tests (fixed: now asserts both valid AND invalid cases)
+// ---------------------------------------------------------------------------
 
 func TestUsernamePasswordSanitation(t *testing.T) {
 	t.Parallel()
+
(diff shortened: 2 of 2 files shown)


pkg/accounts/cli/accounts.go
```diff
@@ -29,10 +29,17 @@ import (
 	"github.com/percona/everest/pkg/accounts"
 	cliutils "github.com/percona/everest/pkg/cli/utils"
 	"github.com/percona/everest/pkg/common"
-	"github.com/percona/everest/pkg/kubernetes"
 	"github.com/percona/everest/pkg/output"
 )
 
+// osExit is a variable so it can be replaced in tests to avoid calling os.Exit.
+var osExit = os.Exit
```
recharte, line 36: Three separate problems with this one, and I'd rather lose the seam than solve any of them.

First, the position I'd like us to hold, stated properly since it isn't written down anywhere yet: we don't add production code to make a test possible, and we don't chase a coverage percentage. Coverage tells you what was executed, not what was checked — I'd take five methods tested well over six tested to move a number. This variable exists only so a test can observe `os.Exit`, and it makes the shipped binary's behaviour reassignable by anything in the package to buy that.

Second, it fails `gochecknoglobals`, which is on for this package (`.golangci.yml` runs `default: all`, and the `gochecknoglobals` exclusion is scoped to `^commands/`). Its allowlist is `_`, `version`, `err`/`Err` error values, `regexp.MustCompile` and `//go:embed` — a `var` holding `os.Exit` isn't any of those, so `make check` will flag it once this rebases.

Third, the test it enables asserts a state that can't occur. With the real `os.Exit` the function never returns; with the stub it does, so `CreateRSAKeyPair` carries on, logs "keys created successfully" and returns `nil` after a failure. The subtest passes on a path production never takes.

So: delete the var, the `osExit(1)` call site and the `kube_error_calls_exit` subtest, and leave `CreateRSAKeyPair` uncovered. If we want its failure path testable later, the honest change is for it to return the error and let `commands/accounts/reset_jwt_keys.go` do the exiting — worth its own PR.

pkg/accounts/cli/accounts.go
```diff
@@ -29,10 +29,17 @@ import (
...
+
+// rsaKeyCreator is the minimal interface needed by Accounts for RSA key pair operations.
+type rsaKeyCreator interface {
```
recharte, line 39: Narrowing the dependency to the one method actually used is a good change on its own merits — it's just a production refactor arriving inside a test PR, which CONTRIBUTING asks us to split ("one concern per pull request", "land preparatory work first").

If the `osExit` seam goes (comment above), this loses its only consumer too: nothing else in the package needs the narrower type, and `CreateRSAKeyPair` stops being tested. So I'd drop it here and keep this PR test-only. Happy to take it separately if you want it.

pkg/accounts/cli/accounts_test.go
```diff
@@ -52,15 +140,336 @@ func TestUsernamePasswordSanitation(t *testing.T) {
...
+		require.Equal(t, "hash123", hash)
+		mgr.AssertExpectations(t)
+	})
+}
+
+// ---------------------------------------------------------------------------
+// CreateRSAKeyPair
+// ---------------------------------------------------------------------------
+
+func TestCreateRSAKeyPair(t *testing.T) {
```
recharte, line 437: I'd drop this whole test, for a different reason per subtest:

- `kube_error_calls_exit` depends on the `osExit` seam, which I've commented on in `accounts.go` — and with the stub in place the function returns and reports success after a failure, which is not a state production can reach.
- `success_not_pretty` asserts `require.NoError` on a function whose only return statement is `return nil`. It can't fail.
- `success_pretty` asserts the same thing as the one above, and to do it swaps the package-level `os.Stdout` from a `t.Parallel()` subtest. That races with the sibling tests that read it — `Delete` and `SetPassword` both write to `os.Stdout` on their `Pretty` path — and `make test` runs with `-race`:

```
$ go test -race -count=10 ./pkg/accounts/cli/...
WARNING: DATA RACE
Read at 0x000005a74268 by goroutine 205:
  ...cli.(*Accounts).Delete()  accounts.go:130
  ...cli.TestDelete.func4()    accounts_test.go:264
Previous write at 0x000005a74268 by goroutine 198:
  ...cli.TestCreateRSAKeyPair.func3.1()  accounts_test.go:468
FAIL	github.com/percona/everest/pkg/accounts/cli	0.145s
```

A single run passes, which is why it looks fine locally — it takes a few iterations before the windows overlap.

Removing the test also removes the need for `mockRSACreator` and for the `rsaKeyCreator` change in `accounts.go`, so the PR becomes test-only.

pkg/accounts/cli/accounts_test.go
```diff
@@ -52,15 +140,336 @@ func TestUsernamePasswordSanitation(t *testing.T) {
...
+		t.Parallel()
+		mgr := &mockAccountManager{}
+		mgr.On("Create", ctx, "validuser", "validpassword").Return(nil)
+		a := newTestAccounts(mgr, nil, false)
+		err := a.Create(ctx, CreateOptions{Username: "validuser", Password: "validpassword"})
+		require.NoError(t, err)
+		mgr.AssertExpectations(t)
+	})
+
+	t.Run("success_pretty", func(t *testing.T) {
```
recharte, line 212: These two subtests are identical apart from `Pretty`, and neither observes what `Pretty` changes — the only difference in the code under test is a line printed to stdout that nothing captures. Same pattern in `TestDelete` and `TestSetPassword`, so it's three duplicated subtests.

They lift the coverage number without adding a way for the suite to fail, which is the kind of test I'd rather not carry — so I'd drop the `_pretty` half of each pair. The alternative, making the output injectable and asserting on it, is a bigger change than this PR should take on.

pkg/accounts/cli/accounts_test.go
```diff
@@ -52,15 +140,336 @@ func TestUsernamePasswordSanitation(t *testing.T) {
...
+		mgr := &mockAccountManager{}
+		listErr := errors.New("list failed")
+		mgr.On("List", ctx).Return(nil, listErr)
+		a := newTestAccounts(mgr, nil, false)
+		err := a.List(ctx, ListOptions{})
+		require.ErrorIs(t, err, listErr)
+		mgr.AssertExpectations(t)
+	})
+
+	t.Run("default_columns_with_headers", func(t *testing.T) {
```
recharte, line 347: `default_columns_with_headers`, `no_headers` and `custom_columns` all assert only `require.NoError`, so all three pass identically whatever the header and column logic does — which is the one thing in `List` with real branching worth testing.

`rodaine/table` renders to `table.DefaultWriter`, so pointing that at a buffer (in a non-parallel test, given the global) and asserting the rendered rows would turn these three into genuine tests. If that's more than you want to take on here, I'd keep `manager_error` and drop the other three rather than leave assertions that can't fail.

pkg/accounts/cli/accounts_test.go
```diff
@@ -1,13 +1,100 @@
...
+
+	"github.com/percona/everest/pkg/accounts"
+	"github.com/percona/everest/pkg/common"
 )
 
+// ---------------------------------------------------------------------------
+// Mocks
+// ---------------------------------------------------------------------------
+
+type mockAccountManager struct {
```
recharte, line 36: These tests mostly assert that the wrapper calls through — `On("Create", ctx, ...)` followed by `AssertExpectations` restates the implementation, so it'll fail on any refactor that keeps the behaviour intact.

A ~30-line in-memory `accounts.Interface` (a `map[string]*accounts.Account` behind a mutex) would let the same subtests assert outcomes instead: the account exists after `Create`, it doesn't after `Delete`, an invalid username never reaches the store, `GetInitAdminPassword` returns the hash and refuses once secure. `pkg/accounts/tests.go` already takes that shape for the real implementations.

Worth knowing for next time too: where we do want a generated mock we use mockery rather than hand-writing one — see `pkg/rbac/gen.go` and `pkg/version_service/gen.go`.

pkg/accounts/cli/accounts_test.go
```diff
@@ -52,15 +140,336 @@ func TestUsernamePasswordSanitation(t *testing.T) {
...
 			})
 		}
 	})
 }
+
+// ---------------------------------------------------------------------------
+// WithAccountManager
+// ---------------------------------------------------------------------------
+
+func TestWithAccountManager(t *testing.T) {
```
recharte, line 161: This one only checks that a setter sets. It can't fail unless the assignment is deleted, in which case every other test in the file fails first — I'd drop it.

pkg/accounts/cli/accounts_test.go
```diff
@@ -27,16 +114,17 @@ func TestUsernamePasswordSanitation(t *testing.T) {
 			t.Run(tc.name, func(t *testing.T) {
 				t.Parallel()
 				err := ValidateUsername(tc.username)
-				if tc.expectedErr == nil {
+				if tc.expectedErr != nil {
```
recharte, line 117: This is the real fix in the PR: the old guard meant every invalid row asserted nothing, so the validation regexes were effectively untested. Good catch.

recharte (changes requested):
Thanks for this, @adity1raut, and sorry it waited so long for a look — I've triaged the issue and the coverage gap is real, so the direction is settled: see #2083 for that part.

The comments below come down to one ask: make this test-only. The two production changes (`osExit`, `rsaKeyCreator`) exist to reach 100% function coverage, and I'd rather not carry them — a percentage isn't the goal, tests that can fail are. Dropping `TestCreateRSAKeyPair` removes the need for both, and takes a reproducible `-race` failure with it.

Three things outside the diff, for whenever you pick this up:

- The branch is behind `main`, and `CreateRSAKeyPair` lost its `error` return in #2531, so `err := a.CreateRSAKeyPair(ctx)` won't compile after a rebase.
- The new file header needs the OpenEverest copyright line as well as the Percona one — `make copyright-headers` adds it.
- `main` is v1 and `release-2.0` is v2; nothing merges between them, and the package is identical on both, so this wants a second PR against `release-2.0` to land properly.

Note that no CI has run here beyond DCO — this is a fork PR, so the workflows are held until a maintainer releases them. I'll do that on your next push so you get real signal.

PR #2086, opened 2026-04-28
Fix token blocklist update error handling

### LINKED ISSUE #2085: Token invalidation can silently succeed even when the blocklist secret update fails, leaving tokens valid after logout.
### What is the issue?

The `Add()` function in `session` package returns `err` not `updateErr` when client fails to update the secret, here `err` will be `nil` when the update fails and `Block()` which wraps `Add()` won't be able to add the token to the `blocklist`.

### Steps to Reproduce

1. Configure a scenario where updating the blocklist secret fails (e.g., revoke permissions or simulate API failure).
2. Trigger token invalidation (logout or password change).
3. Attempt to use the same token against a protected endpoint.

### Relevant Logs / Output

_No response_

### OpenEverest Version

commit: 457bde6

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

blenbot:
## Summary
Fix token blocklist updates so token invalidation does not silently succeed when the secret update fails.

Fixes #2085 

## Context
UpdateSecret failures previously returned a nil error from Add(), which made blocklist operations report success even when no token was added.

### DIFF
--- a/pkg/session/token_store.go
+++ b/pkg/session/token_store.go
@@ -1,5 +1,6 @@
 // everest
 // Copyright (C) 2025 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pkg/session/token_store.go
```diff
@@ -97,7 +98,7 @@ func (ts *tokenStore) Add(ctx context.Context, shortenedToken string) error {
 	_, updateErr := ts.client.UpdateSecret(ctx, secret)
 	if updateErr != nil {
 		ts.l.Errorf("failed to update %s secret in the %s namespace with the %s shortened token, retrying: %v", secret.Name, secret.Namespace, shortenedToken, updateErr)
-		return err
+		return updateErr
```
chilagrow: Is there any reason why you can't use idiomatic `err` naming instead of using `updateErr`?

chilagrow (commented):
Thanks for your contribution!

pkg/session/token_store.go
blenbot: no specific reason, `updateErr` was declared to get error related to update before so I used it, we can assign `err` instead of declaring `updateErr` making use of the idiomatic naming

chilagrow: Let's use idiomatic `err` naming please 🤗.

blenbot: all done @chilagrow 
tested on my end as well

chilagrow (commented):
Please check the failed API test in CI, https://github.com/openeverest/openeverest/actions/runs/25934308549/job/76235708587?pr=2086 🙏

chilagrow (approved):
Thank you for your contribution!

PR #2097, opened 2026-04-30
Add `API` tests for instance and instance backup

chilagrow:
This PR adds APIs test for listing backup of instances and very simple CRUD tests for instances.


api-tests/tests/utils/api.ts
recharte, line 1: This seems like boilerplate that we could replace with some autogenerated stuff. I did a very brief research about which tools could be used for this purpose and 2 stood out:
* [openapi-typescript](https://github.com/openapi-ts/openapi-typescript)
* [orval](https://github.com/orval-labs/orval)

We are already using `openapi-typescript` for generating the types of [http-api.types.ts](https://github.com/openeverest/openeverest/blob/v2/ui/api/http-api.types.ts) so it would be great to leverage to also generate client code. However, I read that even though `openapi-typescript` can generate some client code, it's not its primary focus. The client output tends to be more hook-based (React Query patterns) rather than the traditional function pattern we have here in the `api.ts`.

On the other hand, going with `orval` means introducing one extra dependency but it was purpose-built for client code generation. It generates functions that closely match this pattern but can also generate hooks if needed. I think we could even use this for the core UI code and replace everything in [ui/apps/everest/src/api](https://github.com/openeverest/openeverest/tree/v2/ui/apps/everest/src/api) with autogenerated code from `orval`. @solovevayaroslavna any thoughts about this?

I asked AI what this would look like for both use cases and I think that there's potential for using this to simplify our lives and reduce boilerplate. Here's how this generated code and respective integration is supposed to look for both the API tests and the UI code according to claude:


## orval (traditional client mode)

Generated output (`http-api.client.ts`):
```typescript
// Auto-generated by orval
import type { Instance, CreateInstanceParams } from './http-api.types'

export const createInstance = (data: CreateInstanceParams, options?: RequestInit) =>
  request.post(`/v1/clusters/${CLUSTER_NAME}/namespaces/${EVEREST_CI_NAMESPACE}/instances`, { data, ...options })

export const getInstance = (name: string, options?: RequestInit) =>
  request.get(`/v1/clusters/${CLUSTER_NAME}/namespaces/${EVEREST_CI_NAMESPACE}/instances/${name}`, options)

export const updateInstance = (name: string, data: CreateInstanceParams, options?: RequestInit) =>
  request.put(`/v1/clusters/${CLUSTER_NAME}/namespaces/${EVEREST_CI_NAMESPACE}/instances/${name}`, { data, ...options })

export const deleteInstance = (name: string, options?: RequestInit) =>
  request.delete(`/v1/clusters/${CLUSTER_NAME}/namespaces/${EVEREST_CI_NAMESPACE}/instances/${name}`, options)
```

**In your `api.ts`:**
```typescript
// You can either use generated directly or wrap it:
export { createInstance, getInstance, updateInstance, deleteInstance } from './http-api.client'

// Or add error-checking wrapper:
export const createInstanceWithCheck = async (request, data) => {
  const response = await createInstance(data)
  await checkError(response)
  return response.json()
}
```

---

## orval (React hooks mode)

```typescript
// Auto-generated by orval with hooks option
export const useCreateInstance = (options?: UseMutationOptions<Instance>) => {
  return useMutation({
    mutationFn: (data: CreateInstanceParams) => createInstance(data),
    ...options
  })
}

export const useGetInstance = (name: string, options?: UseQueryOptions<Instance>) => {
  return useQuery({
    queryKey: ['instance', name],
    queryFn: () => getInstance(name),
    ...options
  })
}
```

Then in components:
```typescript
const { mutate: create } = useCreateInstance()
const { data: instance } = useGetInstance(name)
```

@chilagrow @solovevayaroslavna what do you think? Is it worth pursuing this for both API tests and UI code? Is this well suited for the API tests but not for UI? Let me know your thoughts.

chilagrow: This would remove one additional step of creating apis necessary for tests 👍 I think it's sensible for API tests.

solovevayaroslavna: @recharte @chilagrow  Good idea to explore this. For the API tests, I think orval in traditional client mode is a clear win — it would eliminate the manual boilerplate and keep the test client in sync with the OpenAPI spec.

For the core UI code api I'd be more careful. The current API layer has custom logic beyond simple HTTP calls (payload construction, type transformations, axios interceptors), and not all our shared-types don't match the auto-generated OpenAPI types 1:1. On top of that, the v2 Instance API is still evolving, so generated code would break frequently during this phase. So i would suggest to mark it as some "idea" and to keep it untill beta for example.

I think the right approach would be to start with API tests, then once the v2 API stabilizes, try generating a client for the new Instance API endpoints only and see how it fits alongside the existing hooks layer. Replacing the entire src/api/ folder in my opinion would be a large migration with a lot of edge cases and we can thought about it, but not now, and not even before beta i suppose.

recharte: Makes sense, let's start with the API tests and then re-evaluate this in the future for the rest.

chilagrow: Thanks guys!

chilagrow: See how orval works in API tests openeverest/openeverest#2234.

PR #2101, opened 2026-04-30
Fix security insights comment

spron-in:
I hope it is the last fix - adding more repositories into the mix and adding comment mandatory field


recharte (commented):
Should we also add https://github.com/openeverest/openeverest-operator?

spron-in:
> Should we also add https://github.com/openeverest/openeverest-operator?

makes sense. Fixed, @recharte !

recharte (approved):

PR #2103, opened 2026-04-30
feat(ui-generator): add section group type (#1859)

### LINKED ISSUE #1859: UI: Generator: Bordered group (groupParams.bordered)
> **Reframes #1859 "Section".** `groupType: section` is dropped — the bordered card becomes a
> **`groupParams.bordered`** param. Group-level `modes` is split into its own issue. Supersedes PR #2103.

### Task Description

Add **`groupParams.bordered: boolean`**. When `true`, the group renders as a bordered card (reusing
[RoundedBox](https://github.com/openeverest/openeverest/blob/main/ui/apps/everest/src/components/rounded-box/RoundedBox.tsx),
like [FormCard](https://github.com/openeverest/openeverest/blob/main/ui/apps/everest/src/components/form-card/form-card.tsx)):

- **title** (`sectionHeading` typography) from `label`;
- optional **description** (`caption` typography) from `description`;
- **children** in the card body.

```yaml
# 1. bordered with title + description
storageSection:
  uiType: group
  groupParams: { bordered: true }
  label: Storage
  description: Defines the type and performance of storage for your database.
  components:
    storageClass: { uiType: select, path: spec.engine.storage.class }
  componentsOrder: [storageClass]

# 2. bordered without description (description is optional → no subtitle line)
resourcesSection:
  uiType: group
  groupParams: { bordered: true }
  label: Resources
  components:
    cpu: { uiType: number, path: spec.engine.resources.cpu }
    memory: { uiType: number, path: spec.engine.resources.memory }
  componentsOrder: [cpu, memory]

# 3. nesting: a bordered card containing a plain (non-bordered) group
advancedSection:
  uiType: group
  groupParams: { bordered: true }
  label: Advanced
  components:
    inner:
      uiType: group          # no groupParams → a plain container, no border

### PR DESCRIPTION
**Pull request by aabhinavvvvvvv**
Closes #1859

## What this PR does

Adds `section` as a new group type to the ui-generator schema in v2. A section group renders as a bordered card with a title and optional description, allowing plugin developers to visually group and label fields in their UI schemas.

## Changes

- **`GroupType` enum** — added `Section = 'section'`
- **`ComponentGroup` type** — added `GroupModeOverrides` type; added `modes` (hidden/disabled per `FormMode`) and `_disabled` (internal flag set by `apply-mode-overrides`)
- **`SectionWrapper`** — new component in `ui-group-wrappers/section-wrapper/`, reuses `RoundedBox` with `sectionHeading` title and `caption` description
- **`UIGroup`** — fixed to forward `description`, `groupParams`, and `_disabled` to wrapper components (previously dropped)
- **`componentGroupMap`** — registered `SectionWrapper` for `GroupType.Section`
- **`apply-mode-overrides`** — added group-level mode handling: `hidden` sets `uiType: 'hidden'`, `disabled` sets `_disabled: true`
- **Tests** — 6 unit tests for `SectionWrapper`, 5 new tests for group-level mode overrides

## Usage example

```yaml
storageSection:
  uiType: group
  groupType: section
  label: Storage
  description: Defines the type and performance of storage for your database.
  components:
    storageClass:
      uiType: select
      path: spec.components.engine.storage.storageClass
      fieldParams:
        label: Storage Class
```
## Screenshots

**Before** — flat layout, fields directly under Advanced configuration:
image

**After** — fields wrapped in bordered section cards:
image


ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.md
```diff
@@ -0,0 +1,107 @@
+# Section Group
```
solovevayaroslavna: Hey! Can you please check how common documentation for ui-generator was done? We have docs/ui/ui-generator root for that

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.md
```diff
@@ -0,0 +1,107 @@
...
+| `label`           | `string`    | Card title. Rendered with `sectionHeading` typography.              |
+| `description`     | `string`    | Helper text below the title. Rendered with `caption` typography.    |
+| `components`      | `object`    | Child components or nested groups.                                  |
+| `componentsOrder` | `string[]`  | Render order of `components` keys.                                  |
+| `modes`           | `object`    | Group-level mode overrides — `hidden` or `disabled` per `FormMode`. |
+
+## Usage
+
+```yaml
+storageSection:
```
solovevayaroslavna: it would be good to have a ui (screen/video demo of how it's working), for example now we have https://github.com/openeverest/provider-percona-server-mongodb/blob/main/definition/topologies/replicaSet/topology.yaml (the same for topology spec yaml) and there we can modify existing engine parametrs field (this is a screen from v1). We can implement section now without toggle, just to test visual part. + also storage class field also can be wrapped with the section as it was done previously.

Image

Image
in this case we can check and test a lot of staff (like margins between sections, how sections will be rendered and so on.

solovevayaroslavna: so I strongly recommend adding photo and video materials and preparing a second branch of integration with the provider to simplify the review process (like simple switch to the branch of the provider, switch to this branch, and launch openeverest, being able to see live how the sections will work together with ui-generator) at the moment, it is very difficult to conduct a full-fledged review, because I have to manually add all these changes to the ui-generator to check, can you please help with that?

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  it('applies disabled styling when disabled is true', () => {
+    const { container } = render(
+      <SectionWrapper label="Storage" disabled>
+        content
+      </SectionWrapper>
+    );
+    // The outermost percona-rounded-box element should carry opacity styling
+    const box = container.querySelector('.percona-rounded-box');
+    expect(box).toBeInTheDocument();
```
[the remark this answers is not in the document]
solovevayaroslavna: agree with this comment

ui/apps/everest/src/components/ui-generator/utils/preprocess/apply-mode-overrides.ts
```diff
@@ -54,8 +54,15 @@ const applyModeToComponent = (
     'components' in item
   ) {
     const group = item as ComponentGroup;
+    const groupOverrides = group.modes?.[mode];
+
+    if (groupOverrides?.hidden) {
+      return { ...group, uiType: 'hidden' };
+    }
```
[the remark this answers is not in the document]
solovevayaroslavna: Agree with that comment
Once uiType becomes 'hidden', the group will be treated as a leaf component — but it has no path, no fieldParams, and no validation, so the schema builder and renderer will either crash or silently produce incorrect results.

ui/apps/everest/src/components/ui-generator/ui-group/ui-group.tsx
```diff
@@ -28,6 +36,9 @@ const UIGroup = ({
         React.createElement(Component, {
           children,
           label: item?.label,
+          description: item?.description,
+          disabled: item?._disabled,
+          ...groupParams,
```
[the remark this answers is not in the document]
solovevayaroslavna: Currently groupParams is typed as `Record<string, unknown>` and is spread directly into the wrapper component [...groupParams]. Unlike fieldParams, which has discriminated types per field type (NumberFieldParams, TextFieldParams, SelectFieldParams via FieldParamsMap), there is no equivalent GroupParamsMap that maps each GroupType to its specific props type.

This means any arbitrary key in [groupParams] will be passed through without type checking, and as noted in the comment, it can accidentally override computed props like disabled

I'd suggest Introducing a GroupParamsMap (similar to FieldParamsMap) that maps GroupType.Section → SectionGroupParams, GroupType.Accordion → AccordionGroupParams, etc.
Or at minimum, picking only known keys from groupParams before spreading.

Yes, we should have done this refactoring earlier, when adding the first groups, but at that time there was no perfect idea of how this should scale and be supported in the future, so this task is the best time to carry out such a small refactoring as separating the types of groups

ui/apps/everest/src/components/ui-generator/utils/preprocess/apply-mode-overrides.ts
```diff
@@ -54,8 +54,15 @@ const applyModeToComponent = (
...
+
     return {
       ...group,
+      _disabled: groupOverrides?.disabled ?? group._disabled,
```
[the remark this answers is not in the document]
solovevayaroslavna: agree

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { SectionWrapperProps } from './section-wrapper.types';
+
+const SectionWrapper = ({
+  label,
+  description,
+  disabled = false,
+  children,
+}: SectionWrapperProps) => {
+  return (
+    <RoundedBox
```
solovevayaroslavna, line 36: Can you please check how FormCard was done in v1 in main?  SectionWrapper is essentially a simplified FormCard without a controlComponent. This is OK for v2, but it's worth considering that in the future engine parameters (enable/disable switch) will need a controlComponent, so i see much sense in reusing what we have previously, than adding new one

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.tsx
```diff
@@ -0,0 +1,50 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Box, Typography } from '@mui/material';
+import RoundedBox from 'components/rounded-box';
+import { SectionWrapperProps } from './section-wrapper.types';
+
+const SectionWrapper = ({
```
solovevayaroslavna, line 19: Comparing SectionWrapperProps with the existing FormCard component and the other wrapper types, a few properties seem worth considering for future-proofing, some of them can be added lates as toggle enabled/disabled, but next i think should be done with this task:

fieldName — Both AccordionWrapper and StackWrapper accept fieldName in their props type, sectionwrapper doesn't, which breaks interface consistency across wrappers. 
it can be useful for future:
 - For id, data-testid, aria-controls, and aria-labelledby groups.
- For logging, analytics, or debug output for a specific group.
- To display errors at the group level, if group-level validation ever appears.
and so on

+
data-testid — Would make it easier to target specific sections in tests.

solovevayaroslavna (changes requested):
Hey! Thank you for your contribution!

ui/apps/everest/src/components/ui-generator/utils/preprocess/apply-mode-overrides.ts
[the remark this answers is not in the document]
aabhinavvvvvvv: I've fixed by adding an explicit _hidden: true flag instead of modifying uiType. Consumers downstream of that (renderer, schema builder, is-section-editable) now look for _hidden to skip hidden sections.

ui/apps/everest/src/components/ui-generator/utils/preprocess/apply-mode-overrides.ts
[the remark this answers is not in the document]
aabhinavvvvvvv: Fixed - now _disabled: groupOverrides?.disabled === true ? true : undefined, exclusively mode-based, with no fallback carryover.

ui/apps/everest/src/components/ui-generator/ui-group/ui-group.tsx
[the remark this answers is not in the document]
aabhinavvvvvvv: I fixed this issue with the groupParams thing. Now its spread first, which means the computed stuff for label, description, and disabled will always take priority, I think that makes sense. 
Also, I added this new GroupParamsMap...maps ach GroupType to its own specific params type. So now groupParams gets typed as GroupParamsMap[GroupType], not that old Record<string, unknown> which was too loose.

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.test.tsx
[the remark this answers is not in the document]
aabhinavvvvvvv: I've added expect(box).toHaveStyle({ opacity: '0.5', pointerEvents: 'none' }).

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.md
aabhinavvvvvvv: Moved to docs/ui/ui-generator/groups.md following the existing pattern.

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.md
aabhinavvvvvvv: I added some before and after screenshots right into the PR description, so its easier to see the changes. 
The branch named feature/1859-section-group-integration exists as an integration branch in the repository at https://github.com/aabhinavvvvvvv/provider-percona-server-mongodb. The advanced section in definition/topologies/replicaSet/topology.yaml now wraps storageClass and configuration in section groups. You can check both branches together with this branch to see the sections appear live during the Advanced Configurations process.

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.tsx
aabhinavvvvvvv: The FormCard in main now shows SectionWrapper which uses the same header structure as its header and allows for an optional React.ReactNode controlComponent to be added. The SectionWrapper module needs FormCard because it exists in the ui-generator subsystem and FormCard uses a different API which uses cardContent instead of children and requires title to be passed as string instead of React.ReactNode. The requirement for direct reuse makes me willing to change our current project path.

ui/apps/everest/src/components/ui-generator/ui-group-wrappers/section-wrapper/section-wrapper.tsx
aabhinavvvvvvv: Both added to SectionWrapperProps. fieldName is forwarded from render-component.tsx → UIGroup → SectionWrapper and used to auto-generate data-testid as section-{fieldName} when no explicit one is provided.

aabhinavvvvvvv:
@solovevayaroslavna Thanks for the detailed review... I've addressed your comments. The failing checks are infrastructure flakes and not related to code changes in this PR

PR #2104, opened 2026-04-30
fix e2e pg cluster correct schedules restore test

### LINKED ISSUE #2100: E2E test "PG Cluster correct schedules restore" fails due to hardcoded PostgreSQL version 15.13
### What is the issue?

The E2E test PG Cluster correct schedules restore in db-restore-to-new-cluster.e2e.ts consistently fails on all PRs targeting the main branch. The test hardcodes dbVersion: '15.13' when creating a PostgreSQL cluster via the API. Since the pg-operator was upgraded to 2.9.0, version 15.13 is no longer in the available versions list, causing the API to return HTTP 400:
```
invalid request
15.13 is not in available versions list
```

The failure occurs at db-cluster.ts:147 (expect(response.ok()).toBeTruthy()) during createDbClusterFn, where POST /v1/namespaces/everest-ui/database-clusters returns status 400.
A secondary assertion at line 271 (expect(await dbVersionCombobox.textContent()).toBe('15.13')) would also fail for the same reason.

### Steps to Reproduce

- Open any PR targeting the main branch (e.g., PR #1992, PR #2075).
- Wait for the CI workflow dev-fe-gatekeeper → dev-fe-e2e to run.
- Observe the PG Cluster correct schedules restore test failing in all 3 retry attempts within ~1.8s each.
- In the Everest server logs, find: POST /v1/namespaces/everest-ui/database-clusters returning status 400 with body "15.13 is not in available versions list".

### Relevant Logs / Output

_No response_

### OpenEverest Version

main

### Installation Method

Manifests / Manual

### Kubernetes Distribution

Other (please specify below)

### Environment Details

k3d (k3s in Docker), as configured in the CI pipeline

### Possible Solution

The fix should replace the hardcoded version with a dynamically fetched one (e.g., using getEnginesVersions() or the lastVersion pattern already used elsewhere in the test suite).

### Additional Context

_No response_

blenbot:
Fixes: #2100 

Fix PG restore E2E to use an available Postgres version, avoiding failures when older versions are removed from the API.

### DIFF
--- a/ui/apps/everest/.e2e/pr/db-cluster/db-cluster-overview.e2e.ts
+++ b/ui/apps/everest/.e2e/pr/db-cluster/db-cluster-overview.e2e.ts
@@ -34,13 +34,16 @@ test.describe('DB Cluster Overview', async () => {
       request,
       'mysql'
     );
+    // MySQL 8.4.x uses 3GB memory for "Small" preset, other versions use 2GB
+    const isMySQL84x = /^8\.4\./.test(mysqlVersion);
+    const smallMemory = isMySQL84x ? 3 : 2;
     await createDbClusterFn(request, {
       dbName: dbClusterName,
       dbVersion: mysqlVersion,
       dbType: 'mysql',
       numberOfNodes: '1',
       cpu: 1,
-      memory: 2,
+      memory: smallMemory,
       proxyCpu: 0.5,
       proxyMemory: 0.8,
       externalAccess: true,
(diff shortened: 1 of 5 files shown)


solovevayaroslavna:
@blenbot Thank you very much! Yes, let's try adding a helper for other db types as well. It seems that we have already edited the mongo version once in exactly the same situation. So this is a common problem when updating operators. 

Aslo, please, pay attention to the contribution guide to fix the copyright headers pipline (you can also solve it by running `make preflight` locally in the `u`i folder, it performs all the necessary preparations before transferring the PR to the ready for review status

Thanks!

solovevayaroslavna:
@blenbot Hey! I think everything is fine now, let's wait for pipline so I can handle it

solovevayaroslavna:
Hey @blenbot! Thanks for your contribution! I found a problem and fixed in https://github.com/openeverest/openeverest/pull/2104/commits/019c4b91c643529902ed1bad7cbfde0f3b64a1c0, so PR was merged to main

PR #2112, opened 2026-05-02
fix: eject auth interceptor from correct request pipeline

ashnaaseth2325-oss:
## SUMMARY

Fixes incorrect interceptor cleanup in the API layer where the auth interceptor was being ejected from the wrong Axios pipeline.
This change ensures proper removal of the request interceptor during logout.
Primarily affects `ui/apps/everest/src/api/api.ts` in the API module.

---

## FIX

```ts
// Before
api.interceptors.response.eject(authInterceptor);

// After
api.interceptors.request.eject(authInterceptor);
```


ashnaaseth2325-oss:
Hello @spron-in @recharte 
This PR fixes the auth interceptor being removed from the wrong pipeline, so it now cleans up properly and avoids duplicates over time.

solovevayaroslavna (dismissed):
@ashnaaseth2325-oss Thank you very much! You are absolutely right about this. It's strange that we didn't notice it earlier. 
Please pay attention to the contribution guid to fix the issue with copyright headers pipline 🤗

ashnaaseth2325-oss:
Thanks @solovevayaroslavna!
I’ve added the missing copyright headers and updated the PR. 
Could you please take another quick look?

ashnaaseth2325-oss:
Hello @solovevayaroslavna 
Looks like the E2E tests are failing, but this change is limited to interceptor cleanup in the API layer and shouldn’t affect those flows.
Could you please re-run the workflow or take a look? 
Thanks!

Denyme24:
> Hello @solovevayaroslavna Looks like the E2E tests are failing, but this change is limited to interceptor cleanup in the API layer and shouldn’t affect those flows. Could you please re-run the workflow or take a look? Thanks!

hi @ashnaaseth2325-oss, the failing E2E workflow is caused by pre-existing teardown ordering in policies.e2e.ts / global teardown (dependency resources are deleted before DB clusters are fully terminated), dont think its related to your PR

solovevayaroslavna:
Yes, you are absolutely right. We have a couple of fluky tests on the main, one of them is sharding, the other test is caused by another error and we already have a fix for it, but it's still in review status, so i'm going to merge this manually.
@ashnaaseth2325-oss can you please create a cherry-pick PR for the v2 branch? otherwise, we risk losing this edit when switching to v2

PR #2113, opened 2026-05-02
feat: enforce password reset for first-login sessions

### LINKED ISSUE #2037: First login user experience improvement

Right now when user deploys a fresh OpenEverest instance, it is required to fetch the password that was automatically generated (if user did not set any).
It should be done either with everestctl:
```
everestctl accounts initial-admin-password
```
or with long one-liner with kubectl.

This is an additional friction point and can be improved.

The idea is to prompt the user for the password when UI is opened for the first time. 
We will show the default login (admin) and additional field to set the password. That way the user does not need to return to console anymore and just follow the installation.

This change is not the simplest one, but not hard as well.
1. It requires both frontend and backend changes (some API that will set the password and create the proper secret).
2. We will need to change helm and everestctl outputs. Right now they both instruct the user how to fetch the password.
3. We will need to change the docs, installation part.

_No response_

_No response_

### Additional Context

_No response_

AdityaPimpalkar:
Fixes #2037

This PR is first of the few to improve the first install experience for OpenEverest. It implements the JWT token to carry must_change_password and the restricted-session middleware to check if the initial password was reset by the user.

Note that some of lines have been commented out purposely for follow up PR's context.

Feedback is appreciated! :)


internal/server/middlewares.go
```diff
@@ -80,6 +86,46 @@ func (e *EverestServer) shouldAllowRequestDuringEngineUpgrade(c echo.Context) (b
...
+		}
+		claims, ok := token.Claims.(jwt.MapClaims)
+		if !ok {
+			return c.JSON(http.StatusForbidden, api.Error{
+				Message: pointer.ToString("failed to get claims from token"),
+			})
+		}
+		mustChangePassword, ok := claims[session.MustChangePasswordClaim].(bool)
+		if !ok {
+			if !ok {
```
areebahmeddd: duplicate if !ok check

internal/server/middlewares.go
```diff
@@ -80,6 +86,46 @@ func (e *EverestServer) shouldAllowRequestDuringEngineUpgrade(c echo.Context) (b
...
+				return errors.New("failed to parse claim from token")
+			}
+		}
+		_ = mustChangePassword
+		// if mustChangePassword {
+		// 	return c.JSON(http.StatusForbidden, api.Error{
+		// 		Message: pointer.ToString("password change required"),
```
areebahmeddd, line 120: core enforcement block is commented out

internal/server/middlewares.go
```diff
@@ -80,6 +86,46 @@ func (e *EverestServer) shouldAllowRequestDuringEngineUpgrade(c echo.Context) (b
...
+		requestEndpoint := fmt.Sprintf("%s %s", c.Request().Method, c.Path())
+		if slices.Contains(allowedEndpoints, requestEndpoint) {
+			return next(c)
+		}
+
+		token, err := common.ExtractToken(c.Request().Context())
+		if err != nil {
+			return err
+		}
+		claims, ok := token.Claims.(jwt.MapClaims)
```
areebahmeddd, line 107: token is now signed with everestclaims so cast to everestclaims here instead of mapclaims to keep it consistent with the signing side

pkg/session/manager.go
```diff
@@ -149,7 +159,7 @@ func (mgr *Manager) Create(subject string, secondsBeforeExpiry int64, id string)
 	return mgr.signClaims(claims)
 }
 
-func (mgr *Manager) signClaims(claims jwt.Claims) (string, error) {
+func (mgr *Manager) signClaims(claims EverestClaims) (string, error) {
```
areebahmeddd: prefer keeping the jwt.claims interface here narrowing to a concrete type breaks the existing flexibility and couples the internal helper to a single claim struct 🤔

ui/apps/everest/src/api/api.ts
```diff
@@ -53,15 +55,24 @@ export const addApiErrorInterceptor = () => {
             return;
           }
 
+          if (
+            error.response.status === 403 &&
+            message.includes(PASSWORD_CHANGE_REQUIRED_MESSAGE)
+          ) {
+            location.href = '/logout';
```
areebahmeddd: i think we should not use location.href because it bypasses react router and skips logout cleanup

AdityaPimpalkar:
Hi @areebahmeddd, Thank you for the review!

 I'm still working on the PR as their's some cleanup and tests pending (hence the PR is in draft 😃)
 
 I appreciate the comments nonetheless :)

internal/server/middlewares.go
AdityaPimpalkar: This is by design, it will be uncommented in the follow up PR's (once the UI has necessary changes to handle this)

ui/apps/everest/src/api/api.ts
AdityaPimpalkar: Removed as it was not in the scope of this PR. Will be done separately

pkg/session/manager.go
AdityaPimpalkar: I didn't find any other func using signClaims, but I agree lets keep flexibility where we can :)

internal/server/middlewares.go
AdityaPimpalkar: I found out changing `jwt.MapClaims` to `EverestClaims` is not exactly a single line swap because it's currently flowing through multiple places like JWT middleware and RBAC which is beyond the scope of this PR, could be considered doing separately if its required (not worth it IMO, the amount of changes we need to make for a small win) but its upto the maintainers

internal/server/middlewares.go
```diff
@@ -80,6 +86,44 @@ func (e *EverestServer) shouldAllowRequestDuringEngineUpgrade(c echo.Context) (b
...
 }
 
+func (e *EverestServer) validateIfPasswordChangeIsRequired(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		allowedEndpoints := []string{
+			"GET /v1/accounts",
+			"PUT /v1/accounts",
+			"POST /v1/session",
+			"DELETE /v1/session",
+			"GET /v1/settings",
```
AdityaPimpalkar, line 96: I'm not sure if this route should be whitelisted or not. I noticed it was being called when I landed on the login page. Can someone confirm if it's needed?

PR #2114, opened 2026-05-03
feat: improve backup empty state with actionable CTA

ayushshukla1807:
The current 'No backup found' state in the cluster overview is a passive Alert that doesn't help the user. This PR refactors it to use the existing `EmptyState` component and adds a 'Create backup' button.

I've also added query parameter handling in the Backups list so that clicking the CTA automatically opens the backup creation modal on the destination page. This improves the user flow and leverages our existing design patterns.

### DIFF
--- a/ui/apps/everest/src/pages/db-cluster-details/backups/backups-list/backups-list.tsx
+++ b/ui/apps/everest/src/pages/db-cluster-details/backups/backups-list/backups-list.tsx
@@ -28,8 +28,9 @@ import {
 } from 'hooks';
 import { MRT_ColumnDef } from 'material-react-table';
 import { Alert, Typography } from '@mui/material';
+import { useSearchParams } from 'react-router-dom';
 import { RestoreDbModal } from 'modals/index.ts';
-import { useContext, useMemo, useState } from 'react';
+import { useContext, useEffect, useMemo, useState } from 'react';
 import {
   Backup,
   BackupStatus,
@@ -181,6 +182,16 @@ export const BackupsList = () => {
     setScheduleModalMode(WizardMode.New);
     setOpenScheduleModal(true);
   };
+
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  useEffect(() => {
+    if (searchParams.get('createBackup') === 'true') {
+      handleManualBackup();
+      searchParams.delete('createBackup');
+      setSearchParams(searchParams);
+    }
+  }, [searchParams]);
 
   const handleCloseDeleteDialog = () => {
     setOpenDeleteDialog(false);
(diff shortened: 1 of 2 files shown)


areebahmeddd:
1. this pr is identical to #2115
2. your dco is failing. you should sign commits with `git commit -s -m "..."`

ayushshukla1807:
I've just updated this PR to sign all commits with DCO (-s). Ready for review!

ayushshukla1807:
I've updated this PR to sign all commits with DCO (-s). Ready for review!

solovevayaroslavna:
@ayushshukla1807 Thank you so much for your interest in our project and for the contribution. Unfortunately I don't think this change aligns with the purpose of the overview page.

The overview page is designed to show status and state of the database cluster — not to be an action hub for all actions. We already have a dedicated Backups tab that provides the full backup management experience near (schedules, backup list, creation). There's a tab navigation one click away.

Specific concerns:

**UX principle**: Overview = observe, dedicated tabs = act. Adding a "Create backup" CTA here blurs that boundary without solving a real user problem.
**Implementation complexity for minimal gain**: The query param hack (?createBackup=true + useEffect to auto-open a modal) creates a fragile coupling between two pages. It can cause issues with browser history and unexpected re-triggers.
PR #2114 doesn't check permissions — it shows "Create backup" to all users regardless of RBAC. #2115 fixes this but that's a sign the feature wasn't thought through holistically.
**No user signa**l: There's no linked user feedback, research, or product requirement suggesting users struggle to find the backup creation flow.
If we want to improve the empty state in the backup card, a simple info message with a "Go to Backups" link (navigation, not action) would be more appropriate and consistent with overview's purpose.

We strongly recommend to create an issue first, discuss its necessity in github or in our slack channel, to get the response from the community, to discuss ways to implement it, and only then start working on the code. 

To summarize the feedback: I think there are no prerequisites for changing the current behavior in this way. Happy to reconsider if there's product/design alignment on adding actions to overview cards.

ayushshukla1807:
Hey @solovevayaroslavna , thanks for the detailed breakdown. You're completely right. I tunnel-visioned on reducing click-depth and ended up introducing that query param hack; which, in hindsight, is a pretty fragile way to handle modal state across routes.

I completely agree with the 'Overview = observe, Tabs = act' UX principle. Blurring that boundary creates a bad precedent for the rest of the dashboard. Your suggestion to use a simple 'Go to Backups' navigational link for the empty state makes way more sense. It keeps the card read-only while still guiding the user smoothly to the action hub.

Since you recommended opening an issue first for product alignment, I'll draft up a quick ticket specifically for that navigational link idea so we can get a green light from design before any new code is pushed. Really appreciate you taking the time to explain the v2 context and architectural boundaries here!

PR #2115, opened 2026-05-03
refactor: standardize cluster overview layout and resolve EVEREST-1066

ayushshukla1807:
This PR standardizes the minimum width for labels in the `OverviewSectionRow` component across the cluster overview page.

It addresses **EVEREST-1066** by introducing a `OVERVIEW_SECTION_ROW_MIN_WIDTH` constant and removing inconsistent manual overrides in the `BackupsDetails` card. This ensures a unified alignment for all overview sections, regardless of label length.

Changes:
- Created `overview-section-row.constants.ts` to hold layout tokens.
- Updated `OverviewSectionRow` to use the standardized minWidth.
- Removed hardcoded overrides and resolved the TODO in `backups-details.tsx`.

### DIFF
--- a/ui/apps/everest/src/pages/db-cluster-details/cluster-overview/cards/backups-details.tsx
+++ b/ui/apps/everest/src/pages/db-cluster-details/cluster-overview/cards/backups-details.tsx
@@ -50,7 +50,7 @@ export const BackupsDetails = ({
   loading,
   showStorage = true,
 }: BackupsDetailsOverviewCardProps) => {
-  const { canUpdateDb, dbCluster } = useContext(DbClusterContext);
+  const { canUpdateDb, dbCluster, canCreateBackup } = useContext(DbClusterContext);
   const editable =
     canUpdateDb && !shouldDbActionsBeBlocked(dbCluster?.status?.status);
 
@@ -205,8 +205,8 @@ export const BackupsDetails = ({
             contentSlot={
               <Typography variant="body1">{Messages.titles.noData}</Typography>
             }
-            buttonText="Create backup"
-            onButtonClick={handleCreateBackup}
+            buttonText={canCreateBackup ? 'Create backup' : undefined}
+            onButtonClick={canCreateBackup ? handleCreateBackup : undefined}
           />
         )}
         <OverviewSection
@@ -242,10 +242,8 @@ export const BackupsDetails = ({
           showTooltip={editable && !pitrEditable}
           disabledEditTooltipText={getTooltipText()}
         >
-          {/*// TODO EVEREST-1066 the width of the columns on the layouts in different places is limited by a different number (but not by the content), a discussion with Design is required*/}
           <OverviewSectionRow
             dataTestId="pitr-status"
-            labelProps={{ minWidth: '126px' }}
             label={Messages.fields.status}
             content={
               pitrEnabled ? Messages.fields.enabled : Messages.fields.disabled
@@ -254,7 +252,6 @@ export const BackupsDetails = ({
           {showStorage && (
             <OverviewSectionRow
               dataTestId="backup-storage"
-              labelProps={{ minWidth: '126px' }}
               label={Messages.fields.backupStorages}
               content={pitrStorageName}
             />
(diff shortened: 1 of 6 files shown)


ui/apps/everest/src/pages/db-cluster-details/cluster-overview/cards/backups-details.tsx
```diff
@@ -193,7 +201,13 @@ export const BackupsDetails = ({
             />
           </OverviewSection>
         ) : (
-          <Alert severity="info">{Messages.titles.noData}</Alert>
+          <EmptyState
```
Denyme24, line 204: AFAIK, EmptyState is shown for canReadBackups users when there are no backups, so a CTA button for Creating backups should be gated by create permission right?  e.g., canCreate.
we shouldn't show a create CTA to read-only users.

correct me if i am wrong or in case if i'm missing something.

Denyme24 (commented):
also, looks like this commit isn’t DCO-signed
add a signed-off commit to make it compliant:  git commit -s -m <"your commit message">

ayushshukla1807:
Thanks for the feedback! I've just updated the PR to: 1. Sign all commits with DCO (-s). 2. Implement the canCreateBackup permission check in BackupsDetails.tsx. The 'Create backup' button is now correctly gated behind RBAC permissions, so it won't show for read-only users. Ready for another look!

ui/apps/everest/src/pages/db-cluster-details/cluster-overview/cards/backups-details.tsx
ayushshukla1807: Thanks for the feedback! I've updated the PR to: 1. Sign all commits with DCO (-s). 2. Implement the canCreateBackup permission check in BackupsDetails.tsx. The 'Create backup' button is now correctly gated behind RBAC permissions. Ready for another look!

ayushshukla1807:
Thanks for the feedback! I've updated the PR to: 1. Sign all commits with DCO (-s). 2. Implement the canCreateBackup permission check in BackupsDetails.tsx. The 'Create backup' button is now correctly gated behind RBAC permissions. Ready for another look!

solovevayaroslavna:
Hey @ayushshukla1807! We are very pleased to see the involvement in our product and the desire to make it better, but in this case you are combining several different problems under one PR, which should be solved separately in a good way. Regarding backups, I already answered you in [this comment](https://github.com/openeverest/openeverest/pull/2114#issuecomment-4381631134). 

as for EVEREST-1066:
The TODO comment in the code explicitly says "a discussion with Design is required" — this PR removes the TODO and hardcodes a value without that discussion happening.

A few concerns:
- Different cards = different content. The labels "Status", "PITR" and "Backup storage" are shorter than, say, "Connection string" or "Cluster resources". A single MinWidth of 126px may be excessive for some sections and insufficient for others.
- Responsive design. A hard MinWidth at the component level may not behave well on narrow screens. The current approach with labelProps as a prop gives flexibility — each card can adapt the label to its own content.
- The solution removes flexibility. PR removes labelProps={{ MinWidth: '126px' }} from specific locations and hardcodes the same value into the component itself. Now, if some card needs a different MinWidth, you will have to override it again through labelProps, which makes "standardization" meaningless.

Another reason to think about: in v2, we are developing a completely different approach to displaying information that is completely dependent on the provider and dynamically draws content without knowing about the subcomponents that the cards will display. Therefore, in v2 we can get unexpected content, for the correct rendering of which static restriction and unification are not suitable for us.

ayushshukla1807:
Thanks for catching this. I definitely missed the mark here by trying to force code DRYness over actual component flexibility.

Hardcoding the MinWidth at the component level was a bad call on my end; like you said, it completely breaks responsive design for dynamic content, especially with the v2 provider-driven approach coming down the pipeline. The current labelProps implementation is actually the correct architectural choice because it delegates the sizing constraints to the parent that actually knows the content.

I'll drop this approach entirely. Thanks for the rigorous review, it's incredibly helpful context for understanding how the design system needs to scale.

PR #2143, opened 2026-05-04
fix: ensure logout cleanup runs even if session delete fails

ashnaaseth2325-oss:
### SUMMARY

This PR fixes a logout flow issue where local session cleanup was skipped if the API call failed. The change ensures logout always completes on the client side, regardless of server response. The update primarily affects `ui/apps/everest/src/contexts/auth/auth.provider.tsx` within the `logout()` function.

### FIX

```ts
// Before 
await api.delete('/session', { headers: { token } });

// After 
try {
  await api.delete('/session', { headers: { token } });
} catch {
  // 
}
```

### DIFF
--- a/ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
+++ b/ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
@@ -1,4 +1,8 @@
-/// <reference types="vitest" />
+// Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0
+
 
 import { describe, it, vi } from 'vitest';
 
@@ -18,11 +22,12 @@ vi.mock('./auth.provider', async () => {
   const actual = await vi.importActual<typeof import('./auth.provider')>('./auth.provider');
   return {
     ...actual,
-    default: ({ children }: any) => children, // skip auth logic
+    default: ({ children }: { children: ReactNode }) => children
   };
 });
 
 //Now import everything
+import type { ReactNode } from 'react';
 import { render } from '@testing-library/react';
 import AuthProvider from './auth.provider';
 import * as apiModule from '../../api/api';
--- a/ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
+++ b/ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
@@ -1,45 +1,87 @@
+// everest
 // Copyright (C) 2023 Percona LLC
 // Copyright (C) 2026 The OpenEverest Contributors
 //
-// Licensed under the Apache License, Version 2.0
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
-
-import { describe, it, vi } from 'vitest';
-
-//Mock before imports
-vi.mock('oidc-client-ts', () => ({
-  UserManager: vi.fn(() => ({
-    signinRedirect: vi.fn(),
-    signoutRedirect: vi.fn(),
+// Mocks must be declared before imports
+vi.mock('oidc-react', () => {
+  const userManager = {
+    clearStaleState: vi.fn().mockResolvedValue(undefined),
+    removeUser: vi.fn().mockResolvedValue(undefined),
+    signinSilent: vi.fn().mockResolvedValue(null),
+    signinSilentCallback: vi.fn(),
     getUser: vi.fn().mockResolvedValue(null),
-    removeUser: vi.fn(),
-  })),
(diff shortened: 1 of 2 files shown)


ashnaaseth2325-oss:
Hello @solovevayaroslavna @Denyme24 
Kindly review the PR whenever you get time. 
Thanks!

Denyme24 (commented):
good catch!, the changes made are all in place.
just a suggestion, i think we can add a unit test to mock `api.delete` to throw and assert local/session cleanup is still performed by the logout handler. as far as i can tell there are no unit tests in the repo that cover the specific scenario.

cc: @solovevayaroslavna

ashnaaseth2325-oss:
Hello @Denyme24 
I added a unit test where `api.delete` throws, ensuring logout still clears local/session state and the flow completes as expected.
Also mocked the OIDC client to avoid metadata errors in tests. This should cover the failure path discussed
Happy to add more if needed.

ashnaaseth2325-oss:
Hello @Denyme24 @solovevayaroslavna 
Just following up on this.
Kindly revoew when you have time, thanks!

Denyme24:
hey @ashnaaseth2325-oss, appreciate your efforts, but no need to rush.
also, no need to tag me, as i'm not from the core members. 🙂

maintainers will for sure review.

ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+//Mock AuthProvider before importing it
+vi.mock('./auth.provider', async () => {
+  const actual =
+    await vi.importActual<typeof import('./auth.provider')>('./auth.provider');
+  return {
+    ...actual,
+    default: ({ children }: { children: ReactNode }) => children,
+  };
+});
```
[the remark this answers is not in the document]
solovevayaroslavna: @ashnaaseth2325-oss I think edits are needed here.

solovevayaroslavna (changes requested):

ashnaaseth2325-oss:
Hello @solovevayaroslavna 
Updated the test to use the real logout flow instead of mocking the provider.
Thanks!

ashnaaseth2325-oss:
Hello @solovevayaroslavna
Kindly review the PR whenever you have time
Thanks!

recharte:
Hi @ashnaaseth2325-oss — this has been open since 4 May with four unanswered check-ins, and that's a bad miss on our side. Sorry.

It's also the right patch on the right branch: #2236 got fixed on `release-2.0` by an auth rework that replaced the endpoint outright, but `main` still carries the original code, so this is the only open PR that addresses it — and it predates the issue by ten days.

@solovevayaroslavna can you have another go at this so we can merge this?

recharte:
One small thing worth adding while this waits: a `Fixes #2236` line in the description would link the two properly — it puts this PR in the issue's Development sidebar and closes the issue automatically when this merges. Right now they're only cross-referenced by mention, so the issue would need closing by hand.

ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const removeErrorSpy = vi.spyOn(apiModule, 'removeApiErrorInterceptor');
+    const removeAuthSpy = vi.spyOn(apiModule, 'removeApiAuthInterceptor');
+
+    render(
+      <AuthProvider>
+        <LogoutConsumer />
+      </AuthProvider>
+    );
+
+    localStorage.setItem('everestToken', 'test-token');
```
solovevayaroslavna, line 75: localStorage/sessionStorage are seeded after render(), so the provider mounts in a logged-out state and the test relies on statement ordering rather than a real logged-in→logout flow. I suppose we need to put storage before render() so the component starts authenticated.

ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+describe('AuthProvider logout behavior', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+    localStorage.clear();
+    sessionStorage.clear();
+  });
+
+  it('runs cleanup even when api.delete fails', async () => {
+    vi.spyOn(apiModule.api, 'delete').mockRejectedValueOnce(new Error('fail'));
```
solovevayaroslavna, line 65: The test mocks api.delete to reject but never asserts it was invoked, so a refactor that skips the DELETE entirely would still pass. So i think we also should add `.toHaveBeenCalled()` function

ui/apps/everest/src/contexts/auth/auth.provider.test.tsx
solovevayaroslavna, line 75: Small heads-up on this line:

`test-token` isn't a decodable JWT, and the provider runs `jwtDecode` on whatever is in `everestToken`. It's harmless right now only because the token is set after `render`, so the bootstrap effect takes the `!savedToken` branch and never decodes — but once you move the seeding before `render` (or the effect deps change), `jwtDecode` will throw and call `logout()` on its own, so the test could pass/fail for reasons unrelated to the click. Not a blocker, just a robustness nit. 
I suppose to fix it with mocking the decoder, like
`vi.mock('jwt-decode', () => ({ jwtDecode: () => ({ sub: 'user:local', exp: Math.floor(Date.now()/1000) + 3600 }) }))`
or use a realistically-shaped token.

solovevayaroslavna (changes requested):
Hey all! Please excuse the late response; everything is almost perfect, just a couple of minor edits.

PR #2165, opened 2026-05-05
fix: update helm targets to use openeverest instead of percona organization

erikwebb:
Following documentation here: https://github.com/openeverest/helm-charts/tree/main/charts/everest#4-deploy-additional-database-namespaces

Converting `percona/everest-db-namespace` Helm chart to `openeverest/everest-db-namespace`.


recharte (approved):
Thank you @erikwebb

PR #2166, opened 2026-05-05
fix: add unsubscribe to prevent stale observers

ashnaaseth2325-oss:
## 1. SUMMARY

This PR fixes a missing cleanup in `useNamespacePermissionsForResource` that caused subscriptions to `AuthorizerObservable` to accumulate across re-renders. The change ensures callbacks are properly unsubscribed, preventing stale executions and memory growth.

Primary changes are in `ui/apps/everest/src/hooks/rbac/rbac.ts` within the `useNamespacePermissionsForResource` hook.

---

## 2. FIX

```ts
// Before
useEffect(() => {
  AuthorizerObservable.subscribe(checkPermissions);
  checkPermissions();
}, [checkPermissions]);

// After
useEffect(() => {
  AuthorizerObservable.subscribe(checkPermissions);
  checkPermissions();
  return () => AuthorizerObservable.unsubscribe(checkPermissions);
}, [checkPermissions]);
```

### DIFF
--- a/ui/apps/everest/src/hooks/rbac/rbac.ts
+++ b/ui/apps/everest/src/hooks/rbac/rbac.ts
@@ -1,3 +1,17 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @Denyme24 @solovevayaroslavna 
In this PR, I added missing cleanup in useNamespacePermissionsForResource to unsubscribe from AuthorizerObservable. This prevents stale callbacks and aligns with useRBACPermissions.
Thanks!

solovevayaroslavna:
@ashnaaseth2325-oss The cleanup prevents future notifications from invoking stale callbacks, but it does not cancel an async checkPermissions() call that has already started before unmount. That is not introduced by this PR, but it is worth keeping in mind as a separate follow-up if we want to fully eliminate stale async updates. (as an idea for future improves)

solovevayaroslavna:
@ashnaaseth2325-oss can you please create a cherry pick in v2 so that we don't lose it when we will move?

ashnaaseth2325-oss:
Hi @solovevayaroslavna 
Thanks for explaining! I understand now. I’ll keep this in mind for a future improvement.
Also created the v2 cherry-pick PR as requested. 
Thanks!

PR #2181, opened 2026-05-07
fix(cli): honor skip-wizard in namespace updates

### LINKED ISSUE #2123: namespaces update with “skip wizard” still asks about operators
### What is the issue?

while updating a namespace non-interactively and pass the flag meant to skip the wizard, the CLI still tries to prompt for which operators to install. there’s no sensible way to answer in CI or scripted runs, so the command exits with an error even though you asked not to run interactively.

### Steps to Reproduce

1. Pick a namespace that Everest already manages.
2. Run an update without any --operator.* flags but with skip-wizard enabled, e.g.
3. everestctl namespaces update <your-namespace> --skip-wizard --verbose
Observe it fails immediately.

### Relevant Logs / Output

```
2026-05-04T11:06:07.550Z DEBUG everestctl commands/root.go:69 Debug logging enabled
2026-05-04T11:06:07.655Z ERROR everestctl output/command.go:39 can't ask user for operators to install: interactive mode is disabled
...
github.com/percona/everest/commands/namespaces.namespacesUpdatePreRun
/home/naman/openeverest/openeverest/commands/namespaces/update.go:98
```

### OpenEverest Version

v1.15.1

### Possible Solution

treat skip wizard the same way for update as we already do for add: when someone turns that on, the tool should stop asking which database operators to install and just use the choices already implied by the flags (the normal defaults), instead of failing because there’s no interactive session.

Sambhram1:
## Summary
- stop prompting for operators during verestctl namespaces update when --skip-wizard is enabled
- reuse the same operator-prompt decision logic already used by the install and namespace-add flows
- add a focused unit test covering the skip-wizard and explicit-operator cases

## Testing
- go test ./commands/namespaces

Fixes #2123

### DIFF
--- a/commands/namespaces/update.go
+++ b/commands/namespaces/update.go
@@ -87,10 +87,7 @@ func namespacesUpdatePreRun(cmd *cobra.Command, args []string) { //nolint:revive
 	}
 
 	// If user doesn't pass any --operator.* flags - need to ask explicitly.
-	askOperators := !(cmd.Flags().Lookup(cli.FlagOperatorMongoDB).Changed ||
-		cmd.Flags().Lookup(cli.FlagOperatorPostgresql).Changed ||
-		cmd.Flags().Lookup(cli.FlagOperatorXtraDBCluster).Changed ||
-		cmd.Flags().Lookup(cli.FlagOperatorMySQL).Changed)
+	askOperators := shouldAskOperators(cmd, namespacesUpdateCfg.SkipWizard)
 
 	if askOperators {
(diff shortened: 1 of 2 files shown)


areebahmeddd:
Hey, i believe another contributor has already raised a pr 🤔

Sambhram1:
If the other PR already includes that fix, feel free to close this one. If not, I’m also fine
  keeping this open as a narrower alternative.

areebahmeddd:
https://github.com/openeverest/openeverest/pull/2102#issuecomment-4367965222

PR #2186, opened 2026-05-08
Add API for restore

chilagrow:
This PR adds restore API allowing create, get and delete methods.

Manually tested for now.


api/openapi/http-api.yaml
recharte, line 1: Will the `/clusters/{cluster}/instances/{instance}/restores` come in a follow-up PR?

recharte (approved):

api/openapi/http-api.yaml
chilagrow: It was added in an earlier PR #2058 🤗.

PR #2197, opened 2026-05-09
fix(helm): clean up partial cache and retry on corrupt archive

### LINKED ISSUE #2195: bug: corrupted or partial Helm chart cache causes persistent install/upgrade failures
### What is the issue?

When a Helm chart download is interrupted (network drop, Ctrl+C, disk full), a partial .tgz file is left in the everestctl cache directory. On the next run, the cache check only verifies the file exists — it does not validate the file is a complete, valid archive. This causes every subsequent install/upgrade attempt to fail with a cryptic error until the user manually clears the cache directory.

### Steps to Reproduce

1. Run `everestctl install` or `everestctl upgrade`
2. Interrupt the process mid-download (Ctrl+C or network drop)
3. Run the same command again
4. The partial .tgz file is found in cache, download is skipped, loader.LoadArchive() fails with an archive parsing error
5. The error repeats on every run until cache is manually deleted

### Relevant Logs / Output

_No response_

### OpenEverest Version

main branch (current development)

### Installation Method

CLI (everestctl)

### Kubernetes Distribution

Other (please specify below)

### Environment Details

Affected file: pkg/cli/helm/installer.go — newChartFromRemoteWithCache(), lines 312–343

### Possible Solution

1. On pull.Run() failure: delete any partial file written to cacheDir before returning the error
2. On loader.LoadArchive() failure: delete the cached file and retry the download once

### Additional Context

_No response_

Harishrs2006:
newChartFromRemoteWithCache had two gaps:
1. If pull.Run() failed mid-download, a partial .tgz was left in the cache directory. On the next run os.Stat() found the file, skipped the download, and loader.LoadArchive() failed with a cryptic error that repeated on every subsequent invocation.
2. There was no integrity check before loading a cached archive, so a corrupt or partial file caused persistent failures until the user manually cleared the cache.

split the function into three focused helpers:
1) downloadChartToCache: pulls the chart and removes any partial file on failure
2)  loadChartArchive: opens and parses the .tgz
3)   newChartFromRemoteWithCache: orchestrates the two above; if LoadArchive fails on a cached file it removes the bad entry and retries the download once before returning an error

Fixes :   #2195


jtomaszon:
Nice approach — the three-helper split is clean and the retry logic is exactly right.

Two things worth checking before CI runs:

1. `_ = os.Remove(file)` silently discards the error in both `downloadChartToCache` (cleanup on pull failure) and the retry path in `newChartFromRemoteWithCache`. If `os.Remove` fails and Helm won't overwrite an existing file, the retry download will also fail — and `errcheck` may flag `_ = expr` in CI. Consider handling the error explicitly or at minimum using `errors.Is(removeErr, fs.ErrNotExist)` to ignore the already-gone case.

2. In the corrupt-archive retry path, if `downloadChartToCache` also fails, the original `loadChartArchive` error (which explains *why* the cache was corrupt) is lost. Wrapping both with `errors.Join` would preserve the full context for debugging.

Otherwise the logic is solid — this is the right fix.

pkg/cli/helm/installer.go
```diff
@@ -321,19 +322,43 @@ func newChartFromRemoteWithCache(version, name string, repository string) (*char
...
-		pull.Version = version
-		pull.DestDir = cacheDir
-		pull.RepoURL = repository
-		if _, err = pull.Run(name); err != nil {
+	ch, err := loadChartArchive(file)
+	if err != nil {
+		// The cached file may be corrupt or partial (e.g. interrupted download).
+		// Remove it and retry the download once before giving up.
+		_ = os.Remove(file)
+		if err = downloadChartToCache(name, version, repository, cacheDir, file); err != nil {
```
jtomaszon, line 338: If `os.Remove` fails here (e.g. permissions issue or locked file), the error is silently discarded and `downloadChartToCache` will be called with the corrupt file still on disk. Helm may then refuse to overwrite it, so the retry silently fails too. Consider:
```go
if removeErr := os.Remove(file); removeErr != nil && !errors.Is(removeErr, fs.ErrNotExist) {
    return nil, fmt.Errorf("corrupt cached chart and cache eviction failed: %w",
        errors.Join(err, removeErr))
}
```
This also preserves the original `loadChartArchive` error via `errors.Join`, which is lost in the current retry path.

pkg/cli/helm/installer.go
```diff
@@ -321,19 +322,43 @@ func newChartFromRemoteWithCache(version, name string, repository string) (*char
...
+	actionConfig := &action.Configuration{}
+	pull := action.NewPullWithOpts(action.WithConfig(actionConfig))
+	pull.Settings = settings
+	pull.Version = version
+	pull.DestDir = cacheDir
+	pull.RepoURL = repository
+	if _, err := pull.Run(name); err != nil {
+		// Clean up any partial file left by the failed download.
+		_ = os.Remove(file)
+		return err
```
jtomaszon, line 359: Same concern as the retry path above — `os.Remove` errors are silently swallowed. If cleanup fails, the partial archive stays on disk, and `errcheck` may flag `_ = expr` in CI (the project config doesn't set `check-blank: false`). Suggest handling it explicitly:
```go
if removeErr := os.Remove(file); removeErr != nil && !errors.Is(removeErr, fs.ErrNotExist) {
    return errors.Join(err, fmt.Errorf("failed to remove partial download: %w", removeErr))
}
```

pkg/cli/helm/installer.go
Harishrs2006: Fixed  now checks the remove error explicitly and returns a joined error combining the original loadChartArchive failure and the eviction failure, so neither error is lost

pkg/cli/helm/installer.go
Harishrs2006: Same fix applied here. If os.Remove fails, the pull error and the cleanup error are now both returned via errors.Join so the caller sees the full picture. ErrNotExist is skipped since a missing file is not an error in this context

Harishrs2006:
could @jtomaszon  review this

jtomaszon (approved):
Thanks @Harishrs2006, re-reviewed. Both points are addressed cleanly — the corrupt-archive retry path and the partial-file cleanup now both check the `os.Remove` result and surface it via `errors.Join`, so neither the original failure nor an eviction failure gets lost. Skipping `fs.ErrNotExist` is the right call, and the retry is correctly bounded to a single re-download (no recursion). LGTM. 👍

One optional, non-blocking follow-up: a small regression test that seeds a corrupt `.tgz` in the cache dir and asserts recovery-on-retry would lock this behavior in — but I'm fine with it landing as-is.

PR #2201, opened 2026-05-11
feat: add V2 Backup/Restore/PITR architecture documentation

solovevayaroslavna:


docs/proposals/backups-restore-architecture.md
```diff
@@ -0,0 +1,1020 @@
...
+- On submit: POST Instance with `spec.dataSource.dbClusterBackupName` + optional `pitr`
+- User can modify backup/schedule config before creating the new cluster
+
+### Default storage (`main` field)
+
+`main: true` on `InstanceBackupStorage` marks the engine's **default storage**. At most one per instance (not enforced by CEL).
+
+- **Auto-assigned:** First storage bound to instance gets `main: true` automatically
+- **1st iteration:** Default is always the first storage. No manual reassignment UI.
+- **UI (Backups tab):** Shown as filled "Default" chip on the first storage row
```
recharte, line 583: I don't think we should expose this `main` field in our CRs. This is a construct that's very specific of the PSMDB operator, this concept doesn't exists in other operators. I don't think we need this at all but if we do it would need to be part of a custom config.

PR #2202, opened 2026-05-11
Document load-balancer-configs RBAC requirement

blenbot:
Update the permissions example and add a short note to clarify the `load-balancer-configs` RBAC permission requirement.


chilagrow (commented):
Could you please provide a link to a issue please? 🤗

recharte:
@blenbot the most important part of the documentation is here: https://github.com/openeverest/everest-doc/blob/main/docs/administer/rbac.md
This repo :point_up: holds the source that will get rendered onto https://openeverest.io/documentation/current/administer/rbac.html which is what we really want to update.

PR #2225, opened 2026-05-12
fix(server): add HTTP timeout and non-OK logging for telemetry

### LINKED ISSUE #2203: telemetry: add request timeout and log response body on non-200
### What is the issue?

### Steps to Reproduce

returns a non-200 response.

### Relevant Logs / Output

_No response_

### OpenEverest Version

v1.15.0

### Installation Method

Helm

### Kubernetes Distribution

EKS

### Environment Details

_No response_

### Possible Solution

### Additional Context

Vi-shub:
Telemetry POSTs no longer use http.DefaultClient without a bound. Outbound telemetry uses an http.Client with a 30s timeout. On non-200 responses we log status and a truncated response body (up to 4KB) for easier debugging. On 200 we drain the response body before returning.

Why

http.DefaultClient can block for a long time on slow or stuck endpoints, and logging only the status code is often not enough to diagnose telemetry failures.
closes #2203 

@spron-in and @solovevayaroslavna can you please review this. Thankyou for your time :)

### DIFF
--- a/internal/server/telemetry.go
+++ b/internal/server/telemetry.go
@@ -103,11 +103,15 @@ func postTelemetryPayload(
 
 	snippet, readErr := io.ReadAll(io.LimitReader(resp.Body, telemetryMaxErrorBodyLogBytes))
 	if readErr != nil {
-		l.Infow("telemetry non-OK response", "status", resp.StatusCode, "bodyReadErr", readErr)
-		return nil
+		err := fmt.Errorf("telemetry request failed with status %d: %w", resp.StatusCode, readErr)
+		l.Errorw("telemetry non-OK response", "status", resp.StatusCode, "bodyReadErr", readErr)
+		return err
 	}
-	l.Infow("telemetry non-OK response", "status", resp.StatusCode, "bodySnippet", string(snippet))
-	return nil
+	if len(snippet) == telemetryMaxErrorBodyLogBytes {
+		l.Warn("telemetry error response body truncated; original response is longer")
+	}
+	l.Warnw("telemetry non-OK response", "status", resp.StatusCode, "bodySnippet", string(snippet))
+	return fmt.Errorf("telemetry request failed with status %d", resp.StatusCode)
 }
 
 // RunTelemetryJob runs background job for collecting telemetry.
--- a/internal/server/telemetry.go
+++ b/internal/server/telemetry.go
@@ -6,11 +6,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"time"
 
 	"github.com/google/uuid"
+	"go.uber.org/zap"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openeverest/openeverest/v2/cmd/config"
@@ -25,7 +27,15 @@ const (
 	initialMetricsDelay = 5 * time.Minute
 
 	numEngineTypes = 3
+
+	// telemetryHTTPTimeout bounds how long a single telemetry POST may block the caller.
+	telemetryHTTPTimeout = 30 * time.Second
+	// telemetryMaxErrorBodyLogBytes limits how much of a non-OK response body we read for logs.
+	telemetryMaxErrorBodyLogBytes = 4096
 )
+
+// defaultTelemetryHTTPClient is used for outbound telemetry (bounded timeout; avoids hanging on DefaultClient).
+var defaultTelemetryHTTPClient = &http.Client{Timeout: telemetryHTTPTimeout}
 
 // Telemetry is the struct for telemetry reports.
 type Telemetry struct {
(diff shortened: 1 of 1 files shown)


internal/server/telemetry.go
```diff
@@ -1,158 +1,194 @@
...
+		return nil
+	}
+
+	snippet, readErr := io.ReadAll(io.LimitReader(resp.Body, telemetryMaxErrorBodyLogBytes))
+	if readErr != nil {
+		l.Infow("telemetry non-OK response", "status", resp.StatusCode, "bodyReadErr", readErr)
+		return nil
+	}
+	l.Infow("telemetry non-OK response", "status", resp.StatusCode, "bodySnippet", string(snippet))
+	return nil
```
Denyme24: suggestion: instead, should it not return an error if the server replied with a failure? this make it easier for upstream code and operators to handle telemetry delivery failures and should not hide silent errors.

`return fmt.Errorf("telemetry failed! Status: %d", resp.StatusCode)`

internal/server/telemetry.go
```diff
@@ -1,158 +1,194 @@
...
+		return err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode == http.StatusOK {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		return nil
+	}
+
+	snippet, readErr := io.ReadAll(io.LimitReader(resp.Body, telemetryMaxErrorBodyLogBytes))
```
Denyme24, line 104: i think we can consider logging a warning if the error body is truncated so that future readers know some content was not captured. what do you say? 

something like this: `if len(snippet) == telemetryMaxErrorBodyLogBytes {
    l.Warn("error body truncated; original response is longer")
}`

internal/server/telemetry.go
Vi-shub: Done.
postTelemetryPayload now returns an error on non-200 (and on body read failure), so RunTelemetryJob logs it via the existing collectMetrics error path.

internal/server/telemetry.go
Vi-shub: Done.
Added a Warn when the snippet length equals telemetryMaxErrorBodyLogBytes.

Vi-shub:
Thanks @Denyme24
Addressed both points: non-200 now returns an error (so upstream can surface telemetry delivery failures), and we log a warning when the error body hits the 4KB read limit. Happy to adjust further if needed.

jtomaszon:
Good approach overall — extracting `postTelemetryPayload` as a separate function makes it properly testable, and draining the body on 200 with `io.Copy(io.Discard, ...)` is the right practice.

A few things that may need attention before CI passes:

1. **`gochecknoglobals`**: the package-level `var defaultTelemetryHTTPClient` will likely be flagged. The project has this linter enabled (visible in `.golangci.yml`). You'd either need a `//nolint:gochecknoglobals` comment or create the client inline in `report()` on each call (telemetry is low-frequency so the allocation cost is negligible).

2. **`_ = resp.Body.Close()` and `_, _ = io.Copy(...)`**: blank-identifier assignments may still trigger `errcheck` depending on the project's `check-blank` setting. Worth verifying CI passes, or replacing with a `defer func() { if err := ...; err != nil { l.Warn(...) } }()` pattern.

3. **Timeout value**: 30s is conservative — for a background telemetry job that's probably fine, but worth a comment on why 30s was chosen over something tighter like 10-15s.

The refactor to `postTelemetryPayload` is a genuine improvement over a simpler in-place fix — makes the contract testable. Nice work.

internal/server/telemetry.go
```diff
@@ -1,158 +1,198 @@
...
+
+	// telemetryHTTPTimeout bounds how long a single telemetry POST may block the caller.
+	telemetryHTTPTimeout = 30 * time.Second
+	// telemetryMaxErrorBodyLogBytes limits how much of a non-OK response body we read for logs.
+	telemetryMaxErrorBodyLogBytes = 4096
+)
+
+// defaultTelemetryHTTPClient is used for outbound telemetry (bounded timeout; avoids hanging on DefaultClient).
+var defaultTelemetryHTTPClient = &http.Client{Timeout: telemetryHTTPTimeout}
+
```
jtomaszon, line 39: Package-level `var` will be flagged by `gochecknoglobals`, which is enabled in `.golangci.yml`. Two options:

1. Add `//nolint:gochecknoglobals` with a justification comment.
2. Remove the package-level var and create the client inline in `report()` — telemetry fires infrequently so the allocation is negligible, and `postTelemetryPayload` already accepts the client as a parameter so it stays fully testable either way.

internal/server/telemetry.go
```diff
@@ -1,158 +1,198 @@
...
+		httpClient = defaultTelemetryHTTPClient
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		l.Error(errors.Join(err, errors.New("failed to send telemetry request")))
+		return err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
```
jtomaszon, line 98: `_ = resp.Body.Close()` discards the error. `errcheck` may flag blank-identifier assignments depending on the project's `check-blank` setting. Replacing with an explicit handler keeps the linter happy and gives operators a warning if the connection is flaky:
```go
defer func() {
    if err := resp.Body.Close(); err != nil {
        l.Warnw("failed to close telemetry response body", "error", err)
    }
}()
```

internal/server/telemetry.go
```diff
@@ -1,158 +1,198 @@
...
+	if resp.StatusCode == http.StatusOK {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		return nil
```
jtomaszon, line 101: Same `errcheck` concern — `_, _ = io.Copy(...)` uses two blank identifiers. `io.Copy` to `io.Discard` rarely errors in practice, but if the linter catches it you can silence it cleanly with:
```go
if _, err := io.Copy(io.Discard, resp.Body); err != nil {
    l.Warnw("failed to drain telemetry response body", "error", err)
}
```

internal/server/telemetry.go
recharte, line 1: This file came back with CRLF line endings on all 198 lines, which is why the diff reads +198/−158 for what `git diff --ignore-all-space` shows is a 48-line change.

It's the only `.go` file in the repo with CRLF and there's no `.gitattributes` to normalise it, so it will conflict with anything else that touches this file — and `gofmt -l` flags it, so the `Check` job fails on it too.

`gofmt -w internal/server/telemetry.go` and an amended commit should bring the diff back down to the change itself.

internal/server/telemetry.go
```diff
@@ -1,158 +1,198 @@
...
+		e.l.Error(errors.Join(err, errors.New("failed to marshal the telemetry report")))
+		return err
+	}
+
+	return postTelemetryPayload(ctx, defaultTelemetryHTTPClient, baseURL, b, e.l)
+}
+
+// postTelemetryPayload POSTs a JSON payload to the telemetry GenericReport endpoint.
+// httpClient must not be nil in production; tests may pass a custom client (e.g. httptest.Server.Client()).
+func postTelemetryPayload(
```
recharte, line 72: The `httpClient` parameter and its nil fallback exist for a test that isn't in this PR — the one production caller passes `defaultTelemetryHTTPClient` on line 67, so the nil branch is unreachable.

Worth adding the test it was built for: an `httptest.Server` returning 500 with a long body covers the truncation warning and the error return together, which are the two behaviours this PR introduces. Otherwise the parameter is better dropped.

internal/server/telemetry.go
recharte: Confirmed — `gochecknoglobals` is on via `enable-all` and isn't in the disable list, so this will be flagged.

Either option works. For what it's worth the repo tends towards your option 1: `pkg/cli/helm/installer.go:59` and `internal/server/handlers/validation/errors.go` both carry an inline `//nolint:gochecknoglobals`.

internal/server/telemetry.go
recharte: Checked this one — `check-blank` isn't set in `.golangci.yml` and defaults to off, so blank assignments aren't flagged. Merged code leans on that in a fair few places (`_ = cmd.MarkFlagRequired(...)` throughout `commands/`), so nothing needed here.

internal/server/telemetry.go
recharte: Same as the thread above — `check-blank` is off, so this one won't be flagged either.

recharte (changes requested):
CI has never run on this branch — the only checks are DCO and the labeller, from May — so nothing here has been built, linted or tested by us. That's also how the line endings got past two rounds of review; `gofmt` would have caught them on the first push.

@jtomaszon on the 30s question from your review: settled at 30s on #2203, matching the two sibling fixes so the three read the same way. Happy to revisit as a set if you think that's too loose for a background job.

PR #2227, opened 2026-05-13
fix: add unsubscribe to prevent stale observers in v2

ashnaaseth2325-oss:
Cherry-pick of #2166 into v2 so the RBAC observer cleanup fix is not lost during the migration.

Adds the missing unsubscribe cleanup in `useNamespacePermissionsForResource` to prevent stale callbacks from accumulating across re-renders/unmounts.


solovevayaroslavna:
thank you very much!

PR #2228, opened 2026-05-13
fix: Prevent Everest upgrade dialog from reopening after dismissal

ashnaaseth2325-oss:
## SUMMARY

This PR fixes an issue where the Everest upgrade dialog would re-open every 5 seconds after being dismissed by the user. The problem was caused by the upgrade detection logic repeatedly treating the same version change as a new upgrade on every poll.

The changes are focused in `ui/apps/everest/src/contexts/upgrade-everest/upgrade-everest.provider.tsx`, specifically inside the `UpgradeEverestProvider` upgrade detection effect.

---

## FIX

```ts
//Before
if (
  commitVersion.current !== null &&
  commitVersion.current !== apiVersion?.fullCommit
) {
  setOpenReloadEverestDialog(true);
}

//After
if (
  commitVersion.current !== null &&
  commitVersion.current !== apiVersion?.fullCommit
) {
  commitVersion.current = apiVersion!.fullCommit;
  setOpenReloadEverestDialog(true);
}
```


ashnaaseth2325-oss:
Hello @solovevayaroslavna 
Kindly review this PR whenever you have time.
Thanks!

solovevayaroslavna (approved):

PR #2234, opened 2026-05-14
Use Orval client for api tests

chilagrow:
This PR uses orval client for generating API tests.

It also updates Open API parameter names from `/clusters/{cluster}/namespaces/{namespace}/instances/{instance}` to `/clusters/{cluster}/namespaces/{namespace}/instances/{name}`.

Orval does not generate valid code if request body and a parameter name crashes, such as usage of `{instance}` in param and `'crds.gen.yaml#/components/schemas/Instance'` like below:

Param
```
'/clusters/{cluster}/namespaces/{namespace}/instances/{instance}'
```
Body
```
  requestBody:
        description: The instance object to be updated
        required: true
        content:
          application/json:
            schema:
              $ref: 'crds.gen.yaml#/components/schemas/Instance'
```


api-tests/hack/rename-api-params.mjs
```diff
@@ -0,0 +1,47 @@
+/**
```
chilagrow: This is probably one of decider for using orval, it's not nice to rewrite generated code by orval.

api-tests/tests/utils/api.ts
```diff
@@ -346,47 +348,51 @@ export const getBackupStoragePayload = (bsName: string) => {
...
+export const generateBackupStorage = async (request: APIRequestContext, data: BackupStorage) => {
   const response = await createBackupStorageRaw(request, data)
   await checkError(response)
   return (await response.json())
 }
 
-export const createBackupStorageRaw = async (request, data) => {
-  return await request.post(`/v1/clusters/${CLUSTER_NAME}/namespaces/${EVEREST_CI_NAMESPACE}/backup-storages`, {data: data})
+export const createBackupStorageRaw = async (request: APIRequestContext, data: BackupStorage) => {
+  setRequestContext(request)
```
chilagrow, line 267: It's necessary to set context expected by playwrights, this unfortunately makes API noiser and a bit awkward to bridge orval and playwrights.

api-tests/tests/utils/api.ts
```diff
@@ -346,47 +348,51 @@ export const getBackupStoragePayload = (bsName: string) => {
...
+  return await client.createBackupStorage(CLUSTER_NAME, EVEREST_CI_NAMESPACE, data) as any
```
chilagrow, line 268: orval doesn't have native ways to replace path parameters with constants. Hence passed as variables.

api-tests/hack/rename-api-params.mjs
chilagrow: Instead updated API parameter names

chilagrow:
@solovevayaroslavna could you take a look please 🤗

PR #2235, opened 2026-05-14
fix(ui): complete local logout when DELETE /session fails

piyxsh31:
Wrap server session revoke in try/catch so local auth cleanup always runs. Show a warning snackbar on revoke failure; suppress duplicate API toasts for this request.


ui/apps/everest/src/contexts/auth/auth.provider.tsx
```diff
@@ -98,7 +98,20 @@ const AuthProvider = ({ children, isSsoEnabled }: AuthProviderProps) => {
 
   const logout = async () => {
     const token = localStorage.getItem('everestToken');
-    await api.delete('/session', { headers: { token: token } });
+    if (token) {
+      try {
+        await api.delete('/session', {
```
areebahmeddd, line 103: i think the error interceptor is still active when this request runs. in [addApiErrorInterceptor](https://github.com/openeverest/openeverest/blob/main/ui/apps/everest/src/api/api.ts#L53), the 401 branch redirects with `location.href = '/logout'` before the `disableNotifications` check is even reached, so this flow may never hit the `catch` block properly

imo we should remove the api interceptors before calling `api.delete()` since logout itself probably should not go through the normal auth error handling flow

ui/apps/everest/src/contexts/auth/auth.provider.tsx
```diff
@@ -98,7 +98,20 @@ const AuthProvider = ({ children, isSsoEnabled }: AuthProviderProps) => {
...
+          headers: { token },
+          disableNotifications: true,
```
areebahmeddd, line 105: `disableNotifications: true` looks correct, but from what i can see it only skips the snackbar logic. the 401 redirect path in the interceptor still runs before that check

if we move `removeApiErrorInterceptor()` before this request, i think this option may not be needed anymore 🤔

ui/apps/everest/src/contexts/auth/auth.provider.tsx
```diff
@@ -98,7 +98,20 @@ const AuthProvider = ({ children, isSsoEnabled }: AuthProviderProps) => {
...
+        });
+      } catch {
+        enqueueSnackbar(
+          'Signed out here, but the server could not revoke your session. Try again later.',
```
areebahmeddd, line 109: small wording thing here: "try again later" sounds a bit odd since the user is already signed out locally and there’s nothing left for them to retry 😅

maybe we can make this a bit more actionable, especially for shared machines or revoked sessions

areebahmeddd:
also,

1. your dco is failing
2. you should always have an associated issue linked with the pr (+issue assigned to you as well!)
3. you've raised this pr from main instead of a separate branch off main
4. imho we should not rush to open the fastest possible pr. a bit more discussion in slack or the issue thread first would help avoid unnecessary notification spam for the core team 🙂

piyxsh31:
Sorry for these many number of mistakes from my side. I will take care of them from next time 
I think some of them cannot be fixed in this pr so i am closing this one

PR #2238, opened 2026-05-14
fix(ui): complete local logout when DELETE /session fails

### LINKED ISSUE #2236: Complete local sign-out when DELETE /session fails
### What is the issue?

logout() awaits on DELETE /v1/session before clearing local state (token, authStatus, storage). If that request fails (network error, timeout, server error), execution stops before local cleanup, so the user can remain stuck in a logged-in UI with no reliable way to finish signing out.

Expected behavior:

- call DELETE /session when a token exists to revoke the session on the server.
- If the call fails, still complete local sign-out (clear token, auth state, storage, interceptors) so the user is never trapped.
- show a short warning that server-side revoke may have failed.

Related files:
`ui/apps/everest/src/contexts/auth/auth.provider.tsx` — logout

### Steps to Reproduce

Preconditions:

- App running, logged in (manual session with token is enough).

Steps:

- Open DevTools → Network.
- use “Request blocking” / block URL pattern *session*.
- Click Log out in the UI, or open /logout in the address bar.
- Watch the Network panel: DELETE …/v1/session should fail (blocked / failed / (failed)).

### Relevant Logs / Output

_No response_

### OpenEverest Version

v2

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Always finish local sign-out even if DELETE /session fails; show a brief warning when server revoke fails

### Additional Context

_No response_

piyxsh31:
Ensures local logout completes when `DELETE /session` fails; removes the error interceptor before revoke and warns if server session could not be ended.

Closes  #2236

### DIFF
--- a/ui/apps/everest/src/contexts/auth/auth.provider.tsx
+++ b/ui/apps/everest/src/contexts/auth/auth.provider.tsx
@@ -98,7 +98,21 @@ const AuthProvider = ({ children, isSsoEnabled }: AuthProviderProps) => {
 
   const logout = async () => {
     const token = localStorage.getItem('everestToken');
-    await api.delete('/session', { headers: { token: token } });
+    removeApiErrorInterceptor();
+
+    if (token) {
+      try {
+        await api.delete('/session', {
+          headers: { token },
+        });
+      } catch {
+        enqueueSnackbar(
+          'Signed out on this device, but the server did not end your session. If others use this machine, change your password when you can.',
+          { variant: 'warning' }
+        );
+      }
+    }
+
     if (isSsoEnabled) {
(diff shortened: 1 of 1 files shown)


solovevayaroslavna:
Hi @piyxsh31, thanks for the fix and for digging into #2236! 🙏

Closing this one: the issue was already resolved on main as part of the auth rework in #2362. Logout there uses POST /auth/revoke wrapped in a best-effort try/catch, so local session state is always cleared even if the server call fails. Since this PR targets release-2.0 and the old DELETE /session path no longer exists, it can't be forward-ported as-is

Thanks again! 🙌

PR #2239, opened 2026-05-15
fix(ui): prevent UI generator from producing invalid field paths

### LINKED ISSUE #2232: UI Generator produces invalid React Hook Form field paths
### What is the issue?

When a `group` component in a UI Schema has no `path` or `id`, the UI Generator produces invalid React Hook Form field paths like `.myField` or `parent..myField` (leading or double dots). This breaks form data binding silently — inputs render but their values are never connected to the API payload correctly.

### Steps to Reproduce

1. Define a UI Schema section.
2. Create a `group` component that has no `path` or `id` (layout-only group).
3. Add a `text` component inside that group.
4. Inspect the rendered HTML  the `name` attribute of the input will contain a leading or double dot (e.g. `.myField`).

Affected file:
`ui/apps/everest/src/components/ui-generator/utils/component-renderer/render-component.tsx`

Problematic line:
```typescript
const childFieldName = `${fieldName}.${childKey}`;
```
When `fieldName` is empty, this produces `.childKey` instead of `childKey`.

### Relevant Logs / Output

No runtime error is thrown. The bug is visible by inspecting the DOM:

<input name=".myField" />   // leading dot
<input name="parent..myField" />  // double dot

### OpenEverest Version

v2

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Replace the current path concatenation with a safe join:

```typescript
const childFieldName = fieldName ? `${fieldName}.${childKey}` : childKey;
```

This ensures a dot separator is only added when a parent path actually exists.

### Additional Context

The current bug makes pathless groups unusable.

ayushshukla1807:
Fixes #2232 by avoiding leading dots when parent path is empty.

### DIFF
--- a/ui/apps/everest/src/components/ui-generator/utils/component-renderer/render-component.tsx
+++ b/ui/apps/everest/src/components/ui-generator/utils/component-renderer/render-component.tsx
@@ -47,7 +47,7 @@ export const renderComponent = ({
       (item as ComponentGroup).components,
       (item as ComponentGroup).componentsOrder
     ).map(([childKey, childItem]) => {
-      const childFieldName = `${fieldName}.${childKey}`;
+      const childFieldName = fieldName ? `${fieldName}.${childKey}` : childKey;
       return (
         <React.Fragment key={childFieldName}>
           {renderComponent({


areebahmeddd:
Hi, there's already a PR from another contributor for this issue

PR #2248, opened 2026-05-16
fix(cli): replace http.DefaultClient with a timeout-aware client in version service

### LINKED ISSUE #2247: fix(cli): version service HTTP client has no timeout, causing everestctl install/upgrade to hang
### What is the issue?

In `pkg/version_service/client.go`, both `GetSupportedEngineVersions` and `GetEverestMetadata` execute HTTP requests using `http.DefaultClient`, which has `Timeout: 0` (no limit):

```go
// GetSupportedEngineVersions — line 89
res, err := http.DefaultClient.Do(req)

// GetEverestMetadata — line 150
res, err := http.DefaultClient.Do(req)
```

These two methods are called during `everestctl install` (`pkg/cli/install/install.go:270`) and `everestctl upgrade` (`pkg/cli/upgrade/upgrade.go:332`). If the version metadata endpoint (`check.percona.com`) is slow or unreachable, the CLI **hangs indefinitely** with no error and no progress output.

Although the code correctly uses `http.NewRequestWithContext(ctx, ...)`, Cobra commands pass `context.Background()` without a deadline, so the context provides no timeout protection either.

### Steps to Reproduce

1. Point `VersionMetadataURL` at a server that delays responses (or use `iptables` to drop packets to `check.percona.com`)
2. Run `everestctl install`
3. Observe the CLI freezes at the version check step with no timeout or error

### Relevant Logs / Output

No output — the CLI hangs silently.

### OpenEverest Version

main branch

### Installation Method

CLI (everestctl)

### Possible Solution

Replace `http.DefaultClient` with a dedicated `*http.Client` stored on `versionServiceClient`:

```go
const defaultHTTPTimeout = 30 * time.Second

type versionServiceClient struct {
    url        string
    httpClient *http.Client
}

func New(url string) Interface {
    return &versionServiceClient{
        url:        url,
        httpClient: &http.Client{Timeout: defaultHTTPTimeout},
    }
}
```

This is the same pattern that should be applied to `internal/server/telemetry.go` (tracked separately in #2203).

### Additional Context

Also fixes a typo in the package doc comment: `"Perocona"` → `"Percona"`.

abhisheksainimitawa:
## Summary

Fixes #2247

- `GetSupportedEngineVersions` and `GetEverestMetadata` both used `http.DefaultClient` (Timeout: 0). If the version service endpoint is slow or unreachable during `everestctl install` / `everestctl upgrade`, the CLI hung indefinitely with no error.
- Adds a dedicated `*http.Client` with a 30-second timeout to `versionServiceClient`, initialized in `New()`.
- Fixes a typo in the package doc comment: `"Perocona"` → `"Percona"`.
- Adds `client_test.go` with unit tests using `httptest.Server` to verify the timeout fires for both methods.

## Changes

| File | Change |
|------|--------|
| `pkg/version_service/client.go` | Add `httpClient *http.Client` field; initialize with 30s timeout in `New()`; replace both `http.DefaultClient.Do` calls |
| `pkg/version_service/client_test.go` | New — tests that `New()` sets the timeout and that slow servers are rejected within the deadline |

## Test plan

- [ ] `go test ./pkg/version_service/... -v` — all three new tests pass
  - `TestNew_HasTimeout` — asserts `New()` sets `defaultHTTPTimeout`
  - `TestGetEverestMetadata_RespectsClientTimeout` — slow httptest server, 50 ms client timeout → error returned
  - `TestGetSupportedEngineVersions_RespectsClientTimeout` — same pattern for engine versions endpoint
- [ ] `go build ./...` — no compilation errors
- [ ] Existing CI passes

### DIFF
--- a/pkg/version_service/client.go
+++ b/pkg/version_service/client.go
@@ -1,5 +1,6 @@
 // everest
 // Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/version_service/client.go
+++ b/pkg/version_service/client.go
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package versionservice provides an interface for the Perocona version service.
+// Package versionservice provides an interface for the Percona version service.
 package versionservice
 
 import (
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"slices"
 	"strings"
+	"time"
 
 	perconavs "github.com/Percona-Lab/percona-version-service/versionpb"
 	goversion "github.com/hashicorp/go-version"
@@ -58,13 +59,19 @@ type Interface interface {
 	GetEverestMetadata(ctx context.Context) (*perconavs.MetadataResponse, error)
 }
 
+const defaultHTTPTimeout = 30 * time.Second
+
 type versionServiceClient struct {
-	url string
+	url        string
+	httpClient *http.Client
 }
 
 // New returns a new version service client.
 func New(url string) Interface { //nolint:ireturn
-	return &versionServiceClient{url: url}
+	return &versionServiceClient{
+		url:        url,
+		httpClient: &http.Client{Timeout: defaultHTTPTimeout},
+	}
 }
 
 //nolint:gochecknoglobals
@@ -86,7 +93,7 @@ func (c *versionServiceClient) GetSupportedEngineVersions(ctx context.Context, o
 	if err != nil {
 		return nil, errors.Join(err, errors.New("could not create version service request"))
 	}
-	res, err := http.DefaultClient.Do(req)
+	res, err := c.httpClient.Do(req)
 	if err != nil {
 		return nil, errors.Join(err, errors.New("could not retrieve version response"))
 	}
@@ -147,7 +154,7 @@ func (c *versionServiceClient) GetEverestMetadata(ctx context.Context) (*percona
 	if err != nil {
 		return nil, errors.Join(err, errors.New("could not create Everest metadata request"))
 	}
-	res, err := http.DefaultClient.Do(req)
+	res, err := c.httpClient.Do(req)
 	if err != nil {
 		return nil, errors.Join(err, errors.New("could not retrieve Everest metadata"))
 	}
--- a/pkg/version_service/client_test.go
+++ b/pkg/version_service/client_test.go
@@ -1,5 +1,6 @@
 // everest
 // Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/version_service/client_test.go
+++ b/pkg/version_service/client_test.go
@@ -1,0 +1,87 @@
+// everest
+// Copyright (C) 2023 Percona LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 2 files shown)


abhisheksainimitawa:
## Before / After — Demonstrating the Fix

To show the real-world impact of this change, here is a reproducible demo using a local test server that delays responses by 5 seconds (simulating a slow or unreachable `check.percona.com` endpoint).

### Before — `http.DefaultClient` (Timeout: 0)

The CLI **waits the full duration** of the server delay. If the server never responds (e.g., dropped packets, network outage), `everestctl install` hangs indefinitely with no output and no error.

```
$ go run demo/timeout_demo.go

=== BEFORE fix: using http.DefaultClient (Timeout: 0) ===
Server delay: 5s | Client timeout: none
Result after 5.002s: request completed (no timeout fired — CLI would freeze for full 5s+)
```

### After — `&http.Client{Timeout: 30 * time.Second}`

The client returns a **clear error** within the configured timeout. The user knows immediately that the version service is unreachable instead of watching a frozen terminal.

```
=== AFTER fix: using &http.Client{Timeout: 30s} ===
Server delay: 5s | Client timeout: 2s (shortened for demo)
Result after 2.001s: error — Get "http://127.0.0.1:63290": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
→ CLI returns a clear error instead of hanging
```

### Unit tests confirm the fix

```
$ go test ./pkg/version_service/... -v -run "TestNew_HasTimeout|TestGetEverestMetadata|TestGetSupportedEngineVersions"

--- PASS: TestNew_HasTimeout (0.00s)
--- PASS: TestGetEverestMetadata_RespectsClientTimeout (0.20s)
--- PASS: TestGetSupportedEngineVersions_RespectsClientTimeout (0.20s)
PASS
ok      github.com/percona/everest/pkg/version_service  0.722s
```

The tests use `net/http/httptest` with a controlled slow server — no real network or cluster required.

pkg/version_service/client.go
```diff
@@ -58,13 +60,19 @@ type Interface interface {
...
+const defaultHTTPTimeout = 30 * time.Second
+
 type versionServiceClient struct {
-	url string
+	url        string
+	httpClient *http.Client
 }
 
 // New returns a new version service client.
 func New(url string) Interface { //nolint:ireturn
```
recharte, line 71: `New` hard-codes the timeout, so the tests have to build the struct literal directly and bypass the constructor — which is then why `TestNew_HasTimeout` exists at all, to cover the one path nothing else exercises. An option collapses that:

```go
func WithHTTPClient(c *http.Client) Option { ... }

func New(url string, opts ...Option) Interface { ... }
```

Both behavioural tests could then go through `New(srv.URL, WithHTTPClient(&http.Client{Timeout: 50 * time.Millisecond}))`, the nil-client case is covered for free, and `TestNew_HasTimeout` can go. It's a shape the repo already uses — `session.New` and `informer` both take functional options — and `pkg/cli/upgrade/upgrade_test.go:201` already drives this package through `New` against an `httptest` server.

@PutlaTheophila proposed exactly this on #2247 in May and didn't get a reply — credit where it's due.

pkg/version_service/client_test.go
recharte, line 1: Could these use testify? It's what the unit tests elsewhere in the repo reach for, and `require`/`assert` would replace the hand-rolled conditions and drop the `strings` import with them.

One assertion is worth changing on its own merits while you're in here: `strings.Contains(err.Error(), "could not retrieve Everest metadata")` holds for any error out of `Do` — pointing the client at a closed port produces that identical wrapper — so it can go green without the deadline ever firing. The timeout is `context.DeadlineExceeded` underneath, so `require.ErrorIs(t, err, context.DeadlineExceeded)` pins it exactly.

recharte (commented):
Two things with no line to attach to.

CI has never run here — the only checks are DCO and the labeller, from the May run — so there's no build, lint or test evidence on this branch, and it's three months behind `main`. A rebase should trigger the real workflows. Your tests do pass locally, so that gap is ours rather than anything you need to prove.

And `release-2.0` carries the identical unbounded client, so #2247 stays open after this merges — are you up for the companion PR, or would you rather someone ports it?

PR #2263, opened 2026-05-18
fix: replace TEST_ROOT with REPO_ROOT in build-ui and deploy-all also make them quoted.

### LINKED ISSUE #2261: Fix make build-ui target when repository root path is not resolved
### What is the issue?

The commands `make build-ui` and `make deploy-all` fail during local CI style testing because the UI path resolves to `/ui` instead of the repository's `ui` directory.

When running:

```sh
make build-ui
make deploy-all
```
the build fails while trying to initialize the Everest UI:

```
Building Everest UI
make -C /ui init
make[1]: *** /ui: No such file or directory.  Stop.
make: *** [Makefile:173: build-ui] Error 2
```

even `deploy-all` fails for the same reason because it depends on, or eventually invokes, the build-ui target.
The Makefile should run the UI commands from the repositoryc local UI directory, 

### Steps to Reproduce

1. From the root of the repo try running `make deploy-all`
2. Now similarly try running make `build-ui` this too fails with same error.

These both commands fail with same error as `deploy-all `under the hood uses `build-ui`

### Relevant Logs / Output

This is what i found when i ran commands locally -

Image

### OpenEverest Version

v1.15

### Installation Method

Helm

### Kubernetes Distribution

kind

### Environment Details

LATEST
MacOs

### Possible Solution

Update the `build-ui` target to use `REPO_ROOT` consistently instead of an undefined or incorrectly resolved root variable.
Also quote `REPO_ROOT` anywhere it is used as a filesystem path, because repository paths can contain spaces. without quoting, a path like:

```text
/Users/example/Development/open source/openeverest
```
can be split by the shell and interpreted incorrectly.

For example:
```
REPO_ROOT=$(shell dirname "$(realpath $(firstword $(MAKEFILE_LIST)))")

.PHONY: build-ui
build-ui:
	$(info Building Everest UI)
	$(MAKE) -C "$(REPO_ROOT)/ui" init
	$(MAKE) -C "$(REPO_ROOT)/ui" build EVEREST_OUT_DIR="$(REPO_ROOT)/public/dist"
```

This should prevent the UI path from resolving to /ui, and should also make the target work when the repository is located in a directory with spaces in its path.

### Additional Context

_No response_

atharvamhaske:
## Summary

This PR fixes `make build-ui` and `make deploy-all` failing during local development.

The `build-ui` target previously used `TEST_ROOT`, which is not defined anywhere in the Makefile. As a result, Make expanded `${TEST_ROOT}/ui` to `/ui`, causing both `make build-ui` and `make deploy-all` (which depends on `build-ui`) to fail.

This PR replaces `TEST_ROOT` with the existing `REPO_ROOT` variable and quotes `REPO_ROOT` so paths resolve correctly when the repository is in a directory with spaces.

Closes: #2261 

## Problem
 TEST_ROOT is undefined, so Make treats it as empty and resolves the UI path to /ui instead of <repo-root>/ui.

On paths with spaces, REPO_ROOT itself could also be computed incorrectly because dirname and realpath were not quoted

## Changes

- Replace TEST_ROOT with REPO_ROOT in the build-ui target.
- Quote REPO_ROOT computation so dirname receives the full Makefile path as a single argument.
- Quote (REPO_ROOT)/ui and (REPO_ROOT)/public/dist in nested make -C invocations.

### Before:

Screenshot 2026-05-18 at 6 29 14 PM

### After:

1. Running `make deploy-all`
Screenshot 2026-05-18 at 9 13 07 PM

2. Running `make build-ui`
Screenshot 2026-05-18 at 8 51 10 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-REPO_ROOT=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+REPO_ROOT=$(shell dirname "$(realpath $(firstword $(MAKEFILE_LIST)))")
 RELEASE_VERSION ?= v0.0.0-$(shell git rev-parse --short HEAD)
 RELEASE_FULLCOMMIT ?= $(shell git rev-parse HEAD)
 IMAGE_PREFIX ?= ghcr.io/openeverest
@@ -170,8 +170,8 @@ release-cli: ## Build Everest CLI release versions for different OS and ARCH. (U
 .PHONY: build-ui
 build-ui:
 	$(info Building Everest UI)
-	$(MAKE) -C ${TEST_ROOT}/ui init
-	$(MAKE) -C ${TEST_ROOT}/ui build EVEREST_OUT_DIR=${TEST_ROOT}/public/dist
+	$(MAKE) -C "$(REPO_ROOT)/ui" init
+	$(MAKE) -C "$(REPO_ROOT)/ui" build EVEREST_OUT_DIR="$(REPO_ROOT)/public/dist"
 
 .PHONY: docker-build
 docker-build: ## Build docker image with Everest API server.


atharvamhaske:
cc: @recharte fixed!

recharte (approved):
Thank you @atharvamhaske

PR #2264, opened 2026-05-18
Restructure MonitoringConfigSpec using a discriminated union pattern

recharte:
### Problem

The current `MonitoringConfigSpec` has PMM-specific fields (`URL`, `CredentialsSecretName`, `VerifyTLS`) at the root level of the spec. This design makes it difficult to extend support to other monitoring technologies in the future — adding a new backend would require either:
- Making all fields optional (mixing concerns)
- Adding unrelated fields to the same struct (poor cohesion)
- Introducing breaking changes to the API

### Solution

Restructure `MonitoringConfigSpec` using a **discriminated union pattern** — the industry-standard approach for one-of-many configurations in Kubernetes APIs (e.g., `PersistentVolumeSource`, `VolumeSource`). This is already used successfully in OpenEverest's `BackupStorageSpec`.

**New structure:**
```go
MonitoringConfigSpec {
  Type MonitoringType    // discriminator: "pmm" (future: "prometheus", etc.)
  PMM  *PMMMonitoringSpec // nested per-type config
}

PMMMonitoringSpec {
  CredentialsSecretName string
  URL                   string
  VerifyTLS             *bool
}
```

Similarly, status is nested: `MonitoringConfigStatus.PMM.ServerVersion` (was a flat field).

### Benefits

1. **Clean separation of concerns** — PMM-specific configuration is isolated
2. **Forward extensible** — adding new monitoring types requires zero changes to existing fields
3. **Type-safe at schema level** — CEL rules enforce the discriminator↔block relationship (`type == "pmm"` requires `pmm` to be set)
4. **Follows Kubernetes conventions** — aligns with upstream patterns and OpenEverest's own established patterns
5. **No breaking changes** — this is applied to v1alpha2, which is not yet deployed in production

### Changes

- **Types**: Move PMM fields into `PMMMonitoringSpec` struct; add CEL validation rules
- **Controller**: Update field access patterns (`mc.Spec.PMM.URL` etc.; field indexer path to `.spec.pmm.credentialsSecretName`)
- **Webhook**: Guard PMM validation behind `mc.Spec.PMM != nil` check
- **API handler**: Build nested `PMM: &PMMMonitoringSpec{…}` on Create/Update


chilagrow (approved):
Thank you!

PR #2266, opened 2026-05-18
Change MonitoringConfig to v1alpha1

recharte:
We never had a MonitoringConfig v1alpha1 in the monitoring.openeverest.io group so we shouldn't go into v1alpha2.


chilagrow (approved):
Thank you!

PR #2279, opened 2026-05-19
fix: probe S3 access for DataSource at admission time (#2229)

### LINKED ISSUE #2229: No checks done when providing S3 details.
### What is the issue?

When trying to import from external backup using the documented steps:

https://openeverest.io/documentation/1.15.2/backups_and_restore/dataimporters/pg_dataimporter.html#how-to-import-external-postgresql-backups-using-the-openeverest-ui 

Using incorrect/fake information leads to a seemingly successful environment creation. However, the database will be forever stuck at "importing" state:

Image

### Steps to Reproduce

1. use the following:

Image

2. and you can also use a fake path:

Image

3. You will see all green, as if things are ok:

Image

4. Move through the rest of the steps as you normally would, and click on create database at the end.

5. the database will be initializing, and then stuck at importing.

### Relevant Logs / Output

_No response_

### OpenEverest Version

OpenEverest v1.15.2

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

Client Version: v1.36.0
Kustomize Version: v5.8.1
Server Version: v1.35.3-gke.1389000

### Possible Solution

Proposed solution: actually test data that was input by the user to validate if it's a working/valid S3 setup.

I understand the docs warn about it, to some extent:

```
Important
OpenEverest does not validate file paths or verify the existence of files in the specified storage buckets. Make sure that the backup directory path is correct and accessible.
To verify that the specified path exists, you can use the AWS CLI:
aws s3 ls s3://<bucket-name>/<path-to-backup> --region <region>
```

But this should definitely be improved. Can we check credentials at least? Even better, check creds + bucket exists. Why not check path too?

If nothing else, at least include the warning in the UI.

### Additional Context

_No response_

### LINKED ISSUE #2245: IsBase64Encoded heuristic breaks inline S3 credentials
### What is the issue?

When a user supplies inline S3 credentials on a `DataImportJob` (or
`DatabaseCluster.spec.dataSource.dataImport`) — i.e. uses the convenience
fields `source.s3.accessKeyId` / `source.s3.secretAccessKey` — the mutating
admission webhook routes the values into a Kubernetes `Secret` based on a
naive heuristic that conflates "looks like base64" with "is base64-encoded".

This causes two distinct problems:

1. **False rejection.** Any pair of values where one happens to look like
   valid base64 and the other does not is rejected with:

   > `failed to create or update S3 credentials secret: both accessKeyID and secretAccessKey must be either base64 encoded or not`

   This is extremely common because typical S3-compatible secret access keys
   are 32 or 40 random alphanumerics (length divisible by 4, all chars in the
   base64 alphabet => flagged as base64 even though they were typed as plain
   text).

2. **Silent credential corruption.** When *both* values happen to look like
   base64, the webhook stores them into `secret.Data` raw rather than
   base64-decoding them. The downstream consumer then reads the *base64
   string itself* as the access key/secret. Authentication fails at runtime
   against S3 instead of at admission time.

The intended behavior is presumably "users may pass either plain values or
pre-base64-encoded values"; the heuristic is the wrong way to detect that.

### Steps to Reproduce

Use the "Import" functionality, and add details for importing from S3-compatible storage in "Provide S3 details". See 1 and 2 above to see which keys trigger these issues.

Where the bug lives:

```go
// utils/validators.go
// IsBase64Encoded checks if the given string is a valid base64 encoded string.
func IsBase64Encoded(s string) bool {
  _, err := base64.StdEncoding.DecodeString(s)
  return len(s)%4 == 0 && len(s) > 0 && err == nil
}
```

This returns `true` for *any* string that is non-empty, has length divisible
by 4, and uses only characters in the standard base64 alphabet
(`A-Z`, `a-z`, `0-9`, `+`, `/`, `=`). It does not distinguish "user typed
base64" from "user typed something that happens to look like base64".

### Relevant Logs / Output

_No response_

### OpenEverest Version

v1.15.2

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Two possible solutions follow. 

- **Drop the heuristic entirely.** Treat the inline fields as plain values
  always; always populate `StringData`. Document that values must be
  passed in plain form.
- **Make the encoding explicit.** Add a sibling field such as
  `credentialsEncoding: plain | base64` (default `plain`), or rename the
  base64 path to a separate field (`accessKeyIdBase64` /
  `secretAccessKeyBase64`), and add a selector in the UI.

### Additional Context

_No response_

PutlaTheophila:
## Summary

Today, providing fake/invalid inline S3 credentials or a non-existent backup path on `DatabaseCluster.spec.dataSource` (either `BackupSource` or `DataImport.Source.S3`) is accepted by the API server with no checks. The DB then creates successfully and gets stuck forever in `importing` state because the failure surfaces only at runtime against S3 — there is no surface-level error for the user.

## Fix

Extend `validateDataSource` to:

1. Require the `DataImport` structural fields up front (`source`, `path`, `s3`, `bucket`, `region`, `credentialsSecretName`) — these were previously not checked at the API layer.
2. Run a synchronous read-only S3 probe at admission time via `validateDataSourceS3Access` for both `BackupSource` and `DataImport.Source.S3`:
   - `HeadBucket` to verify credentials + bucket + region + endpoint
   - `ListObjectsV2` with `Prefix=path, MaxKeys=1` to verify the path exists

Credential resolution:
- For `BackupSource`, credentials come from the referenced `BackupStorage`'s credentials Secret.
- For `DataImport.Source.S3`, credentials come from the Secret named by `credentialsSecretName` — which is where the everest-operator's mutating webhook moves the inline `AccessKeyID` / `SecretAccessKey` fields. By the time we reach this handler the inline fields are typically empty, so we read the Secret.

The probe is gated through a package-level seam (`s3ReadOnlyAccessFn`) and short-circuits when `config.Debug` is set, mirroring the existing pattern in `backup_storage.go`.

## Tests

- 7 structural subtests for `DataImport` (missing `source` / `path` / `s3` / `bucket` / `region` / `credentialsSecretName` / ok).
- `TestValidateDataSourceS3Access_DataImport` and `TestValidateDataSourceS3Access_BackupSource` swap the seam via `t.Cleanup` to capture the arguments passed to S3 without making real network calls.
- Both new tests use a 32-char plain-alphanumeric key — the input class the operator's `IsBase64Encoded` heuristic misclassifies (see openeverest/openeverest-operator#955).

## Related

Companion PR in the operator removes the broken `IsBase64Encoded` heuristic so user-supplied credentials are stored verbatim in the credentials Secret: **openeverest/openeverest-operator#955** (Fixes #2245).

Together the two PRs make the inline-credentials path correct end-to-end: the operator stores them as plain strings, and this handler validates them against S3 before the resource is admitted.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/server/handlers/validation/...` pass
- [x] DCO sign-off

Fixes #2229


PutlaTheophila:
@recharte @chilagrow — friendly bump on this one. CI is green (DCO + Labeler), no conflicts with `main`, and the changes are scoped to `internal/server/handlers/validation/` to add admission-time S3 reachability checks for `DataSource`. Addresses issue #2229.

Happy to iterate on anything — let me know if you'd prefer the probe surfaced as a separate validator function or wired the way it currently is.

internal/server/handlers/validation/database_cluster.go
```diff
@@ -430,6 +461,131 @@ func validateDataSource(dataSource *everestv1alpha1.DataSource) error {
...
+		return nil
+	}
+	secret, err := h.kubeConnector.GetSecret(ctx, types.NamespacedName{Namespace: namespace, Name: bs.Spec.CredentialsSecretName})
+	if err != nil {
+		return fmt.Errorf("failed to GetSecret '%s' for BackupStorage '%s': %w", bs.Spec.CredentialsSecretName, src.BackupStorageName, err)
+	}
+	accessKey := string(secret.Data[awsAccessKeyIDField])
+	secretKey := string(secret.Data[awsSecretAccessKeyField])
+
+	var endpoint *string
```
chilagrow:
```suggestion
	var endpoint string
```

Let's use idiomatic approach 🤗

internal/server/handlers/validation/backup_storage.go
```diff
@@ -276,6 +277,75 @@ func s3Access(
...
+//
+//nolint:gochecknoglobals
+var s3ReadOnlyAccessFn = s3ReadOnlyAccess
+
+// s3ReadOnlyAccess validates that the given S3 credentials can reach the bucket
+// and that at least one object exists under the given prefix. It performs only
+// non-mutating operations (HeadBucket + ListObjectsV2 with MaxKeys=1) and is
+// intended for credential paths where the caller has read-only intent — e.g.
+// validating an import source or a backup-restore source.
+func s3ReadOnlyAccess(
```
chilagrow, line 301: There are a lot of duplicated logic from `s3Access()` function. Could you unify the duplicated logic please? 🤗

internal/server/handlers/validation/backup_storage.go
```diff
@@ -276,6 +277,75 @@ func s3Access(
 	return nil
 }
 
+// s3ReadOnlyAccessFn is the indirection used by callers; tests overwrite this
+// to assert which credentials were resolved without making real S3 calls.
+//
+//nolint:gochecknoglobals
+var s3ReadOnlyAccessFn = s3ReadOnlyAccess
```
chilagrow: Let's not do this, using global variable for this case is not recommended. This becomes difficult to maintain and has potential for race.

Instead of testing with mock by injecting fake `s3ReadOnlyAccessFn`, can you test it via API tests https://github.com/openeverest/openeverest/tree/main/api-tests/tests?

chilagrow (commented):
Thanks for your contribution, sorry for delayed review.

PutlaTheophila:
Thanks for the review @chilagrow! Addressed in d1fd0707:

- **Idiomatic endpoint type**: `s3Access` / `s3ReadOnlyAccess` now take `endpoint string` (no more `*string`). Call sites in `database_cluster.go` dropped the nil/empty pointer dance, and `backup_storage.go` callers use `pointer.Get(...)` on the API params.
- **Unify duplicated logic**: extracted `newS3Client(endpoint, accessKey, secretKey, region, verifyTLS, forcePathStyle)` shared by both `s3Access` and `s3ReadOnlyAccess` — the session/HTTP client setup is no longer duplicated.
- **Drop the package-level seam**: removed `s3ReadOnlyAccessFn`; `database_cluster.go` calls `s3ReadOnlyAccess` directly. Replaced the two seam-based unit tests with API tests at `api-tests/tests/datasource-validation.spec.ts` that exercise the validator's credential-resolution path end-to-end (missing Secret, missing BackupStorage, and inline-creds happy path).
- Also added the missing Apache 2.0 copyright headers on the 4 files the Copyright Headers job flagged.

internal/server/handlers/validation/backup_storage.go
PutlaTheophila: Done in d1fd0707 — removed s3ReadOnlyAccessFn entirely. database_cluster.go now calls             
   s3ReadOnlyAccess directly, and the seam-based unit tests are replaced by API tests in             
   api-tests/tests/datasource-validation.spec.ts that cover missing CredentialsSecret, missing       
   BackupStorage, and the inline-creds happy path. Thanks for the catch!

internal/server/handlers/validation/backup_storage.go
PutlaTheophila: Done in d1fd0707 — extracted newS3Client(endpoint, accessKey, secretKey, region, verifyTLS,       
forcePathStyle) shared by both s3Access and s3ReadOnlyAccess. The HTTP client + session + s3.New  
setup now lives in one place, so each function only contains its own probe semantics (read-write  
probe vs. HeadBucket + ListObjectsV2 read-only probe). Thanks!

internal/server/handlers/validation/database_cluster.go
PutlaTheophila: Done in d1fd0707 — s3Access / s3ReadOnlyAccess now take endpoint string instead of *string, and   
both probeBackupSourceAccess / probeDataImportS3Access drop the pointer dance and pass 
bs.Spec.EndpointURL / s3src.EndpointURL directly. Thanks!

api-tests/tests/datasource-validation.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    const resp = await th.createDBClusterWithDataRaw(request, payload)
+    expect(resp.status()).toBeGreaterThanOrEqual(400)
+    expect(resp.status()).toBeLessThan(500)
+    const body = await resp.json()
+    expect(body.message ?? '').toMatch(/BackupStorage/i)
+  })
+
+  test('accept: DataImport with inline credentials skips Secret lookup', async ({request}) => {
```
chilagrow, line 74: Thanks for testing this out.

I was hoping that the test actually tests validates with local s3 bucket such as minio to check newly added code can list objects from s2. This would require creating a databasecluser, then creating a backup in local s3 bucket, and finally creating another cluster with dataimporter to check `validateDataSourceS3Access`. The limitation is there are a lot of setup and creating backup, database cluster takes time. You can find postgresql backup test for example.

I think current tests are not quite appropriate for API tests, let's remove API tests and let's go back unit tests. Testing validation failure case is better suited for unit tests. And please test manually for the success path.

 For unit tests, you can use mock client instead of replacing global variable. 🤗

PR #2283, opened 2026-05-21
Remove hardcoded `everest-system` namespace

chilagrow:
This PR removes hardcoded `everest-system` namespace and infer it dynamically. This is a large PR, I would like the concept of what has changed below to be reviewed.

Ref openeverest/helm-charts#45.

1. At installation, users specify namespace `everestctl install --system-namespace openeverest-system`, with `openeverest-system` being default. I kept `--system-namespace` as hidden flag, using default value is perfectly fine for majority of users.

2. For remaining usage of `everestctl`, it scans deployed Helm releases to find the OpenEverest chart's namespace.
* `everestctl upgrade` when it does not use `--in-cluster` flag
* `everestctl uninstall`
* `everestctl status`
* ...

3. For everest server and `everestctl upgrade --in-cluster`, it reads `NAMESPACE` env var, injected by the Helm chart.

`everestctl` run on the user's machine with access to a kubeconfig. They can query the cluster for Helm releases to find which namespace OpenEverest is installed in. This is the only reliable source since the namespace is user-chosen at install time.

The server and `everestctl upgrade --in-cluster` run as pods inside the cluster. They need the namespace to build a Kubernetes client, but they can't query Helm releases without a client first. The Helm chart solves this by injecting the pod's own namespace via the Kubernetes downward API.

For CI tests, CLI integration test uses `--system-namespace openeverest-system` with explicit OpenEverest installation namespace. While other tests such as `FE E2E Workflow` uses implicit default value when `--system-namespace` is not specified (which is actually also `openeverest-system`).


recharte:
Honestly, I'm not convinced that we should change the default namespace name from `everest-system` to `openeverest-system`. Forcing engineers to type `openeverest-system` every time they interact with `kubectl` introduces administrative spam. Similarly, I think we should keep calling the deployments `everest-server` and `everest-controller`, avoiding the `open` prefix.

Everything else looks good.

chilagrow:
> Honestly, I'm not convinced that we should change the default namespace name from `everest-system` to `openeverest-system`. Forcing engineers to type `openeverest-system` every time they interact with `kubectl` introduces administrative spam. Similarly, I think we should keep calling the deployments `everest-server` and `everest-controller`, avoiding the `open` prefix.
> 
> Everything else looks good.

That's a valid point, let's keep the default as `everest-system`.

PR #2285, opened 2026-05-21
Remove v1 API resources

recharte:


api/openapi/http-api.yaml
chilagrow, line 1: Let's remove unused tags around L16 🤗

chilagrow (commented):
Looks good.

api/openapi/http-api.yaml
recharte: Thanks, I initially left that out so that I wouldn't forget to add tags for the new endpoints and ended up forgetting both adding new ones and removing old ones :sweat_smile:

PR #2320, opened 2026-05-29
Fix storage location edit

solovevayaroslavna:
Fixes two issues with the backup storage edit form that appeared after the v1→v2 API migration:

##  "Edit" button always disabled on open
The v2 API never returns credentials accessKey/secretKey — they are write-only fields cleared by the backend webhook. The edit form default values fell back to "" for these fields, which Zod .nonempty() rejects even when the fields are made .partial(). 
- Fixed by removing the ?? '' fallback so the value remains undefined, which .partial() correctly accepts.

##  422 error on save:
The v1 API used PATCH with partial params — the server did GET→merge→update internally. The v2 API uses PUT with a full CRD body and delegates to k8s.Update() directly, which requires resourceVersion for optimistic concurrency. Fixed by implementing read-modify-write in useEditBackupStorage: GET the current resource (obtaining resourceVersion), merge the edited s3 spec fields, then PUT the full merged object. Includes 409 conflict retry via TanStack Query retry/retryDelay


chilagrow (dismissed):
Looks fine to me.

recharte:
Completely changed the initial approach of this PR, we added the PATCH method with a server side apply and a controller to manage the owner references of the credentials secret. No UI changes are required.

PR #2341, opened 2026-06-07
fix: add timeouts to HTTP clients and ensure local sign-out

### LINKED ISSUE #2203: telemetry: add request timeout and log response body on non-200

### LINKED ISSUE #2236: Complete local sign-out when DELETE /session fails

### LINKED ISSUE #2247: fix(cli): version service HTTP client has no timeout, causing everestctl install/upgrade to hang

ayushshukla1807:
**Fixes #2247**: Replaced the default HTTP client in version_service/client.go with a dedicated *http.Client that has a strict 30-second timeout. This resolves the critical issue where everestctl would hang indefinitely during install/upgrade if the version server was unreachable.
**Fixes #2203**: Added the same 30-second timeout to the background telemetry HTTP client (internal/server/telemetry.go). Also added robust response body logging for non-200 responses to aid debugging.
**Fixes #2236**: Wrapped the DELETE /session call in auth.provider.tsx with a try/catch block. This ensures that even if the API throws an error (like 500 Internal Server Error), the local application state will still be successfully cleared and the user will be properly signed out.

### DIFF
--- a/internal/server/telemetry.go
+++ b/internal/server/telemetry.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"time"
@@ -62,14 +63,19 @@ func (e *EverestServer) report(ctx context.Context, baseURL string, data Telemet
 	}
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Accept", "application/json")
-	resp, err := http.DefaultClient.Do(req)
+	
+	client := &http.Client{
+		Timeout: 30 * time.Second,
+	}
+	resp, err := client.Do(req)
 	if err != nil {
 		e.l.Error(errors.Join(err, errors.New("failed to send telemetry request")))
 		return err
 	}
 	defer resp.Body.Close() //nolint:errcheck
 	if resp.StatusCode != http.StatusOK {
-		e.l.Info("Telemetry service responded with http status ", resp.StatusCode)
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		e.l.Infof("Telemetry service responded with http status %d, body: %s", resp.StatusCode, string(bodyBytes))
 	}
 	return nil
 }
--- a/pkg/version_service/client.go
+++ b/pkg/version_service/client.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"slices"
 	"strings"
+	"time"
 
 	perconavs "github.com/Percona-Lab/percona-version-service/versionpb"
 	goversion "github.com/hashicorp/go-version"
@@ -58,13 +59,19 @@ type Interface interface {
 	GetEverestMetadata(ctx context.Context) (*perconavs.MetadataResponse, error)
 }
 
+const defaultHTTPTimeout = 30 * time.Second
+
 type versionServiceClient struct {
-	url string
+	url        string
+	httpClient *http.Client
 }
 
 // New returns a new version service client.
 func New(url string) Interface { //nolint:ireturn
-	return &versionServiceClient{url: url}
+	return &versionServiceClient{
+		url:        url,
+		httpClient: &http.Client{Timeout: defaultHTTPTimeout},
+	}
 }
 
 //nolint:gochecknoglobals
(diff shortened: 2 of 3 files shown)


recharte:
Hi @ayushshukla1807, thanks for this, and sorry it sat two months without a word.

Both Go changes here lose on ordering rather than on merit. #2248 is moving forward for the version service — open since the minute #2247 was filed, and it carries tests — and #2225 for telemetry, which comes from that issue's reporter and has been through two rounds of review already. The `version_service` hunks are in fact identical: you and @abhisheksainimitawa both wrote what the issue proposed.

The local sign-out fix is the part that exists only in this PR. #2236 needs a triage pass before I can give you a straight answer on it, and I'll come back here once it has one.

Nothing for you to change meanwhile — please don't spend time polishing this in the interim.

recharte:
Coming back on the sign-out half as promised, and correcting myself while I'm here: I said this morning that it was the one part unique to this PR. That was wrong. #2236 is already fixed on `release-2.0`, and on `main` @ashnaaseth2325-oss's #2143 has been open since 4 May with tests — a month before this PR — so that part is superseded as well.

Which leaves nothing here that isn't covered elsewhere. I'd suggest closing it, but I'd rather that was your call than mine. None of the three was a judgement on your code — it's ordering in each case.

PR #2353, opened 2026-06-10
Backup storage e2e tests

solovevayaroslavna:
Backup Storage E2E Tests
This PR adds and improves end-to-end tests for backup storage features

Main changes:
Uses one shared namespace (everest)  instead of separate namespaces for each provider type
Enabled backup storage tests: Uncommented and activated backup storage tests in the CI pipeline.
Better backup storage API: Updated API paths to include cluster information for consistency.

Refactored data source logic: Moved field value reconciliation into a separate utility with dedicated tests for better code organization
Optimistic UI updates: Added smart data synchronization so UI updates immediately without waiting for server response, then rolls back if something fails.

### DIFF
--- a/.github/workflows/dev-fe-e2e.yaml
+++ b/.github/workflows/dev-fe-e2e.yaml
@@ -138,40 +138,14 @@ jobs:
           --helm.set operator.image=ghcr.io/openeverest/openeverest-operator-dev \
           --helm.set olm.catalogSourceImage=ghcr.io/openeverest/openeverest-catalog-dev
 
-      - name: Add everest-ui namespace
+      - name: Add everest namespace
         shell: bash
         run: |
-          ./bin/everestctl namespaces add everest-ui -v \
+          ./bin/everestctl namespaces add everest -v \
           --operator.mongodb=true \
           --operator.postgresql=true \
           --operator.mysql=true \
-          --skip-wizard
-
-      - name: Adding psmdb namespace
-        shell: bash
-        run: |
-          ./bin/everestctl namespaces add psmdb-only -v \
-          --operator.mongodb=true \
-          --operator.postgresql=false \
-          --operator.mysql=false \
-          --skip-wizard
-
-      - name: Adding pxc namespace
-        shell: bash
-        run: |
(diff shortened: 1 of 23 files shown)


.github/workflows/dev-fe-e2e.yaml
```diff
@@ -225,9 +199,9 @@ jobs:
           kubectl -n everest-system describe pods
           kubectl -n everest-monitoring describe pods
           kubectl -n everest-system logs deploy/everest-server
-          kubectl -n pxc-only describe pods
-          kubectl -n psmdb-only describe pods
-          kubectl -n pg-only describe pods
+          kubectl -n pxc-only describe pods || true
+          kubectl -n psmdb-only describe pods || true
+          kubectl -n pg-only describe pods || true
```
chilagrow:
```suggestion
```

CI no longer installs database operator, could we remove describing pods that won't exist?

Makefile
```diff
@@ -360,33 +360,30 @@ undeploy: build-cli-debug ## Undeploy Everest from K8S cluster using Everest CLI
...
-	$(LOCALBIN)/everestctl namespaces add $(DB_NAMESPACES) -v \
-	--operator.mongodb=false \
-	--operator.postgresql=false \
+	--operator.postgresql=true \
 	--operator.mysql=true \
+	--take-ownership \
 	--skip-wizard
 
+.PHONY: cleanup-legacy-provider-namespaces
+cleanup-legacy-provider-namespaces: ## LEGACY manual one-time migration cleanup for upgraded environments.
```
chilagrow: I don't quite understand why we need this, there is no longer code to add namespaces such as `psmdb-only` so why is it needed? What is the upgraded environment?

chilagrow (commented):
Thanks for adding the test 🤗 Just a couple of comment on CI and makefile

Makefile
solovevayaroslavna: This is the reason why I wanted you to take a look, in fact it is only needed for manual deletion. So I'll delete it! I'm glad we agreed on the same thing

chilagrow (approved):

Issue #2352, opened 2026-06-10
Create database selection populates different database name in v2


chilagrow:
### What is the issue?

In OpenEverest v2, when you add more than 1 provider, there is a drop down selection for choice of the database. It shows 2 `PostgreSQL`s.

<img width="318" height="200" alt="Image" src="https://github.com/user-attachments/assets/2c533221-53f2-4e57-8f5f-cb7c2445efc1" />

### Steps to Reproduce

I used tilt to configure PXC using this PR https://github.com/openeverest/openeverest/pull/2346/changes. There were couple tweaks I made on top of the branch, which is found in the comments of PR. Then I started tilt environment and clicked to create a database.

```
make dev-up
```

### OpenEverest Version

v2.0.0-dev.1

### Installation Method

Manifests / Manual

### Kubernetes Distribution

kind

AdityaPimpalkar:
Hi @chilagrow 

I ran into the same behaviour while testing CloudNativePG. I made a few temporary changes to help me with exactly which provider I was selecting but we should fix this properly for better UX.

The provider dropdown label is derived here:

https://github.com/openeverest/openeverest/blob/fb525a0433caf3defb783244b894eeb383fa4fb8/ui/apps/everest/src/components/create-db-button/create-db-button.tsx#L197-L199

It passes `provider.metadata.name` into `dbEngineToDbType()`:

https://github.com/openeverest/openeverest/blob/fb525a0433caf3defb783244b894eeb383fa4fb8/ui/packages/utils/src/db/db.ts#L14-L23

That function expects `DbEngineType` values (e.g. `percona-server-mongodb`, `pxc`, `postgresql`), but provider names are plugin-defined and often differ - for example `percona-mongodb-server` or `provider-cloudnative-pg`. When there’s no match, the default branch returns `DbType.Postresql`, which `humanizeDbType()` renders as `PostgreSQL`. That’s why multiple providers can show up with the same label.

There’s already a `// TODO rewrite for provider logic at the call site`, which lines up with this.

**Suggestions:**

1. **Short term:** Render `provider.metadata.name` directly in the dropdown (or map known provider names to the correct `DbEngineType` values).
2. **Better UX:** Add a `displayName` field to the Provider CRD (similar to BackupClass/Plugin) and render that in the UI, keeping `metadata.name` as the stable identifier.

CC @solovevayaroslavna would like to know your thoughts on this

Happy to help with a PR if useful.

alokkumardalei-wq:
[...]

PR #2354, opened 2026-06-10
UI-2352 fix of create database button menu (provider list) label duplication

solovevayaroslavna:
This pull request fixes the #2352 issue and primarily removes the usage of utility functions like `beautifyDbTypeName`, `dbEngineToDbType`, and `humanizeDbType` from legacy parts of the UI codebase, because in v2 we don't need them.


chilagrow (commented):
We don't enforce provider name length as far as I'm aware.

Image

Image

solovevayaroslavna:
@chilagrow thanks for testing! Yeah, let me add some fixes to solve this

PR #2361, opened 2026-06-11
Implement API auth token management with refresh tokens

recharte:
Implements the following design spec (except PATs): https://github.com/openeverest/specs/blob/main/specs/005-authentication-system-redesign.md


chilagrow (approved):
Looks good 👍

PR #2366, opened 2026-06-12
fix: provider status message silently dropped

### LINKED ISSUE #2359: Provider status messages are silently dropped
### What is the issue?

Provider `Status()` methods can return human-readable messages via helpers like `controller.Provisioning("waiting for CloudNativePG cluster to initialize")`, but those messages never reach the `Instance` CR, provider logs, or the UI. Operators only see a bare phase (e.g. `Provisioning`) with no explanation of what the provider is waiting on.

### Steps to Reproduce

1. Implement a provider `Status()` that returns a descriptive message, e.g.:
    `return controller.Provisioning("waiting for CloudNativePG cluster to initialize"), nil`
2. Create an `Instance` that remains in `Provisioning` while waiting on an underlying engine resource.
3. Observe: `kubectl get instance <name> -o yaml` -> `status.phase: Provisioning, no message field`

### Root cause 

`ToV2Alpha1()` drops `Status.Message`
https://github.com/openeverest/openeverest/blob/5d8f7f19b21e40e967995dd8138dee6fae64f43d/provider-runtime/controller/common.go#L421-L426

### OpenEverest Version

v2

### Possible Solution

1. Add `message` to `InstanceStatus` (API + CRD + generated clients).
2. Update `ToV2Alpha1()` to copy Message

Confirmed with @chilagrow here - https://cloud-native.slack.com/archives/C09RRGZL2UX/p1780879288377799?thread_ts=1780409575.582159&cid=C09RRGZL2UX

I will raise a PR for this 👍

AdityaPimpalkar:
Closes #2359

CC @chilagrow


provider-runtime/reconciler/provider.go
```diff
@@ -480,7 +480,9 @@ func (r *ProviderReconciler) Reconcile(ctx context.Context, req reconcile.Reques
 		return reconcile.Result{}, err
 	}
 
-	in.Status = status.ToV2Alpha1()
```
chilagrow, line 483: I'm curious why this needed to change? 🤔

chilagrow (commented):
Thanks for your contribution! 🤗

provider-runtime/reconciler/provider.go
AdityaPimpalkar: @chilagrow because `ToV2Alpha1()` returns a new `InstanceStatus` struct with only `Phase` and `Message` mapped to it, keeping it as `in.Status = status.ToV2Alpha1()` wipes the fields set for `InstanceStatus` earlier in the same reconcile pass by `BackupConfigured` / `DataSourceReady` conditions (lines 457–472)

Unless that's by design?

AdityaPimpalkar: Also, there seem to be unit tests missing for controllers and reconcilers logic. Would be nice to add some coverage for those. LMK if that would be useful as well 👍

chilagrow: I see it now, thank you. This was an bug that I was unaware. It was wiping conditions before 🙈

chilagrow (approved):
Thanks for your contribution!

One last step for updating CR is updating https://github.com/openeverest/helm-charts `v2` branch with new change. You are welcome to contribute to it 🤗. 

https://github.com/openeverest/helm-charts/blob/v2/CONTRIBUTING.md#4-prepare-your-pr

chilagrow:
Actually don't worry about helm-charts for this case, I just created a PR https://github.com/openeverest/helm-charts/pull/60. There were couple other things that needed updating 🤗

PR #2367, opened 2026-06-12
V2 hub implementation

### LINKED ISSUE #2476: [Core] Support provider discovery
### Task Description

https://github.com/openeverest/specs/blob/main/specs/007-extension-hub.md

### Reasoning / Benefit

TBA

### Additional Context

_No response_

spron-in:
Removed PluginInstallation  CR from OpenEverest and added InstalledExtension  as per the spec. Both Plugins and Providers can use it. It does not break existing plugins and providers - they continue to work. But it is a breaking change as I just removed PluginInstallation resource without the fallback. PR is quite big.

Removed hardcoded /plugins UI url that was showing plugins to replace it with the thing below.


ui/apps/everest/src/hooks/api/plugins/useAllPlugins.ts
```diff
@@ -1,39 +0,0 @@
...
-  const resp = await window.fetch('/v1/plugins', { headers });
-  if (!resp.ok) return [];
-  return resp.json();
-}
-
-/**
- * Returns the list of all enabled plugins the current user can access.
- * Used by the Plugins overview page.
- */
-export const useAllPlugins = () => {
```
solovevayaroslavna, line 34: What is the intended UI replacement for the removed /plugins overview page and useAllPlugins hook? Is it not ready yet or was it not planned in this PR?

ui/apps/everest/src/pages/db-cluster-details/db-cluster-details.tsx
```diff
@@ -70,13 +70,13 @@ const WithPermissionDetails = ({
...
-  // 2. namespace: only plugins with an active PluginInstallation in this namespace
+  // 2. namespace: only plugins enabled for this namespace via an InstalledExtension
   const { plugins } = usePlugins();
   const { data: nsPlugins } = usePluginsForNamespace(namespace);
   const engineType = instance?.spec?.provider;
 
   // Build a set of plugin names enabled in this namespace.
-  // When no PluginInstallation CRs exist in the namespace, skip filtering entirely.
+  // When no InstalledExtension entry covers this namespace, skip filtering entirely.
   const enabledInNs = nsPlugins?.length
```
solovevayaroslavna: Could we rename nsPlugins to pluginsEnabledInNamespace (or enabledPluginsForNamespace)?
After switching from PluginInstallation to InstalledExtension, this variable now represents plugins enabled by InstalledExtension namespace rules, so a more explicit name would improve readability and reduce semantic drift.

ui/apps/everest/src/hooks/api/plugins/usePluginsForNamespace.ts
```diff
@@ -48,9 +48,10 @@ async function fetchPluginsForNamespace(
 }
 
 /**
- * Returns the list of plugins with an active PluginInstallation in the given
- * namespace. Used to filter cluster-detail tabs to only plugins the namespace
- * operator has explicitly enabled.
+ * Returns the list of plugins enabled for the given namespace via an
```
solovevayaroslavna, line 52: Could we revisit the PluginDescriptor interface naming/documentation to align with the InstalledExtension model?
The runtime behavior is correct, but the interface context still reads as plugin-installation-era semantics. A small naming/doc update would make the contract clearer and reduce confusion for future maintainers.

ui/apps/everest/src/hooks/api/plugins/usePluginsForNamespace.ts
```diff
@@ -48,9 +48,10 @@ async function fetchPluginsForNamespace(
...
+ * InstalledExtension entry (spec.plugin.namespaces[] or scope: Cluster with
```
solovevayaroslavna: Will we update the architecture docs to replace PluginInstallation terminology with InstalledExtension where applicable? (I mean docs/ui/architecture/plugins-architecture.md.)

controller
recharte, line 1: Please don't commit build binaries.

PROJECT
```diff
@@ -75,20 +75,20 @@ resources:
...
     crdVersion: v1
-    namespaced: true
+    namespaced: false
   controller: true
   domain: openeverest.io
-  group: plugin
-  kind: PluginInstallation
-  path: github.com/openeverest/openeverest/v2/api/plugin/v1alpha1
+  group: core
+  kind: InstalledExtension
```
recharte: Do you think that `Plugin` and `InstalledExtension` should be part of the `core` API group? IMO I think we should move them to an `extension` (or similar) group.

A couple of reasons why I think a separate group makes more sense:
* The hub/extension system is entirely optional — a cluster can run providers installed by raw helm install with no hub at all.
* Independent, faster version cadence — the extension/hub machinery defined in the spec has a lot of points to be added in multiple phases meaning that it will likely churn at `v1alpha1` for a while. Isolating the `core` means we can work towards a stable `v1` without being dragged by an unstable group.

commands/extension/namespace.go
```diff
@@ -0,0 +1,112 @@
...
+		Use:     "remove [flags]",
+		Args:    cobra.NoArgs,
+		Example: "everestctl extension namespace remove --name hello -n my-namespace",
+		Long:    "Disable an extension in a namespace by removing it from spec.plugin.namespaces[]",
+		Short:   "Disable an extension in a namespace",
+		PreRun:  namespaceRemovePreRun,
+		Run:     namespaceRemoveRun,
+	}
+	namespaceRemoveCfg = &cliext.DisableConfig{}
+)
```
recharte, line 58: This is a question I should have asked in the initial plugin PR/spec but better late than never...
I don't quite understand why we need to manage plugin's namespaces. It seems to me like this introduces a lot of complexity in the controller to manage RBAC policies. This can also be a security concern since it requires the controller to have elevated permissions.
@spron-in can you please explain the use cases that we want to support that require this architecture?

pkg/cli/plugins/disable.go
recharte, line 1: Shouldn't this be completely replaced by `pkg/cli/extension/disable.go`?

pkg/cli/plugins/enable.go
recharte, line 1: Shouldn't this be completely replaced by `pkg/cli/extension/enable.go`?

pkg/cli/plugins/install.go
recharte, line 1: Shouldn't this be completely replaced by `pkg/cli/extension/install.go`?

ui/apps/everest/src/hooks/api/plugins/useAllPlugins.ts
spron-in: It is the new https://github.com/openeverest/plugin-hub as I showed in a demo in our slack.

commands/extension/namespace.go
spron-in: I'm not sure I get your point. It think it is actually the other way around: if generic plugin is namespace scoped - it gets less permissions. 
And for generic plugins it is normal to be namespace scoped. Users might want to keep them isolated per namespace. Keep secrets isolated, keep permissions isolates, etc.

Maybe I misunderstood.

ui/apps/everest/src/pages/db-cluster-details/db-cluster-details.tsx
spron-in: Fixed.

commands/extension/namespace.go
recharte: Okay maybe my concern isn't explained properly. The thing that concerns me the most is having to create and manage Roles and RoleBindings that are defined in the Plugin CR. Maybe it would be easier for me to understand if you try to explain how are Plugin's RBAC supposed to work and how that ties to namespace enabling/disabling.

ui/apps/everest/src/hooks/api/plugins/usePluginsForNamespace.ts
spron-in: Fixed.

ui/apps/everest/src/hooks/api/plugins/usePluginsForNamespace.ts
spron-in: @solovevayaroslavna we have updated the spec and generic-public-architecture.md in docs. I actually think to keep the spec only later on.

pkg/cli/plugins/disable.go
spron-in: you are right, fixed.

pkg/cli/plugins/enable.go
spron-in: you are right, fixed.

pkg/cli/plugins/install.go
spron-in: you are right, fixed.

PROJECT
spron-in: I don't have strong opinion here. 
Moving it to `plugin` or `extension` looks ok. I like the idea of independence for the API.

Also generic plugins might allow users to add their own CRs. We will need to keep a separate group for those. 
So yeah, let me go with `extension` group.

commands/extension/namespace.go
spron-in: `everestctl extension namespace add/remove` does only one thing: mutate `spec.plugin.namespaces[]` on the InstalledExtension. The reconciler then diffs that list against the currently-bound per-namespace Role/RoleBinding set (filtered by the `installedExtensionLabel`) and creates / deltes accordingly.

There is no separate RBAC API the user touches, no per-namespace install flow, no "register plugin in namespace" CR. The namespace list is the source of truth and everything RBAC-shaped is derived state.

One of the possible solutions here is to do everything cluster scope, which might be worse. It breaks tenant-isolation. 

But you are right that controller holds elevated permissions. It is kinda of a trade off. 

Let me know if it makes sense.

PROJECT
spron-in: I did the update

commands/extension/namespace.go
chilagrow: Let me see if I understand this in the context of provider installation. Does it mean providers such as psmdb will be installed in namespace-scope rather than cluster-scope?

recharte: I don't think this touches providers, they would always be installed cluster-wide.

recharte (approved):

PR #2368, opened 2026-06-12
feat(auth): synchronize logout state across browser tabs

ashnaaseth2325-oss:
## SUMMARY

Many OpenEverest users work across multiple tabs at the same time. Previously, logging out from one tab only cleaned up that tab, leaving other tabs with stale authenticated state.

This PR adds cross-tab logout synchronization and centralizes logout cleanup, ensuring all open tabs clear cached data, stop background activity, and stay in sync when a user signs out.

---

### Before

```ts
const logout = async () => {
  await api.delete('/session');
  await setLogoutStatus();
};
```

### After

```ts
const logout = async () => {
  try {
    await api.delete('/session');
  } catch {}

  broadcastLogout();
  await runLogoutCleanup();
};
```

### DIFF
--- a/ui/apps/everest/src/contexts/auth/auth.provider.tsx
+++ b/ui/apps/everest/src/contexts/auth/auth.provider.tsx
@@ -1,3 +1,17 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


ashnaaseth2325-oss:
Hello @solovevayaroslavna 
Kindly review the PR whenever possible. 
Thank You!

PR #2372, opened 2026-06-15
cli: add everestctl auth login

VijetaPriya47:
Adds the CLI side of the authentication flow from [spec 005](https://github.com/openeverest/specs/blob/main/specs/005-authentication-system-redesign.md). Users can now run `everestctl auth login --server <url>` to exchange credentials for an access JWT and refresh token, persisted to
~/.config/everest/config.yaml (mode 0600) iin a three-section format modelled on kubeconfig — servers, users, and contexts are kept separate, with a context acting as a thin pointer linking a named server to a named user. This avoids duplicating server URLs or tokens when multiple accounts share the same server, makes token rotation atomic (one user entry to update), and sets up clean future commands like config get-servers and config get-users.

By default, context names are derived as username@host:port (overrideable via --context-name).

Also adds a Refresh() helper in pkg/cli/auth for token rotation, intended for use by future data-plane commands.


pkg/cli/config/config.go
```diff
@@ -0,0 +1,101 @@
...
+		return "", fmt.Errorf("could not determine home directory: %w", err)
+	}
+	return filepath.Join(home, configDir, configFile), nil
+}
+
+// Load reads the config file at path. Returns an empty Config if the file does not exist.
+func Load(path string) (*Config, error) {
+	data, err := os.ReadFile(path)
+	if os.IsNotExist(err) {
+		return &Config{APIVersion: "v1", Kind: "Config"}, nil
```
recharte: I think we should avoid using `kind: Config` or `apiVersion: v1`. These generic names will conflict with Kubernetes kubeconfigs. What do you think about using the following?
```
apiVersion: config.openeverest.io/v1alpha1
kind: ClientConfig
```

pkg/cli/config/config.go
```diff
@@ -0,0 +1,101 @@
...
+	ExpiresAt    time.Time `yaml:"expiresAt"`
+}
+
+// DefaultPath returns the default path to the config file (~/.config/everest/config.yaml).
+func DefaultPath() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("could not determine home directory: %w", err)
+	}
+	return filepath.Join(home, configDir, configFile), nil
```
recharte: `XDG_CONFIG_HOME` is usually `~/.config` but it might not be. We should use `osUserConfigDir()` to get the `XDG_CONFIG_HOME` and then just join `everest/config.yaml`.

pkg/cli/config/config.go
```diff
@@ -0,0 +1,101 @@
...
+	Context Context `yaml:"context"`
+}
+
+// Context holds the server URL and credentials for one Everest server.
+type Context struct {
+	Server       string    `yaml:"server"`
+	AccessToken  string    `yaml:"accessToken"`
+	RefreshToken string    `yaml:"refreshToken"`
+	ExpiresAt    time.Time `yaml:"expiresAt"`
+}
```
recharte, line 72: I know I was the one suggesting to include the user name in the context name to allow for managing multiple user accounts at the same time but after thinking about this deeply I now see that that was a mistake. If we hardcode tokens directly inside the context, we lose the ability to reuse configurations cleanly, e.g. if we have a single development server but want to switch between your admin token and a developer token, a combined format forces you to duplicate the server URL over and over.

Kubernetes solves context management in a much more elegant way that I think we should copy. A Context should simply link a specific User to a specific Server, where users and servers are defined separately. A few benefits of this architecture:
* Avoid data duplication:  If we have 3 distinct users (e.g., `admin`, `developer`, `read-only`) who all need access to 3 different servers (e.g., `dev`, `staging`, `prod`), a tightly coupled format requires us to duplicate server URLs or user tokens repeatedly. By separating them, we write the 3 servers once and the 3 users once. You then create lean contexts that point to them like pointers in code.
* Atomic Credential Rotation: Tokens expire frequently, but server endpoints rarely change. When an access token expires, we only need to find and overwrite one specific string under the users block.
* Multi-Cluster Impersonation: An advanced user or administrator often wants to run commands on a specific cluster using a completely different identity or service account token for debugging. With a decoupled architecture, a user can instantly create a temporary context mapping their local production server to a restricted test user identity without affecting their primary administrative configurations.
* Integration with External IdPs: The UI/API already integrates with external OIDC providers, we'll eventually need to do the same for the CLI. By isolating users into their own block, we can easily swap a static `accessToken` field for an external credential plugin configuration.
* Cleaner UX: Decoupling allows us to build highly specific management commands that make sense to the user. We can easily offer distinct command sub-trees like:
  * `everestctl config get-servers` (Lists all known API endpoints)
  * `everestctl config get-users` (Lists all saved user sessions and identities)
  * `everestctl config use-context <name>` (Instantly switches the active environment)

If everything were bundled into a single object, filtering and presenting this information to your users on the command line would become highly messy.

The only caveat it that we'll need to be careful generating user and server names to ensure we don't have collisions.

e.g.
```
apiVersion: config.openeverest.io/v1alpha1
kind: ClientConfig
currentContext: dev-admin
contexts:
  - name: dev-admin
    context:
      server: dev-server
      user: admin-dev
  - name: prod-admin
    context:
      server: prod-server
      user: admin-prod
servers:
  - name: dev-server
    server:
      url: https://dev.xyz
  - name: prod-server
    server:
      url: https://prod.abc
users:
  - name: admin-dev
    user:
      accessToken: token_for_dev_server_xyz
      refreshToken: token_for_dev_server_xyz
      expiresAt: expiry_time
  - name: admin-prod
    user:
      accessToken: token_for_prod_server_abc
      refreshToken: token_for_prod_server_abc
      expiresAt: expiry_time
```

recharte (changes requested):
Great job @VijetaPriya47, this is definitely heading in the right direction but I think that there are a few things we can improve. Namely, I suggest we change the config format to be more like kubeconfig. Please see the comments below for more details.

(This is still a partial review, I haven't finished the refresh mechanism yet but thought it would be better to submit this feedback ASAP since it will lead to a considerable change.)

pkg/cli/config/config.go
recharte: Regarding auto-generating user and context names I think they can both be `user@server` by default.
Let's say that the user ran the `everestctl login` command twice with the `admin` user name both times and used `https://dev.xyz` in the first run and `https://prod.abc` in the second one. If we rework the above example for auto-generated names we can have something like:
```
apiVersion: config.openeverest.io/v1alpha1
kind: ClientConfig
currentContext: dev-admin
contexts:
  - name: admin@dev.xyz
    context:
      server: dev.xyz
      user: admin@dev.xyz
  - name: admin@prod.abc
    context:
      server: prod.abc
      user: admin@prod.abc
servers:
  - name: dev.xyz
    server:
      url: https://dev.xyz
  - name: prod.abc
    server:
      url: https://prod.abc
users:
  - name: admin@dev.xyz
    user:
      accessToken: token_for_dev_server_xyz
      refreshToken: token_for_dev_server_xyz
      expiresAt: expiry_time
  - name: admin@prod.abc
    user:
      accessToken: token_for_prod_server_abc
      refreshToken: token_for_prod_server_abc
      expiresAt: expiry_time
```

commands/auth/login.go
```diff
@@ -0,0 +1,106 @@
...
+		username, err := accountscli.PopulateUsername(cmd.Context())
+		if err != nil {
+			output.PrintError(err, logger.GetLogger(), loginCfg.Pretty)
+			os.Exit(1)
+		}
+		loginOpts.Username = username
+	}
+
+	if loginOpts.Password == "" {
+		password, err := accountscli.PopulatePassword(cmd.Context())
```
recharte: I think we should create a new method for prompting passwords instead of using `accountscli.PopulatePassword`. 
`accountscli.PopulatePassword` was meant for the command where users change passwords. It calls [ValidatePassword](https://github.com/VijetaPriya47/openeverest/blob/911a8f548bec3f007e17e291e3360e7328ea58cc/pkg/accounts/cli/utils.go#L59-L65) which enforces password requirements. There's no point in running this check for existing passwords. While this isn't an issue in production, it breaks in local development where the admin password is `admin` (less than 6 characters):
```
go run cmd/cli/main.go auth login
❓ Everest server URL: http://localhost:8080
❓ Provide username: admin
❓ Provide password: *****
❌ password may contain only letters, numbers and specific special characters (@*#$%^&+=!_-), and must be at least 6 characters long

exit status 1
```

Passing the password explicitly via the flag (`-p admin`) bypasses the validation and succeeds, so this issue only affects interactive mode prompts.

pkg/cli/auth/login.go
```diff
@@ -0,0 +1,154 @@
...
+
+	cfg, err := config.Load(cfgPath)
+	if err != nil {
+		return err
+	}
+	cfg.UpsertServer(srvName, config.Server{URL: opts.Server})
+	cfg.UpsertUser(userName, config.User{
+		AccessToken:  tokenResp.AccessToken,
+		RefreshToken: tokenResp.RefreshToken,
+		ExpiresAt:    time.Now().Add(time.Duration(tokenResp.ExpiresIn) * time.Second),
```
recharte, line 112: This is unrelated to your changes but this makes me question if the API is designed correctly... Maybe it should return an absolute timestamp (`ExpiresAt`) instead of `ExpiresIn` 🤔

recharte (commented):
Well done @VijetaPriya47, it looks great. Just a small thing on the password prompt in interactive mode.

commands/auth/login.go
VijetaPriya47: then what about using  tui.NewInputPassword() instead of accountcli.PopulatePassword?

recharte: Yes, let's use that directly, no need to create a new wrapper for it.

pkg/cli/auth/login.go
VijetaPriya47: but what about different timestamps for every user?

recharte: The backend already [calculates the expiration time](https://github.com/openeverest/openeverest/blob/32b986887b1bbd0489d9971adad4bc79aaf712a9/pkg/session/manager.go#L145-L146) and stores it in the accessToken's [`exp` claim](https://www.rfc-editor.org/info/rfc7519/#section-4.1.4). Since the JWT already tracks this, keeping `ExpiresIn` or `ExpiresAt` is actually redundant. 

I propose we make the JWT our sole source of truth and remove those fields. We can update the client config file in this PR, and tackle the wider API cleanup in a future PR. For the refresh mechanism, we will just need to parse the token and extract the `exp` claim. This should be super straightforward using our existing [JWT library](https://pkg.go.dev/github.com/golang-jwt/jwt/v5#Claims.GetExpirationTime). 

*Note: Everything is handled in UTC per the RFC standard.*

recharte: TLDR: remove the `ExpiresAt` from the config file.
In the future, when we integrate with the refresh mechanism we should check the expiration timestamp encoded in the AccessToken.

recharte: Actually, I'm mixing stuff up... Please wait, let me properly address this...

recharte: I was mixing stuff up because we have a bug in the API server :see_no_evil: 
https://github.com/openeverest/openeverest/blob/d2897c0ff4fd7199320db8985d1e8de054e53f93/internal/server/auth.go#L170
The `ExpiresIn` in the response relates to the `RefreshToken` and not the `AccessToken`. This is precisely because the `RefreshToken` is an opaque token instead of a JWT.
Having said this, your question about different timezones is very relevant but the format you are using to store the timestamp already encodes the timezone data in the timestamp so we are fine. Using a standard UTC timestamp would also work but there's no need to take on this change right now, it's good as is.
We'll just need to fix the API server to return the refresh token's TTL instead of the access token's TTL.

recharte (approved):
🚀

VijetaPriya47:
should I fix the linter issues also, or is it fine to merge?

recharte:
> should I fix the linter issues also, or is it fine to merge?

The `release-2.0` branch fails the `Check` CI pipeline, it's fine.

PR #2375, opened 2026-06-15
Support load balancer in Instance spec

chilagrow:
It also adds helper function used for common cases such as `LoadBalancer` config.

Refs https://github.com/openeverest/helm-charts/pull/57.
Refs https://github.com/openeverest/provider-percona-server-mongodb/pull/55.


api/core/v1alpha1/instance_types.go
```diff
@@ -282,12 +284,44 @@ type ComponentSpec struct {
...
 }
 
+type Service struct {
+	// ServiceType defines how the component is exposed.
+	// Common ServiceTypes include standard Kubernetes types "ClusterIP", "LoadBalancer"
+	// and "NodePort".
+	// The provider ultimately decides which service types are supported, and may
+	// support additional ServiceTypes.
+	// +kubebuilder:default:=ClusterIP
+	ServiceType string `json:"serviceType,omitempty"`
```
recharte: Should we use `corev1.ServiceType` instead?

api/core/v1alpha1/instance_types.go
```diff
@@ -282,12 +284,44 @@ type ComponentSpec struct {
...
+}
+
+type SourceRanges []string
+
+type LoadBalancerService struct {
+	// Annotations is a map of key-value pairs for annotating the Service.
+	// Commonly used to configure cloud provider settings
+	// (e.g., AWS ELB annotations, GCP load balancer settings).
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
```
recharte: Annotations (and labels which we might need to support as well) are generic metadata that could be applied to any service type, I'm not sure it should be inside the LoadBalancerService. WDYT?

api/core/v1alpha1/instance_types.go
chilagrow: We support existing types in `corev1.ServiceType` except `ExternalName`, so initially I wanted to have separate thing but I agree it make sense to use `corev1.ServiceType`. Still each provider is responsible for deciding what is supported within `corev1.ServiceType` and validating it 🤗

api/core/v1alpha1/instance_types.go
chilagrow: <s>Can do. I wanted to make it declarative that if you chose loadbalancer, you may fill `SourceRanges` and `Annotations`. If you choose ClusterIP or Node Port, you don't have options to fill. It's a bit like how we did MonitoringConfig PMM spec and BackupStorage S3 spec, what do you think?</s>

chilagrow: Agreed. To keep annotation on top level 👍

recharte (dismissed):

PR #2407, opened 2026-06-18
Implement Preset

### LINKED ISSUE #2539: API: add InstancePreset CRD

chilagrow:
Spec https://github.com/openeverest/specs/pull/6.

This PR implements phase 1 of Instance Preset spec.
It includes simple API tests for ease of testing resolving default values from the given namespace.

Refs https://github.com/openeverest/openeverest/issues/1820.
Refs https://github.com/openeverest/helm-charts/pull/58.


internal/server/handlers/k8s/instance_preset.go
```diff
@@ -0,0 +1,350 @@
...
+			}
+		}
+	}
+
+	if len(filtered) == 0 {
+		return nil, nil
+	}
+
+	return getMostRecentlyCreated(convertMonitoringConfigsToObjects(filtered)), nil
+}
```
recharte, line 318: It's fine for now, you don't need to change it in this PR but I'm wondering if we should introduce this kind of default getter as a property of each model, e.g. `findDefaultMonitoringConfig` would be defined somewhere as part of a `MonitoringConfig` package or something like that. I'm not sure exactly how this would look like but since we might need this default logic in other places maybe it would make sense to extract it from the instance preset.

internal/server/handlers/rbac/instance_preset.go
```diff
@@ -0,0 +1,68 @@
...
+// ResolveInstancePreset returns an instance preset with namespace defaults, gated by RBAC.
+func (h *rbacHandler) ResolveInstancePreset(ctx context.Context, cluster, name, namespace string) (*corev1alpha1.InstancePreset, error) {
+	object := rbac.ClusterObjectName(cluster, name)
+	if err := h.enforce(ctx, rbac.ResourceInstancePresets, rbac.ActionRead, object); err != nil {
+		return nil, err
+	}
+	// Also check namespace access
+	nsObject := rbac.ClusterObjectName(cluster, namespace)
+	if err := h.enforce(ctx, rbac.ResourceNamespaces, rbac.ActionRead, nsObject); err != nil {
+		return nil, err
```
recharte, line 65: 🤔 I didn't consider this. What exactly are we trying to protect here?

recharte (approved):
Great job 👏

internal/server/handlers/rbac/instance_preset.go
chilagrow: Imagine if the user without read access to `prod` namespace try to access a preset with namespace defaults, this rbac check rejects it.

internal/server/handlers/k8s/instance_preset.go
chilagrow: Extracting would be useful when we need this default logic in other places 👍.

recharte: https://github.com/openeverest/openeverest/issues/2422 created for tracking this

PR #2408, opened 2026-06-18
fix: robust async cleanup of MonitoringConfig finalizers and secrets

### LINKED ISSUE #2317: Make monitoringconfig controller reconciliation more robust
### Task Description

see failing CI https://github.com/openeverest/openeverest/actions/runs/26623705436/job/78455300871?pr=2315

It's failing because if some secrets are already deleted, it returns error and requeued with exponential back off. Default timeout of chainsaw is 5s. The monitoring config controller should handle not found error, and requeue request appropriately. Rather than relying on reconciliation to requeue upon error. Also increase the timeout of test execution.

https://github.com/openeverest/openeverest/blob/release-2.0/internal/controller/monitoring/monitoringconfig_controller.go#L559

### Reasoning / Benefit

Make controller and tests more stable.

### Additional Context

The MonitoringConfig controller's deletion flow in OpenEverest v2 has issues with finalizer handling and secret cleanup that can cause resources not being deleted in single reconciliation. The current implementation doesn't properly handle async cleanup operations and doesn't signal when reconciliation needs to be requeued.

Update `cleanupSecrets` signature
```
func (r *MonitoringConfigReconciler) cleanupSecrets(ctx context.Context, mc *monitoringv1alpha1.MonitoringConfig) (bool, error)
```
* Returns boolean true when all secrets are deleted (`len(secrets.Items) == 0`)
* Add error handling for `client.IgnoreNotFound(err)` to continue deleting other secrets

Extract finalizer handling

```
func (r *MonitoringConfigReconciler) handleFinalizers(ctx context.Context, mc *monitoringv1alpha1.MonitoringConfig) (bool, error)
```
* Should handle deleting `cleanupSecretsFinalizer` and `vmagentFinalizer` if they exist
* It should call `cleanupSecret`
* Returns boolean indicating whether all finalizers have been fully removed

Update `reconcileVMAgent` signature with bool indicating reconciliation needs requeuing
```
func (r *MonitoringConfigReconciler) reconcileVMAgent(ctx context.Context) (bool, error)
```
* If a `MonitoringConfig` is being deleted, call `handleFinalizers` before `ensureVMAgentResources` in the loop and track requeuing needs
* If any finalizer is not yet handled return early after the loop with `true, nil` indicating requeuing is needed

Update `Reconcile` function to requeue requests with `ctrl.Result{RequeueAfter: 5 * time.Second }` when requeuing is requested with `true` return value  from `r.reconcileVMAgent(ctx)`

Update https://github.com/openeverest/openeverest/blob/release-2.0/test/integration/.monitoring.yaml with `exec` timeout of 30sec.

alokkumardalei-wq:
## Description
Fixes #2317.
Fixes an issue where the `MonitoringConfig` controller would get stuck in a requeue loop during deletion. The controller now properly ignores `NotFound` errors when cleaning up secrets concurrently, and strictly orchestrates the deletion of the `vmagentFinalizer` and `cleanupSecretsFinalizer`.

## Changes Made
- Updated `cleanupSecrets` to safely ignore `NotFound` errors and signal completion.
- Extracted `handleFinalizers` to cleanly manage the sequence of finalizer removal.
- Adjusted `reconcileVMAgent` to track when a requeue is needed and skip updating `VMAgent` specs during deletion.
- Added a 5-second backoff requeue in the main `Reconcile` loop to wait for async cleanup.

## Testing & Validation
- ✅ Validated that the Go codebase successfully compiles (`go build` & `go test`).
- ✅ Successfully built the Docker image `ghcr.io/openeverest/openeverest-controller-dev:0.0.0` locally.
- ✅ Executed the full K3D integration test suite (`make test-integration-monitoring`). It successfully spun up the test cluster and passed the entire `chainsaw` end-to-end testing suite (verifying that the deletion race conditions and hangs are completely resolved).


internal/controller/monitoring/monitoringconfig_controller.go
```diff
@@ -540,8 +551,35 @@ func (r *MonitoringConfigReconciler) monitoringSecretName(mc *monitoringv1alpha1
...
+			return false, fmt.Errorf("could not remove finalizers: %w", err)
+		}
+	}
+
+	return true, nil
+}
+
 // cleanupSecrets deletes all secrets in the monitoring namespace that belong to the given MonitoringConfig.
-func (r *MonitoringConfigReconciler) cleanupSecrets(ctx context.Context, mc *monitoringv1alpha1.MonitoringConfig) error {
+func (r *MonitoringConfigReconciler) cleanupSecrets(ctx context.Context, mc *monitoringv1alpha1.MonitoringConfig) (bool, error) {
```
chilagrow, line 583: Could you update function description about boolean return value please?

internal/controller/monitoring/monitoringconfig_controller.go
```diff
@@ -374,16 +395,6 @@ func (r *MonitoringConfigReconciler) ensureVMAgentResources(ctx context.Context,
 	}
 
 	if !mc.GetDeletionTimestamp().IsZero() {
```
chilagrow: Do you need this condition?

internal/controller/monitoring/monitoringconfig_controller.go
```diff
@@ -540,8 +551,35 @@ func (r *MonitoringConfigReconciler) monitoringSecretName(mc *monitoringv1alpha1
 	return mc.Spec.PMM.CredentialsSecretName + "-" + mc.GetNamespace()
 }
 
+// handleFinalizers manages the removal of finalizers and secret cleanup.
+func (r *MonitoringConfigReconciler) handleFinalizers(ctx context.Context, mc *monitoringv1alpha1.MonitoringConfig) (bool, error) {
+	if mc.Spec.Type != monitoringv1alpha1.PMMMonitoringType {
+		return true, nil
+	}
+
+	secretsCleaned, err := r.cleanupSecrets(ctx, mc)
```
chilagrow: Could you check if `cleanupSecretsFinalizer` exists and only remove secret if it doesn't exist? 🤗

internal/controller/monitoring/monitoringconfig_controller.go
```diff
@@ -540,8 +551,35 @@ func (r *MonitoringConfigReconciler) monitoringSecretName(mc *monitoringv1alpha1
 	return mc.Spec.PMM.CredentialsSecretName + "-" + mc.GetNamespace()
 }
 
+// handleFinalizers manages the removal of finalizers and secret cleanup.
```
chilagrow, line 551: Could you add description about boolean return value please?

chilagrow (commented):
Thank you for your contribution!

alokkumardalei-wq:
Thanks @chilagrow  for the thorough review and helpful feedback! 🤗 I've just pushed an update addressing all your comments:

1. Boolean Return Descriptions: Added clear comment descriptions explaining the boolean return values for both handleFinalizers and cleanupSecrets.
2. Simplified Deletion Check: Good catch on the ensureVMAgentResources condition! I removed the unneeded deletion timestamp check there and instead added a clean continue in reconcileVMAgent when handling deleting resources.
3. Finalizer Existence Check: Updated handleFinalizers to explicitly check if cleanupSecretsFinalizer exists on the MonitoringConfig before calling cleanupSecrets.

Let me know if anything else is needed! 🙌

Thank you!

chilagrow (approved):
Thanks for your contribution and sorry for the delayed review 🙏

I removed env specific file .DS_Store, I recommend keeping such file in your global gitignore file.

PR #2413, opened 2026-06-18
fix: validate oldVersion semver before slicing in version upgrade check

ashnaaseth2325-oss:
### SUMMARY

This PR adds validation for both the current and target database engine versions before performing semver-based upgrade checks. It prevents a potential runtime panic when an existing cluster contains an invalid or empty engine version and ensures the request fails gracefully with a validation error.

Primary changes are in `internal/server/handlers/validation/database_cluster.go`, specifically within `validateDBEngineVersionUpgrade()`.

### FIX

```go
//Before
if !semver.IsValid(newVersion) {
    return errInvalidVersion
}

//After
if !semver.IsValid(newVersion) || !semver.IsValid(oldVersion) {
    return errInvalidVersion
}
```


spron-in (approved):
lgtm

PR #2416, opened 2026-06-19
Adding generic plugin new extension point in the provider's form

spron-in:
This PR adds the new extension point for generic plugins in the provider's creation and edit form.


spron-in:
Fixes #2412

PR #2417, opened 2026-06-20
Remove circular dependency (#2334) and add strict eslint rule

### LINKED ISSUE #2334: UI: Break the ui-lib checkbox circular dependency
### Task Description

Break the `ui-lib` checkbox circular dependency by replacing the package-root barrel import with a direct internal import, and add a guard to prevent the pattern from coming back.

### Reasoning / Benefit

- Removes a real circular dependency in the UI library
- Makes dependency direction explicit and easier to reason about
- Reduces the risk of build fragility and hidden coupling between leaf components and the package API surface

### Additional Context

The current cycle goes through checkbox.tsx, checkbox/index.ts, and ui-lib/src/index.ts.
The checkbox component currently imports LabeledContent through the root barrel.

Acceptance criteria:
- `checkbox.tsx` no longer imports `LabeledContent` through the package root barrel.
- The circular dependency between `checkbox.tsx`, `checkbox/index.ts`, and `ui-lib/src/index.ts` is removed.
- A dependency guard is added to prevent internal `ui-lib` imports through the package root barrel.
- No runtime UI behavior changes.

StepanovPlaton:
## Summary

- Replace barrel imports (`../../..`) in `checkbox.tsx` and `checkbox.types.ts` with direct imports from `../../../labeled-content`, matching other form inputs.
- Add an ESLint `no-restricted-imports` rule to prevent internal `ui-lib` imports through the package root barrel.

Fixes #2334


ui/packages/ui-lib/.eslintrc.cjs
```diff
@@ -1,4 +1,25 @@
 module.exports = {
   root: true,
   extends: ['@percona/eslint-config-react', 'plugin:storybook/recommended'],
+  rules: {
+    // Prevent internal ui-lib components from importing through the package root barrel.
+    // Use direct relative paths (e.g. '../../../labeled-content') instead.
+    'no-restricted-imports': [
```
solovevayaroslavna: The rule catches only home and ../../../index, so only files at a depth of src/X/Y/Z/. If files at other nesting levels appear in the future (for example, src/X/file.ts with import from '..'), the rule won't catch them.

The alternative is to use patterns with glob like:
```
'no-restricted-imports': [
  'error',
  {
    patterns: [
      {
        group: ['@percona/ui-lib'],
        message: 'Do not import from the package itself inside src/.',
      },
    ],
  },
],
```
It's not a blocker, but it's better to fix it.

StepanovPlaton: Yes, thanks — that's a valid point. Using no-restricted-imports with group: ['@percona/ui-lib'] only blocks imports by package name; it does not catch relative barrel imports such as ../../...

I've applied a fix and suggest adding eslint-plugin-import to handle these more complex cases. The import/no-restricted-paths rule resolves import paths and forbids imports that resolve to src/index.ts, regardless of nesting depth.

A similar approach could be implemented with regex in no-restricted-imports, but that option is not available in the ESLint version we're using (ESLint 8).

solovevayaroslavna (approved):

PR #2424, opened 2026-06-22
shift-right - moving permissions to plugins

spron-in:
As we discussed with Chi - we are shifting permissions to plugins. Instead of OpenEverest handling kube permissions, we are allowing plugins to do that. This makes sense, as:
1. There is no reason to overcomplicate OpenEverest and extend its permissions
2. Plugins come with their own logic and predicting the permissions is not always possible, especially for the future extensibility


api/extensions/v1alpha1/installedextension_types.go
recharte: I believe this should also be removed. Managing namespaces here is fundamentally incompatible with Everest's core RBAC model. Whenever we add the ability to customize each user's plugin access this will get in the way and make for unsolvable edge cases. We had this exact issue a couple of years ago when we introduced RBAC for the first time because the BackupStorage and MonitoringConfig CRs had a similar namespace-management model to this.

chilagrow: I think it can be shifted to plugin itself rather than OpenEverest as well.

From my understanding, this is intended for usage where Plugin can generate CRD within Plugin CR and the namespaces are used for limiting which it can create CRD https://github.com/openeverest/specs/blob/main/specs/003-generic-plugins.md#host-reconciliation-model.

In that use case, it will have backend service + pods of plugin so I think having CRD bundled with plugin makes it natural 🤔

recharte (approved):

api/extensions/v1alpha1/installedextension_types.go
spron-in: Resolved.

PR #2443, opened 2026-06-23
auth: implement everestctl auth logout

VijetaPriya47:
Revokes the current session on the Everest server (POST /v1/auth/revoke) and removes the context, user, and orphaned server entries from the local config file.

Key details:
- Sends the stored refresh token in the request body and the access token as a Bearer header (required by the JWT middleware on /v1/auth/revoke)
- Automatically refreshes the access token if it has expired before attempting revocation (don't know if this is the correct approach, since, we are gonna logout, why should we make a new access token for this to be valid for 15 mins. One alternative is that we could revoke using the refresh token instead)
- (And what of the case when, network call fails or something, what should be the backup ?)
- ( one idea is that :- if the access token is expired and refresh fails, just wipe the local config without calling the server. The refresh token will eventually expire server-side anyway. is doing this a good practice?)
- Cleans up server entry only when no other context still references it, preserving multi-server configs
- Adds RemoveContext/Server/User and IsServerReferenced/IsUserReferenced helpers to config.Config following the same pattern as the existing Upsert/Get helpers

### DIFF
--- a/pkg/cli/auth/logout.go
+++ b/pkg/cli/auth/logout.go
@@ -1,0 +1,118 @@
+// everest
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/openeverest/openeverest/v2/client"
+	"github.com/openeverest/openeverest/v2/pkg/cli/config"
+)
+
+// Logout revokes the current session on the server and removes its context, user,
+// and (if unreferenced) server entry from the local config.
+func (lo *Login) Logout(ctx context.Context, cfgPath string) error {
+	// If the access token is expired, refresh it so the revoke endpoint accepts it.
+	cfg, err := config.Load(cfgPath)
+	if err != nil {
+		return err
+	}
+
+	currentCtx, ok := cfg.GetCurrentContext()
+	if !ok {
+		return fmt.Errorf("no active context %q found in config", cfg.CurrentContext)
+	}
+
+	usr, ok := cfg.GetUser(currentCtx.User)
+	if !ok {
+		return fmt.Errorf("user %q not found in config", currentCtx.User)
+	}
+
+	if time.Now().After(usr.ExpiresAt) {
+		if err := lo.Refresh(ctx, cfgPath); err != nil {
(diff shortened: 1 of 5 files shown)


recharte:
Hey @VijetaPriya47 regarding the questions you raised.

### Automatically refreshes the access token if it has expired before attempting revocation
We are not following the OAuth 2.0 strictly but we still want to be as close to it as possible. In this case, we should follow the spec described in [RFC 7009](https://datatracker.ietf.org/doc/html/rfc7009).

According to the standard, the refresh token is the session credential. Its presence in the request is sufficient proof of ownership — requiring a valid access token to revoke it creates the exact chicken-and-egg problem you've identified. The idiomatic answer is: we should make `/auth/revoke` unauthenticated, and treat the refresh token as the credential for that endpoint. With a 15-minute access token TTL, the blocklisting of the access JWT is a nice-to-have, not a security requirement. If the caller does present an access token, we can blocklist it opportunistically. If they don't (e.g. expired), we can skip that step and still return 204.

TLDR: We require some changes in the API server but you should not exchange the refresh token for an access token before logging out. The body of the revocation request should always include the refresh token. If you have a valid access token you include it as an authorization header.

### And what of the case when, network call fails or something, what should be the backup?
Always clear local credentials and proceed — never block logout on a network failure.

The primary purpose of logout is to protect the user on this device. If the network is down, the server is unreachable, or the token was already expired, the user still needs to be able to log out. Blocking them creates a worse security outcome — a frustrated user who closes the browser tab with their session still "active" in the UI.

The server-side revocation (blocklisting the access JWT, deleting the refresh token from the registry) is a defence-in-depth measure against token theft, not the primary logout mechanism. The access token's 15-minute TTL limits the blast radius of a failed revocation to a bounded window.

TLDR: Clear the credentials/context from config file and log a warning.

recharte:
Here's the PR that changes the API server's behavior according to the previous comment: https://github.com/openeverest/openeverest/pull/2448

VijetaPriya47:
@recharte I think, the final changes are done, now you can review it for merge.

pkg/cli/auth/logout.go
```diff
@@ -79,19 +65,28 @@ func (lo *Login) Logout(ctx context.Context, cfgPath string) error {
 		return fmt.Errorf("failed to create API client: %w", err)
 	}
 
+	// Include the access token only if it is still valid so the server can
+	// blocklist it. If expired, omit it — the refresh token alone is enough.
+	var reqEditors []client.RequestEditorFn
+	if time.Now().Before(usr.ExpiresAt) {
```
recharte, line 71: As discussed in #2372, the API server is incorrectly returning the expiry date of the access token instead of the refresh token. We'll fix that soon but you should already take that into account. In order to check the expiry date of the access token you should parse the JWT and get the `exp` claim.

recharte: This should be super straightforward using our existing [JWT library](https://pkg.go.dev/github.com/golang-jwt/jwt/v5#Claims.GetExpirationTime).

recharte: Please take into account that the JWT expiration date is in UTC.

recharte: Actually, it looks like I'm mixing stuff up again 🙈 
I just took another look at [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2) and this is correct, it should be indeed the access token's TTL. Please disregard my previous comment.

recharte (approved):

PR #2252, opened 2026-05-16
fix(cli): fix typo 'namesapce' in namespace existence check error message

abhisheksainimitawa:
## Summary

Fixes #2251

Fixes a transposed-character typo in the error message returned by `namespaceExists()` in `pkg/cli/namespaces/utils.go`. This message surfaces to users when `everestctl namespaces add` or `everestctl namespaces update` fails to check whether a namespace exists in Kubernetes.

## Change

```diff
- return false, false, fmt.Errorf("cannot check if namesapce exists: %w", err)
+ return false, false, fmt.Errorf("cannot check if namespace exists: %w", err)
```

Also adds the missing OpenEverest copyright line to the file header.

## Test plan

- [x] `go build ./pkg/cli/namespaces/...` — compiles cleanly
- [ ] Existing CI passes

PR #2444, opened 2026-06-23
fix(cli): correct 'namesapce' typo in namespace check messages

### LINKED ISSUE #2251: fix(cli): typo 'namesapce' in namespace existence check error message
### What is the issue?

There is a typo in `pkg/cli/namespaces/utils.go` inside the `namespaceExists()` function — `"namesapce"` instead of `"namespace"`:

```go
// pkg/cli/namespaces/utils.go:94
return false, false, fmt.Errorf("cannot check if namesapce exists: %w", err)
```

This error surfaces to the user when `everestctl namespaces add` or `everestctl namespaces update` cannot check whether a namespace exists in Kubernetes — making it harder to read and understand.

### Possible Solution

```go
return false, false, fmt.Errorf("cannot check if namespace exists: %w", err)
```

sahilforkshere:
Fixes #2251

Corrects the namesapce → namespace typo in the namespace-existence check error messages (pkg/cli/namespaces/utils.go, pkg/cli/install/install.go) and a related code comment in pkg/cli/upgrade/steps.go.


areebahmeddd:
Hi, i believe there's already a pr associated to this issue! 🙂

recharte:
Hi @sahilforkshere, thanks for this — and sorry you've been waiting on it since June with only a fellow contributor's heads-up for an answer.

#2252 was open five weeks before this one, so we're treating that as primary; the reasoning and the timeline are on #2251, so let's keep the discussion there. This isn't a judgement on your patch — it's about order, and about us leaving a green pull request unreviewed for three months.

Your two extra occurrences in `install.go` and `steps.go` aren't in #2252, so this stays open for those. The `utils.go` line is identical in both patches, so nothing should need redoing on your side.

PR #2446, opened 2026-06-24
fix: resources showing large numbers in database overview

### LINKED ISSUE #2423: UI: Resource details showing large numbers for Memory requests and limits in overview
### What is the issue?

For v2, we can allow users to set requests and limits separately for more flexibility. If set >1, Kubernetes returns the value in `milli-bytes`

### Steps to Reproduce

1. Install cloudnative-pg plugin
2. Click `Create database` -> `provider-cloudnative-pg`
3. Go to `Resources details` section and Memory requests and Memory limits to values >1 (eg. `0.4` or `0.3`)
4. Click `Create database` and go to database overview page and check `Resources details` section

### Relevant Logs / Output

Image

Resources values returned from the backend in API response from Kubernetes -

`
"resources": {
    "limits": {
        "cpu": "1",
        "memory": "644245094400m"
    },
    "requests": {
        "cpu": "1",
        "memory": "214748364800m"
    }
},
`

### OpenEverest Version

v2

### Possible Solution

`K8ResourceParses` on the frontend needs to handle value conversion for `mili-bytes`

I can submit a PR for the bug fix :)

AdityaPimpalkar:
Closes #2423


solovevayaroslavna:
Hey @AdityaPimpalkar — I owe you an apology. You opened this back in June with a clean, well-tested fix, and we left it without a response for so long that you ended up closing it yourself. That's on us, not you, and I'm sorry — it's not the experience a contributor should have here.

For what it's worth: this was a good PR. The approach was right and the tests were thorough, and the work isn't wasted — I'm making sure it's carried forward so it doesn't get lost. Thank you for the effort and the patience 🙇

PR #2448, opened 2026-06-25
Make `POST /v1/auth/revoke` unauthenticated

recharte:
The revoke endpoint previously required a valid access JWT, which created a chicken-and-egg problem: a client with an expired access token could not log out without first exchanging its refresh token for a new access token — generating a credential purely to immediately invalidate it.

The refresh token is itself proof of session ownership. Accepting it directly as the revocation credential (per [RFC 7009](https://datatracker.ietf.org/doc/html/rfc7009)) removes the dependency on the access token entirely. Clients in any state — valid access token, expired access token, or no access token at all — can now log out with a single request.

Access token blocklisting is retained as a best-effort operation: if a valid `Authorization` header is present the token is still blocklisted, but its absence no longer fails the request. Given the 15-minute access token TTL the security impact of skipping blocklisting is bounded.


chilagrow (commented):
Could you check API tests CI please? I think one of the failure is related to logout. 🤔

recharte:
> Could you check API tests CI please? I think one of the failure is related to logout. 🤔

I'm so glad we had these tests :sweat_smile: It's indeed a bug... The root cause is the JWT middleware. When `security: []` makes the skipper bypass the endpoint, the middleware never parses the token or stores it in the echo context — so `common.ExtractToken` returns `nil` even though an Authorization header is present. Blocklisting is silently skipped, and the access token remains valid.
I'll fix it.

Issue #2422, opened 2026-06-22
Extract default namespace resource handling for preset


chilagrow:
### Task Description

In https://github.com/openeverest/openeverest/pull/2407, support for preset was added. There is logic related to handling default resource for the given namespace for namespace-scoped resource such as MonitoringConfig.

Extract computing default logic out from internal/server/handlers/rbac. The suggestion is define it with `MonitoringConfig` resource definition. More details here: https://github.com/openeverest/openeverest/pull/2407#discussion_r3442427724

### Reasoning / Benefit

We may use default resource logic more than one place, extracting enable reusability.

alokkumardalei-wq:
Hi! @chilagrow just raised a pr for this , please have a look when you get time, happy to get any feedback on this .

Thank you!

PR #2449, opened 2026-06-25
Extract default MonitoringConfig computation logic to resource definition

alokkumardalei-wq:
## Description
Fixes: #2422
This PR addresses the extraction of the default resource computation logic introduced in #2407. Rather than keeping the filtering and resolution logic embedded directly within the server handlers, it has been moved to the `MonitoringConfigList` resource definition.

## Changes Proposed:

1. Added Default(annotationKey string) `*MonitoringConfig` method to `MonitoringConfigList` in `api/monitoring/v1alpha1/monitoringconfig_types.go`.
2. Refactored `findDefaultMonitoringConfig` in `internal/server/handlers/k8s/instance_preset.go` to use the new Default method and cleaned up unused conversion helpers and imports.
3. Added comprehensive unit tests in `monitoringconfig_types_test.go` to verify correct filtering and timestamp tie-breaking.

**Reasoning / Benefit**: Extracting this logic directly into the resource definition promotes reusability across the project, allowing any component working with `MonitoringConfigList` to resolve defaults cleanly without duplicating the annotation matching logic.


spron-in:
@alokkumardalei-wq have you tested it? The change in instance_preset.go will not do any good.

alokkumardalei-wq:
> @alokkumardalei-wq have you tested it? The change in instance_preset.go will not do any good.

Hi @spron-in, I tested the logic thoroughly via unit tests and it behaves correctly. However, if you mean that this change doesn't do any good because findDefaultSecret still requires the old duplicated logic, I completely agree. Would you prefer I extract this into a truly generic helper function (e.g., in `pkg/kubernetes`) that can handle finding the default for any `ctrlclient.Object` list, rather than tying it specifically to `MonitoringConfigList`?

recharte:
Thanks for picking this up, but I'm going to close it.

#2422 is assigned to @chilagrow — per CONTRIBUTING.md please don't open a PR against an issue that's already assigned to someone else.

On the substance: the point of the issue is reuse, and a method on `MonitoringConfigList` can only ever cover types we define. `findDefaultSecret` runs the same filter over `corev1.SecretList`, and #2454 has since added `findDefaultStorageClass` over `storagev1.StorageClassList` — both upstream types we can't hang a method on. So this covers one of the three call sites and the other two keep their copy of the loop, which is what @spron-in meant above.

Your follow-up question answers itself, IMO: a generic helper taking a list of objects and an annotation key is the shape that works. That's a different change from this one, on an issue that's already claimed.

PR #2453, opened 2026-06-25
Ui/tiles view create

### LINKED ISSUE #2476: [Core] Support provider discovery

spron-in:
Instead of a dropdown we have tiles now.
If hub plugin is not installed, we show the name from the provider output.
If hub is installed, we make it a bit more beautifyl with icon and tooltip.
Hub is not mandatory, but makes things better.
In the videos below I show the difference.

We are not using InstalledExtension yet to track if plugin or provider is installed. 


spron-in:
Screencast 1: without hub, just tiles:
https://github.com/user-attachments/assets/bae7b461-2845-40a7-9b39-fbf383183a61

Screencast 2: with hub, nicer view
https://github.com/user-attachments/assets/4bb80c84-a171-408f-a173-924da2a0f6ba

AdityaPimpalkar:
This looks nice!👌🏼

ui/apps/everest/src/pages/databases/DbClusterView.tsx
```diff
@@ -227,65 +227,63 @@ export const DbClusterView = () => {
...
-                )} */}
-                <CreateDbButton />
-              </Box>
-            )
-          }
-          hideExpandAllIcon
-        />
+        {!instancesLoading && !loadingNamespaces && tableData.length === 0 ? (
+          namespaces.length > 0 ? (
+            <EmptyStateDatabases />
```
solovevayaroslavna: RBAC verification is lost, this is the critical issue

ui/apps/everest/src/components/empty-state-databases/empty-state-databases.tsx
```diff
@@ -12,53 +12,76 @@
...
 const EmptyStateDatabases = ({
-  showCreationButton,
-  hasCreatePermission,
-}: {
-  showCreationButton: boolean;
-  hasCreatePermission: boolean;
-}) => {
-  return (
-    <>
+  hasCreatePermission = true,
```
solovevayaroslavna: Granting rights is the task of the RBAC layer (its configuration), not the UI. The UI never "hands out" rights. When RBAC is turned off, the admin has everything by default → the permission check returns true. 
Therefore, it is really not necessary to "hardcode/grant rights to the UI".

I would even say that we shouldn't set any defaults other than false, because this will lead to the fact that in the absence of real rights, we will give access where it cannot be. The key rule here is that if there is no right, then we don't show anything.

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
```diff
@@ -0,0 +1,157 @@
...
+            key={name}
+            elevation={0}
+            sx={{
+              flex: '1 1 200px',
+              minWidth: 200,
+              maxWidth: 220,
+              display: 'flex',
+              borderRadius: 2,
+              border: '1px solid',
+              borderColor: (theme) =>
```
solovevayaroslavna: Please de-duplicate the provider item UI (tiles vs. create-DB drawer)

The provider tile in provider-tiles.tsx and the provider row inside create-db-button.tsx render the same thing — avatar + label + (optional) description — with copy-pasted markup. The logic has already started to drift, so I'd like to move out the shared part before it goes further. 

Please keep the two layouts as separate components though — a card in a grid and a row in a drawer are genuinely different UI, so this is not "merge them into one".

Goal: one shared presentational atom + shared data helper; the two containers stay separate and just compose the atom.

1. New shared atom (follow the existing component conventions — separate .types.ts, .utils.ts, index.ts, inline sx, presentational only, no hooks inside
`components/provider-identity/
  provider-identity.tsx        // Avatar + label + optional description tooltip
  provider-identity.types.ts   // props type
  provider-identity.utils.ts   // resolveProviderDisplay(provider, getMeta) -> { name, meta, label }
  index.ts`

suggested props:
`type ProviderIdentityProps = {
  label: string;
  meta?: ResolvedExtensionMeta;   // avatar only renders when meta is present (current behaviour)
  showDescription?: boolean;      // tiles: true, drawer row: false
  ellipsis?: boolean;             // drawer row: true, tiles: false
  labelVariant?: TypographyProps['variant'];
};`

2. Keep two separate containers, each just composing the atom
- provider-tiles.tsx→ keeps the Card/ CardActionArea grid, renders <ProviderIdentity showDescription /> inside.
- Move the drawer's provider list out of create-db-button.tsx into its own small component 

so i expect: no duplicated markup, drift removed, the two layouts remain clearly separate components, and it matches the existing .types.ts / .utils.ts / presentational-wrapper style in the codebase.

ui/apps/everest/src/components/create-db-button/create-db-button.tsx
```diff
@@ -103,13 +104,12 @@ export const CreateDbButton = ({
...
       sx={buttonStyle}
       aria-controls={
-        open
-          ? `${createFromImport ? 'import' : 'add'}
-            -db-cluster-button-menu`
+        drawerOpen
+          ? `${createFromImport ? 'import' : 'add'}-db-cluster-button-menu`
           : undefined
       }
       aria-haspopup="true"
```
solovevayaroslavna: the trigger button still has aria-haspopup="true", but it now opens a Drawer, not a menu — should be aria-haspopup="dialog"

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
```diff
@@ -0,0 +1,157 @@
...
+                >
+                  {label}
+                </Typography>
+                {meta?.description && (
+                  <Tooltip title={meta.description} placement="top" arrow>
+                    <HelpOutlineIcon
+                      fontSize="small"
+                      sx={{ flexShrink: 0, color: 'text.secondary' }}
+                      // Prevent the tooltip icon from triggering navigation.
+                      onClick={(e) => {
```
solovevayaroslavna: the provider description is only reachable by hovering the HelpOutline icon, so it's not available to keyboard/screen-reader users. Please expose it via aria-label

ui/apps/everest/src/shared-types/extension-catalog.types.ts
```diff
@@ -0,0 +1,92 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Types describing the OpenEverest Extension Hub index (`hub.openeverest.io/v1`).
+// These mirror the JSON returned by the plugin-hub plugin at
+// `/v1/plugins/plugin-hub/api/summary`. They are hand-written because the hub
```
solovevayaroslavna, line 17: I would suggest several improvements of this situation:

- Check schemaVersion (cheapest). Read metadata.schemaVersion when loading the catalog; if it's outside the range this frontend was built for, log a warning and fall back to raw names. Doesn't validate fields, but catches major breaks loudly instead of silently.
or
- Generate the types from a shared schema. Have the hub publish an OpenAPI/JSON-Schema for /summary and generate these FE types from it, so both sides derive from one source and drift is caught at build time. (It's a long time, but it's the right way.)

what we should always do is not lie to the compiler.

Right now entry.type: ExtensionType = 'provider' | 'plugin' is required and unguarded. If the hub sends a new type , the comparison just won't match — better to make it explicit:
`if (entry?.type !== 'provider') continue;   // tolerate unknown types instead of crashing`

And type the incoming data as "untrusted" (unknown / Partial) where you actually want TypeScript to force you to guard. (If you keep api.get<ExtensionIndex>(), TS treats the shape as guaranteed — ?. 

I think this is critical. When we work with the code inside, we need to know the truth. Whether we can really rely on some data or not. And if you can guarantee that the contract on the other side does not change, then it's better to add a cross-referenced comment on the hub side so that in case of a breakdown you can jump to the hub repository and see the line and location and the latest changes there quickly.
And the developers of the hub would have been warned that in the event of a contract change, it is necessary to change the UI in this repository as well.

solovevayaroslavna: we can also do the next thing to simplify the checks across the app. we can do the check in one place instead of `?.` in different places.

`const normalizeEntry = (raw: Partial<ExtensionCatalogEntry>): ResolvedExtensionMeta | null => {
  if (!raw?.name || raw.type !== 'provider') return null;   // guard
  return {
    name: raw.name,
    displayName: raw.displayName || raw.name,
    icon: resolveIconSrc(raw.icon),
    description: raw.description,
    version: resolveVersion(raw as ExtensionCatalogEntry),
  };
};`

ui/apps/everest/src/hooks/api/extension-catalog/useExtensionCatalog.ts
```diff
@@ -0,0 +1,111 @@
...
+
+// The plugin-hub backend rewrites cross-origin icon URLs to a path relative to
+// its own mount (e.g. `api/icon/<sha256>`) to avoid CSP failures. Prepend the
+// plugin's mount prefix so the resulting URL works as an `<img src>`.
+const resolveIconSrc = (rawIcon?: string): string | undefined => {
+  if (!rawIcon) {
+    return undefined;
+  }
+  if (
+    rawIcon.startsWith('http://') ||
```
solovevayaroslavna: resolveIconSrc returns http:// and https:// icon URLs unchanged, and the value ends up in <Avatar src=...> (an ). But the app's CSP is img-src 'self' data:, so the browser will block any external image URL — the icon just won't load and you get a CSP error in the console.

The comment right above even says the hub backend already rewrites cross-origin icons to a relative api/icon/... path to avoid exactly this. 

So: why keep the http(s)://] pass-through? If the hub always sends relative paths, that branch is dead; and if it ever does send an absolute URL, CSP guarantees a broken image. Should we just drop the http(s) case and rely on the relative rewrite?

solovevayaroslavna (changes requested):

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
```diff
@@ -0,0 +1,157 @@
...
+  CardContent,
+  Tooltip,
+  Typography,
+} from '@mui/material';
+import HelpOutlineIcon from '@mui/icons-material/HelpOutlineOutlined';
+import { Link } from 'react-router-dom';
+import { useExtensionCatalog } from 'hooks/api/extension-catalog';
+import type { Provider } from 'shared-types/api.types';
+
+type ProviderTilesProps = {
```
solovevayaroslavna: We keep prop types in a sibling *.types.ts see ActionableAlert.types.ts, status-field.types.ts. Could you move these out for consistency? (Note extension-catalog.types.ts already follows this correctly)

ui/apps/everest/src/components/create-db-button/create-db-button.tsx
```diff
@@ -131,63 +131,144 @@ export const CreateDbButton = ({
...
+                alignItems: 'center',
+                justifyContent: 'space-between',
+                px: 3,
+                py: 2,
+                borderBottom: (theme) => `1px solid ${theme.palette.divider}`,
+              }}
+            >
+              <Box>
+                <Typography variant="h6">
+                  {createFromImport ? 'Import database' : 'Create database'}
```
solovevayaroslavna: We centralize copy in a messages.ts (you already did this in empty-state-databases/messages.ts). Could these strings move there too? (and all others across PR)

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
```diff
@@ -0,0 +1,157 @@
...
+        maxWidth: 900,
+      }}
+    >
+      {providers.map((provider) => {
+        const name = provider?.metadata?.name ?? '';
+        const meta = getProviderMeta(name);
+        const label = meta?.displayName || name;
+        return (
+          <Card
+            key={name}
```
solovevayaroslavna: We tend to pull big style objects into const xxxSx = {…} above the component (you already do it for buttonStyle/skeletonStyle. Same treatment here would make the JSX readable.

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
```diff
@@ -0,0 +1,157 @@
...
+            elevation={0}
+            sx={{
+              flex: '1 1 200px',
+              minWidth: 200,
+              maxWidth: 220,
+              display: 'flex',
+              borderRadius: 2,
+              border: '1px solid',
+              borderColor: (theme) =>
+                theme.palette.dividers?.divider ?? theme.palette.divider,
```
solovevayaroslavna: These custom palette keys exist in the theme — if they're typed as required we can drop the ?. ?? fallback; if optional, worth a small helper. Right now it's repeated noise.

ui/apps/everest/src/components/create-db-button/create-db-button.tsx
```diff
@@ -131,63 +131,144 @@ export const CreateDbButton = ({
         <Skeleton variant="rounded" sx={skeletonStyle} />
       )}
       {providers.length > 1 && (
-        <Menu
+        <Drawer
```
solovevayaroslavna, line 132: Whole file (~275 lines): trigger button + skeleton + tooltip + Drawer header + banner + provider list. Please consider extracting the provider list / drawer body into its own small component (like EmptyState delegates via slots). Keeps responsibilities separated and shrinks this file.

I wrote a similar comment somewhere in more detail.

ui/apps/everest/.e2e/pr/db-cluster/db-wizard/db-wizard-utils.ts
```diff
@@ -207,46 +207,100 @@ export const openCreateScheduleDialogFromDBWizard = async (page: Page) => {
...
+  const toolbarBtn = page.getByTestId('add-db-cluster-button');
+  const tileFallback = dbType
+    ? page.getByTestId(`provider-tile-${dbType}`)
+    : page.locator('[data-testid^="provider-tile-"]').first();
+
+  await expect(toolbarBtn.or(tileFallback).first()).toBeVisible({
+    timeout: TIMEOUTS.ThirtySeconds,
+  });
+
+  if (await toolbarBtn.isVisible().catch(() => false)) {
```
solovevayaroslavna: if (await toolbarBtn.isVisible().catch(() => false)) … else tile pattern is repeated in clickAddDbClusterBtn, checkAmountOfDbEngines, selectDbEngine, openDbCreationForm. A single helper would remove the repetition.

ui/apps/everest/src/pages/databases/DbClusterView.tsx
spron-in: fixed

ui/apps/everest/src/components/empty-state-databases/empty-state-databases.tsx
spron-in: fixed

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
spron-in: fixed

ui/apps/everest/src/components/create-db-button/create-db-button.tsx
spron-in: fixed

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
spron-in: fixed

ui/apps/everest/src/shared-types/extension-catalog.types.ts
spron-in: fixed

ui/apps/everest/src/hooks/api/extension-catalog/useExtensionCatalog.ts
spron-in: fixed

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
spron-in: fixed

ui/apps/everest/src/components/create-db-button/create-db-button.tsx
spron-in: done

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
spron-in: fixed

ui/apps/everest/src/components/empty-state-databases/provider-tiles.tsx
spron-in: done

ui/apps/everest/src/components/create-db-button/create-db-button.tsx
spron-in: yep, fixed

ui/apps/everest/.e2e/pr/db-cluster/db-wizard/db-wizard-utils.ts
spron-in: done

solovevayaroslavna:
image

We need to figure out how the plugin should pick up the theme and recommend that all developers write a UI with support for two themes

Plugin Hub ignores the host's light/dark mode because self-contained plugin bundles run on their own React/MUI instance and never see the host's ColorModeContext. As an incremental fix, we can extend PluginApi with the current colorMode plus an onColorModeChange subscription sourced from the host's ColorModeContext, so a plugin can build its own ThemeProvider in the matching mode and re-theme on toggle. This passes a value rather than React context, so it works across the plugin's separate React instance without requiring the full singleton-sharing setup.

spron-in:
> Plugin Hub ignores the host's light/dark mode because self-contained plugin bundles run on their own React/MUI instance and never see the host's ColorModeContext. As an incremental fix, we can extend PluginApi with the current colorMode plus an onColorModeChange subscription sourced from the host's ColorModeContext, so a plugin can build its own ThemeProvider in the matching mode and re-theme on toggle. This passes a value rather than React context, so it works across the plugin's separate React instance without requiring the full singleton-sharing setup.

it is known problem, but not for this PR to fix

solovevayaroslavna:
Screenshot from 2026-07-20 12-21-47
Screenshot from 2026-07-20 12-21-56
Screenshot from 2026-07-20 12-23-38
Screenshot from 2026-07-20 12-23-55


[Screencast from 2026-07-20 11-29-01.webm](https://github.com/user-attachments/assets/142ef215-b364-45b1-8b52-073c2c0683c5)

recharte (approved):

PR #2457, opened 2026-06-26
Preset selection for instance creation flow

### LINKED ISSUE #1844: UI: Creating a database from a preset

As a user, I want to be able to create a database using presets.

After selecting the database type, the system should present a pre-filled form based on the chosen template. If the preset lacks certain required data, openEverest should automatically apply default values. The user's only task should be naming the database and clicking 'Create'.
Our goal is to automate as much as possible to reduce user effort. 
All auto-filled parameters remain fully editable for total control.

_No response_

_No response_

### Additional Context

_No response_

chilagrow:
This PR adds Preset selection for instance creation flow, implementing phase 1 of https://github.com/openeverest/specs/blob/main/specs/006-preset.md.

* Select Preset in Instance creation flow
* If Preset is selected, only namespace and instance name are editable. This is equivalent of newly introduced `deploy` permission on RBAC.
* If Preset is unselected, keeping the value as it is.


chilagrow:
@solovevayaroslavna this demonstrates using Preset to populate Instance in creation flow. Feel free to scrap it all.
Current implementation reflects minimal permission to create an instance without allowing modification. This will be controlled by RBAC set on each user in the later phase.

PR #2461, opened 2026-06-26
Allow separate CPU/Memory requests and limits for database engines and proxies

### LINKED ISSUE #2360: Allow to set requests and limits separately

Right now when the user is creating a database in v1, there are fields to set CPU and RAM for database pods. Setting these fields is translated into `requests` and `limits`. These are set equal to ensure guaranteed QoS as per the best practices for stateful workloads.
There are still cases where users might need to set requests a bit lower and users are fine with these Pods to be evicted or killed. Example, non-production workloads. 

It would be cool if API/UI allow users to set requests and limits separately. Maybe default to single value, but enable some advanced settings where these values can be set separately. 

UI/API, later on CLI as well.

_No response_

### Additional Context

It is interesting that even v2 unblocks this, it might be disregarded by Provider builders. It is okay, but would be great to have it in there as well to ensure proper flexibility.

recharte:
Follow-up to https://github.com/openeverest/openeverest-operator/pull/961 and https://github.com/openeverest/helm-charts/pull/62

ref: https://github.com/openeverest/openeverest/issues/2360


chilagrow (dismissed):
Once everest-operator is merged and go mod is updated, code should compile 👍 

Is there UI changes needed to support setting `requests` in OpenEverest UI?

Issue #2059, opened 2026-04-27
UI: Plugin Developer Playground: Interactive UI Schema Editor with Live Preview (V2)


solovevayaroslavna:
### Is your feature request related to a problem?

OpenEverest uses a plugin architecture where database providers define their UI through declarative YAML schemas (UISchema in Provider CRD). Currently, plugin developers have no integrated tool for developing and testing these schemas — the only way to see the rendered result is to deploy a full Provider CRD into a running Kubernetes cluster. This makes the development cycle slow and error-prone.

An existing proof-of-concept (`/ui-generator-builder`) demonstrates the viability of the "live editor + form preview" approach, but has security issues: the Monaco Editor it uses loads scripts from an external CDN (jsdelivr.net) and requires `unsafe-eval` in Content Security Policy for its parser to function. OpenEverest enforces a strict CSP policy on the server (`script-src 'self'`, nonce-based `style-src`, `object-src 'none'`, `frame-ancestors 'none'`, etc.), so the current POC **does not work in production** — the browser blocks loading of external scripts and execution of `new Function()`.

### Describe the solution you'd like

Improve the existing POC or rewrite it from scratch, creating a fully functional Plugin Developer Playground — a tool that allows plugin developers to write, test, and debug UI schemas directly in the browser without deploying to a cluster.

The Playground is a page at the `/plugin-developer` route, running inside a deployed OpenEverest instance.

## Functional Requirements:

1. **YAML editor with syntax highlighting**
   - Split-pane layout: schema editor on the left, live form preview on the right
   - YAML syntax highlighting
   - Resizable panel divider

2. **Real-time schema validation**
   - Display YAML parsing errors (syntax errors)
   - Structural validation: verify that the entered schema conforms to the `TopologyUISchemas` type
   - Inline error display in the editor (markers, underlines)

3. **Live form rendering**
   - Use the existing `UIGenerator` component to render the form from the schema
   - Stepper navigation through schema sections
   - Topology switching (if the schema defines multiple topologies)

4. **Dynamic field support**
   - Fields of type `openEverestApiProvider` (e.g., monitoring configs, storage classes) load data from the running OpenEverest API by default — this is the primary mode of operation
   - If a plugin developer defines a new provider type in the schema that doesn't yet exist in the API, the playground should provide a way to mock data for that field so they can see how it will render

5. **Output panel**
   - Display the JSON payload generated after form data post-processing (form values → postprocess → API shape)
[...]

PR #2465, opened 2026-06-28
ui: add plugin developer playground

Denyme24:
part of  #2059 

adds a playground under /ui-generator-builder where plugin developers can paste a topology UI schema as YAML and see the generated form render live, with validation feedback.

### done so far
- YAML editor - CodeMirror 6 with syntax highlighting, split pane (schema left, preview right) with a draggable divider. 
- Real-time validation - YAML syntax errors, then a structural check against TopologyUISchemas, shown inline on the offending key.
- Live form rendering - renders through the existing UIGenerator, with stepper navigation and topology switching.
- Dynamic fields - openEverestApiProvider fields load from the running API (namespace-gated); unknown provider types can be mocked with static options: in the schema.

Testing

unit tests cover the protocol, srcdoc, Zod schema, and validation (vitest). also ran it against a live Everest backend-  schemas render, validation errors show inline, and registered providers resolve.

below is a screen-recording for reference.

https://github.com/user-attachments/assets/aba96611-c83c-4da2-9fc6-9ef2fd27f529

### yet to do
- Output panel
- Schema persistence and sharing


Denyme24:
cc: @solovevayaroslavna @spron-in

ui/apps/everest/src/pages/ui-generator-builder/utils/topology-ui-schema.zod.ts
```diff
@@ -0,0 +1,60 @@
...
+import { z } from 'zod';
+
+// Runtime schema for the parts of `TopologyUISchemas` the renderer depends on:
+// the topology -> sections -> components/groups skeleton and a known `uiType`
+// on every leaf. Everything else passes through, so authors can use any field
+// params without the validator getting in the way.
+
+// The allowed leaf field uiTypes. Exported so error messages can name them
+// without re-hardcoding the list (keeps the validator and its messages in sync).
+export const FIELD_UI_TYPES = ['number', 'select', 'text', 'hidden'] as const;
```
solovevayaroslavna: Can we improt this from existing enum of ui-generator?  FieldTypes from ui/apps/everest/src/components/ui-generator/ui-generator.types.ts?

ui/apps/everest/package.json
```diff
@@ -4,10 +4,11 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "dev": "vite --port 3000",
+    "dev": "pnpm build:editor && vite --port 3000",
```
solovevayaroslavna: the editor runs the full Vite build (parsing, tree-shaking, CM6 bundle minification) at each:
pnpm dev (every restart of the dev server)
pnpm test:unit (every test run)
pnpm build (CI build)
CM6 bundle changes only when codemirror dependencies are updated or cm-editor-entry.ts is edited. In the remaining 99% of cases, this is ~2-5 seconds of wasted time, so i would suggest to fix this with check if file exist or not.

ui/apps/everest/src/pages/ui-generator-builder/utils/topology-ui-schema.zod.ts
Denyme24: done, used FieldType from ui-generator instead of hardcoding the list.

ui/apps/everest/package.json
Denyme24: @solovevayaroslavna, yeah i get the point on the wasted time. 
but here's the thing on why it's wired this way: editor-bundle.generated.js is gitignored and CodeMirrorEditor.tsx imports it statically, so a clean checkout has to generate it or Vite can't resolve the import, that's why build:editor is chained into dev/build/test:unit instead of being a manual step. a plain "skip if the file exists" check would fix the speed, but it'd also keep a stale bundle around whenever cm-editor-entry.ts changes, with no signal that it's out of date.

solovevayaroslavna:
Hey! I’ve taken a look — great job. Both refactors are done exactly as I wanted, and thanks for fixing the editor build issues at the same time.

Before removing it from the draft, I really want the tests we talked about — for the page and for the three hooks. That’s what the split was for. 
Also: please rename the folder to the new name, and take a look at why the checks aren’t green on CI.

And let’s break it down into parts: First, we’ll merge the core with the tests, and the mock panel and result output will be handled in separate PRs — this one is already quite large.

Denyme24:
> Hey! I’ve taken a look — great job. Both refactors are done exactly as I wanted, and thanks for fixing the editor build issues at the same time.
> 
> Before removing it from the draft, I really want the tests we talked about — for the page and for the three hooks. That’s what the split was for. Also: please rename the folder to the new name, and take a look at why the checks aren’t green on CI.
> 
> And let’s break it down into parts: First, we’ll merge the core with the tests, and the mock panel and result output will be handled in separate PRs — this one is already quite large.

thanks! all three done:
 - Tests ->  page + the three hooks (use-schema-validation, use-preview-namespace, use-split-pane), all green.
- Rename ->  ui-generator-builder → plugin-developer
 - CI ->  green now
Also fixed a stale-selection bug in useTopology i hit while testing (topology didn't update on live schema edits). it's in the shared ui-generator, but the only caller is the playground — the DB creation form passes topology in via use-form-engine, so no regression there (confirmed against its tests).

yep i will keep the mock panel and result output as separate PRs.

i think it's ready to be merged now.

solovevayaroslavna (approved):

PR #2516, opened 2026-07-06
Add fuzztest for rbac

spron-in:
Start adding fuzztest where it makes sense + to comply with OpenSSF best practices. This ome adds BoBuiltInFuzzer and fuzz test for RBAC.

Need to cherry-pick to v2 later on.


chilagrow (approved):
Thank you!

PR #2518, opened 2026-07-06
everestctl: add --preset flag to instance create

### LINKED ISSUE #2505: everestctl instance create: add --preset flag to deploy an instance from an InstancePreset

Today `everestctl instance create` requires users to assemble the full Instance spec by hand — pick the provider, then configure every component through repeated `--set` flags or a `-f values.yaml` file (see `commands/instance/create.go` and `pkg/cli/instance/create.go`). Providers already ship bundled `InstancePreset` CRs — cluster-scoped templates that mirror the Instance spec with known-valid, sensible defaults — but the CLI has no way to consume them. Users must copy preset values manually, which is error-prone and defeats the one-click deployment story described in `specs/006-preset.md`.

Add a `--preset` flag to `everestctl instance create`. When set, the CLI fetches the **resolved** preset (namespace-scoped defaults such as `MonitoringConfig`/`Secret` pre-filled), uses its spec as the base for the new Instance, and lets `--set` / `-f` overrides layer on top. The created Instance is self-contained (values copied from the preset, no live reference) and carries the `openeverest.io/instance-preset: <name>` annotation for tracking/RBAC, matching the Instance CR shape in the spec.

The server API and generated client already exist, so this is mostly CLI-side plumbing:

- `GET /clusters/{cluster}/instance-presets?provider={provider}` → `ListInstancePresets`
- `GET /clusters/{cluster}/instance-presets/{name}` → `GetInstancePreset`
- `GET /clusters/{cluster}/instance-presets/{name}/resolve?namespace={ns}` → `ResolveInstancePreset`

CLI flow mirroring the spec's UI flow:

```bash
# Deploy from a preset (one-click equivalent)
everestctl instance create --name my-mongo --namespace prod \
  --provider percona-server-mongodb --preset default-replicaset

# Preset as a base, override specific fields on top
everestctl instance create --name my-mongo --namespace prod \
  --provider percona-server-mongodb --preset default-replicaset \
```

Behavior:

1. Resolve the preset for the target namespace (`/resolve?namespace=`) so namespace defaults are filled in.
2. Preset spec becomes the base; `-f` file then `--set` merge on top (preset < file < set), reusing the existing merge logic.
3. The preset's `provider` must match `--provider`; mismatch is rejected.
4. Instance is created with the `openeverest.io/instance-preset` annotation.
5. (Optional / follow-up) discovery of available presets, e.g. `everestctl instance presets list --provider ...` using the existing list endpoint.

- Manually copying preset YAML into a `-f values.yaml` file — works today, but is manual, skips namespace-default resolution, and doesn't set the tracking annotation.
- Client-side templating of defaults — rejected in the spec in favor of server-side resolution.

### Additional Context

- Spec: `specs/006-preset.md` (InstancePreset CR, `resolve` endpoint, annotation, `deploy` RBAC permission).
- The generated client currently exposes the raw `ListInstancePresets` / `GetInstancePreset` / `ResolveInstancePreset` methods but **not** the `*WithResponse` typed variants the CLI uses elsewhere — a small client regen or manual decode may be needed.
- Out of scope: `deploy` RBAC enforcement and editable-preset UI (Phase 2 in the spec).

VijetaPriya47:
Adds support for bootstrapping a new instance from a named InstancePreset. When --preset is given, the CLI resolves the preset via the API (passing --namespace so the server can fill in namespace-scoped defaults such as secretRef), uses the resolved spec as the base, then layers -f file and --set overrides on top in that order.

- New --preset flag on `instance create`; --provider stays required and must match the preset's provider field
- Version and topology are inherited from the preset when not supplied explicitly via --version / --topology
- Instances created from a preset are annotated with openeverest.io/instance-preset so the origin is traceable
- topology.config from the preset is preserved when --topology overrides only the type
- Unit tests covering the happy path, version/topology inheritance, --set override priority, provider mismatch, 404 not-found, namespace query param forwarding, and explicit-flag precedence 
- closes #2505

### DIFF
--- a/commands/instance/create.go
+++ b/commands/instance/create.go
@@ -42,6 +42,13 @@ to override any Instance spec field using dot-notation paths rooted at the spec
 		Example: `  # Minimal: server defaults for all components
   everestctl instance create --name my-mongo --namespace everest --provider percona-server-mongodb
 
+  # Bootstrap from a preset (provider inferred; version/topology/components from preset)
+  everestctl instance create --name my-mongo --namespace everest --preset mongodb-production
+
+  # Preset as base, override a field on top
+  everestctl instance create --name my-mongo --namespace everest --preset mongodb-production \
+    --set components.engine.replicas=5
+
   # Override component fields via --set
   everestctl instance create --name my-mongo --namespace everest --provider percona-server-mongodb \
     --set components.engine.replicas=3 \
@@ -73,17 +80,17 @@ to override any Instance spec field using dot-notation paths rooted at the spec
 func init() {
 	createCmd.Flags().StringVar(&createOpts.Name, cli.FlagInstanceName, "", "Instance name (required)")
 	createCmd.Flags().StringVar(&createOpts.Namespace, cli.FlagInstanceNamespace, "", "Namespace to create the instance in (required)")
-	createCmd.Flags().StringVar(&createOpts.Provider, cli.FlagInstanceProvider, "", "Provider name, e.g. percona-server-mongodb, percona-xtradb-cluster (required)")
+	createCmd.Flags().StringVar(&createOpts.Provider, cli.FlagInstanceProvider, "", "Provider name, e.g. percona-server-mongodb (required unless --preset is given)")
+	createCmd.Flags().StringVar(&createOpts.Preset, cli.FlagInstancePreset, "", "InstancePreset name to bootstrap from; --provider is inferred from the preset when omitted")
 	createCmd.Flags().StringVar(&createOpts.Cluster, cli.FlagInstanceCluster, "main", "Cluster name")
-	createCmd.Flags().StringVar(&createOpts.Version, cli.FlagInstanceVersion, "", "Version bundle name (default: provider's default bundle)")
-	createCmd.Flags().StringVar(&createOpts.Topology, cli.FlagInstanceTopology, "", "Topology name (default: provider's first topology)")
+	createCmd.Flags().StringVar(&createOpts.Version, cli.FlagInstanceVersion, "", "Provider version bundle to use (default: taken from preset if --preset is given, otherwise provider's default)")
+	createCmd.Flags().StringVar(&createOpts.Topology, cli.FlagInstanceTopology, "", "Topology name; cannot be used with --preset (default: provider's first topology)")
 	createCmd.Flags().StringVar(&createOpts.Context, cli.FlagInstanceContext, "", "Context to use (default: current context)")
 	createCmd.Flags().StringVarP(&createOpts.ValuesFile, cli.FlagInstanceFile, "f", "", "Path to a YAML file with component overrides (--set takes precedence)")
 	createCmd.Flags().StringArrayVar(&createOpts.Set, cli.FlagInstanceSet, nil, "Set a spec field: --set components.engine.replicas=3 or --set backup.enabled=true (repeatable)")
 
 	_ = createCmd.MarkFlagRequired(cli.FlagInstanceName)
 	_ = createCmd.MarkFlagRequired(cli.FlagInstanceNamespace)
-	_ = createCmd.MarkFlagRequired(cli.FlagInstanceProvider)
 }
 
 func createPreRun(cmd *cobra.Command, _ []string) { //nolint:revive
--- a/pkg/cli/instance/create.go
+++ b/pkg/cli/instance/create.go
@@ -43,12 +43,13 @@ type CreateOptions struct {
 	Name       string
 	Namespace  string
 	Provider   string
+	Preset     string // InstancePreset name; provider must match --provider
 	Cluster    string
 	Version    string
 	Topology   string
 	Context    string   // overrides the active context when set
-	ValuesFile string   // path to a YAML file with spec-level overrides (optional)
-	Set        []string // each entry: "specField.subfield=value" e.g. "components.engine.replicas=3" — takes precedence over ValuesFile
+	ValuesFile string   // path to a YAML values file with spec-level overrides (optional)
+	Set        []string // dot-notation overrides e.g. "components.engine.replicas=3"; takes precedence over ValuesFile
 }
 
 type InstanceCreator struct {
@@ -65,9 +66,32 @@ func NewInstanceCreator(cfg Config, l *zap.SugaredLogger) *InstanceCreator {
 }
 
 func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath string) error {
+	if opts.Provider == "" && opts.Preset == "" {
+		return fmt.Errorf("--provider is required when --preset is not given")
+	}
+	if opts.Preset != "" && opts.Topology != "" {
+		return fmt.Errorf("--topology cannot be combined with --preset")
+	}
+
 	c, err := authcli.NewAPIClient(authcli.Config{Pretty: ic.config.Pretty}, ic.l.Desugar().Sugar(), cfgPath, opts.Context)
 	if err != nil {
 		return err
+	}
+
+	var (
+		presetSpecBase map[string]any
+		annotations    map[string]string
+	)
+	if opts.Preset != "" {
+		pr, err := ic.resolvePreset(ctx, c, opts.Preset, opts.Cluster, opts.Namespace)
+		if err != nil {
+			return err
+		}
+		if err := applyPresetDefaults(opts.Preset, &opts, pr); err != nil {
+			return err
+		}
+		presetSpecBase = pr.specMap
+		annotations = map[string]string{"openeverest.io/instance-preset": opts.Preset}
 	}
 
 	provResp, err := c.GetProviderWithResponse(ctx, opts.Cluster, opts.Provider)
@@ -103,12 +127,19 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath 
 		}
 	}
 
+	// Merge order: preset < -f file < --set flags.
 	specOverrides, err := buildSpecOverrides(opts.ValuesFile, opts.Set)
 	if err != nil {
 		return err
 	}
+	if presetSpecBase != nil {
+		if specOverrides != nil {
+			deepMerge(presetSpecBase, specOverrides)
+		}
+		specOverrides = presetSpecBase
+	}
 
-	payload := buildPayload(opts.Name, opts.Provider, resolvedVersion, resolvedTopology, specOverrides)
+	payload := buildPayload(opts.Name, opts.Provider, resolvedVersion, resolvedTopology, specOverrides, annotations)
 
 	body, err := json.Marshal(payload)
 	if err != nil {
@@ -142,6 +173,82 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath 
 	}
 
 	return nil
+}
+
+// applyPresetDefaults copies provider/version/topology from the preset into opts,
+// validating that an explicit --provider matches the preset's provider.
+func applyPresetDefaults(preset string, opts *CreateOptions, pr *presetResult) error {
+	if pr.provider != "" && opts.Provider != "" && pr.provider != opts.Provider {
+		return fmt.Errorf("--provider %q does not match preset %q provider %q", opts.Provider, preset, pr.provider)
+	}
+	if opts.Provider == "" {
+		if pr.provider == "" {
+			return fmt.Errorf("preset %q does not specify a provider; use --provider to set one", preset)
+		}
+		opts.Provider = pr.provider
+	}
+	if opts.Version == "" {
+		opts.Version = pr.version
+	}
+	if pr.topology != "" {
+		opts.Topology = pr.topology
+	}
+	return nil
+}
+
+type presetResult struct {
+	specMap  map[string]any
+	provider string
+	version  string
+	topology string
+}
+
+func (ic *InstanceCreator) resolvePreset(ctx context.Context, c *client.ClientWithResponses, preset, cluster, namespace string) (*presetResult, error) {
+	resp, err := c.ResolveInstancePresetWithResponse(
+		ctx,
+		cluster,
+		preset,
+		&client.ResolveInstancePresetParams{Namespace: namespace},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve preset %q: %w", preset, err)
+	}
+	if resp.StatusCode() == http.StatusNotFound {
+		return nil, fmt.Errorf("preset %q not found in cluster %q", preset, cluster)
+	}
+	if resp.StatusCode() != http.StatusOK || resp.JSON200 == nil {
+		return nil, fmt.Errorf("unexpected response resolving preset %q: %s", preset, resp.Status())
+	}
+
+	p := resp.JSON200
+	pr := &presetResult{}
+	if p.Spec.Provider != nil {
+		pr.provider = *p.Spec.Provider
+	}
+	if p.Spec.Version != nil {
+		pr.version = *p.Spec.Version
+	}
+	if p.Spec.Topology != nil && p.Spec.Topology.Type != nil {
+		pr.topology = *p.Spec.Topology.Type
+	}
+	specMap, err := presetSpecToMap(p)
+	if err != nil {
+		return nil, err
+	}
+	pr.specMap = specMap
+	return pr, nil
+}
+
+func presetSpecToMap(p *client.InstancePreset) (map[string]any, error) {
+	b, err := json.Marshal(p.Spec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal preset spec: %w", err)
+	}
+	var m map[string]any
+	if err := json.Unmarshal(b, &m); err != nil {
+		return nil, fmt.Errorf("failed to parse preset spec: %w", err)
+	}
+	return m, nil
 }
 
 func defaultVersion(prov *client.Provider) string {
@@ -264,7 +371,7 @@ func providerName(prov *client.Provider) string {
 	if prov.Metadata == nil {
 		return "<unknown>"
 	}
-	if name, ok := (*prov.Metadata)["name"]; ok {
+	if name, ok := (*prov.Metadata)[metadataKeyName]; ok {
 		if s, ok := name.(string); ok {
 			return s
 		}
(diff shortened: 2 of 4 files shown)


pkg/cli/instance/create.go
```diff
@@ -91,6 +92,19 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath
...
+	var (
+		presetSpecBase map[string]any
+		annotations    map[string]string
+	)
+	if opts.Preset != "" {
+		presetSpecBase, err = ic.resolvePreset(ctx, c, token, &opts)
+		if err != nil {
+			return err
+		}
+		annotations = map[string]string{"openeverest.io/instance-preset": opts.Preset}
```
recharte, line 94: In order to make RBAC work properly, this annotation should only be set if there are no edits to the preset. If the user overwrites some field using `-f` or `--set` then this annotation should not be applied.
Reference: https://github.com/openeverest/specs/blob/main/specs/006-preset.md#prevent-users-from-modifying-defaults-on-one-click-deployment

recharte (changes requested):

commands/instance/create.go
```diff
@@ -74,9 +83,10 @@ func init() {
 	createCmd.Flags().StringVar(&createOpts.Name, cli.FlagInstanceName, "", "Instance name (required)")
 	createCmd.Flags().StringVar(&createOpts.Namespace, cli.FlagInstanceNamespace, "", "Namespace to create the instance in (required)")
 	createCmd.Flags().StringVar(&createOpts.Provider, cli.FlagInstanceProvider, "", "Provider name, e.g. percona-server-mongodb, percona-xtradb-cluster (required)")
+	createCmd.Flags().StringVar(&createOpts.Preset, cli.FlagInstancePreset, "", "InstancePreset name to bootstrap from; provider must match the preset's provider")
```
chilagrow: Should provider flag be required if `--preset` is required? InstancePreset CR must already define provider so I feel provider flag is redundant this case. 🤔

pkg/cli/instance/create.go
```diff
@@ -91,6 +92,23 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath
...
+		presetSpecBase map[string]any
+		annotations    map[string]string
+	)
+	if opts.Preset != "" {
+		presetSpecBase, err = ic.resolvePreset(ctx, c, token, &opts)
+		if err != nil {
+			return err
+		}
+		// Stamp the annotation only when the spec is unmodified; the server uses it
+		// to enforce the deploy RBAC tier and rejects any deviation from preset defaults.
```
chilagrow: I'm not sure if this comment is accurate. Just to clarify, does `everestctl` user always have `deploy` RBAC permission and non other?

In API server, a user who can create an instance may have `deploy` or `create`.

pkg/cli/instance/create.go
```diff
@@ -166,6 +197,54 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath
...
+	// Provider in the preset must match the --provider flag.
+	if p.Spec.Provider != nil && *p.Spec.Provider != opts.Provider {
+		return nil, fmt.Errorf("--provider %q does not match preset %q provider %q", opts.Provider, opts.Preset, *p.Spec.Provider)
+	}
+
+	// Preset fills in version/topology only when the caller didn't supply them.
+	if opts.Version == "" && p.Spec.Version != nil {
+		opts.Version = *p.Spec.Version
+	}
+	if opts.Topology == "" && p.Spec.Topology != nil && p.Spec.Topology.Type != nil {
```
chilagrow: This should be handled with care. `InstancePreset` without topology indicates it should use default topology, just like how it's defined in `Instance` https://github.com/openeverest/openeverest/blob/release-2.0/api/core/v1alpha1/instance_types.go#L244.

Overriding empty topology may create invalid instance. Imagine, you use flag to set topology to `sharded` while Preset has empty topology (default replicaset topology), this creates invalid Instance spec. In fact, I can't find a use case where we want to use topology flag in combination with preset flag. Should we reject topology flag if preset flag is set? 🤔

pkg/cli/instance/create.go
```diff
@@ -124,12 +142,25 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath
...
+		}
+		if userOverrides != nil {
+			deepMerge(presetSpecBase, userOverrides)
+		}
+		specOverrides = presetSpecBase
+	} else {
+		specOverrides, err = buildSpecOverrides(opts.ValuesFile, opts.Set)
+		if err != nil {
+			return err
+		}
```
chilagrow: Can it be simplified?
```suggestion
	specOverrides, err := buildSpecOverrides(opts.ValuesFile, opts.Set)
	if err != nil {
		return err
	}
	if presetSpecBase != nil {
		if specOverrides != nil {
			deepMerge(presetSpecBase, specOverrides)
		}
		specOverrides = presetSpecBase
	}
```

pkg/cli/instance/create.go
chilagrow: If a user have `create` RBAC permission on instance, they are free to modify the instance as they like even if they have this annotation.

However, if a user have `deploy` RBAC permission on instance, they must have this annotation, and any modification on instance will return RBAC error.

commands/instance/create.go
VijetaPriya47: yup that is true, and provider is not required, but as an extra safety check, I kept it required in every command when creating databases.

pkg/cli/instance/create.go
VijetaPriya47: yup okay, i've made the changes.

pkg/cli/instance/create.go
VijetaPriya47: no, it has create user permission also, but here it means, to use annotations only when the spec is unmodified......create users does not have any effect of it's presence and absence. and it's a mandatory command for deploy users

pkg/cli/instance/create.go
VijetaPriya47: made the changes. the annotation is only stamped when no -f/--set overrides are given (per the spec). This means deploy-permission users who pass overrides won't have the annotation, so the server will reject their request. create-permission users are unaffected either way.

VijetaPriya47:
@chilagrow  resolved all the comments. could you review again?

pkg/cli/instance/create.go
chilagrow: > the annotation is only stamped when no -f/--set overrides are given (per the spec)

I don't quite understand where in the spec it specifies? 🤔 The spec mentions `deploy` RBAC permission user must set annotation. I think `create` RBAC permission user should also keep the annotation just for the future tracking purpose.

commands/instance/create.go
```diff
@@ -73,17 +80,17 @@ to override any Instance spec field using dot-notation paths rooted at the spec
...
 	createCmd.Flags().StringVar(&createOpts.Namespace, cli.FlagInstanceNamespace, "", "Namespace to create the instance in (required)")
-	createCmd.Flags().StringVar(&createOpts.Provider, cli.FlagInstanceProvider, "", "Provider name, e.g. percona-server-mongodb, percona-xtradb-cluster (required)")
+	createCmd.Flags().StringVar(&createOpts.Provider, cli.FlagInstanceProvider, "", "Provider name, e.g. percona-server-mongodb (required unless --preset is given)")
+	createCmd.Flags().StringVar(&createOpts.Preset, cli.FlagInstancePreset, "", "InstancePreset name to bootstrap from; --provider is inferred from the preset when omitted")
 	createCmd.Flags().StringVar(&createOpts.Cluster, cli.FlagInstanceCluster, "main", "Cluster name")
-	createCmd.Flags().StringVar(&createOpts.Version, cli.FlagInstanceVersion, "", "Version bundle name (default: provider's default bundle)")
-	createCmd.Flags().StringVar(&createOpts.Topology, cli.FlagInstanceTopology, "", "Topology name (default: provider's first topology)")
+	createCmd.Flags().StringVar(&createOpts.Version, cli.FlagInstanceVersion, "", "Version bundle name (default: preset's version, then provider's default bundle)")
```
chilagrow: Preset also defines provider's version. Not preset's version 🤗

Preset is just like an Instance, but it's a template for Instance.

pkg/cli/instance/create.go
```diff
@@ -166,6 +197,60 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath
 	return nil
 }
 
+func (ic *InstanceCreator) resolvePreset(ctx context.Context, c *client.ClientWithResponses, token client.RequestEditorFn, opts *CreateOptions) (map[string]any, error) {
```
chilagrow: It's a bit more global design question, when we have a typed Go struct such as `InstanceSpec`, why do we use generic `map[string]any` for `presetSpecBase`? Maybe there is a reason for this usage that I'm unaware 🤗. It makes code harder to read.
https://github.com/openeverest/openeverest/blob/release-2.0/api/core/v1alpha1/instance_types.go#L31

pkg/cli/instance/create.go
```diff
@@ -426,11 +511,21 @@ func buildPayload(name, provider, version, topology string, specOverrides map[st
 		specOverrides["version"] = version
 	}
 	if topology != "" {
-		specOverrides["topology"] = map[string]any{"type": topology}
+		// Merge type into existing topology object to preserve topology.config from preset.
+		if existing, ok := specOverrides["topology"].(map[string]any); ok {
+			existing["type"] = topology
```
chilagrow, line 518: Is `existing` used anywhere? I see it being set but I don't see it's usage. 🤔

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,366 @@ func TestRun_UnknownContext(t *testing.T) {
...
+	}`)
+}
+
+func TestRun_WithPreset_HappyPath(t *testing.T) {
+	t.Parallel()
+
+	var capturedBody []byte
+	presetHandler := func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write(presetJSON())
```
chilagrow: Could you check error please using `assert.NoError`? It's a good practice to check errors 🤗 Please apply it to other places too.

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,366 @@ func TestRun_UnknownContext(t *testing.T) {
...
+	ic := NewInstanceCreator(Config{}, zap.NewNop().Sugar())
+	// No --version or --topology; both should come from the preset.
+	err := ic.Run(context.Background(), CreateOptions{
+		Name:      "my-db",
+		Namespace: "everest",
+		Provider:  "psmdb",
+		Cluster:   "main",
+		Preset:    "my-preset",
+	}, cfgPath)
+	require.NoError(t, err)
```
chilagrow: There are a lot of duplicated code, the difference appears to be `CreateOptions` and expected response. Could you use table test for readability and easier maintenace? See a simple table test example https://github.com/openeverest/openeverest/blob/release-2.0/internal/server/handlers/rbac/backup_test.go#L50 🤗

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,366 @@ func TestRun_UnknownContext(t *testing.T) {
...
+
+	var payload map[string]any
+	require.NoError(t, json.Unmarshal(capturedBody, &payload))
```
chilagrow: There is function in testify which allows you to compare JSON strings  https://pkg.go.dev/github.com/stretchr/testify@v1.11.1/require#JSONEq. Could you compare the expected and actual response using them? Is there a reason why we don't compare entire JSON? 🤔

For test variable naming, it's common to use `expected` and `actual` or `want` and `got` within tests for easier readability 🤗

chilagrow (commented):
Thanks for applying changes 🤗

As for requesting review, don't forget to re-request it. It will notify reviewers

image

pkg/cli/instance/create.go
chilagrow: What do you think @recharte? I think we should keep the annotation for future tracking.

VijetaPriya47: here, `create` can also set annotations, but only when the spec is unmodified......if the preset is modified using overrides, it will omit annotations.
Basically here annotations represent the preset having not modified. This is what is done in new changes. @chilagrow

pkg/cli/instance/create_test.go
VijetaPriya47: yeah sure done :)

pkg/cli/instance/create.go
chilagrow: Hmm but I don't understand why you would omit annotations. Could you give me a bit more explanations please? I think I'm missing the context here 🤔

chilagrow: Hey sorry about the confusion. Let's keep the annotation if it originated from a preset regardless of `-f`/`--set` override is used. There are a couple of reasons. If we want to keep the annotation only when the instance is exactly the same as preset, we need added mechanism in controller. For example if the instance created by preset is edited later on, we need to remove the annotation. If the preset changes, we also need to remove the annotation. To avoid these complexities, let annotation have the meaning that _an instance was initially scaffolded using preset_ 🤗.

recharte: Sorry for originating this mess, let's go with what Chi just said.

pkg/cli/instance/create.go
VijetaPriya47: `existing` holds a reference to the topology map. Setting existing["type"], updates it in place since Go maps are passed by reference, so there's no need to write it back to specOverrides.

pkg/cli/instance/create.go
VijetaPriya47: the generated client uses anonymous struct fields that don't map directly to InstanceSpec, so we can't use it without an explicit conversion. 
Also, deepMerge must operate on map[string]any because --set/-f overrides are parsed into generic maps, and to merge preset + user overrides into one structure, everything needs to be the same type. Using InstanceSpec as an intermediate would still require a marshal/unmarshal round-trip to map[string]any before the merge. So it might becme more complex. Open to suggestions if there's a cleaner approach though!

chilagrow (changes requested):
Could you check failed CI for linter please 🙏

pkg/cli/instance/create.go
chilagrow: Thanks for the explanation, I think what's confusing is that this function is called `resolvePreset` but it appears to do way more.

For example it mutates *CreateOptions as well as returning a map. I think function should document such cases in doc string, but mixing mutating parameter and having return value is weird. Can the mutation of *CreatOptions done outside this function?

pkg/cli/instance/create.go
chilagrow: Yes that's right. It just reads weird but I recall this is the idiomatic way for using map.

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,225 @@ func TestRun_UnknownContext(t *testing.T) {
...
+	testCases := []struct {
+		desc        string
+		opts        CreateOptions
+		wantPayload string
+	}{
+		{
+			desc: "happy path: annotation stamped, version and topology from preset",
+			opts: CreateOptions{
+				Name: "my-db", Namespace: "everest", Provider: "psmdb",
+				Cluster: "main", Preset: "my-preset",
```
chilagrow:
```suggestion
				Name: "my-db", 
				Namespace: "everest",
				Provider: "psmdb",
				Cluster: "main",
				Preset: "my-preset",
```

No need to be conservative of lines 😉

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,225 @@ func TestRun_UnknownContext(t *testing.T) {
...
+			},
+			wantPayload: `{
+				"metadata": {"name":"my-db","annotations":{"openeverest.io/instance-preset":"my-preset"}},
+				"spec": {"provider":"psmdb","version":"8.0","topology":{"type":"replicaset"},"components":{"engine":{"replicas":3}}}
+			}`,
```
chilagrow:
```suggestion
			wantPayload: `{
   "metadata":{
      "name":"my-db",
      "annotations":{
         "openeverest.io/instance-preset":"my-preset"
      }
   },
   "spec":{
      "provider":"psmdb",
      "version":"8.0",
      "topology":{
         "type":"replicaset"
      },
      "components":{
         "engine":{
            "replicas":3
         }
      }
   }
}`,
```
We can format the JSON standard way.

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,315 @@ func TestRun_UnknownContext(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+
+			var got []byte
+			srv := newRunServerWithPreset(t, okProviderHandler,
+				func(w http.ResponseWriter, r *http.Request) {
+					got, _ = io.ReadAll(r.Body)
```
chilagrow:
```suggestion
					got, err = io.ReadAll(r.Body)
					assert.NoError(t, err)
```
Let's check errors 🤗 You'd need to declare error prior though

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,315 @@ func TestRun_UnknownContext(t *testing.T) {
...
+
+			cfgPath := filepath.Join(t.TempDir(), "config.yaml")
+			require.NoError(t, newTestConfig(srv.URL).Save(cfgPath))
+
+			ic := NewInstanceCreator(Config{}, zap.NewNop().Sugar())
+			err := ic.Run(context.Background(), tc.opts, cfgPath)
+			require.Error(t, err)
+			for _, s := range tc.wantErrContains {
+				assert.Contains(t, err.Error(), s)
+			}
```
chilagrow: Let's use https://pkg.go.dev/github.com/stretchr/testify/require#ErrorContains.

This way you can match expected error string rather than using an array `wantErrContains`

pkg/cli/instance/create_test.go
```diff
@@ -688,3 +689,315 @@ func TestRun_UnknownContext(t *testing.T) {
...
+	ic := NewInstanceCreator(Config{}, zap.NewNop().Sugar())
+	require.NoError(t, ic.Run(context.Background(), CreateOptions{
+		Name:      "my-db",
+		Namespace: "prod",
+		Provider:  "psmdb",
+		Cluster:   "main",
+		Preset:    "my-preset",
+	}, cfgPath))
+
+	assert.Contains(t, capturedURL, "namespace=prod")
```
chilagrow: Could you assert full URL for readability please? 🙏 

something like 
```
assert.Contains(t, capturedURL, "/v1/clusters/main/instance-presets/resolve?namespace=prod")
```

pkg/cli/instance/create.go
chilagrow: I still don't think `deepMerge` using map is the correct way to operate on merging two Instance specs, because it's derived from strongly typed `Instance`. But let's not handle this yet, not in this PR definitely. 🙏

chilagrow (commented):
Great work, there's few minor handling in the tests that can be improved 💪

chilagrow (approved):
Thank you! Great work 🤗

Issue #2506, opened 2026-07-01
everestctl instance status: add -w/--watch flag for continuous polling


recharte:
### Is your feature request related to a problem?

`everestctl instance status` fetches an instance's status once and exits (see `commands/instance/status.go` and `pkg/cli/instance/status.go`). While provisioning or troubleshooting, users must re-run the command repeatedly — or wrap it in a shell `watch` loop — to observe phase/component/condition changes over time. That's cumbersome, especially while waiting for an instance to reach the `Running` phase.

### Describe the solution you'd like

Add a `-w/--watch` flag to `everestctl instance status`. When set, the command does not exit after the first fetch; instead it polls the Everest API on an interval and re-renders the status (analogous to `kubectl get ... -w`). It runs until interrupted (Ctrl-C / context cancellation).

### How would users interact with this feature?

```bash
# Continuously watch status until Ctrl-C
everestctl instance status --name my-mongo --namespace prod --watch

# Custom poll interval (optional)
everestctl instance status --name my-mongo --namespace prod -w --interval 5s
```

Behavior:

- Default poll interval (e.g. 2s), optionally configurable via `--interval`.
- Re-render the formatted table each tick, reusing `printInstanceStatus`.
- Decide `--json` interaction: stream JSON snapshots per poll, or restrict `--watch` to pretty output.
- Stop cleanly on Ctrl-C; handle transient API errors gracefully (log and keep polling rather than exiting).
- (Optional / follow-up) auto-exit once a terminal phase (e.g. `Running`) is reached — could be a separate `--wait` flag.

### Alternative solutions considered

- Shell `watch everestctl instance status ...` — works, but re-establishes a session/token on every run, causes flicker, and isn't discoverable.
- A dedicated `wait` command that blocks until a condition — different semantics; could complement `--watch` later.

### Additional Context

- Current command: `commands/instance/status.go`, logic in `pkg/cli/instance/status.go` (`InstanceStatusRunner.Run`).
- The runner already refreshes the auth token near expiry; a long-running watch must keep doing so on each poll.
- Reuse `printInstanceStatus` in `pkg/cli/instance/status.go` for rendering each frame.

recharte:
@VijetaPriya47 please let me know your thoughts.

VijetaPriya47:
okay I thought about it.
[...]

PR #2519, opened 2026-07-06
everestctl instance status: add --watch/-w flag

VijetaPriya47:
Poll the Everest API continuously on a configurable interval (default 2s) until Ctrl-C, token expiry, or instance deletion. The HTTP client is created once in Run and reused across all ticks — no per-tick allocation.

Behaviour:
- Pretty mode: --- separator between each render, output scrolls
- JSON mode: one JSON object per line per tick (NDJSON), pipe-friendly
- Transient errors (5xx, network): logged as warnings, polling continues
- Token near expiry (within 30s): exits cleanly with a message
- Instance deleted (404): exits cleanly with a message
- Ctrl-C / context cancel: exits with no error

#2506 does not close this, need refresh iteration for mid-watch rotation


pkg/cli/instance/status.go
```diff
@@ -84,23 +86,113 @@ func (is *InstanceStatusRunner) Run(ctx context.Context, opts StatusOptions, cfg
...
+			// Proactive rotation — refresh before the token expires.
+			if time.Now().After(sess.User.ExpiresAt.Add(-30 * time.Second)) {
+				lo := authcli.NewLogin(authcli.Config{Pretty: is.config.Pretty}, is.l.Desugar().Sugar())
+				newSess, err := lo.Refresh(ctx, cfgPath)
+				if err != nil {
+					return fmt.Errorf("access token expired and refresh failed: %w", err)
+				}
+				sess = newSess
+				token = cli.BearerToken(sess.User.AccessToken)
+			}
```
recharte: The token refresh logic is currently duplicated in several places — proactive refresh in `Run()`, again inside the watch loop, and a third copy for the 401 refresh-and-retry path (and `create.go` repeats the same preamble). Every future command will need to copy this too.

I think we should move this into the HTTP client itself, as client-side auth middleware. This is the standard way to solve it in Go:

- `golang.org/x/oauth2` does exactly this: `TokenSource` + a `Transport` that injects/refreshes tokens transparently
- kubernetes client-go has `transport.NewBearerAuthWithRefreshRoundTripper`
- AWS/GCP SDKs use credential providers with caching + auto-renewal

Our generated oapi-codegen client already has the hook for it: `client.WithHTTPClient(doer)`. A `RoundTripper` is better than a `RequestEditorFn` here because editors run before the request and can't observe a 401 response — the round tripper can also handle the reactive refresh-and-retry.

Rough shape (new file in `pkg/cli/auth`, e.g. `transport.go`):

```go
// TokenSource yields valid access tokens, refreshing (and persisting via
// Login.Refresh) when within 30s of expiry. Mutex-guarded.
type TokenSource struct { /* login, cfgPath, mu, sess */ }

func (ts *TokenSource) Token(ctx context.Context) (string, error)        // proactive refresh
func (ts *TokenSource) ForceRefresh(ctx context.Context) (string, error) // after a 401

// Transport is client-side auth middleware: injects the Bearer token on
// every request, and on 401 force-refreshes and retries once.
type Transport struct {
    source *TokenSource
    base   http.RoundTripper
}

func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
    token, err := t.source.Token(req.Context())
    // clone req, set Authorization, t.base.RoundTrip(...)
    // on 401: ForceRefresh + retry once (replay body via req.GetBody)
}

// NewAPIClient wires it all together so commands need zero token plumbing.
func NewAPIClient(cfg Config, l *zap.SugaredLogger, cfgPath, contextName string) (*client.ClientWithResponses, error) {
    sess, err := cli.LoadSession(cfgPath, contextName)
    // ...
    return client.NewClientWithResponses(
        client.WithHTTPClient(&http.Client{Transport: &Transport{source: ts, base: http.DefaultTransport}}),
    )
}
```

Commands then reduce to:

```go
c, err := authcli.NewAPIClient(authcli.Config{Pretty: is.config.Pretty}, is.l.Desugar().Sugar(), cfgPath, opts.Context)
// ...
resp, err := c.GetInstanceWithResponse(ctx, opts.Cluster, opts.Namespace, opts.Name)
```

The watch loop loses the entire 401/refresh branch. A couple of details worth handling:

1. **Distinguishing terminal auth failures from transient errors in the watch loop** — export a sentinel (`var ErrTokenRefresh = errors.New("access token expired and refresh failed")`) wrapped by the transport, so the loop can `errors.Is(err, authcli.ErrTokenRefresh)` → stop watching, vs. log-and-retry for transient network errors. (`errors.Is` unwraps through the `*url.Error` that `http.Client` adds.)
2. **Request body replay on the 401 retry** — only retry when `req.GetBody != nil` (it's set automatically for `bytes.Reader` bodies, which is what the generated client uses); otherwise surface the 401.
3. **Don't mutate the original request** in `RoundTrip` — clone it before setting the header (required by the `RoundTripper` contract).

Side note this refactor surfaces: `Login.Refresh(ctx, cfgPath)` always refreshes the current context, but commands accept `--context` overrides — so a refresh triggered while targeting a non-current context would update the wrong user's tokens. Probably worth threading `contextName` through `Refresh` as part of this.

I'd rather we land this before a third command copies the pattern.

recharte (changes requested):

pkg/cli/instance/status.go
VijetaPriya47: yup sure. It's good you pointed that out. This definitely needs a helper function. will complete this by tomorrow.

recharte:
Hi @VijetaPriya47 please fix the linting errors of https://github.com/openeverest/openeverest/actions/runs/29046175034/job/86215285565?pr=2519

pkg/cli/instance/status.go
```diff
@@ -54,53 +56,88 @@ func NewInstanceStatusRunner(cfg Config, l *zap.SugaredLogger) *InstanceStatusRu
...
+			switch resp.StatusCode() {
+			case http.StatusOK:
+				if resp.JSON200 == nil {
+					is.l.Warnf("empty response body — retrying in %s", interval)
+					continue
+				}
+				if is.config.Pretty {
+					fmt.Fprintln(os.Stdout, "---")
+				}
+				is.render(resp.JSON200, opts.Namespace)
```
recharte: We should be smarter about this. We should not print the status if nothing changed in the Instance CR.
Right now, there's no functional difference between running `everestctl instance status -w` and running `watch everestctl instance status`. We should behave just like kubectl that only output something new if there was an actual change.

There are two options for comparison keys:
1. `metadata.resourceVersion` — cheap and never misses a change, but it's noisy: it bumps on any write to the object (annotations, labels, spec edits, managedFields churn), not just on the status fields you render.
1. Diff what you actually display — this is what I'd recommend. Since `render()` is deterministic, the simplest correct implementation is: render to a buffer, compare against the last rendered string, print only on change. It works identically for pretty and JSON modes and matches the user-visible definition of "an actual change"

WDYT?

recharte (changes requested):

pkg/cli/instance/status.go
```diff
@@ -54,53 +56,88 @@ func NewInstanceStatusRunner(cfg Config, l *zap.SugaredLogger) *InstanceStatusRu
...
+			case http.StatusNotFound:
+				return fmt.Errorf("instance %q has been deleted", opts.Name)
+			default:
```
recharte, line 139: Now that we have the retry logic within the client we should add a case here that will return an error if it still got a 401. Something like:
```go
case http.StatusUnauthorized:
    return fmt.Errorf("server rejected credentials — run 'everestctl auth login' again")
```

pkg/cli/auth/transport.go
```diff
@@ -0,0 +1,150 @@
...
+		cfgPath:     cfgPath,
+		contextName: contextName,
+	}
+
+	return client.NewClientWithResponses(
+		cli.NormalizeServerURL(sess.Server.URL),
+		client.WithHTTPClient(&http.Client{
+			Transport: &authTransport{source: ts, base: http.DefaultTransport},
+		}),
+	)
```
recharte: We should wrap the error in some added context otherwise we are propagating something without a clear indication of where it happened.

```suggestion
    c, err := client.NewClientWithResponses(
        cli.NormalizeServerURL(sess.Server.URL),
        client.WithHTTPClient(&http.Client{
            Transport: &authTransport{source: ts, base: http.DefaultTransport},
        }),
    )
    if err != nil {
        return nil, fmt.Errorf("failed to create API client: %w", err)
    }
    return c, nil
```

pkg/cli/auth/transport.go
```diff
@@ -0,0 +1,150 @@
...
+	defer ts.mu.Unlock()
+
+	newSess, err := ts.login.Refresh(ctx, ts.cfgPath, ts.contextName)
+	if err != nil {
+		return "", fmt.Errorf("%w: %w", ErrTokenRefresh, err)
+	}
+	ts.sess = newSess
+
+	return ts.sess.User.AccessToken, nil
+}
```
recharte, line 75: Up to you but maybe we could extract the shared logic of refresh-and-store into it's own function to avoid duplication. This is just a nitpick, so like I said, it's up to you.
```go
func (ts *tokenSource) token(ctx context.Context) (string, error) {
    ts.mu.Lock()
    defer ts.mu.Unlock()

    if time.Now().After(ts.sess.User.ExpiresAt.Add(-30 * time.Second)) {
        if err := ts.refreshLocked(ctx); err != nil {
            return "", err
        }
    }
    return ts.sess.User.AccessToken, nil
}

func (ts *tokenSource) forceRefresh(ctx context.Context) (string, error) {
    ts.mu.Lock()
    defer ts.mu.Unlock()

    if err := ts.refreshLocked(ctx); err != nil {
        return "", err
    }
    return ts.sess.User.AccessToken, nil
}

func (ts *tokenSource) refreshLocked(ctx context.Context) error {
    newSess, err := ts.login.Refresh(ctx, ts.cfgPath, ts.contextName)
    if err != nil {
        return fmt.Errorf("%w: %w", ErrTokenRefresh, err)
    }
    ts.sess = newSess
    return nil
}
```

pkg/cli/auth/transport_test.go
```diff
@@ -0,0 +1,195 @@
...
+	ts := loadTokenSource(t, cfgPath, NewLogin(Config{}, zap.NewNop().Sugar()))
+	tr := &authTransport{source: ts, base: http.DefaultTransport}
+
+	req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet, srv.URL, nil)
+	_, err := tr.RoundTrip(req)
+	require.NoError(t, err)
+	assert.Equal(t, "Bearer refreshed-access", gotAuth)
+}
+
+func TestTransport_On401_RefreshAndRetry(t *testing.T) {
```
recharte, line 123: This retry test uses a GET, so the body-replay branch (`req.GetBody` handling) is never actually exercised — both `GetBody` and `Body` are nil.

The caller that depends on it is instance create (POST). Bodies are one-shot readers, so if the replay logic regresses, a 401 during create would retry with an empty body — and that only happens when token expiry races a POST, so nothing else would catch it.

Suggestion: add a POST variant of this test where the server 401s the first call and asserts the second call arrives with the original payload, e.g.:

```go
func TestTransport_On401_RetryReplaysBody(t *testing.T) {
    t.Parallel()

    var bodies []string
    srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        if r.URL.Path == "/v1/session" { // refresh endpoint
            _ = json.NewEncoder(w).Encode(refreshResp)
            return
        }
        b, _ := io.ReadAll(r.Body)
        bodies = append(bodies, string(b))
        if len(bodies) == 1 {
            w.WriteHeader(http.StatusUnauthorized)
            return
        }
        w.WriteHeader(http.StatusOK)
    }))
    defer srv.Close()

    // ...same tokenSource/transport setup as the other tests...

    payload := `{"metadata":{"name":"my-mongo"}}`
    req, _ := http.NewRequestWithContext(t.Context(), http.MethodPost, srv.URL+"/instances", bytes.NewReader([]byte(payload)))
    resp, err := tr.RoundTrip(req)
    require.NoError(t, err)
    assert.Equal(t, http.StatusOK, resp.StatusCode)
    require.Len(t, bodies, 2)
    assert.Equal(t, payload, bodies[1], "retried request must carry the original body")
}
```

pkg/cli/instance/status.go
VijetaPriya47: I think the second option is much simpler for the users to understand.....but it might be stuck on one output also, if let's say the if there is an error in creation and and the status is not changing. there would be one output and then silence. but this is a tradeoff that's implemented in the kubectl also

recharte (approved):
Well done 🚀

PR #2520, opened 2026-07-07
Implement `InstancePreset` API server

### LINKED ISSUE #2540: API: add InstancePreset CRUD operations

chilagrow:
Spec updated https://github.com/openeverest/specs/pull/10.

This PR implements Create, Update and Delete on `InstancePreset`, and drafting preset on `Instance`. Drafting preset simply returns namespace scoped references removed from the spec. Then that spec is used to create `InstancePreset`.

`POST /clusters/{cluster}/instance-presets`
`PUT /clusters/{cluster}/instance-presets/{name}`
`DELETE /clusters/{cluster}/instance-presets/{name}`
`GET /clusters/{cluster}/namespaces/{namespace}/instances/{instance}/draft-preset`

It also rewrites the resolving preset implementation by extracting it to a separate package for reusability. In short, `internal/preset` package was introduce and it walks through `Spec` and sets or unsets namespace scoped fields as needed. It's similar to `WalkDir` in `filepath` go standard library.


api/openapi/http-api.yaml
```diff
@@ -475,6 +475,82 @@ paths:
...
+            application/json:
+              schema:
+                $ref: 'crds.gen.yaml#/components/schemas/InstancePreset'
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  '/clusters/{cluster}/instance-presets/from-instance':
```
recharte, line 511: I'd love to open up a discussion about this API design, mostly because I've been going back and forth on this myself and I'm genuinely not sure what the best answer is.

The main thing I'm curious about: did you consider mirroring the `resolve` pattern we already have? We introduced `GET /instance-presets/{name}/resolve?namespace=...` which takes a preset and returns a derived, non-persisted object (preset + namespace defaults → concrete spec). What we're building here is basically the exact inverse (instance − namespace specifics → preset), so my instinct was something like:

1. `GET .../instances/{instance}/draft-preset` — pure function, returns the sanitized preset without persisting anything (strips secret refs, storage class, etc. — the same fields `resolve` fills in, just in reverse)
2. `POST /clusters/{cluster}/instance-presets` — plain generic create, body is a full `InstancePreset`

A few things nudged me that way, though none of them feel like slam dunks:

- **The `from-instance` segment sits in the `{name}` position** of `GET /instance-presets/{name}`, so it effectively reserves a preset name and makes the routing/spec a bit ambiguous.
- **We are also introducing a generic `POST /instance-presets`** meaning that we'd have two creation paths with potentially diverging validation.
- **RBAC gets a bit simpler**: the GET is purely an `instances` read, the POST purely an `instance-presets` create, instead of one endpoint enforcing both.
- **The UI flow probably wants the two-step anyway** — the preset is cluster-scoped so the user has to pick a name (can't safely inherit the instance name), and they may want to review what got stripped before saving.

The flip side, and where your approach clearly wins: it's atomic, it's one call, and it's simpler for headless/scripted use. So I can absolutely see the case for it.

What are your thoughts?

chilagrow: You highlight the exact point I was not quite happy myself 😅.

I think not introducing an independent payload required for `POST /clusters/{cluster}/instance-presets/from-instance` is big enough reason for this case using 2 requests. Let's do it 👍

PR #2526, opened 2026-07-07
fix: ignore stale RBAC permission check results

ashnaaseth2325-oss:
## SUMMARY

This PR prevents stale permission checks from overwriting newer RBAC state when multiple permission requests are in flight. It updates `useRBACPermissions` to ignore outdated async results, ensuring the UI always reflects the most recent permission evaluation.

Primarily affected module:

* `ui/apps/everest/src/hooks/rbac/rbac.ts`

## FIX

```ts
// Before
const canRead = await can(...);
setPermissions({ read: canRead, ... });

// After
const requestId = ++latestRequestIdRef.current;
const canRead = await can(...);

if (requestId !== latestRequestIdRef.current) return;

setPermissions({ read: canRead, ... });
```

### DIFF
--- a/ui/apps/everest/src/hooks/rbac/rbac.ts
+++ b/ui/apps/everest/src/hooks/rbac/rbac.ts
@@ -94,7 +94,11 @@ export const useNamespacePermissionsForResource = (
   const queryResult = useNamespaces();
   const { data: namespaces } = queryResult;
 
+  const latestRequestIdRef = useRef(0);
+
   const checkPermissions = useCallback(async () => {
+    const requestId = ++latestRequestIdRef.current;
+
     const newPermissions: Record<RBACAction, string[]> = {
       read: [],
       update: [],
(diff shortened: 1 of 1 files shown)


solovevayaroslavna:
Hey @ashnaaseth2325-oss! Thanks for your help! The fix is correct. A few notes:

- The same race exists in a sibling hook. useNamespacePermissionsForResource in the same file has the identical issue: multiple concurrent checkPermissions runs, then an await Promise.all(...) followed by an unconditional setPermissions(newPermissions). If the bug is real here, it's real there too. It'd be good to apply the same guard for consistency — or note why only this hook needs it
- setState after unmount isn't covered. The request-ID guard doesn't prevent setPermissions from being called on an unmounted component (late-resolving promise). This isn't a regression from this PR  but the "cancel in-flight requests" case remains open. An isMountedRef checked in the effect cleanup would close it if desired.

ashnaaseth2325-oss:
Hello @solovevayaroslavna 
Thanks for the review! I've updated the PR to address both comments. The same request id guard is now applied in useNamespacePermissionsForResource for consistency, and I also added an isMountedRef check to prevent state updates from late-resolving async requests after the component has unmounted. Thanks again for the helpful suggestions!

PR #2534, opened 2026-07-09
Secret management API

chilagrow:
This PR implements secret API https://github.com/openeverest/specs/pull/12.

* Updated provider CR spec to have secret definitions which contain UI schema and Open API schema.
* Added secret validation against schema in `provider-runtime`, it will be used by provider's validation webhook (when supported), or during the reconciliation of Instance.
* API to create secret also validates the secret against its schema. 
* Added API tests.

Failing CI is due to using helm-charts in https://github.com/openeverest/helm-charts/pull/65, API tests need this to pass.

Refs https://github.com/openeverest/helm-charts/pull/65.


internal/server/handlers/validation/secret.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// CreateSecret validates the secret request and proxies it to the next handler.
+func (h *validateHandler) CreateSecret(ctx context.Context, cluster, namespace string, secret *corev1.Secret) (*corev1.Secret, error) {
+	if err := utils.ValidateEverestResourceName(secret.GetName(), "name"); err != nil {
+		return nil, errors.Join(ErrInvalidRequest, err)
+	}
+
+	labels := secret.GetLabels()
+	if labels == nil || labels[common.OpenEverestCategoryLabel] == "" {
```
chilagrow: Secret is generic hence without `category` becomes difficult to understand what it is used for. API rejects without category

internal/server/handlers/validation/secret.go
```diff
@@ -0,0 +1,55 @@
...
+// CreateSecret validates the secret request and proxies it to the next handler.
+func (h *validateHandler) CreateSecret(ctx context.Context, cluster, namespace string, secret *corev1.Secret) (*corev1.Secret, error) {
+	if err := utils.ValidateEverestResourceName(secret.GetName(), "name"); err != nil {
+		return nil, errors.Join(ErrInvalidRequest, err)
+	}
+
+	labels := secret.GetLabels()
+	if labels == nil || labels[common.OpenEverestDefinitionLabel] == "" {
+		return nil, errors.Join(ErrInvalidRequest, errors.New("secret must have a definition label"))
+	}
```
recharte: Besides enforcing that the label is set shouldn't we also enforce that the secret data itself complies with the schema definition?

chilagrow: Very good point, I was initially intending each provider to validate schema but we can do this earlier here!

api/core/v1alpha1/provider_types.go
```diff
@@ -92,6 +97,27 @@ type TopologyComponent struct {
...
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	OpenAPIV3Schema *apiextensionsv1.JSONSchemaProps `json:"openAPIV3Schema,omitempty"` //nolint:tagliatelle
+
+	// Shared indicates this secret definition is shared across multiple providers.
+	// When true, secrets with the same definition can be used by other providers.
+	// Each provider still explicitly declares its secret schema,
+	// it would cause issues if the schema is declared differently by different providers.
+	// +optional
+	Shared bool `json:"shared,omitempty"`
```
chilagrow: I've been thinking about the use cases where secrets are used by multiple providers. Each provider define secret schema, so this is allowing exceptional case for allowing secret use.

We don't yet have a use case, however, this will allow usage similar to MonitoringConfig without making the change in the core. But this can be fragile because those who are adding secrets definition in multiple providers must make sure secret schemas are the same.

I can remove sharing secret until solid use case comes up as well.

recharte: Won't this complicate the logic a lot? We rely on labeling secrets with the provider and definition. If we now allow multiple providers to share the same secret that can quickly become very messy 🤔

chilagrow: I vito'd it. We don't need it now, so no need to have it yet 🙏

recharte: Thanks, let's follow YAGNI as much as possible :wink:

recharte (approved):
Well done 🚀

PR #2545, opened 2026-07-11
fix: add panic recovery middleware to echo server

### LINKED ISSUE #2544: [Bug]: Missing Panic Recovery Middleware in API Server
### What is the issue?

In Go web servers built with the Echo framework, it is considered a critical best practice to inject the `echomiddleware.Recover()` middleware into the global handler chain.

Currently, `NewEverestServer` in `internal/server/everest.go` configures the server middlewares (like `requestLoggerMiddleware` and `RemoveTrailingSlash`) but completely omits the panic recovery middleware.

If any API endpoint handler, downstream library, or database connector inadvertently calls `panic()` or dereferences a nil pointer, the entire `openeverest` API server process will crash. Kubernetes will then restart the pod, resulting in dropped connections and temporary denial of service for all users on the cluster.

A panic in an individual API endpoint should be gracefully caught by the framework. The specific HTTP request should fail with an `HTTP 500 Internal Server Error`, a stack trace should be logged to the server logs for debugging, but the API server process itself must remain alive to continue serving other users.

### Steps to Reproduce

1. Introduce a deliberate `panic("test")` into any API handler (e.g. `internal/server/database_cluster.go`).
2. Run the API server.
3. Call the API endpoint.
4. The entire API server process crashes and exits.

### Relevant Logs / Output

```bash
# Example log output when a panic occurs without the middleware
panic: test

goroutine 1 [running]:
...
exit status 2
```

### OpenEverest Version

main / latest

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

Host OS: All
Cloud Provider: All
Kubernetes Version: All
(This is an architectural codebase bug that affects all environments)

### Possible Solution

Add `e.echo.Use(echomiddleware.Recover())` to the top of the middleware chain in `initHTTPServer` (`internal/server/everest.go`). I am opening a PR for this now!

### Additional Context

N/A

shellyco-code:
### Description
This PR addresses Issue #2544

The OpenEverest API server relies on the Echo framework, but it was missing the `echomiddleware.Recover()` middleware in its handler chain (`internal/server/everest.go`).

While Go's standard `net/http` server does have a built-in panic recovery mechanism that prevents panics from crashing the process, it only aborts the active connection — the client receives a connection reset with **no HTTP response**, and **no error is logged** to the server logs. This makes panics in handlers effectively silent failures that are difficult to diagnose.

This PR adds the standard `Recover()` middleware to the top of the Echo middleware chain so that unhandled panics are:
- Caught at the Echo layer **before** `net/http`'s blunt recovery kicks in
- Converted into proper **HTTP 500 Internal Server Error** responses to the client
- **Logged with a full stack trace** for debugging

The server process was not at risk of crashing, but without this middleware, panicking requests were silently dropped with no observability.

### Changes Made
- Added `e.echo.Use(echomiddleware.Recover())` to `initHTTPServer` in `internal/server/everest.go`.

### Testing
- [x] Tested locally via `make test` / `go test`.
- [x] Verified middleware chain order.

### Related Issues
Fixes #2544

### DIFF
--- a/internal/server/everest.go
+++ b/internal/server/everest.go
@@ -213,6 +213,7 @@ func (e *EverestServer) initHTTPServer(ctx context.Context) error {
 	e.echo.GET("/static/*", echo.WrapHandler(staticFilesHandler), e.securityHeaders())
 
 	// Middlewares
+	e.echo.Use(echomiddleware.Recover())
 	e.echo.Use(e.requestLoggerMiddleware())
 	e.echo.Pre(echomiddleware.RemoveTrailingSlash())
 


Denyme24:
the description isn't quite accurate because as far as i can tell, a panic wouldn't crash the pod since net/http already recovers them, it just drops the connection. change is still useful though. 

>This PR addresses Issue #[INSERT_ISSUE_NUMBER_HERE]. 

@shellyco-code  also please fix this.🙂

jtomaszon:
Thanks for picking this up @shellyco-code — and good catch reframing the issue. You're right that the process doesn't actually crash: Go's `net/http` recovers panics per-connection, so #2544's "process exits / pod restarts" premise is off. The real gap is the silently-dropped request with no proper 500 and no useful log entry, which is worth fixing.

A couple of things we ran into when we added `Recover()` to our own Echo-based providers that are worth a look here:

**1. Ordering vs. the request logger — the structured log still won't capture panics.**
`Recover()` is registered here *outside* `requestLoggerMiddleware`. Because `requestLoggerMiddleware` logs via straight-line code after `err := next(c)` (no `defer`), a handler panic unwinds *through* the logger before it can log, and `Recover()` catches it one layer out — logging via Echo's default logger, not the app's structured (zap) logger. Net effect: panicking requests get a 500, but produce **no structured request-log entry** (no request_id, status, latency), which is the observability this PR is aiming for.

Two ways to close that:
- Register the logger first, then `Recover()` (so Recover sits *inside* the logger), or
- Use `middleware.RecoverWithConfig` with a `LogErrorFunc` that routes the panic + stack through the app's structured logger at ERROR level.

The second option is nice because the stack lands in the same queryable log stream as everything else.

**2. `Pre()` middleware isn't covered.**
Since `RemoveTrailingSlash` is registered via `Pre()`, `Recover()` (a `Use()` middleware) executes *inside* it — so a panic in any `Pre` hook wouldn't be recovered. Low risk today, but if the intent is "Recover catches everything," registering it via `Pre()` first would make it truly outermost.

**3. A small test would lock this in.**
`internal/server/middlewares_test.go` already has the `echo.New()` / `httptest` patterns for this. A table test that registers a panicking handler, drives it through the full chain, and asserts (a) 500, (b) the goroutine survives, and (c) a structured log entry was emitted would catch exactly the ordering issue above.

Happy to help if useful — this is a good fix to have in.

shellyco-code:
> Thanks for picking this up @shellyco-code — and good catch reframing the issue. You're right that the process doesn't actually crash: Go's `net/http` recovers panics per-connection, so #2544's "process exits / pod restarts" premise is off. The real gap is the silently-dropped request with no proper 500 and no useful log entry, which is worth fixing.
> 
> A couple of things we ran into when we added `Recover()` to our own Echo-based providers that are worth a look here:
> 
> **1. Ordering vs. the request logger — the structured log still won't capture panics.** `Recover()` is registered here _outside_ `requestLoggerMiddleware`. Because `requestLoggerMiddleware` logs via straight-line code after `err := next(c)` (no `defer`), a handler panic unwinds _through_ the logger before it can log, and `Recover()` catches it one layer out — logging via Echo's default logger, not the app's structured (zap) logger. Net effect: panicking requests get a 500, but produce **no structured request-log entry** (no request_id, status, latency), which is the observability this PR is aiming for.
> 
> Two ways to close that:
> 
> * Register the logger first, then `Recover()` (so Recover sits _inside_ the logger), or
> * Use `middleware.RecoverWithConfig` with a `LogErrorFunc` that routes the panic + stack through the app's structured logger at ERROR level.
> 
> The second option is nice because the stack lands in the same queryable log stream as everything else.
> 
> **2. `Pre()` middleware isn't covered.** Since `RemoveTrailingSlash` is registered via `Pre()`, `Recover()` (a `Use()` middleware) executes _inside_ it — so a panic in any `Pre` hook wouldn't be recovered. Low risk today, but if the intent is "Recover catches everything," registering it via `Pre()` first would make it truly outermost.
> 
> **3. A small test would lock this in.** `internal/server/middlewares_test.go` already has the `echo.New()` / `httptest` patterns for this. A table test that registers a panicking handler, drives it through the full chain, and asserts (a) 500, (b) the goroutine survives, and (c) a structured log entry was emitted would catch exactly the ordering issue above.
> 
> Happy to help if useful — this is a good fix to have in.

I'll work on it . thankyou

recharte:
Direction's settled on #2544 — we do want this in. The reasoning, and the constraints any implementation has to meet, are on the issue; sorry it took us this long to get you an answer.

@jtomaszon's review above is the most useful thing on this issue, and points 1 and 3 are now part of those constraints, so it's worth working straight from it. Two things to add on top:

- The middleware needs to end up outside the `Pre()` chain, not just ahead of the rate limiters — that's his point 2.
- Whatever `RecoverWithConfig` passes to `c.Error` has to keep the response body generic. `everestErrorHandler`'s `default` branch sets `Message: err.Error()`, so a plain `Recover()` hands the caller `{"message":"runtime error: invalid memory address ..."}`.

@shellyco-code — you said back in July you'd pick this up, and this is the PR we'd like to take forward. Do you have time to rework it along those lines? A yes or a no are equally welcome, I'd just rather know than leave it sitting. If I don't hear back in a week or so I'll assume you're busy and find another way to land it, no hard feelings either way.

Once it's ready I'll take the code review from there.

recharte:
@shellyco-code — nothing here since July, so I'm doing what I said I would and passing the first go to #2614. No blame in it: three weeks of our silence came first.

I'll close this once #2614 is moving, so we're not carrying two versions of the same change — say the word if you'd rather I didn't. You'd be welcome back any time.

recharte:
@shellyco-code — the fix landed today: #3200 on `main` (v2) and #3201 on `v1.x`. Closing this one, so I'd rather say plainly why it isn't yours that merged.

Not because anything was wrong with it. You reported the gap and had a patch up the same day, and every version since has been the same change you proposed here. What grew on top — `RecoverWithConfig` instead of `Recover()`, `Pre()` instead of `Use()`, a sanitised body — came out of the review thread on *this* PR, which is the thread that worked out what the fix had to do. And the reframing you took on board in it (net/http already recovers the panic, so this is observability rather than availability) is what got the issue described accurately in the end; the original report had us fixing a crash that wasn't happening.

The reason it isn't this PR is simply that it went quiet in July while we still hadn't answered you, and by the time we did, the thing had sat long enough that I went looking for whatever would land it. Six weeks of that wait was ours. You replied to @jtomaszon's review inside four minutes.

So: thank you, genuinely, and please do come back — for something in the API server or anywhere else. If you want a next one, say the word on an issue and I'll make sure you get an answer in days rather than weeks this time.

PR #2547, opened 2026-07-12
Group backup class job specs in discriminated union pattern

### LINKED ISSUE #2528: Update BackupClass CR
### Task Description

Currently `BackupClass` CR has a lot of things in top level. Some top level fields such as `Config` and `Job` comes together. This PR is about re-organizing `BackupClass` CR fields

https://github.com/openeverest/openeverest/blob/cdc96e804f8904864e76f6fe350a371c2acea271/api/backup/v1alpha1/backupclass_types.go#L52-L102

### Reasoning / Benefit

Tidier CR binding relevant fields.

### Additional Context

_No response_

recharte:
Closes #2528


api/backup/v1alpha1/backupclass_types.go
```diff
@@ -91,15 +95,23 @@ type BackupClassSpec struct {
...
 	// +optional
-	RestoreJob *JobExecution `json:"restoreJob,omitempty"`
+	Job *JobModeSpec `json:"job,omitempty"`
+}
+
+// JobModeSpec bundles everything the in-tree controller needs to run backup
+// and restore operations as Kubernetes Jobs in ExecutionMode="Job".
+type JobModeSpec struct {
+	// Backup describes the job spawned per Backup CR.
+	// +kubebuilder:validation:Required
```
chilagrow, line 109: Let me know if I understand this correctly, Backup is required because Restore via job is not supported yet?

recharte: Not quite — restore via job is fully supported (the restore controller spawns it from `spec.job.restore`). The split is about capabilities: `backup` is required because a backup class that can't take backups makes no sense — that's the whole point of the class — so the schema forces it. `restore` is optional because being able to restore is a real capability that not every tool has: some are backup/export-only, or ship restore support later. Leaving `restore` unset is how a class says "I don't do restores", and if someone creates a Restore CR against it anyway, the controller fails it cleanly with "BackupClass does not define spec.job.restore.jobSpec" instead of blowing up. Same idea as `supportsPITR` on the ProviderManaged side — backup is assumed, recovery features are opt-in.

chilagrow: Thanks!

chilagrow (approved):
Great!

PR #2561, opened 2026-07-14
Support import from external source with no Backup CR

chilagrow:
Refs https://github.com/openeverest/provider-percona-server-mongodb/pull/62.

This PR implements importing data from external source with no Backup CR. https://github.com/openeverest/specs/pull/11

* `BackupClass` handles `ProviderManaged` or `Job` mode import, and holds open API schema for validating parameters.
* `Restore` CR extends import dataSource.
* `provider-runtime` handles creating `Restore` CR for import and resolves `BackupClass`
* `restore_controller.go` reconciles `Restore` CR for import

Refs https://github.com/openeverest/helm-charts/pull/86.


api/backup/v1alpha1/restore_types.go
```diff
@@ -90,6 +94,29 @@ type DataSource struct {
...
+}
+
+// DataSourceExternal references an external storage location for import.
+type DataSourceExternal struct {
+	// BackupClassRef references the BackupClass that defines the import method.
+	// The BackupClass must either:
+	// - Have executionMode=ProviderManaged with providerManaged.supportsImport=true, OR
+	// - Have executionMode=Job with importJob set
+	// +kubebuilder:validation:Required
+	BackupClassRef corev1.ObjectReference `json:"backupClassRef"`
```
chilagrow: TODO: update to common.ObjectRef after https://github.com/openeverest/openeverest/pull/2570

api/backup/v1alpha1/restore_types.go
```diff
@@ -90,6 +94,29 @@ type DataSource struct {
...
+	// StorageRef references a BackupStorage CR in the same namespace
+	// that contains S3 credentials and endpoint configuration.
+	// +kubebuilder:validation:Required
+	StorageRef corev1.ObjectReference `json:"storageRef"`
```
chilagrow: TODO: update to common.ObjectRef after #2570.

api/backup/v1alpha1/restore_types.go
```diff
@@ -128,6 +131,27 @@ type DataSource struct {
...
+	// +optional
+	ClassRef *common.ObjectRef `json:"classRef,omitempty"`
+	// StorageRef references a BackupStorage.
+	// +kubebuilder:validation:Required
+	StorageRef common.ObjectRef `json:"storageRef"`
+	// Parameters contains all import configuration.
+	// Validated against BackupClass.spec.importParameterSchema.
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Required
+	Parameters *runtime.RawExtension `json:"parameters"`
```
recharte, line 154: I'm not sure about this approach. I've been trying to thinking about how a user mentally walks through this, and I want to share my thought process — I'm not claiming I have the right answer here, this is just how I see the flow.

When someone wants to create an instance from existing data, I think they make three decisions, in order:

1. **"I want this instance seeded with data"** → that points them to `dataSource`.
2. **"What state do I want back?"** → the state captured by a specific backup, or the state at some point in time. That's what `type` answers today, and it feels like the natural first question.
3. **"Where does that backup / PITR stream actually live?"** → an existing Backup CR in the cluster, or something external with no CR.

My hesitation with `type: Import` is that it doesn't answer question 2 at all — it jumps straight to question 3. It says "the data is external" but says nothing about whether I'm restoring a full backup or rolling forward to a point in time. So it mixes two different decisions into one field.

It also means we end up with two ways of expressing "external" — this one, and `PointInTime`'s stream source — that look nothing alike.

Maybe the fix is just naming, or maybe "where does it live" should be a nested choice inside `Backup`/`PointInTime` rather than a sibling of them. I honestly don't know yet — but I wanted to kickstart the discussion.

chilagrow: The separation of 2 and 3 is something I didn't consider but worth noting 🤔. Current approach handles "how do I seed the instance", and `type: Import` currently implicitly assumes full backup, which is shortcoming. Additionally the difference for `type: Import` is the lack of Backup CR, both PITR and Backup requires it.

I think allowing `DataSourceBackup` to declare external source in addition to using Backup CR is most sensible. Using full backup is explicit this case and aligns with full backup either from CR or external storage.

```go
// DataSourceBackup identifies the backup to restore.
type DataSourceBackup struct {
	// BackupRef references the Backup CR in the same namespace.
	BackupRef *common.ObjectRef `json:"backupRef"`
	External *ExternalSource `json:"external,omitempty"` // new
}

type ExternalSource struct {
	// ClassRef references a BackupClass. Required for Job mode.
	// ProviderManaged mode uses the Instance's backup class (spec.backup.classRef)
	// and backup must be enabled.
	ClassRef *common.ObjectRef `json:"classRef,omitempty"`
	// StorageRef references a BackupStorage.
	StorageRef common.ObjectRef `json:"storageRef"`
	// Parameters contains all import configuration.
	Parameters *runtime.RawExtension `json:"parameters"`
}
```

Renaming could work as well. That case I would consider something like following.

```go
type DataSource struct {
    // +kubebuilder:validation:Required
    Type DataSourceType `json:"type"`
    Backup *DataSourceBackup `json:"backup,omitempty"`
    ExternalBackup *ExternalSource `json:"externalBackup,omitempty"`
    PointInTime *DataSourcePointInTime `json:"pointInTime,omitempty"`
}
```

While data importer in v1 didn't support PITR equivalent and (not in scope of this work), it's possible that operators support PITR without having Backup CR. That case, restructuring CR rather than renaming is more trivial to extend to support such case. 

WDYT?

recharte: Honestly, I'm leaning towards the first option as it feels more unified and easier for a new user to understand, but it's not something I'm 100% confident about.

PR #2562, opened 2026-07-14
Add PITR support plumbing: validation, restore gating, and per-storage recovery-window status

### LINKED ISSUE #2468: [Core] Support PITR in v2
### Task Description

TBA

### Reasoning / Benefit

TBA

### Additional Context

_No response_

recharte:
This PR completes the runtime-side support for point-in-time recovery on ProviderManaged backup classes:

- **PITR config validation** — the per-storage PITR config on an Instance (`spec.backup.storages[].pitr.config`) is now validated against the OpenAPI schema the BackupClass declares in `providerManaged.pitrConfigSchema`. Invalid or unknown fields are rejected both at admission time (provider validation webhook) and during reconcile, where the violation is surfaced on the `BackupConfigured` condition without failing the Instance.
- **Restore PITR gating** — a Restore that requests PITR is now rejected when the resolved BackupClass does not advertise `providerManaged.supportsPITR`. This is enforced in two places: the API server returns a 400 on creation, and the runtime restore reconciler marks the Restore terminally `Failed` before any provider code runs (covering Restores created directly against the Kubernetes API).
- **CEL rule on `DataSourcePITR`** — `date` is now required at admission when `pitr.type` is `date`.
- **Recovery-window observability** — new `instance.status.backup.storages[]` status block reporting `latestRestorableTime` per configured backup storage. Providers opt in via the new `InstanceBackupStatusReporter` interface in the provider-runtime; the runtime publishes the reported values after every successful Sync. Errors while collecting this data are logged but never fail reconciliation.


chilagrow (dismissed):
Great work!

PR #2564, opened 2026-07-15
Add provider list subcommand to Everestctl

### LINKED ISSUE #2554: Add `everestctl provider list` subcommand

Providers are the entry point for everything in OpenEverest — users must know which providers are installed before creating an instance. Today there is no CLI way to discover installed providers; users have to inspect Provider CRs with kubectl or browse the UI.

Add an `everestctl provider list` subcommand that lists installed Provider resources via the Everest API. The table output should include NAME and AGE. Details such as supported topologies, version catalog, and schemas remain available via `--json`. It should follow the existing CLI conventions: `--json`/`--verbose` output modes, `--cluster`, and `--context` flags. The `provider` command group should register the alias `prov`, and the `list` subcommand the alias `ls`.

```
# List installed providers
everestctl provider list

# Using aliases
everestctl prov ls

# JSON output — inspect the full version catalog
everestctl provider list --json | jq '.[].spec.versions'
```

`kubectl get providers.core.openeverest.io` — requires direct cluster access and doesn't surface provider details in a readable form.

### Additional Context

Part of a set of `list` subcommands for core resources (instances, backup storages, backup classes, providers). This would introduce the `provider` command group, which could later host `install`/`upgrade` subcommands. Columns like READY were deliberately left out since the Provider CR status is not currently populated by any controller; additional columns (topologies, default version) can be added later if they prove useful.

VijetaPriya47:
Lists Provider resources via the Everest API with NAME and AGE columns. READY is omitted since the Provider CR status is not yet populated by any controller. Supports --json/--verbose for raw output, --cluster/--context selection, the `provider`/`prov` command group, and the `list`/`ls` aliases, following instance list conventions.
Closes #2554

### DIFF
--- a/commands/provider.go
+++ b/commands/provider.go
@@ -1,0 +1,36 @@
+// everest
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package commands ...
+package commands
+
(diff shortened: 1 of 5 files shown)


VijetaPriya47:
@recharte could you review this too :>

recharte (approved):
🚀

PR #2570, opened 2026-07-16
Unify cross-resource reference mechanisms across all CRDs

recharte:
## Problem

Our CRDs accumulated **seven different mechanisms** for referencing other objects, with no rule for when to use which:

1. Plain `fooName` strings (`Backup.spec.instanceName`, `backupClassName`, `credentialsSecretName`, …)
2. A bare string with no suffix at all (`Instance.spec.provider`)
3. `corev1.LocalObjectReference` (`storages[].storageRef`, `Config.secretRef/configMapRef`)
4. A custom one-field struct (`BackupClassReference`)
5. `corev1.TypedLocalObjectReference` (status `operatorBackupRef`/`operatorRestoreRef`)
6. A custom namespace+name struct, plus a **string** field misleadingly named `credentialsSecretRef` (Plugin)
7. A logical-name indirection: `Backup.spec.storageName` did **not** reference the BackupStorage CR — it referenced the logical key `Instance.spec.backup.storages[].name`, while its own doc comment claimed otherwise

Beyond the inconsistency, the corev1 reference types carry no validation (an empty `storageRef: {}` passed admission), can't be documented per-usage (`kubectl explain` showed upstream's generic TODO-laden comments), and the logical-name indirection was a standing source of confusion.

## What we changed

**One convention everywhere:** every cross-object reference is a struct field named `{something}Ref`, typed by a small set of purpose-built, project-owned reference types in the new `api/common/v1alpha1` package:

| Type | Shape | Used for |
|---|---|---|
| `ObjectRef` | `{name}` | CR→CR refs (`instanceRef`, `classRef`, `storageRef`, `providerRef`, `backupRef`, …) |
| `SecretRef` | `{name}` | whole-Secret refs (`credentialsSecretRef`, `connectionSecretRef`) |
| `NamespacedObjectRef` | `{name, namespace}` | cluster-scoped referrer → namespaced target only |
| `TypedObjectRef` | `{group, kind, name}` | status refs to operator-native resources |

All types carry uniform validation (required, `MaxLength=253`, DNS-subdomain pattern) and `+structType=atomic` so Server-Side Apply treats each reference as a single owned value. Bare string fields now exist **only** for intra-object logical keys (e.g. `scheduleName`), never for object references.

**The logical storage name is gone.** `Instance.spec.backup.storages[]` entries are keyed directly by the referenced BackupStorage's name (`storageRef.name`, CEL-enforced unique), and `Backup.spec.storageRef` now points at the BackupStorage CR itself. One name, one meaning.

## Why this approach

- The Kubernetes API conventions' [Object references](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references) section recommends purpose-built `{field}Ref` structs whose schemas are "designed to enable purely additive fields" — a `fooRef: {name}` struct can later grow `key`, `namespace`, or `kind` without a breaking change, whereas a `fooName` string cannot.
- The corev1 reference types' own godoc says **not** to use them: *"Instead of using this type, create a locally provided and used type that is well-focused on your reference"* (they can't carry validation markers, per-usage docs, or evolve safely).
- This is where the ecosystem converged: Gateway API (`parentRefs`/`backendRefs`), cert-manager (`issuerRef`), Cluster API. The shared-package layout mirrors `k8s.io/apimachinery/pkg/apis/meta/v1` and cert-manager's `apis/meta/v1`: it is *not* an API group (no kinds, no scheme, no CRDs) — the types are embedded and versioned on the wire by each parent's `apiVersion`; the package version segment is a Go-level evolution seam.
- A **fork-on-divergence policy** (documented in the package) keeps the shared types honest: when one usage needs extra fields, it forks into its own purpose-built type — a wire-invisible move, since Go type identity is not part of the API contract.

## Simplified: component `config` is a plain inline string

While unifying the references we re-examined `ComponentSpec.Config`, which was a `{secretRef, configMapRef, key}` structure. Instead of porting that shape to the new reference types, we **removed the Secret/ConfigMap indirection entirely**:

```yaml
components:
  engine:
    config: |            # plain string — the engine's config file content
      max_connections=100
```

Rationale:

- **v1 evidence.** v1 supported inline-only engine config for years with zero requests for Secret/ConfigMap-sourced config; the ref arm was speculative v2 design with no driving use case.
- **Evolution asymmetry.** Adding a reference mechanism later is purely additive (an optional `configFrom` sibling, following the `env`/`envFrom` precedent); *removing* one later would be a breaking change. YAGNI therefore costs nothing here.
- **Sharing across Instances** is already served by InstancePresets — and keeping config inline means edits bump the Instance's `generation` and flow through the normal reconcile path, staying visible to GitOps diffing (referenced-object edits would be invisible to both).
- It plays cleanly with the platform's secret/configmap management APIs: no key-selector coupling between user-created objects and Instance fields.

The field is capped at 64 KiB and documented as *not encrypted at rest* — credentials do not belong in it. Structured, schema-validated payloads (`backup.config`, `global`, `topology.config`, `customSpec`) likewise stay inline `RawExtension` — a reference arm there would defeat admission-time schema validation. Credentials stay ref-only (the BackupStorage write-only `accessKeyId`/`secretAccessKey` webhook flow is unchanged).

A follow-up PR will unify the *structured* config fields (`global`/`customSpec`/`config` naming and schema declaration) across the CRDs.

## Breaking changes

No backward compatibility is provided (all APIs are `v1alpha1`):

- All renamed spec/status fields listed above; existing CRs must be migrated.
- `ComponentSpec.Config` changed from an object to a plain string; the InstancePreset namespace-default Secret resolution for config was removed along with it.
- Field indexes changed (`spec.instanceName` → `spec.instanceRef.name`).
- The provider-runtime SDK surface changed (`BackupExecutionStatus`/`RestoreExecutionStatus` now use `TypedObjectRef`; the `Context.ComponentConfig()` helper was removed — read `component.Config` directly); **provider repos (psmdb/pxc/pg) require a follow-up migration** before they compile against this runtime.


api/core/v1alpha1/instance_types.go
```diff
@@ -324,10 +320,38 @@ type Storage struct {
...
+	ValueFrom *ConfigSource `json:"valueFrom,omitempty"`
+}
+
+// ConfigSource represents a source for the content of a Config.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.secretKeyRef) != has(self.configMapKeyRef)",message="exactly one of secretKeyRef or configMapKeyRef must be set"
+type ConfigSource struct {
+	// SecretKeyRef selects a key of a Secret in the same namespace.
+	// +optional
+	SecretKeyRef *common.SecretKeyRef `json:"secretKeyRef,omitempty"`
```
chilagrow: With SecretKeyRef, do I understand it references specific key within Secret's data? If multiple keys are needed what's the recommended way to set component spec? Let's use split horizon as an example, where secret data contains TLS certificate and key where both are needed in components. 🤔 I feel using SecretRef will allow Split horizon use case.

recharte: Thanks for challenging this. It got me thinking that we don't even have a good use case for this secret/configmap-based config. In v1 it was always inlined in the DBC CR and we never had any indication that we might need to move it to a secret/configmap so I believe this is just a premature change that introduces a behavior we might not even need. I suggest dropping this entirely and moving Config to be a simple string. We can always add it later if we need to.

chilagrow (dismissed):
Great!

PR #2573, opened 2026-07-17
Add backup-storage list subcommand to everestctl

### LINKED ISSUE #2552: Add `everestctl backup-storage list` subcommand

Backup storages can be managed through the API (`/v1/backup-storages`) and the UI, but `everestctl` has no command to view them. When configuring backups from the terminal, users cannot verify which storages exist or which namespaces they are available in without switching to the UI or raw kubectl/API calls.

Add an `everestctl backup-storage list` subcommand that lists configured backup storages via the Everest API. The table output should include NAME, NAMESPACE, TYPE, and AGE. Type-specific details (e.g. S3 bucket, region, endpoint) are available via `--json`; credentials must never be printed. It should support namespace filtering and the standard `--json`/`--verbose`, `--cluster`, and `--context` flags. The `backup-storage` command group should register the aliases `backupstorage` and `bs`, and the `list` subcommand the alias `ls`.

```
# List backup storages in a namespace
everestctl backup-storage list --namespace everest

# Using aliases
everestctl bs ls --namespace everest

# JSON output — inspect S3 details
everestctl backup-storage list --namespace everest --json | jq '.[].spec.s3'
```

Querying `/v1/backup-storages` with curl and a JWT token — functional but requires manual token handling and offers no formatted output.

### Additional Context

Part of a set of `list` subcommands for core resources (instances, backup storages, backup classes, providers). This would introduce the `backup-storage` command group, which could later host `create`/`delete` subcommands. Although only the `s3` type exists today, the TYPE column reflects the `spec.type` discriminator already present in the schema, so the table format stays stable as new storage types are added.

VijetaPriya47:
Adds `everestctl backup-storage list` to view configured backup storages via the Everest API, mirroring the existing provider list pattern. Supports namespace filtering (-n/--namespace), --json output, and the backupstorage/bs and list/ls aliases. Credentials are never included in the output.
Closes #2552

### DIFF
--- a/commands/backupstorage/list.go
+++ b/commands/backupstorage/list.go
@@ -42,8 +42,8 @@ Credentials are never included in the output.`,
 		Example: `  # List backup storages in a namespace
   everestctl backup-storage list --namespace everest
 
-  # Using aliases
-  everestctl bs ls --namespace everest
+  # List across all namespaces
+  everestctl bs ls --all-namespaces
 
   # JSON output — inspect S3 details
   everestctl backup-storage list --namespace everest --json | jq '.[].spec.s3'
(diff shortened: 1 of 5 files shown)


commands/backupstorage/list.go
```diff
@@ -0,0 +1,89 @@
...
+  everestctl backup-storage list --namespace everest --cluster staging --context staging-ctx`,
+		PreRun: listPreRun,
+		Run:    listRun,
+	}
+	listCfg  = &backupstoragecli.Config{}
+	listOpts = &backupstoragecli.ListOptions{}
+)
+
+func init() {
+	listCmd.Flags().StringVarP(&listOpts.Namespace, cli.FlagBackupStorageNamespace, "n", "", "Namespace to list backup storages from (required)")
```
recharte: Let's also add `--all-namespaces/-A` for consistency with other commands.

recharte (changes requested):
Well done @VijetaPriya47. Please fix the conflicts and add the all-namespaces and we'll be good to merge 😉

recharte (approved):

PR #2574, opened 2026-07-17
Fix dynamic select default option resolution (#2566)

### LINKED ISSUE #2566: UI-generator select ignores default: true
### What is the issue?

**Title:** UI-generator select ignores `default: true` on dynamically-resolved options (version picker always selects first list item)

**Area:** UI / ui-generator (v2 schema-driven form)

## Description

When a `select` component resolves its options dynamically via `optionsPath` (e.g. the database version picker with `optionsPath: spec.versions`), the default selected option is always the **first item in the array**, regardless of the `default: true` flag set on an individual option in the provider spec.

## Expected

The version marked with `default: true` (`8.0.8`) is pre-selected.

## Actual

The first item in the list (`8.0.12`) is pre-selected. The `default: true` flag has no effect.

Additional confirmation:
- Moving `default: true` to a lower item does not change the pre-selected value.
- Reordering the list so a different version is first *does* change the pre-selected value.
- It currently "works" for PSMDB only by coincidence — the `default: true` version happens to be first in `spec.versions`.

### Steps to Reproduce

1. Use a provider whose spec exposes a versions list via `optionsPath`, e.g. PSMDB `spec.versions`.
2. In the provider spec, mark a version that is **not first** in the list with `default: true` (e.g. `8.0.8`), while the first list item (`8.0.12`) has no default flag.
3. Open the "Create database" wizard (v2 form).

### Relevant Logs / Output

_No response_

### OpenEverest Version

release-2.0

### Installation Method

Helm

### Kubernetes Distribution

minikube

### Environment Details

_No response_

### Possible Solution

1. Extend `optionsPathConfig` with an optional `defaultPath` (e.g. `defaultPath: default`) so the resolver knows which field marks the default option.
2. Have `resolveSelectOptions` carry an `isDefault` flag per option (or return the resolved default value).
3. In `preprocess-schema.ts`, when `defaultValue` is not explicitly set in the schema, prefer the option marked as default and fall back to `options[0].value` only if none is marked.

### Additional Context

- `utils/preprocess/preprocess-schema.ts`: default value falls back to `options[0].value` instead of the option flagged as default.
- `ui-component/utils/select-component-handler.tsx`: `resolveSelectOptions` maps each resolved item to `{ label, value }` only via `labelPath`/`valuePath`, dropping any `default` field.
- `ui-generator.types.ts`: neither `SelectOptionsItem` nor `optionsPathConfig` carry any "default" information, so the flag can never reach the preprocessing step.

- Should also work for other dynamically-resolved selects, not only version.
- Static-`options` selects can keep using explicit `fieldParams.defaultValue`.

Anuragkumar-687:
## Summary

Fixes #2566 by preserving provider-defined default values for dynamically populated select fields while maintaining the existing fallback behavior.

## Changes

- Add optional `defaultPath` to `optionsPathConfig`
- Preserve provider-defined default information while resolving dynamic select options
- Prefer the provider-defined default option before falling back to the first option
- Preserve existing behavior when `fieldParams.defaultValue` is explicitly provided
- Update unit tests

## Testing

- [x] Updated unit tests
- [x] Targeted unit tests pass
- [x] Lint passes
- [x] Copyright check passes

### DIFF
--- a/ui/apps/everest/src/components/ui-generator/ui-component/utils/select-component-handler.tsx
+++ b/ui/apps/everest/src/components/ui-generator/ui-component/utils/select-component-handler.tsx
@@ -18,6 +18,7 @@ import {
   FieldType,
   Component,
   SelectFieldParams,
+  SelectOptionsItem,
 } from '../../ui-generator.types';
 import { Provider } from 'shared-types/api.types';
 import { getValueByPath } from './get-value-by-path';
@@ -31,7 +32,7 @@ export const isSelectComponent = (item: Component): item is SelectComponent => {
 export const resolveSelectOptions = (
   selectParams: SelectFieldParams,
   providerObject?: Provider
-): { label: string; value: string }[] => {
+): SelectOptionsItem[] => {
   if ('options' in selectParams && selectParams.options) {
     return selectParams.options;
   }
@@ -46,7 +47,7 @@ export const resolveSelectOptions = (
     const rawData = getValueByPath(providerObject, optionsPath);
 
     if (Array.isArray(rawData)) {
-      const { labelPath, valuePath } = optionsPathConfig;
+      const { labelPath, valuePath, defaultPath } = optionsPathConfig;
       return rawData.map((item) => ({
         label: String(
           getValueByPath(item as Record<string, unknown>, labelPath) ?? ''
@@ -54,6 +55,11 @@ export const resolveSelectOptions = (
         value: String(
           getValueByPath(item as Record<string, unknown>, valuePath) ?? ''
         ),
+        isDefault: defaultPath
+          ? Boolean(
+              getValueByPath(item as Record<string, unknown>, defaultPath)
+            )
+          : false,
       }));
     }
   }
--- a/ui/apps/everest/src/components/ui-generator/ui-component/utils/tests/select-component-handler.test.tsx
+++ b/ui/apps/everest/src/components/ui-generator/ui-component/utils/tests/select-component-handler.test.tsx
@@ -47,20 +47,21 @@ describe('select-component-handler utils', () => {
         optionsPathConfig: {
           labelPath: 'version',
           valuePath: 'version',
+          defaultPath: 'default',
         },
       },
       {
         spec: {
           availableVersions: {
-            engine: [{ version: '8.0' }, { version: '8.1' }],
+            engine: [{ version: '8.0' }, { version: '8.1', default: true }],
           },
         },
       } as never
     );
 
     expect(options).toEqual([
-      { label: '8.0', value: '8.0' },
-      { label: '8.1', value: '8.1' },
+      { label: '8.0', value: '8.0', isDefault: false,},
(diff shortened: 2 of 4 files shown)


Anuragkumar-687:
Hi @solovevayaroslavna , just a friendly follow-up on this PR. When you have a chance, could you please take a look? I'd really appreciate your review. Thanks!

Anuragkumar-687:
Hi @solovevayaroslavna! Just a friendly follow-up on this PR. It's been open for a couple of weeks now, and I'd really appreciate your review whenever you have a chance. I can also update/rebase the branch if needed to get the remaining CI checks running again. Thanks!

vickysharma-prog (changes requested):
The three steps @solovevayaroslavna sketched in #2566 are all here, in the three files the issue named, and the shape is right: `defaultPath` on `optionsPathConfig`, `isDefault` carried through `resolveSelectOptions`, and `defaultOption?.value ?? options[0].value` in `preprocess-schema.ts`. The fallback keeps the old behaviour when nothing is marked, so it is safe to land.

Two things before it does.

**Nothing turns it on yet.** `defaultPath` is optional, and across the provider repos there are 17 files using `optionsPathConfig` — the `definition/topologies/*/topology.yaml` sources and the `charts/*/generated/provider-spec.yaml` they produce — and none of them sets it. The version select in `provider-percona-xtradb-cluster/definition/topologies/cluster/topology.yaml` is representative:

```yaml
optionsPathConfig:
  labelPath: "name"
  valuePath: "name"
```

`provider-sdk/PROVIDER_DEVELOPMENT.md:627` documents the same two keys and no third. So `defaultPath` is `undefined`, `isDefault` is `false` on every option, `defaultOption` is `undefined`, and `defaultValue` still falls back to `options[0].value`, which is the symptom in #2566.

The data side is already there and waiting: `definition/versions.yaml` in the provider repos carries `default: true` (PSMDB marks `8.0.12-4`), and that file's header says "Mark exactly one version per type as `default: true`". Only the schema key connecting the two is missing, and the issue's own plan is UI-only.

So this is the UI half of a two-part change. Either the provider `topology.yaml` files and `PROVIDER_DEVELOPMENT.md` come with it, or `Fixes #2566` becomes `Part of #2566` with a note on what is still needed. Otherwise merging closes an issue whose symptom is unchanged.

**CI is red on `format-check`.** `prettier . --check` reports `Code style issues found in 3 files`. `lint-check` passed, so the checklist is accurate on that; this is Prettier rather than ESLint. `build` and `test` are marked cancelled rather than failed, since the matrix stopped them once format-check went down, so their state is unknown rather than broken.

Running the repo's format script over the four changed files should clear it: the trailing whitespace after `const defaultOption = ...`, the `isDefault: false,},` spacing in the test expectations, and the two type lines that now run past the print width.

Anuragkumar-687:
Hi @solovevayaroslavna , just a friendly follow-up on this PR. When you have a chance, could you please take a look? I'd really appreciate your review. Thanks!

PR #2583, opened 2026-07-20
Unify structured configuration under a single `parameters` concept

recharte:
## Problem

After the reference unification (#2570), the same concept — "a structured, inline payload that core treats as opaque and validates against a schema declared by the Provider or BackupClass" — still went by three different names depending on where it appeared:

| Field | Name used |
|---|---|
| `Instance.spec.global` | "global" |
| `Instance.spec.components[].customSpec` | "customSpec" |
| `Instance.spec.topology.config`, `Backup/Restore.spec.config`, `schedules[].config`, `pitr.config` | "config" |

The schema-declaration side was split across two mechanisms: BackupClass declared typed `{openAPIV3Schema}` wrappers with a shared `Validate()`, while the Provider CRD declared bare `RawExtension` schemas with ad-hoc validation paths (PITR even needed a conversion shim to reuse the validator). And `config` — the most overloaded word in the ecosystem — was simultaneously used for a different thing on components (the engine configuration file), inviting a "what goes where?" confusion.

## What we changed

**One name, one mechanism.** Every provider/class-interpreted structured payload is now called `parameters`, at every scope:

- `Instance.spec.parameters` (was `global`) — provider-level
- `Instance.spec.topology.parameters` — topology-level
- `Instance.spec.components[].parameters` (was `customSpec`) — component-level
- `Backup.spec.parameters`, `Restore.spec.parameters`, `schedules[].parameters`,
  `pitr.parameters` — backup-class-level

And every schema declaration uses one shared type, `common.ParametersSchema` (`{openAPIV3Schema}` + a single `Validate()`), promoted from the BackupClass pattern: `Provider.spec.parametersSchema`, `topologies[].parametersSchema`, `components[].parametersSchema`, `BackupClass.spec.parametersSchema` / `.restoreParametersSchema` / `.pitrParametersSchema`. The PITR validation shim is gone; runtime decode helpers follow the same vocabulary (`DecodeParameters`, `DecodeComponentParameters`, `DecodeTopologyParameters`).

The rule an API user or provider author now holds in their head: **`parameters` = structured, schema-validated, always inline; schemas are declared as `parametersSchema` by the thing that interprets them.**

## Why `parameters`

The name is the established term in both lineages this platform combines:

- **Open Service Broker API** — the original DBaaS-platform spec: provisioning requests carry `parameters`, validated against the plan's `schemas.service_instance.create.parameters` JSON Schema. Structurally identical to our Provider/BackupClass schema mechanism, down to the naming of both halves.
- **The Kubernetes class pattern** — StorageClass, IngressClass, VolumeSnapshotClass `parameters`, and Gateway API's `parametersRef`: implementation-interpreted settings that core carries but does not interpret.

On shape: our `parameters` are nested objects rather than StorageClass's flat `map[string]string` — deliberately so. The flat map predates structural schemas and is a documented pain point; every newer instance of the pattern (OSB objects with non-string values, Gateway API's structured `parametersRef` targets, CAPI variables) went structured, and flattening would discard our schema-validation machinery. `parameters` denotes the field's role, not its shape.

This also completes the story from the previous PR, which removed the engine-config `value`/`valueFrom` union: the engine configuration file is no longer a core concern at all — it is an ordinary property in the provider's component `parametersSchema` (recommended name: `configuration`, mirroring the upstream Percona operators verbatim), keeping the boundary clean: **core types what the platform interprets; provider schemas own what only providers interpret.**

## Breaking changes

No backward compatibility (all APIs are `v1alpha1`):

- All field renames listed above; existing CRs must be migrated.
- `BackupClassConfig` replaced by `common.ParametersSchema`; Provider schema fields changed type from `RawExtension` to `ParametersSchema`.
- Provider-runtime SDK: decode helper renames; providers require a follow-up migration (together with `definition/` files and uiSchema hint paths, which still reference `customSpec.*`, and the provider-sdk's `providerbuilder generate` templates).


chilagrow (dismissed):
Thanks for standardized them 💪

PR #2588, opened 2026-07-21
feat(api): add TLSSpec to DatabaseClusterSpec with validation and contract rules

### LINKED ISSUE #2576: add a shared `tls` field to `InstanceSpec`

TLS is a universal database concern, but every operator defines its own shape for it. Users have to learn a different TLS config per provider, and there's no common way to "connect over TLS."

I'd like to propose adding a small, generic `tls` field to the shared `InstanceSpec` so it would work across all providers

We already have a precedent `Backup ` in `InstanceSpec` which provides a way to unified generic way to handle backups rather than handling each provider's custom spec. TLS/certificates is the same kind of thing, every engine (Postgres, Mongo ) has server certs, a CA, client certs, a verification mode.

Benefits:

- One schema users learn once, works across providers.
- Each provider translates it into its native operator config.
- The `CustomSpec` stays as the escape hatch for engine-specific config.

```
type TLSSpec struct {
    // require | verify-ca | verify-full
    Mode string `json:"mode,omitempty"`

    // CA used to verify the server (and to issue certs if the engine self-signs)
    CASecretRef string `json:"caSecretRef,omitempty"`

    // Optional bring-your-own server cert
    ServerCertSecretRef string `json:"serverCertSecretRef,omitempty"`

    // Optional bring-your-own client cert
    ClientCertSecretRef string `json:"clientCertSecretRef,omitempty"`
}
```

Let's keep the shape to a minimal common denominator fields, I think these should be present for starters we can  discuss if different fields are preferred.

UI, CLI

_No response_

### Additional Context

_No response_

Nachiket-Roy:
## Description

This PR implements the API foundation for #2576 by introducing a unified `spec.tls` API for `DatabaseCluster` and adding the validation required to ensure consistent configuration across supported database engines.

### What this PR changes

* Add a new `TLSSpec` API with support for:

  * `mode`
  * `caSecretRef`
  * `serverCertSecretRef`
  * `clientCertSecretRef`
* Add `spec.tls` to `DatabaseClusterSpec`.
* Default `spec.tls.mode` to `require` when `spec.tls` is present and no mode is specified.
* Add validation for:

  * Required secret data keys (`ca.crt`, `tls.crt`, `tls.key`)
  * Conflicts between `spec.tls` and engine-specific TLS configuration
  * Provider-specific TLS mode restrictions
* Add conversion support, CLI flags, UI form mappings, and API documentation for the new field.
* Add unit tests covering defaulting, validation, secret contracts, conflict detection, and backward compatibility when `spec.tls` is not specified.

## Related Issue

Closes #2576

## How I Tested This

* `go test ./pkg/tests/...`
* `go build ./...`
* `make format`

Added unit tests covering:

* TLS defaulting
* Secret validation
* Conflict detection
* Provider validation
* Backward compatibility


Nachiket-Roy:
closing in favour of Broader secret management system being developed in `openeverest/specs#12

Issue #2582, opened 2026-07-19
Event stream: since parameter is not honored subscriber reconnects permanently lose events


onkar717:
The `GET /v1/events` endpoint is supposed to take a `since=<resourceVersion>` param spec 003 §16 Q2 lists it as part of the minimal backend surface, and plugin-audit already does its half of the contract: it saves the last resourceVersion it processed and sends it back every time it reconnects. On the host side though, `eventsHandler` (`internal/server/events.go`) never looks at `since`. The only params it reads are `types` and `namespaces`:

```go
func (e *EverestServer) eventsHandler(c echo.Context) error {
	var typeFilter []events.Type
	if raw := c.QueryParam("types"); raw != "" { ... }
	var nsFilter []string
	if raw := c.QueryParam("namespaces"); raw != "" { ... }
	ch, cancel := e.eventHub.Subscribe(typeFilter, nsFilter)
	...
```

Same story one level down: `Hub.Subscribe` (`pkg/events/hub.go`) has no cursor parameter at all, just types and namespaces. So a reconnecting subscriber gets a fresh live feed starting at "now", and everything that happened while it was away is simply gone.

I ran into this while working on the "make audit rock solid" task for plugin-audit, and verified it with a live setup rather than just reading the code. Setup was envtest instead of a full kind cluster `NewEverestServer` wants a real kube API for the CRD watches and RBAC bootstrap, and envtest (etcd + kube-apiserver) was the lightest thing that would actually boot it. Then:

1. Core + plugin-audit running, plugin connected.
2. Created 20 `Instance` CRs  20/20 rows in the plugin's SQLite.
3. `kill -9` on plugin-audit, created 20 more Instances during a ~15s outage, brought it back up.
4. 20 more Instances after reconnect.

60 generated, 40 persisted. The missing 20 are exactly the outage-window ones, and none of them ever came back. From the plugin's log, it did its part correctly:

```
event stream connected (cursor="")
[... killed, ~15s outage, 20 Instance creates fired ...]
event stream connected (cursor="455")
```

`since=455` went out on the wire. The host accepted the connection and streamed from "now" anyway.

[...]

PR #2591, opened 2026-07-21
events: replay buffer so reconnecting subscribers don't lose events

onkar717:
Implements the design from #2582.

**What changes**
- `Event` gains `seq` (monotonic, assigned in the Hub) and `epoch` (set at Hub startup)
- New bounded ring buffer behind an interface, bounded by count and age; in-memory impl only. The interface is so valkey/redis can drop in later without touching callers.
- `Hub` stamps seq, appends to the buffer before fan-out, and `Subscribe` now takes a cursor
- `eventsHandler` parses `since`, replays from the buffer, or returns 410 when it can't serve the gap
- Buffer size and max age are configurable

**Cursor format** is now `epoch:seq` as one opaque string, carried in the SSE `id:` field. Old clients sending a bare resourceVersion get a 410, reset, and fall back to live-only - same behaviour they have today, no crash. Verified with a test using the exact format the current plugin-audit client sends.

**Two decisions worth your eye**
1. `Since(n)` treats a cursor immediately left of the retention boundary as valid: if seq 3-5 are retained, `Since(2)` succeeds (nothing between last-seen and oldest-retained is missing) and `Since(1)` 410s. Easy to get off by one either way - all four boundary cases are pinned down in one test against one buffer state.
2. Moving channel sends outside the Hub lock (required so a slow subscriber can't block publishing) means `Cancel()` could close a channel mid-send. Handled by never closing `sub.ch` and closing a separate `done` channel via `sync.Once`; `Subscription` gains a `Dropped` field so the handler stops reading.

**Not in this PR:** metrics, valkey/redis backend, and the plugin-side snapshot fallback (separate repo).

Tests: buffer unit tests incl. eviction, wrap-around, and all four replay-boundary cases; Hub tests asserting no gaps or duplicates across reconnect under concurrent publishers with `-race`; handler tests for 410, replay, and legacy-cursor backward compat.


onkar717:
Two things I noticed while re-reading this that are worth flagging myself rather than waiting to be asked - neither blocks this PR, both affect the plan going forward.

**1. The valkey/redis backend will need the locking reworked, not just a new `Buffer` impl.** `broadcast()` calls `h.buffer.Append` while holding `h.mu` (`hub.go`) - fine for an in-memory slice, but for a network-backed buffer that's a round-trip inside the global publish lock. Every event in the system would serialize behind one redis call, so a slow/unavailable redis stalls the whole hub, not just event delivery. The `Buffer` interface boundary is right, but the locking boundary will need to move - likely assign `seq` under `h.mu` (has to stay synchronous, it's the ordering guarantee) but do the actual `Append` after releasing it, or make it async via a write queue. Wanted this on record before the valkey PR so it's not a surprise mid-review.

**2. Filtered subscribers can 410 without having actually missed anything.** The cursor is a single global `seq` shared across all event types, but a client only ever advances its cursor from events matching its own filter. A subscriber watching only `user.login` sitting on cursor 7 while unrelated high-volume events (say `instance.created`) push the global seq to 5000 will find its cursor aged out of the buffer on reconnect - a 410 - even though zero `user.login` events were lost. Kubernetes solves this exact problem with **watch bookmarks**: periodic no-op events carrying the current resourceVersion so an idle-but-connected watcher's cursor stays fresh without needing a matching event to advance it. Same fix would apply here (a periodic bookmark broadcast that every subscriber receives regardless of filter, carrying just seq/epoch). Not fixing it in this PR since it's a distinct piece of work - happy to open a follow-up issue for it if that's useful.

spron-in:
@onkar717 looks good, thanks!

Couple of things:
1. TTL of 1 minute seems aggressive. What's the reasoning? If a pod restarts or a network issue lasts 90s, subscribers will 410 and lose events. Should this be configurable?
2. Does `events.NewRingBuffer` validate `maxLen`? If someone passes 0 or negative, does it panic or silently fail? If it doesn't validate internally, we should guard at the call site.

onkar717:
@spron-in thanks - both fair, answered below.

**1. TTL:** actual default is 5 minutes, not 1 - `EventBufferMaxAge`, `cmd/config/config.go`, `EVENT_BUFFER_MAX_AGE` env var. Already configurable, matches the kube watch-cache default we've been anchoring the rest of this design to (§16 Q6 in the original spec discussion), so a plugin that's been gone longer than that was always going to need a resync regardless of this buffer. If you want a longer default given real restart/network-blip durations you've seen in practice, it's a one-line change - your call on the number.

**2. Good catch, this was a real gap.** `NewRingBuffer(0, ...)` (or a negative value) previously meant `unbounded` for that dimension - not a panic, but arguably worse: a misconfigured/empty env var would let the buffer grow without limit instead of failing safely. Fixed in the latest commit: non-positive `maxLen`/`maxAge` now clamp to the same defaults (4096 events / 5m) instead of disabling the bound. Added tests confirming eviction still happens under `0`/negative input rather than growing unbounded.

Also merged in your `release-2.0` update from earlier - branch is current.

onkar717:
cc @spron-in

spron-in (approved):

PR #2596, opened 2026-07-22
feat(ui): convert memory resource units from GB to Gi

Nachiket-Roy:
### Description
This change updates all database and proxy memory resource fields in the UI from `GB`/`G` (decimal Gigabytes) to `Gi` (binary Gibibytes) for limits, requests, and spec serialization.

- Updates the serialization logic to write memory specifications with the `Gi` suffix.
- Changes UI labels, end-suffixes, and tooltips to display `Gi` instead of `GB`.
- Aligns resource matching and synchronization functions to use `Gi` as the target unit.
- Updates unit test suites to assert the corrected `Gi` formatting and values.

### DIFF
--- a/ui/apps/everest/src/components/cluster-form/resources/constants.ts
+++ b/ui/apps/everest/src/components/cluster-form/resources/constants.ts
@@ -64,9 +64,9 @@ export const matchFieldsValueToResourceSize = (
     return ResourceSize.custom;
   }
 
-  const memory = memoryParser(extractedMemory.toString(), 'G');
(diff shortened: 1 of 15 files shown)


spron-in:
Any reason for this change? @Nachiket-Roy 
I feel that GB makes more sense in the UI. Gi might be confusing for non-k8s folks.

Nachiket-Roy:
@spron-in it was mainly for accuracy but yes could cause confusion to non-k8s ppls

PR #2599, opened 2026-07-23
everestctl: add backup create subcommand

### LINKED ISSUE #2578: Add `everestctl backup create` subcommand

Backups can be triggered through the API and the UI, but there is no way to take an on-demand backup of an instance from the terminal. Operators scripting maintenance workflows (e.g. backup before an upgrade) have to hand-craft Backup CRs with kubectl or call the API with curl and a JWT token.

Add an `everestctl backup create` subcommand that creates an on-demand Backup for an instance via the Everest API. Required flags: `--instance` and `--namespace` to identify the target instance, `--class` and `--storage` to select the BackupClass and backup storage (mapping to `spec.backupClassName` / `spec.storageName` on the Backup CR — the `backup` prefix is omitted from the flags since the command group already provides that context). Optional flags: `--name` (generated from the instance name + timestamp when omitted) and `--deletion-policy`. The command should print the created backup's name and follow existing CLI conventions: `--json`/`--verbose` output modes, `--cluster`, and `--context` flags. Consider a `--wait` flag to block until the backup reaches a terminal state.

```
# Take an on-demand backup
everestctl backup create --instance my-mongo --namespace everest \
  --class percona-backup-mongodb --storage my-s3

# With explicit name and wait for completion
everestctl backup create --instance my-mongo --namespace everest \
  --class percona-backup-mongodb --storage my-s3 --name pre-upgrade --wait
```

Applying a Backup CR with `kubectl apply` — works but bypasses Everest API authentication/RBAC and requires knowing the CRD schema by heart.

### Additional Context

Complements the `list` subcommand set (#2551–#2554). This would introduce the `backup` command group, which also hosts `backup list` (separate issue) and could later host `delete`/`restore` subcommands.

VijetaPriya47:
Operators scripting maintenance workflows (e.g. backup before an upgrade) previously had to hand-craft Backup CRs with kubectl or call the API directly with curl and a JWT token, bypassing Everest's own RBAC. This adds `everestctl backup create --instance --namespace --class --storage`, with an optional --name (falls back to metadata.generateName so scripted runs can't collide on a timestamp), --deletion-policy, and --wait/--timeout to block until the backup reaches a terminal state.
Closes #2578


VijetaPriya47:
let's first merge the other two pull requests, then merge this one.

pkg/cli/backup/create.go
```diff
@@ -0,0 +1,207 @@
...
+		Metadata: &map[string]interface{}{
+			"namespace": opts.Namespace,
+		},
+	}
+	if opts.Name != "" {
+		(*backup.Metadata)["name"] = opts.Name
+	} else {
+		(*backup.Metadata)["generateName"] = opts.Instance + "-"
+	}
+	backup.Spec.InstanceName = opts.Instance
```
recharte: @VijetaPriya47 we have recently changed how the resource references work. You can check [this PR](https://github.com/openeverest/openeverest/pull/2570) for how you should be referencing the instance in the backup CR.
TLDR is that you need to set `.spec.instanceRef.name` instead.

recharte (changes requested):

pkg/cli/backup/create.go
recharte: same is true for the class ref and the backup storage ref

VijetaPriya47:
@recharte check this too. If this gets merged today, we can work on new tickets, and we can discuss new tickets in the sync call.

recharte:
@VijetaPriya47 In the last round of review of #2600 I left [some suggestions](https://github.com/openeverest/openeverest/pull/2600#pullrequestreview-4777292435) that should also be applied to this PR.

commands/instance/create.go
```diff
@@ -101,13 +130,64 @@ func createRun(cmd *cobra.Command, _ []string) { //nolint:revive
...
+			fmt.Errorf("timed out after %s waiting for instance %q to become ready; it is still provisioning", opts.Timeout, opts.Name),
+			logger.GetLogger(), pretty,
+		)
+		return exitTimeout
+	default:
+		// *wait.FailedError (Failed phase / deleted mid-wait) and everything else.
+		output.PrintError(err, logger.GetLogger(), pretty)
+		return exitFailed
 	}
 }
```
recharte: I believe that this and the exit consts defined in the beginning of this file should be extracted to a common helper that can be used by both instance and backup create commands. Thus, avoiding the drift I just mentioned in the previous comment.

recharte (approved):

PR #2600, opened 2026-07-23
Add --wait flag to instance create

### LINKED ISSUE #2580: Add `--wait` flag to `everestctl instance create`

`everestctl instance create` returns immediately after the Instance is accepted by the API, while provisioning continues in the background. Scripts that need a usable database (CI pipelines, demo setups, integration tests) must implement their own polling loop around `instance status --watch`, which is awkward to script since `--watch` runs until Ctrl-C rather than exiting on readiness.

Add a `--wait` flag to `everestctl instance create` that blocks until the instance reaches the Running phase, then exits 0. If the instance enters a failed state, exit non-zero with the failure message. Add a companion `--timeout` flag (e.g. default 10m, only valid with `--wait`) that bounds the wait and exits non-zero on expiry. While waiting, pretty mode should show progress (phase + per-component ready counts, reusing the `instance status` rendering); `--json` mode should stay quiet and print the final instance state as a single JSON object. Polling can reuse the existing status machinery from `instance status --watch`.

```
# Create and block until the instance is ready
everestctl instance create --name my-mongo --namespace everest \
  --provider percona-server-mongodb --wait

# Bounded wait for CI
everestctl instance create --name my-mongo --namespace everest \
  --provider percona-server-mongodb --wait --timeout 15m

# Scripting: create, wait, then grab connection details
everestctl instance create --name my-mongo --namespace everest \
  --provider percona-server-mongodb --wait --json | jq '.status.phase'
```

Wrapping `everestctl instance status --watch` in a shell loop with grep/timeout — fragile, duplicated in every script, and `--watch` has no exit-on-ready semantics.

### Additional Context

The same `--wait`/`--timeout` pattern is proposed for `backup create` (#2578), so the polling helper could be shared. Analogous to `kubectl wait` and `helm --wait` semantics.

VijetaPriya47:
Scripting a blocking create today means wrapping `instance status --watch` in a shell loop, which is awkward because --watch never exits on its own. This adds --wait so a create can block until the instance is actually ready.

With --wait the command polls until the instance reaches the Ready phase and exits 0, or exits non-zero if it hits the Failed phase, gets deleted, or the --timeout elapses. Exit codes are distinct so CI can tell the cases apart: 1 for a failed instance, 124 for a timeout (matching timeout(1)), and 130 when someone Ctrl-C's out. Cancelling only stops the wait -- the instance keeps provisioning server-side, and we say so on the way out.

Pretty mode streams progress a line at a time and prints the final status; --json stays quiet and emits a single instance object at the end. While here, `create --json` without --wait now emits the created object too, so the JSON contract no longer depends on whether --wait was passed.

The polling itself lives in a small reusable pkg/cli/wait helper (poll func + condition + timeout/exit-code handling) so backup create can share it later rather than growing a second copy that drifts.
closes #2580

### DIFF
--- a/commands/instance/create.go
+++ b/commands/instance/create.go
@@ -133,30 +133,34 @@ func createRun(cmd *cobra.Command, _ []string) { //nolint:revive
 		os.Exit(exitFailed)
 	}
 
-	if err := validateWaitFlags(cmd); err != nil {
+	if err := validateWaitFlags(cmd, createOpts); err != nil {
 		output.PrintError(err, logger.GetLogger(), createCfg.Pretty)
 		os.Exit(exitFailed)
 	}
 
 	ctx := cmd.Context()
+	stop := func() {}
 	if createOpts.Wait {
 		// Ctrl-C cancels only the wait; the instance keeps provisioning.
-		var stop context.CancelFunc
-		ctx, stop = signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
-		defer stop()
+		var cancel context.CancelFunc
+		ctx, cancel = signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
+		stop = cancel
 	}
 
 	ic := instancecli.NewInstanceCreator(*createCfg, logger.GetLogger())
-	os.Exit(createExitCode(ic.Run(ctx, *createOpts, cfgPath)))
+	code := createExitCode(ic.Run(ctx, *createOpts, cfgPath), createOpts, createCfg.Pretty)
+	// Release the signal handler before os.Exit (which skips defers).
+	stop()
+	os.Exit(code)
 }
 
 // validateWaitFlags rejects --timeout without --wait and non-positive timeouts.
 // (cobra's MarkFlagsRequiredTogether would wrongly reject a bare --wait.)
-func validateWaitFlags(cmd *cobra.Command) error {
-	if cmd.Flags().Changed(cli.FlagInstanceTimeout) && !createOpts.Wait {
+func validateWaitFlags(cmd *cobra.Command, opts *instancecli.CreateOptions) error {
+	if cmd.Flags().Changed(cli.FlagInstanceTimeout) && !opts.Wait {
 		return errors.New("--timeout is only valid together with --wait")
 	}
-	if createOpts.Wait && createOpts.Timeout <= 0 {
+	if opts.Wait && opts.Timeout <= 0 {
 		return errors.New("--timeout must be a positive duration (use e.g. --timeout 10m)")
 	}
 	return nil
@@ -164,25 +168,25 @@ func validateWaitFlags(cmd *cobra.Command) error {
 
 // createExitCode maps the create/wait result to an exit code and prints the
 // matching message. See the exit* constants for the contract.
-func createExitCode(err error) int {
+func createExitCode(err error, opts *instancecli.CreateOptions, pretty bool) int {
 	switch {
 	case err == nil:
 		return 0
 	case errors.Is(err, context.Canceled):
 		_, _ = fmt.Fprint(os.Stderr, output.Warn(
 			"wait cancelled; instance %q is still provisioning — check with 'everestctl instance status'",
-			createOpts.Name,
+			opts.Name,
 		))
 		return exitCanceled
 	case errors.Is(err, wait.ErrTimeout):
 		output.PrintError(
-			fmt.Errorf("timed out after %s waiting for instance %q to become ready; it is still provisioning", createOpts.Timeout, createOpts.Name),
-			logger.GetLogger(), createCfg.Pretty,
+			fmt.Errorf("timed out after %s waiting for instance %q to become ready; it is still provisioning", opts.Timeout, opts.Name),
+			logger.GetLogger(), pretty,
 		)
 		return exitTimeout
 	default:
 		// *wait.FailedError (Failed phase / deleted mid-wait) and everything else.
-		output.PrintError(err, logger.GetLogger(), createCfg.Pretty)
+		output.PrintError(err, logger.GetLogger(), pretty)
 		return exitFailed
 	}
 }
--- a/commands/instance/create.go
+++ b/commands/instance/create.go
@@ -17,15 +17,35 @@
 package instance
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 
 	"github.com/openeverest/openeverest/v2/pkg/cli"
 	"github.com/openeverest/openeverest/v2/pkg/cli/config"
 	instancecli "github.com/openeverest/openeverest/v2/pkg/cli/instance"
+	"github.com/openeverest/openeverest/v2/pkg/cli/wait"
 	"github.com/openeverest/openeverest/v2/pkg/logger"
 	"github.com/openeverest/openeverest/v2/pkg/output"
+)
+
+// Exit codes for `instance create`, chosen so CI scripts can tell the failure
+// modes apart without scraping stderr:
+//
+//	0   success
+//	1   the instance reached the Failed phase (or any other error)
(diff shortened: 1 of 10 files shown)


VijetaPriya47:
@recharte could you check it out

VijetaPriya47:

recharte:
@VijetaPriya47 please fix the linting errors

VijetaPriya47:
@recharte  i think there is some helm charts issues

recharte:
> @recharte i think there is some helm charts issues

Just fixed that, if you rebase/merge `release-2.0` it should be fixed now.

recharte:
Hi @VijetaPriya47 I'm sorry, there was a bug with the latest helm chart update which [has now been fixed](https://github.com/openeverest/openeverest/pull/2607). Please pull the latest changes from `release-2.0`.

VijetaPriya47:
@recharte hey, could you checkout the error, if you get time?

recharte:
> @recharte hey, could you checkout the error, if you get time?

Hey @VijetaPriya47 the error seemed to be just a temporary network issue, re-running the job worked. I'll submit a proper review later today.

pkg/cli/instance/create.go
```diff
@@ -168,10 +172,87 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath
...
+	})
+
+	final := latest
+	if final == nil {
+		final = created
+	}
+
+	if err != nil {
+		return err
+	}
```
recharte: Just a style nitpick but it's more idiomatic to check the error right next to the call. Moreover, `final` is only used on the success path, so computing it before the error check just pads the distance.

```suggestion
	if err := wait.Until(ctx, poll, instanceCondition, wait.Options{
		Timeout:  opts.Timeout,
		OnUpdate: onUpdate,
	}); err != nil {
		return err
	}

	final := latest
	if final == nil {
		final = created
	}
```

pkg/cli/instance/wait.go
```diff
@@ -0,0 +1,122 @@
...
+		switch resp.StatusCode() {
+		case http.StatusOK:
+			if resp.JSON200 == nil {
+				return nil, fmt.Errorf("empty response body fetching instance %q", name)
+			}
+			return resp.JSON200, nil
+		case http.StatusNotFound:
+			return nil, fmt.Errorf("instance %q was deleted while waiting", name)
+		default:
+			return nil, fmt.Errorf("unexpected response fetching instance %q: %s", name, resp.Status())
```
recharte: This makes any error a terminal state. The `status --watch` command handles this in a better way, it retries transient fetch failures and unexpected statuses and only treats 404 as terminal.

commands/instance/create.go
```diff
@@ -101,13 +130,64 @@ func createRun(cmd *cobra.Command, _ []string) { //nolint:revive
...
+// matching message. See the exit* constants for the contract.
+func createExitCode(err error, opts *instancecli.CreateOptions, pretty bool) int {
+	switch {
+	case err == nil:
+		return 0
+	case errors.Is(err, context.Canceled):
+		_, _ = fmt.Fprint(os.Stderr, output.Warn(
+			"wait cancelled; instance %q is still provisioning — check with 'everestctl instance status'",
+			opts.Name,
+		))
```
recharte: This path doesn't respect `--json` mode. Every other outcome goes through `output.PrintError(err, logger, pretty)`, which pretty-prints in pretty mode and falls back to the structured logger in `--json` mode. The `context.Canceled` case bypasses that and always emits the emoji-styled `output.Warn` to stderr, even when the user asked for `--json`:

I suggest we add a `PrintWarn` sibling to `PrintError` in `pkg/output/command.go` so the warn path follows the same contract:

```go
// PrintWarn formats and prints a warning, respecting pretty/JSON mode.
func PrintWarn(msg string, l *zap.SugaredLogger, prettyPrint bool) {
    if prettyPrint {
        _, _ = fmt.Fprint(os.Stderr, Warn("%s", msg))
    } else {
        l.Warn(msg)
    }
}
```

and then use it here like

```go
case errors.Is(err, context.Canceled):
    output.PrintWarn(
        fmt.Sprintf("wait cancelled; instance %q is still provisioning — check with 'everestctl instance status'", opts.Name),
        logger.GetLogger(), pretty,
    )
    return exitCanceled
```

pkg/cli/instance/create.go
```diff
@@ -168,10 +172,87 @@ func (ic *InstanceCreator) Run(ctx context.Context, opts CreateOptions, cfgPath
 	}
 
 	ic.l.Infof("created instance %q in namespace %q", opts.Name, opts.Namespace)
+
+	// The API returns the accepted Instance as JSON201 (or JSON200).
+	created := resp.JSON201
+	if created == nil {
+		created = resp.JSON200
+	}
```
recharte, line 180: If the server returns 200/201 but the body doesn't deserialize (`JSON201`/`JSON200` both nil) we'll have an empty stdout. Consider returning an error (or re-fetching the instance) when both are nil.

pkg/cli/wait/wait.go
```diff
@@ -0,0 +1,145 @@
...
+type PollFunc[T any] func(ctx context.Context) (T, error)
+
+// Options tunes the waiter.
+type Options struct {
+	// Interval between polls. Internal knob (no CLI flag); the zero value means
+	// unset and falls back to DefaultInterval, so a bad value can't be typed.
+	Interval time.Duration
+	// Timeout bounds the total wait and must be positive — the command layer
+	// rejects zero/negative.
+	Timeout time.Duration
```
recharte: Setting `Timeout: 0` means "time out immediately", not "no timeout".

`Interval` is defensive about its zero value (falls back to `DefaultInterval`), but `Timeout` isn't: an unset `Timeout` goes straight into `context.WithTimeout(ctx, 0)`, which is already expired — `Until` polls once and returns `ErrTimeout` right away.

The instance-create path is safe because the `--timeout` flag defaults to 10m and `validateWaitFlags` rejects non-positive values. But since this package is meant to be reused by `backup create --wait` later, the next caller writing `wait.Options{}` gets instant timeouts unless they remember to replicate the same flag validation.

Can we make this self-defending? Any of these would work:

1. Treat `Timeout <= 0` as "no timeout" (only wrap in `context.WithTimeout` when positive) — matches kubectl's `--timeout=0` semantics and how `Interval` already behaves.
2. Fall back to a `DefaultTimeout`, mirroring `DefaultInterval`.
3. Have `Until` return an explicit error for non-positive timeouts so misuse fails loudly instead of looking like a real timeout.

pkg/cli/instance/wait.go
recharte: Also, a 401 mid-wait loses the actionable hint that `status --watch` gives.

If the session token expires during a long `--wait`, the poll lands in this `default:` branch and the user gets:

```
unexpected response fetching instance "my-db": 401 Unauthorized
```

`status --watch` handles the same situation explicitly with:

```go
case http.StatusUnauthorized:
    return fmt.Errorf("server rejected credentials — run 'everestctl auth login' again")
```

I think we should mirror this for consistency.

The exit code stays 1 either way; this is purely about giving the user the fix instead of a bare status code.

recharte (changes requested):
Overall great job @VijetaPriya47
Leaving a few suggestions to polish it a little bit.

pkg/cli/wait/wait.go
VijetaPriya47: Went with option 1.  Until now only wraps in context.WithTimeout when Timeout > 0 and a zero/negative timeout polls against the parent ctx directly, so it means "no timeout" instead of "expire immediately"

commands/instance/create.go
VijetaPriya47: Done. Added PrintWarn to pkg/output/command.go with the same pretty/JSON contract as PrintError, and switched the context.Canceled case to use it. The cancel warning now goes through the structured logger in --json mode instead of always emitting the styled line to stderr.

VijetaPriya47:
@recharte could you review again? resolved all the comments.

recharte (approved):
Great job 💪 well done

PR #2611, opened 2026-07-24
Fix provider-runtime reporting readiness before manager is ready

### LINKED ISSUE #2598: provider-runtime reports readiness before the controller manager is ready
### What is the issue?

In `provider-runtime/reconciler/provider.go` (lines 289–293 and 308–312 at `6fec2595`), both `Start()` and `StartWithSignalHandler()` call `r.server.SetReady(true)` before the blocking `r.manager.Start(ctx)` call. A matching comment in `provider-runtime/server/server.go:121` ("SetClient ... called by reconciler after manager is ready") is likewise contradicted.

The surrounding comments indicate the server should only be marked ready once the manager is ready, but because `manager.Start` blocks, the current execution order always marks the server ready first.

At runtime this allows `/readyz` to return 200 OK before the controller manager has started, and requests to `/validate` during this window may fail because the cache-backed client is not yet available. In the captured output below, the infrastructure error is returned as a validation denial (`"allowed": false`) to the user.

I reproduced this on the current release-2.0 branch using a provider configured with an unreachable API server to keep the manager from becoming ready.

Note: the unreachable API server makes the not-ready state deterministic for the reproduction. In a real cluster the equivalent window is pod startup (until cache sync), so real-world exposure is rollout-sized rather than permanent — the ordering itself is environment-independent.

### Steps to Reproduce

1. Build a minimal provider using provider-runtime on the current release-2.0 branch.
2. Configure KUBECONFIG to point to an unreachable API server (https://127.0.0.1:59999) so the manager never becomes ready.
3. Start the provider.
4. Poll GET /readyz.
5. Send a request to POST /validate.

### Relevant Logs / Output

```
/readyz returned 200 after 52.51ms (manager cannot become ready in this setup)
Start() was still blocked in manager.Start() at that moment.

/validate HTTP 200 after 948µs
Response:
{
  "allowed": false,
  "message": "fetching provider for version validation: failed to get server groups: Get \"https://127.0.0.1:59999/api\": dial tcp 127.0.0.1:59999: connect: connection refused"
}
```

### OpenEverest Version

release-2.0 @ 6fec2595

### Installation Method

Manifests / Manual

### Kubernetes Distribution

Other (please specify below)

### Environment Details

Built from source; no cluster required for the reproduction (reproduced in-process).

### Possible Solution

- Gate readiness the controller-runtime way: register a manager.Runnable (or wait on mgr.Elected() / cache sync) that calls SetReady(true) once the manager is up, and reset it on shutdown. This would also allow deduplicating Start() and StartWithSignalHandler(), which currently repeat the same block.
- Separately worth deciding: whether handleValidation should distinguish infrastructure errors (5xx) from validation denials (allowed: false) — in the captured output a connection refused becomes a denial message shown to the user. I can split this into its own issue if preferred.

### Additional Context

Happy to implement the fix with tests if this is confirmed as a bug.

Anuragkumar-687:
## What

Gates the provider-runtime validation server's readiness on the manager's caches being synced, instead of marking it ready before `manager.Start()`.

Fixes #2598.

## Why

`Start()` / `StartWithSignalHandler()` called `server.SetReady(true)` before the blocking `manager.Start(ctx)`. As a result `/readyz` returned 200 before the controller manager (and its cache-backed client) were up. A probe or automation that waits on `/readyz` and then creates an Instance could reach `/validate` before the client could serve reads, and the error surfaced as a validation denial (`allowed: false`) rather than being kept away from a not-ready endpoint.

## How

- Extracted the server startup into a single `startServer` helper, removing the duplicated block across `Start` and `StartWithSignalHandler`.
- Readiness is now set from a `manager.Runnable`. It waits for the informers the validation path reads — `Instance` and `Provider` — to sync before calling `SetReady(true)`, and clears it again on shutdown.
- The Runnable is wrapped in `leaderElectionFree` so it runs on every replica rather than only the elected leader. Readiness is a per-pod property: a plain `manager.RunnableFunc` lands in controller-runtime's leader-election group, which would leave every non-leader replica permanently not-ready — pulled out of the Service that fronts `/validate`, and unable to complete a rollout — the day a provider enables leader election.

Two notes on the approach:

- A plain `mgr.GetCache().WaitForCacheSync(ctx)` is not enough: at the time the manager starts its runnables no informer has been requested yet, so it returns `true` vacuously and readiness is not actually gated. Waiting on specific informers avoids that.
- `GetInformer` returns an error while the API server is unreachable, so it is retried with a short backoff; otherwise a transient blip at startup would leave the server not-ready until the process restarts.

Out of scope: `handleValidation` currently returns infrastructure errors as `allowed: false`. That is a separate concern and can be addressed in its own issue.

## Testing

- `provider-runtime/server/server_test.go` — `/readyz` reflects `SetReady`, `/healthz` is independent of readiness.
- `provider-runtime/reconciler/readiness_test.go` — regression test: with an unreachable API server the caches never sync, so `/readyz` never reports ready while `/healthz` does. Fails against the previous behavior and passes with this change.

`go test ./provider-runtime/...` passes, including `-race`. Copyright headers verified with `make copyright-check`.

### DIFF
--- a/provider-runtime/reconciler/provider.go
+++ b/provider-runtime/reconciler/provider.go
@@ -18,11 +18,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	toolscache "k8s.io/client-go/tools/cache"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -30,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -278,37 +281,18 @@ func (r *ProviderReconciler) setupServer(p providerAdapter) error {
 
 // Start starts the reconciler and server (blocking).
 func (r *ProviderReconciler) Start(ctx context.Context) error {
-	// Start server if configured
-	if r.server != nil {
-		r.server.SetClient(r.Client)
-		go func() {
-			if err := r.server.Start(ctx); err != nil {
-				log.FromContext(ctx).Error(err, "Server error")
-			}
-		}()
-		// Mark server as ready once manager is ready
-		r.server.SetReady(true)
+	if err := r.startServer(ctx); err != nil {
+		return err
 	}
-
 	return r.manager.Start(ctx)
 }
 
 // StartWithSignalHandler starts the reconciler and server with OS signal handling.
 func (r *ProviderReconciler) StartWithSignalHandler() error {
 	ctx := ctrl.SetupSignalHandler()
-
-	// Start server if configured
-	if r.server != nil {
-		r.server.SetClient(r.Client)
-		go func() {
-			if err := r.server.Start(ctx); err != nil {
-				log.FromContext(ctx).Error(err, "Server error")
-			}
-		}()
-		// Mark server as ready once manager is ready
-		r.server.SetReady(true)
+	if err := r.startServer(ctx); err != nil {
+		return err
 	}
-
 	return r.manager.Start(ctx)
 }
 
@@ -540,6 +524,73 @@ func (r *ProviderReconciler) Reconcile(ctx context.Context, req reconcile.Reques
 
 	logger.Info("Reconciliation complete", "phase", in.Status.Phase)
 	return reconcile.Result{}, nil
+}
+
+// leaderElectionFree marks a Runnable as one that must run on every replica,
+// not only the elected leader: readiness is a per-pod property.
+type leaderElectionFree struct{ manager.Runnable }
+
+func (leaderElectionFree) NeedLeaderElection() bool { return false }
+
+// startServer wires the validation server into the manager lifecycle. It is a
+// no-op when no server is configured.
+//
+// Readiness is gated on the manager's caches: the server is marked ready from a
+// manager Runnable only after the informers the validation path reads (Instance
+// and Provider) have synced, so /readyz reports ready only once the cache-backed
+// client is usable. This keeps probes from routing validation traffic to the
+// server before it can serve it. Readiness is cleared again on shutdown.
+//
+// The Runnable is wrapped in leaderElectionFree so it runs on every replica, not
+// only the elected leader. Must be called before manager.Start.
+func (r *ProviderReconciler) startServer(ctx context.Context) error {
+	if r.server == nil {
+		return nil
+	}
+
+	r.server.SetClient(r.Client)
+	go func() {
+		if err := r.server.Start(ctx); err != nil {
+			log.FromContext(ctx).Error(err, "Server error")
+		}
+	}()
+
+	return r.manager.Add(leaderElectionFree{manager.RunnableFunc(func(runnableCtx context.Context) error {
+		// Report ready only once the informers the validation path reads have
+		// synced, so the cache-backed client is usable before /readyz turns
+		// green. Stay not-ready if the context is cancelled first (shutdown).
+		for _, obj := range []client.Object{&v1alpha1.Instance{}, &v1alpha1.Provider{}} {
+			if !r.waitForCacheSync(runnableCtx, obj) {
+				return nil
+			}
+		}
+		r.server.SetReady(true)
+		<-runnableCtx.Done()
+		r.server.SetReady(false)
+		return nil
+	})})
+}
+
+// waitForCacheSync blocks until the informer for obj has been built and synced,
+// returning true on success or false if ctx is cancelled first. GetInformer
+// returns an error while the API server is unreachable, so it is retried with a
+// short backoff to tolerate a transient blip at startup rather than latching
+// not-ready until the process restarts.
+func (r *ProviderReconciler) waitForCacheSync(ctx context.Context, obj client.Object) bool {
+	for {
+		informer, err := r.manager.GetCache().GetInformer(ctx, obj)
+		if err == nil {
+			return toolscache.WaitForCacheSync(ctx.Done(), informer.HasSynced)
+		}
+		// The API server can be briefly unreachable at startup; retry until it
+		// is reachable (controller-runtime logs the underlying failure) or the
+		// context is cancelled.
+		select {
+		case <-ctx.Done():
+			return false
+		case <-time.After(time.Second):
+		}
+	}
 }
 
 func (r *ProviderReconciler) handleDeletion(
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/oapi-codegen/echo-middleware v1.1.0
 	github.com/oapi-codegen/runtime v1.6.0
-	github.com/openeverest/helm-charts/charts/everest v0.0.0-20260802124859-30258bdcf192
+	github.com/openeverest/helm-charts/charts/everest v0.0.0-20260808040649-ebf9682bfe5a
 	github.com/operator-framework/api v0.45.0
 	github.com/percona/everest-operator v0.6.0-dev1.0.20260429065444-70caa52c384a
 	github.com/rodaine/table v1.3.1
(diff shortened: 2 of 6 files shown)


Anuragkumar-687:
Hi @recharte! Whenever you have a chance, could you please take a look at this PR? It addresses the readiness race described in #2598, adds regression tests, and is currently waiting review. I'd really appreciate your feedback. Thanks!

recharte:
Hi @Anuragkumar-687! Sorry for the delay on this. We'll get to your PR as soon as possible, so please bear with us. Every time we apply for the LFX Mentorship program, we get a huge surge of new contributor issues and PRs. It takes us a bit of time to clear the queue while also handling critical v2 architecture work. Thanks for understanding!

Anuragkumar-687:
Thanks for the update, @recharte! No worries at all, I completely understand. I know it's a busy period with the LFX Mentorship applications and ongoing v2 work. Thanks for letting me know, and please take your time. I appreciate it!

provider-runtime/reconciler/provider.go
```diff
@@ -278,38 +281,75 @@ func (r *ProviderReconciler) setupServer(p providerAdapter) error {
...
-		r.server.SetReady(true)
+// startServer wires the validation server into the manager lifecycle. It is a
+// no-op when no server is configured.
+//
+// Readiness is gated on the manager's caches: the server is marked ready from a
+// manager Runnable only after WaitForCacheSync succeeds, so /readyz reports
+// ready only once the cache-backed client is usable. This keeps probes from
+// routing validation traffic to the server before it can serve it. Readiness is
+// cleared again on shutdown. Must be called before manager.Start.
+func (r *ProviderReconciler) startServer(ctx context.Context) error {
```
recharte: Good catch on this. A bare `mgr.GetCache().WaitForCacheSync(ctx)` really would have been a no-op: the controller's sources don't request their informers until the controller itself starts, so at the point the manager runs its runnables the cache has nothing to wait on and returns true vacuously. That's the kind of fix that reads as correct in review and silently does nothing in production. Folding the duplicated block into `startServer` while you were in there was the right call too.

provider-runtime/reconciler/provider.go
```diff
@@ -278,38 +281,75 @@ func (r *ProviderReconciler) setupServer(p providerAdapter) error {
...
 
-	return r.manager.Start(ctx)
+	r.server.SetClient(r.Client)
+	go func() {
+		if err := r.server.Start(ctx); err != nil {
+			log.FromContext(ctx).Error(err, "Server error")
+		}
+	}()
+
+	return r.manager.Add(manager.RunnableFunc(func(runnableCtx context.Context) error {
```
recharte: `manager.RunnableFunc` doesn't implement `LeaderElectionRunnable`, so this lands in the leader-election group: in controller-runtime v0.24.0, `runnables.Add` falls through to `default:` and calls `r.LeaderElection.Add(fn, nil)`. It works today only because `newReconciler` never sets `LeaderElection` on the manager options, so the manager treats "no election configured" as "elected" and starts the group unconditionally.

The day any provider turns leader election on, every non-leader replica would sit at `/readyz` 503 for its whole life — pulled out of the Service that fronts `/validate`, and a rollout could never complete. Readiness is a per-pod property and shouldn't depend on winning an election; it's the same reason controller-runtime deliberately starts its `HTTPServers` and `Webhooks` groups before leader election, with a warning comment saying exactly that.

`RunnableFunc` can't declare it, but a three-line wrapper can, and it routes the runnable to the `Others` group — still after the caches, still before leader election:

```go
// leaderElectionFree marks a Runnable as one that must run on every replica,
// not only the elected leader: readiness is a per-pod property.
type leaderElectionFree struct{ manager.Runnable }

func (leaderElectionFree) NeedLeaderElection() bool { return false }
```

Then `r.manager.Add(leaderElectionFree{manager.RunnableFunc(...)})`.

provider-runtime/reconciler/provider.go
```diff
@@ -278,38 +281,75 @@ func (r *ProviderReconciler) setupServer(p providerAdapter) error {
...
+}
+
+// waitForInstanceCacheSync blocks until the Instance informer has been built
+// and synced, returning true on success or false if ctx is cancelled first.
+// GetInformer returns an error while the API server is unreachable, so it is
+// retried with a short backoff to tolerate a transient blip at startup rather
+// than latching not-ready until the process restarts.
+func (r *ProviderReconciler) waitForInstanceCacheSync(ctx context.Context) bool {
+	for {
+		informer, err := r.manager.GetCache().GetInformer(ctx, &v1alpha1.Instance{})
```
recharte: The informer this waits on isn't the one the validation path reads first. `handleValidation` → `validateVersionBundle` starts with `c.Get(ctx, ..., &v1alpha1.Provider{})` — that's the exact call in the traceback on #2598 ("fetching provider for version validation: failed to get server groups"). The `Instance` informer is the one the controller builds, not the one `/validate` needs.

The gap is narrow in practice — once the `Instance` informer has synced the API server is plainly reachable, and the `Provider` informer will sync a moment later — but a cache-backed `Get` on a type with no started informer blocks while that informer starts and syncs, so the first `/validate` after the pod goes Ready can still stall. And `startServer`'s doc comment ("so `/readyz` reports ready only once the cache-backed client is usable") is then claiming more than the code delivers.

Could you warm both? The loop already generalises:

```go
for _, obj := range []client.Object{&v1alpha1.Instance{}, &v1alpha1.Provider{}} {
	if !r.waitForCacheSync(ctx, obj) {
		return false
	}
}
```

recharte (changes requested):
Thanks for this — the diagnosis on #2598 was right and the fix is the right shape. Direction is settled there, so this is purely about the code.

The `Check` job is red on 8 lint issues from the new lines. Could you clear these?

```
provider-runtime/reconciler/provider.go:307:1: unexported method "startServer" for struct "ProviderReconciler" should be placed after the exported method "Reconcile" (funcorder)
provider-runtime/reconciler/provider.go:338:1: unexported method "waitForInstanceCacheSync" for struct "ProviderReconciler" should be placed after the exported method "Reconcile" (funcorder)
provider-runtime/reconciler/readiness_test.go:67:22: net.Listen must not be called. use (*net.ListenConfig).Listen (noctx)
provider-runtime/reconciler/readiness_test.go:70:9: type assertion must be checked (forcetypeassert)
provider-runtime/reconciler/readiness_test.go:99:17: testingcontext: context.WithCancel can be modernized using t.Context (modernize)
provider-runtime/server/server_test.go:32:22: net.Listen must not be called. use (*net.ListenConfig).Listen (noctx)
provider-runtime/server/server_test.go:35:9: type assertion must be checked (forcetypeassert)
provider-runtime/server/server_test.go:57:17: testingcontext: context.WithCancel can be modernized using t.Context (modernize)
```

The two `funcorder` ones just want `startServer` and `waitForInstanceCacheSync` moved below `Reconcile`; the other six are the duplicated `freePort`/`statusCode` helpers and the context setup in the two new test files. You can reproduce the job locally with `go tool golangci-lint run --new-from-rev=$(git merge-base origin/release-2.0 HEAD) ./provider-runtime/...`.

The two red `Controller Integration Tests` jobs are not yours — they fail at job setup on an action-pinning policy that `release-2.0` has since fixed, so a rebase onto current `release-2.0` clears both and gets the branch out of its BEHIND state at the same time. Everything else is green.

Two design points inline, both worth your view.

Anuragkumar-687:
Thanks for the detailed review @recharte! I addressed all of the requested changes:

- Fixed all reported `golangci-lint` issues.
- Wrapped the readiness runnable so it runs independently of leader election.
- Generalized the cache sync helper and now wait for both the `Instance` and `Provider` informers before marking the server ready.
- Updated the comments to reflect the implementation.
- Rebased onto the latest `release-2.0`.

I also re-ran `go test ./provider-runtime/... -race` and `golangci-lint` (`--new-from-rev`), and both pass cleanly.

Thanks again for the thoughtful review!

vickysharma-prog (approved):
Went through @recharte's review point by point against the current head, since re-verifying that is probably the slowest part of picking this back up.

All five landed as asked:

- `leaderElectionFree` is implemented as sketched (`provider.go:529-533`) and the runnable is wrapped at `:558`, so it routes to `Others` rather than the leader-election group.
- The sync loop now warms both informers, `for _, obj := range []client.Object{&v1alpha1.Instance{}, &v1alpha1.Provider{}}` at `:562`, which covers the `Provider` read that `validateVersionBundle` does first.
- `startServer`'s doc comment now says "the informers the validation path reads (Instance and Provider)", so it no longer claims more than the code delivers.
- `funcorder`: `startServer` at `:546` and `waitForCacheSync` at `:579` are both below `Reconcile` at `:345`.
- The six test lint issues are gone: `(&net.ListenConfig{}).Listen(t.Context(), ...)`, the `l.Addr().(*net.TCPAddr)` assertion is comma-ok, and `t.Context()` replaces the manual `context.WithCancel`. Checks are green and the branch is a single rebased commit.

One thing I would add, and it is the part of this change most likely to be lost later.

`leaderElectionFree` has no test. `readiness_test.go` has `TestServerReadinessGatedOnManager` and `server_test.go` has `TestServerReadinessEndpoint`; neither touches `NeedLeaderElection`. So the failure @recharte described, non-leader replicas sitting at `/readyz` 503 for their whole life, pulled out of the Service fronting `/validate`, rollouts unable to complete, is not pinned by anything. Someone refactoring `startServer` could drop the wrapper, or swap it back for a bare `manager.RunnableFunc`, and the suite would stay green.

It is a one-liner to close. `readiness_test.go` is already `package reconciler`, so the unexported type is reachable, and it needs nothing beyond adding the `manager` import:

```go
func TestReadinessRunnableIsLeaderElectionFree(t *testing.T) {
	var r manager.LeaderElectionRunnable = leaderElectionFree{}
	require.False(t, r.NeedLeaderElection())
}
```

The declared type does as much work as the assertion here: if the interface is ever dropped from the wrapper, that line stops compiling rather than the runnable quietly falling back into the leader-election group.

recharte:
Thanks for the careful pass, @vickysharma-prog — your verification matches mine.

On the wrapper: you're right that nothing pins it, but I don't think the test would either. It asserts that a method returning `false` returns `false`, and if someone later hands a bare `manager.RunnableFunc` to `manager.Add`, the type still exists and still satisfies the interface — so it stays green. Catching that would mean a seam in `startServer` purely for the test, which I've already decided against elsewhere in this review, so I'll leave it.

recharte (approved):
Thanks for working through all the rounds on this one, @Anuragkumar-687 — nice job.

Anuragkumar-687:
Thanks so much for the thorough review and guidance throughout this PR, @recharte. I really appreciate the detailed feedback. It helped improve both the implementation and the tests. Thanks again for your time and for the approval!

PR #2612, opened 2026-07-25
chore: handle ignored errors on defer resp.Body.Close()/Sync() (#2109)

### LINKED ISSUE #2109: Fix ignored errors on defer resp.Body.Close() across multiple files
### Task Description

Multiple files in the codebase suppress errors from resp.Body.Close() using //nolint:errcheck or similar patterns:
- cmd/main.go:41 - defer logger.Sync() error ignored
- pkg/version_service/client.go:93, 154 - HTTP response body close errors suppressed
- pkg/oidc/oidc.go:67 - HTTP response body close errors suppressed
- internal/server/telemetry.go:70 - HTTP response body close errors suppressed
Replace these with proper error handling that logs warnings when close fails.
Current State:
Errors from Close() calls are explicitly ignored with nolint directives. While common in Go, this could mask underlying connection issues and lead to subtle connection leaks in high-throughput scenarios.
Proposed Change:
Wrap defer calls to log errors:
```
defer func() {
    if err := resp.Body.Close(); err != nil {
        logger.Warn("failed to close response body", zap.Error(err))
    }
}()
```
Estimated Complexity: Simple
Breaking Change Risk: No risk
Acceptance Criteria:
- [ ] All linters pass (make check)
- [ ] No //nolint:errcheck on Close() calls remain
- [ ] Unit tests pass (make test)

### Reasoning / Benefit

- Prevents potential connection pool leaks under high load
- Improves observability by logging underlying connection issues
- Follows Go best practices for error handling
- Helps debug production connection problems

### Additional Context

_No response_

jtomaszon:
Closes #2109.

Replaces the `//nolint:errcheck` suppressions on `defer resp.Body.Close()` / `logger.Sync()` with deferred closures that check and log the error at warn level.

**Files**
- `cmd/main.go` — `logger.Sync()` warn-logged; the closure captures `logger`/`l` by value at defer-registration time so it syncs the originally initialized logger. Note: `Sync()` commonly returns `EINVAL` on stdout/stderr/pipes, so the shutdown warning is expected/by-design.
- `internal/server/telemetry.go` — `resp.Body.Close()` via the injected receiver logger (`e.l`).
- `pkg/oidc/oidc.go` — `resp.Body.Close()` via the package logger (`NewProviderConfig` has no receiver/ctx/param logger in scope; documented inline).
- `pkg/version_service/client.go` — `res.Body.Close()` (x2) via the package logger (`versionServiceClient` has no logger field; documented inline).

**Verification**
- `go build ./...`, `go vet`, and `go test` (`./cmd`, `./pkg/oidc`, `./pkg/version_service`, `./internal/server`) all pass.
- `golangci-lint run ./cmd ./pkg/oidc ./pkg/version_service` reports 0 issues.

Scoped intentionally to the errcheck fix — threading a logger through the `oidc`/`version_service` call chains would be a signature-changing refactor across multiple callers and is left out of this chore.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -38,8 +38,16 @@ const (
 
 func main() {
 	logger := logger.MustInitLogger(true, "everest")
-	defer logger.Sync() //nolint:errcheck
 	l := logger.Sugar()
+	// Capture logger/l by value at defer-registration time so this syncs the
+	// originally initialized logger, even though both are reassigned below.
+	// Sync() commonly returns EINVAL on stdout/stderr/pipes; the warn-log on
+	// shutdown here is intentional and expected.
+	defer func(logger *zap.Logger, l *zap.SugaredLogger) {
+		if err := logger.Sync(); err != nil {
+			l.Warnf("failed to sync logger: %s", err)
+		}
+	}(logger, l)
 
 	// This is required because controller-runtime requires a logger
 	// to be set within 30 seconds of the program initialization.
--- a/internal/server/telemetry.go
+++ b/internal/server/telemetry.go
@@ -67,7 +67,11 @@ func (e *EverestServer) report(ctx context.Context, baseURL string, data Telemet
 		e.l.Error(errors.Join(err, errors.New("failed to send telemetry request")))
 		return err
 	}
-	defer resp.Body.Close() //nolint:errcheck
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			e.l.Warnf("failed to close telemetry response body: %s", err)
+		}
+	}()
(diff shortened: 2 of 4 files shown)


recharte:
@jtomaszon please sign your commits in order to make the DCO compliance check pass.

recharte:
Thanks for picking this up, @jtomaszon, and sorry we reached the issue after you'd already written the code.

We've decided not to make this change — the reasoning is on #2109, so let's keep it in one place. The call is about direction, not about your patch: the suppressed errors aren't actionable, and the `logger.Sync()` case would put a spurious warning on every clean shutdown, which your description already anticipates.

PR #2613, opened 2026-07-25
fix(version_service): guard nil Version Service Matrix to prevent panic (#2593)

### LINKED ISSUE #2593: fix(version_service): prevent nil pointer panic when Version Service Matrix is nil
### Summary
In `pkg/version_service/client.go`, `GetSupportedEngineVersions` parses the Version Service response. While it checks `len(response.GetVersions()) == 0`, it calls `.GetMatrix()` and accesses operator version maps directly without validating whether `GetMatrix()` is non-nil.

If the upstream Version Service returns a response where `matrix` is missing or omitted (`{"versions": [{}]}`), calling `.GetPxc()`, `.GetMongod()`, or `.GetPostgresql()` triggers a runtime nil pointer dereference panic. Additionally, unsupported operator strings pass silently without an error.

### Code Location
- File: `pkg/version_service/client.go`
### Expected Behavior
`GetSupportedEngineVersions` should return an explicit error when `GetMatrix()` is `nil` or when an unsupported operator is supplied.

### Fix Details & PR Ready
A fix and unit tests have been prepared on local branch `fix/version-service-nil-matrix-panic` adding:
- `if firstVer == nil || firstVer.GetMatrix() == nil` guard check returning an error.
- - A `default:` branch in `switch operator` returning an unsupported operator error.
- - - Comprehensive unit tests in `pkg/version_service/client_test.go`.

### Steps to Reproduce

1. Call `GetSupportedEngineVersions` against a Version Service returning `{"versions": [{}]}`.
2. 2. Observe panic: `runtime error: invalid memory address or nil pointer dereference`.

### Relevant Logs / Output

_No response_

### OpenEverest Version

main

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

jtomaszon:
Closes #2593.

The `GetSupportedEngineVersions` method accesses `GetMatrix()` maps without validating the matrix is non-nil. When the upstream Version Service returns a response with a missing/omitted matrix, the function silently returns an empty slice with no error — a wrong-success correctness bug.

**Fix:** Extracts a `versionsForOperator` helper that nil-checks both the `OperatorVersion` entry and its `Matrix` before selecting the operator's version map. Returns an explicit error if either is nil or the operator is unrecognized.

**Files:**
- `pkg/version_service/client.go` — new `versionsForOperator` helper + caller updated.
- `pkg/version_service/client_test.go` — new test file with two cases: nil-matrix → error, unsupported operator → error.

**Note:** The generated protobuf getters (`GetMatrix()`, `GetPxc()`, etc.) are nil-safe in the current pinned proto version, so this doesn't reproduce as a literal panic. The real bug is the silent wrong-success — no error signal when data is missing. The guard is added for both correctness and defense-in-depth.

**Verification:** `go build ./...`, `go vet`, `go test ./pkg/version_service/...`, `golangci-lint run ./pkg/version_service` — all pass, 0 issues.

### DIFF
--- a/pkg/version_service/client.go
+++ b/pkg/version_service/client.go
@@ -75,6 +75,27 @@ var defaultUnmarshal = protojson.UnmarshalOptions{
 	DiscardUnknown: true,
 }
 
+// versionsForOperator returns the version map for the given operator from the
+// first operator entry in a Version Service response. It returns an error if
+// the entry has no matrix (rather than panicking on a nil dereference) or if
+// the operator is not recognized.
+func versionsForOperator(firstVer *perconavs.OperatorVersion, operator string) (map[string]*perconavs.Version, error) {
+	if firstVer == nil || firstVer.GetMatrix() == nil {
+		return nil, errors.New("version service response is missing a version matrix")
+	}
+
+	switch operator {
+	case PXCOperatorName:
+		return firstVer.GetMatrix().GetPxc(), nil
+	case PSMDBOperatorName:
+		return firstVer.GetMatrix().GetMongod(), nil
+	case PGOperatorName:
+		return firstVer.GetMatrix().GetPostgresql(), nil
+	default:
+		return nil, fmt.Errorf("unsupported operator %q", operator)
+	}
+}
+
 // GetSupportedEngineVersions returns a list of supported versions for a given operator and version.
 func (c *versionServiceClient) GetSupportedEngineVersions(ctx context.Context, operator, version string) ([]string, error) {
 	p, err := url.Parse(c.url)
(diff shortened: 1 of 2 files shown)


recharte:
Thanks for the patch. I've set out on #2593 why the reported panic isn't reachable — the accessors are generated protobuf getters and are nil-safe by construction, and the caller already resolves the operator through a validated map, so the `default:` arm is dead code.

Your description is right that it "doesn't reproduce as a literal panic". The reframing to "silent wrong-success" doesn't quite land either, though: the empty list reaches `preflightCheckDBEngineVersion`, which returns `no minimum supported versions found for <operator>`. There's no missing error signal, just one raised a layer up.

Worth flagging too that #2618 covers the same ground, and the issue author had said they had a fix prepared.

Not a comment on the code — the call is about whether the change is needed. Thanks for the time you spent on it.

PR #2614, opened 2026-07-25
fix(server): add panic recovery middleware to prevent server crash on handler panic (#2544)

### LINKED ISSUE #2544: [Bug]: Missing Panic Recovery Middleware in API Server

jtomaszon:
Closes #2544.

`NewEverestServer` configures several Echo middlewares (RateLimiter, requestLogger, RemoveTrailingSlash, OAPI validator, JWT, blocklist) but omits `echomiddleware.Recover()`. A handler panic kills the entire server process instead of being caught and returned as a 500.

**Fix:** Adds `echoServer.Use(echomiddleware.Recover())` as the first global middleware after `echo.New()`, before RateLimiter — so it catches panics from all subsequent handlers and middlewares.

**Files:**
- `internal/server/everest.go` — one line added (Recover middleware registration).
- `internal/server/middlewares_test.go` — new test proving a panicking handler returns 500 instead of crashing the process. Standalone Echo instance (avoids Kubernetes/OIDC/session deps).

**Verification:** `go build ./...`, `go vet`, `go test ./internal/server/...`, `golangci-lint run ./internal/server` — all pass, 0 issues.

### DIFF
--- a/internal/server/everest.go
+++ b/internal/server/everest.go
@@ -111,6 +111,7 @@ func NewEverestServer(ctx context.Context, c *config.EverestConfig, l *zap.Sugar
 	}
 
 	echoServer := echo.New()
+	echoServer.Use(echomiddleware.Recover())
 	echoServer.Use(echomiddleware.RateLimiter(echomiddleware.NewRateLimiterMemoryStore(rate.Limit(c.APIRequestsRateLimit))))
 	middleware, store := sessionRateLimiter(c.CreateSessionRateLimit)
 	echoServer.Use(middleware)
--- a/internal/server/middlewares_test.go
+++ b/internal/server/middlewares_test.go
@@ -24,11 +24,13 @@ package server
 
 import (
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"testing"
 	"time"
 
 	"github.com/labstack/echo/v4"
+	echomiddleware "github.com/labstack/echo/v4/middleware"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
@@ -39,6 +41,29 @@ import (
 	everestv1alpha1 "github.com/percona/everest-operator/api/everest/v1alpha1"
 	"github.com/percona/everest/pkg/kubernetes"
 )
+
+// TestPanicRecoveryMiddleware ensures that a handler panic is caught and
+// converted into an HTTP 500 response, instead of crashing the process.
+// See issue #2544. This mirrors the echomiddleware.Recover() registration
+// done in NewEverestServer without needing to stand up a full EverestServer
+// (which requires a live Kubernetes cluster, OIDC provider, session manager, etc).
+func TestPanicRecoveryMiddleware(t *testing.T) {
+	t.Parallel()
+
+	echoServer := echo.New()
+	echoServer.Use(echomiddleware.Recover())
+	echoServer.GET("/panic", func(echo.Context) error {
+		panic("test panic")
+	})
+
+	req := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/panic", nil)
+	rec := httptest.NewRecorder()
+
+	require.NotPanics(t, func() {
+		echoServer.ServeHTTP(rec, req)
+	})
+	assert.Equal(t, http.StatusInternalServerError, rec.Code)
+}
 
 func TestShouldAllowRequestDuringEngineUpgrade(t *testing.T) {
 	lockedAt := time.Now().Format(time.RFC3339)


recharte:
Thanks for this, and genuinely — the review you left on #2545 is the sharpest thing on this issue, and two of its three points are now recorded on #2544 as constraints on the fix.

We're giving #2545 the first go: it came first, it's from the reporter, and its author had already said they'd pick your feedback up. I'd rather not have two implementations of the same change in review at once. That's a queue decision, not a comment on this code.

I've asked over there whether they have time to rework it, with a week or so to reply. I'm leaving this one open in the meantime rather than closing it — if that thread stalls, we'll come back here. So nothing to polish for now; and if you'd like to stay involved sooner, reviewing the reworked version on #2545 is the most useful place to put it.

internal/server/everest.go
```diff
@@ -111,6 +111,7 @@ func NewEverestServer(ctx context.Context, c *config.EverestConfig, l *zap.Sugar
 	}
 
 	echoServer := echo.New()
+	echoServer.Use(echomiddleware.Recover())
```
recharte, line 114: Two problems with this line, and one rewrite fixes both.

**It returns the panic text to the caller.** Echo's `Recover()` passes the raw panic value to `c.Error()`, and `everestErrorHandler`'s `default` branch sets `Message: err.Error()`. So a nil-pointer dereference in any handler answers the client with `{"message":"runtime error: invalid memory address or nil pointer dereference"}` — internal detail, over the wire, to whoever triggered it.

**It's registered with `Use()`, so `Pre()` middleware isn't covered.** `RemoveTrailingSlash` goes in via `Pre()` at line 218, which runs outside every `Use()` middleware — this is your own point 2 from #2545, and it applies here too.

`RecoverWithConfig` with a `LogErrorFunc` closes both, and also gets the stack into zap rather than Echo's default logger (constraint 2 on the issue):

```go
echoServer.Pre(echomiddleware.RecoverWithConfig(echomiddleware.RecoverConfig{
	LogErrorFunc: func(c echo.Context, err error, stack []byte) error {
		l.Errorw("panic recovered while serving request",
			"error", err,
			"method", c.Request().Method,
			"uri", c.Request().RequestURI,
			"stack", string(stack),
		)
		return echo.NewHTTPError(http.StatusInternalServerError)
	},
}))
```

Returning `echo.NewHTTPError(http.StatusInternalServerError)` is what sanitises the response: with no message argument it defaults to `"Internal Server Error"`, and `everestErrorHandler` passes `*echo.HTTPError` through untouched via its first case.

One trap worth knowing: don't add `DisablePrintStack: true` to that config. It reads as "we're logging the stack ourselves, don't also print it", but in Echo v4.15.4 it gates whether the stack is *captured at all* — set it and `LogErrorFunc` receives an empty slice. Leaving it false is correct, because the `LogErrorFunc != nil` branch already stops Echo printing a second copy.

internal/server/middlewares_test.go
```diff
@@ -40,6 +42,29 @@ import (
 	"github.com/percona/everest/pkg/kubernetes"
 )
 
+// TestPanicRecoveryMiddleware ensures that a handler panic is caught and
```
recharte, line 45: I'd drop this test rather than extend it.

As written it builds a fresh `echo.New()`, registers `Recover()` on it, and asserts that Echo's own middleware works — delete line 114 of `everest.go` and it still passes. The name and the `See issue #2544` reference make that worse than a gap rather than better: the next person looking for coverage of this finds a green test and stops looking.

The version that would earn its place asserts the response body is the generic `"Internal Server Error"` rather than the panic text, since that's the real regression risk. But getting there means either standing up a full `EverestServer` — which your comment correctly says isn't possible here — or pulling the registration into a helper purely so a test can reach it, which bends production shape to suit a test. Neither is worth it for a six-line middleware registration.

So: no test rather than a misleading one. It also makes this the one-liner the change actually is.

recharte (changes requested):
Thanks for picking this back up, and sorry for the wait — you stood down when we asked over on #2545 and then heard nothing for three weeks. That thread has stayed quiet, so this is the one we're taking forward, and I've assigned #2544 to you.

Two comments inline. The first is the blocking one: as registered, a handler panic returns its own text to the caller, which is exactly constraint 1 on the issue.

Two things outside the diff:

- **DCO is red.** `c2b82a2` carries a `Co-Authored-By` trailer but no `Signed-off-by`. It's a single commit, so `git rebase HEAD~1 --signoff && git push --force-with-lease` clears it.
- **CI hasn't actually run.** Five workflows are held awaiting maintainer approval because this is your first PR here from a fork, so the near-empty check list isn't a pass. We'll get those released.

recharte:
@jtomaszon — the fix is in: #3200 on `main` (v2) and #3201 on `v1.x`. Closing this one, and it's worth being specific about what of it merged.

Open either diff and you'll recognise it. `RecoverWithConfig` with a `LogErrorFunc` routing the panic and stack into the structured logger, registered with `Pre()` so it sits outside `RemoveTrailingSlash` — that's points 1 and 2 of the review you left on #2545, and it's the entire shape of the landed change. You wrote that review in July, on somebody else's PR, about a codebase you'd hit the same problem in elsewhere. It did more to settle this than anything we wrote.

What got added on top was small: returning a bare `echo.NewHTTPError(http.StatusInternalServerError)` so `everestErrorHandler` doesn't hand the caller the panic text, and knowing not to reach for `DisablePrintStack: true` (in echo v4.15.4 it gates whether the stack is captured at all, so `LogErrorFunc` would receive an empty slice).

Your point 3 — the test — is the one I talked myself out of, and I may be wrong. The reasoning is in the body of #3200; if you think there's a version that asserts something real without either standing up a full `EverestServer` or bending the registration into a helper to make it reachable, I'd like to read it.

On the process: I asked you to hold in August while #2545 had its go, then didn't come back here when that thread stayed quiet, and the stale bot unassigned you three weeks later. That's our failure to close a loop we opened, not yours. Sorry.

You're welcome on anything here, and if you'd like something with more room in it than a six-line middleware registration, say so and I'll point you at one.

PR #2616, opened 2026-07-27
Secret definitions in `Provider` CRD uses parameters schema

chilagrow:
This is a breaking change. This PR updates Provider's secret schema to use standardized `parametersSchema`.

* Provider CRD's `spec.secrets.openAPIV3Schema` is replaced with `spec.secrets.parametersSchema.openAPIV3Schema` and uses standardized validation.
* Update secrets API test

Refs https://github.com/openeverest/helm-charts/pull/76.


provider-runtime/controller/secret_validation.go
```diff
@@ -62,7 +64,12 @@ func ValidateSecretSchema(secret *corev1.Secret, providerSpec *corev1alpha1.Prov
 	// stringData takes precedence (same as Kubernetes behavior).
 	data := extractSecretData(secret)
 
-	if err := secretDef.Validate(data); err != nil {
+	dataJSON, err := json.Marshal(data)
+	if err != nil {
+		return fmt.Errorf("failed to marshal secret data: %w", errors.Join(ErrSecretSchemaValidation, err))
+	}
+
+	if err := secretDef.ParametersSchema.Validate(&runtime.RawExtension{Raw: dataJSON}); err != nil {
```
recharte, line 77: Shouldn't we short-circuit here?

```suggestion
	if secretDef.ParametersSchema == nil {
		return nil
	}

	if err := secretDef.ParametersSchema.Validate(&runtime.RawExtension{Raw: dataJSON}); err != nil {
```

`ParametersSchema.Validate` rejects "no schema declared, but a payload supplied", and we always pass a non-nil `params` here (`extractSecretData` returns an initialised map, so `dataJSON` is at worst `{}`). Since `parametersSchema` is `+optional`, a definition that declares only a `uiSchema` would reject every secret with `secret does not conform to schema: schema validation failed`.

`ValidateConfigMapSchema` has the same shape, so it probably wants the same treatment.

If you agree and make the change, could you add a test case for it? The table in `secret_validation_test.go` currently has no definition-without-schema case.

chilagrow: Good point, for optional schema skipping validation is sensible when schema is absent.

recharte (dismissed):
Looks good 😉

PR #2618, opened 2026-07-27
fix(version_service): prevent panic on nil matrix and handle unsupported operator

vedant21-oss:
Added nil check for version matrix response in GetSupportedEngineVersions to prevent panic when matrix payload is empty or nil. Added explicit error handling for unsupported operators.

### DIFF
--- a/pkg/version_service/client.go
+++ b/pkg/version_service/client.go
@@ -108,14 +108,21 @@ func (c *versionServiceClient) GetSupportedEngineVersions(ctx context.Context, o
 		return nil, errors.New("no versions found")
 	}
 
+	firstVer := response.GetVersions()[0]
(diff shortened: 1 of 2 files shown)


recharte:
Thanks for putting a patch together alongside the report. I've explained on #2593 why the panic isn't reachable — the accessors are generated protobuf getters, which are nil-safe by construction, and the unsupported-operator arm can't be reached from the only caller.

That makes this change inert: it doesn't alter observable behaviour, and the test name would leave future readers believing a nil dereference was once possible here. So I'd rather not take it as-is.

This is a call about direction, not about the quality of your code — the patch does what it says. If you'd like to pursue the narrower consistency point (missing matrix → explicit error, matching the `no versions found` check just above it), that would be a welcome standalone PR with framing that matches what it actually does.

PR #2626, opened 2026-07-27
Add generic provider upgrade preflight library to provider-runtime

recharte:
## What

Second PR of the provider upgrades feature ([spec 009](https://github.com/openeverest/specs/blob/main/specs/009-provider-upgrades.md), §6.1–§6.2, §6.6): the provider-agnostic preflight library that decides whether upgrading a provider release would strand a running Instance. Builds on the API fields from #2604.

**`PreflightUpgrade` — catalog membership + deprecation (R1).** For every Instance, resolves the effective version of each component exactly as the reconciler does (explicit component version → `spec.version` → frozen `status.version` → target default bundle) and checks it against the target catalog:
- version absent, or `removedInVersion` ≤ the target release → **blocking error**
- `deprecated` (or removal scheduled for a later release) → **warning** with the remediation runway
- Instances being deleted are skipped; reports are deterministically ordered.

**`CheckUpgradePath` — release floor (R2).** Blocks a jump when the installed `release.version` is below the target's `release.minUpgradableFrom` (e.g. an operator that only upgrades one minor at a time). A missing or unparseable installed version degrades to a warning so first adoption isn't blocked.

**`UpgradeProvider`** — new optional provider interface (alongside `WatchProvider`/`BackupProvider`) for checks the generic library can't know, e.g. the R4 skew-window/bounded-deferral check. `RunUpgradePreflight` orchestrates all three; `HasBlockingIssues` feeds the hook's exit code.

## Notes

- Pure library: nothing calls it in-tree yet. The Helm pre-upgrade hook runner that wires it into provider charts is the next PR in the stack.
- Version comparisons use `hashicorp/go-version` (existing direct dependency), which accepts short release versions like `0.3`.


chilagrow (approved):
Great!

PR #2648, opened 2026-07-28
Add restore management commands and list functionality

### LINKED ISSUE #2621: Add `everestctl restore list` subcommand

There is no CLI visibility into restores. After triggering a restore, users cannot check from the terminal whether it is running, succeeded, or failed — during an incident this means switching to the UI or kubectl at the worst possible moment.

Add an `everestctl restore list` subcommand that lists Restore resources via the Everest API. The table output should include NAME, NAMESPACE, INSTANCE, BACKUP, STATE, and AGE (from `spec.instanceRef`, `spec.dataSource.backup.backupRef`, and `status.state` — one of Pending, Running, Succeeded, Failed, Error). It should support filtering by namespace and by instance (`--instance`). Follow existing CLI conventions: `--json`/`--verbose` output modes, `--cluster`, and `--context` flags. The `list` subcommand should register the alias `ls`.

```
# List all restores in a namespace
everestctl restore list --namespace everest

# Only restores targeting one instance
everestctl restore ls --namespace everest --instance my-mongo

# JSON output — check the state of the most recent restore
everestctl restore list --namespace everest --instance my-mongo --json \
  | jq 'last | .status.state'
```

`kubectl get restores.backup.openeverest.io -n everest` — requires direct cluster access and bypasses Everest RBAC.

### Additional Context

Part of the `list` subcommand set (#2551–#2554, #2579) and complements `restore create` (#2620).

VijetaPriya47:
closes #2621

### DIFF
--- a/pkg/cli/restore/list.go
+++ b/pkg/cli/restore/list.go
@@ -1,0 +1,208 @@
+// everest
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package restore provides CLI business logic for restore management.
+package restore
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"sort"
+	"time"
+
+	"github.com/rodaine/table"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/util/duration"
+
+	"github.com/openeverest/openeverest/v2/client"
+	authcli "github.com/openeverest/openeverest/v2/pkg/cli/auth"
+)
+
+// Config holds the shared configuration for restore CLI runners.
+type Config struct {
+	Pretty bool
+}
+
+// ListOptions holds the inputs for the list command.
+type ListOptions struct {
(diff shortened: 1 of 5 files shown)


VijetaPriya47:
@recharte could u check it out?

pkg/cli/restore/list.go
```diff
@@ -0,0 +1,208 @@
...
+	v, ok := (*r.Metadata)[key]
+	if !ok {
+		return "-"
+	}
+	s, ok := v.(string)
+	if !ok || s == "" {
+		return "-"
+	}
+	return s
+}
```
recharte, line 208: This is now the third copy of the metadataStringField / *CreationTime / sortByRecency trio (backup list has identical ones, instance has close cousins). Copying was a reasonable call since the generated client types have no methods, so there's no clean constraint to hang generics on, but I think it's probably time to extract.

A small `pkg/cli/clientmeta/clientmeta.go` (stdlib-only, so anything under pkg/cli can import it) where only the sort is generic and the metadata map is accessed via a closure, since that's the one thing all the generated types share:

```go
// String returns the string value of a metadata key, or "" when missing.
func String(md *map[string]any, key string) string

// CreationTime parses metadata.creationTimestamp (RFC 3339).
func CreationTime(md *map[string]any) (time.Time, bool)

// SortByRecency orders items newest-first by creationTimestamp, falling
// back to name for stability when timestamps tie or are missing.
func SortByRecency[T any](items []T, meta func(*T) *map[string]any)
```

Call sites stay almost as tidy as today:

```go
clientmeta.SortByRecency(restores, func(r *client.Restore) *map[string]any { return r.Metadata })
```

One deliberate tweak in those signatures: `String` should return `""` for missing rather than `"-"`. Right now the helpers return the display sentinel, which is why `restoreCreationTime` has to compare against `"-"` to detect absence — the shared version should speak data ("empty means missing") and leave the `-` rendering to a tiny `orDash` in each table. The sort-tiebreak tests would move into `clientmeta` once instead of living in every `list_test.go`.

But maybe let's do this in a separate PR that handles the migration of all resources. I'll open a new issue for this.

recharte (approved):
Well done 🚀

PR #2652, opened 2026-07-28
Type CRD metadata as ObjectMeta in the OpenAPI spec and generated clients

### LINKED ISSUE #2649: Extract shared metadata helpers for CLI list commands into pkg/cli/clientmeta
### Task Description

~~Extract the `metadataStringField` / `*CreationTime` / `sortByRecency` helper trio into a shared `pkg/cli/clientmeta` package.~~

**Superseded by a root-cause fix.** During design discussion we realized the duplicated helpers were downstream damage from a spec deficiency: Kubernetes forbids CRD structural schemas from describing `metadata`, so the extracted OpenAPI spec declared it as a bare `type: object`, which `oapi-codegen` rendered as `*map[string]any` (and openapi-typescript as the unusable `Record<string, never>`). Every CLI command then re-implemented string-map spelunking to read names and timestamps back out.

The implemented fix instead types the metadata at the source:

1. **`gen-crds-openapi` emits a shared `ObjectMeta` component schema** — a curated subset of `metav1.ObjectMeta` (`name`, `generateName`, `namespace`, `labels`, `annotations`, `creationTimestamp`, `deletionTimestamp`, `resourceVersion`, `uid`; machinery fields like `managedFields`/`ownerReferences` are omitted but still accepted as additional properties) — and every CRD kind's `metadata` property `$ref`s it. This mirrors how Kubernetes publishes `ObjectMeta` in its own OpenAPI spec: the API server owns the metadata description that CRDs are forbidden from providing.
2. **The schema carries `x-go-type`/`x-go-type-import` extensions**, so the generated Go client and server models alias it to the canonical type (`type ObjectMeta = metav1.ObjectMeta`) rather than a structural look-alike. No DTO, no mapping layer: the wire format is unchanged and objects round-trip byte-identical.
3. **All five copies of `metadataStringField` and the RFC3339-parsing age helpers were deleted** — direct field access replaces them, with timestamp parsing handled by `metav1.Time`.
4. **TypeScript clients (UI, api-tests) get typed metadata for free** from the same spec change.
5. **The one helper with residual value — the newest-first recency sort with its tiebreaks — moved to `pkg/cli/clientmeta.SortByRecency`**, with the tiebreak test matrix, for reuse by `restore list` (#2648) and future `<resource> list` commands.

### Reasoning / Benefit

- Fixes the cause (untyped spec) instead of the symptom (duplicated map helpers); Go, TypeScript, and any future SDK all benefit from one change.
- Zero drift by construction in Go (type alias) and TS (generated from the spec). The one hand-written artifact — the curated schema — is guarded by `TestObjectMetaSchemaMatchesType`, which reflects over `metav1.ObjectMeta`'s json tags and fails on any property-name or type mismatch.
- The "no DTO" API design principle is preserved: the schema is a *description* of the upstream wire contract, not a translation layer; `additionalProperties` remain allowed so full ObjectMeta round-trips untouched.

### Additional Context

- Implementation branch: `client-typed-objectmeta` (PR to follow), targeting `release-2.0`.
- `pkg/cli/restore/list.go` in #2648 will need the same mechanical migration on rebase (drop `metadataStringField`/`restoreCreationTime`, use `clientmeta.SortByRecency`).
- Discussion trail: originally scoped as a helper extraction (see edit history); superseded after evaluating `x-go-type`-only vs. spec-level typing.

recharte:
The extracted OpenAPI spec declared each CRD's `metadata` as a bare `type: object` — Kubernetes forbids CRD structural schemas from describing ObjectMeta, so the hole propagated into every generated client: `oapi-codegen` rendered `Metadata *map[string]any`, openapi-typescript rendered the unusable `Record<string, never>`, and every CLI list/create command re-implemented string-map helpers (`metadataStringField`, `*CreationTime`, RFC3339 parsing) to read names and timestamps back out.

This PR fixes the cause instead of the symptom:

- `gen-crds-openapi` now emits a shared `ObjectMeta` component schema — a curated subset of `metav1.ObjectMeta` (name, generateName, namespace, labels, annotations, creationTimestamp, deletionTimestamp, resourceVersion, uid) — and every CRD kind's `metadata` `$ref`s it, mirroring how kube-apiserver publishes ObjectMeta in its own OpenAPI doc. Machinery fields (managedFields, ownerReferences, finalizers) are deliberately not advertised but still pass validation via additionalProperties, so full objects round-trip byte-identical: this is a description of the wire contract, not a DTO/mapping layer.
- The schema carries `x-go-type`/`x-go-type-import`, so the Go client and server models alias the canonical type (`type ObjectMeta = metav1.ObjectMeta`) rather than a look-alike. All six `metadataStringField` copies (backup, restore, instance, provider, backupclass, backupstorage) and the timestamp-parsing helpers are deleted in favor of direct field access.
- TypeScript clients (UI, api-tests) get typed metadata from the same spec change; the full UI builds cleanly against the new types.
- `TestObjectMetaSchemaMatchesType` guards the hand-curated schema against drift: it reflects over `metav1.ObjectMeta`'s json tags and fails on any property-name or type mismatch.
- Second commit: the one helper with residual value — the newest-first recency sort with its tiebreaks — moves to `pkg/cli/clientmeta.SortByRecency`, reused by backup and restore list.

Wire format is unchanged (apimachinery v0.36's `omitzero` prevents `creationTimestamp: null` in requests). List-kind metadata is intentionally untouched.

Closes #2649


ui/api/http-api.types.ts
```diff
@@ -1367,7 +1398,7 @@ export interface components {
              *     More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
              */
             kind?: string;
-            metadata?: Record<string, never>;
+            metadata?: components["schemas"]["ObjectMeta"];
```
recharte, line 1401: @solovevayaroslavna can this change have any harmful side-effect in the UI?

VijetaPriya47 (approved):
I understood the backend side of work, it will affect restore create also, so would raise the PR for create after merging this one.

ui/api/http-api.types.ts
solovevayaroslavna: @recharte No harmful side-effects on the UI. This is a type-level change only — openapi-typescript output isn't shipped at runtime, and the actual payloads already carried full k8s metadata, so runtime behavior is unchanged. On the type side it's strictly an improvement: the UI only reads curated fields name, namespace, labels, annotations, creationTimestamp on the affected CRD types, all present in ObjectMeta, so it compiles cleanly.

recharte: Thanks 🤗

PR #2656, opened 2026-07-29
everestctl: add restore create subcommand

### LINKED ISSUE #2620: Add `everestctl restore create` subcommand

Restores can be triggered through the API and the UI, but there is no way to restore an instance from a backup via the terminal. Recovery is precisely the scenario where operators are most likely to be working from a shell under pressure, yet today it requires hand-crafting a Restore CR with kubectl or raw API calls.

Add an `everestctl restore create` subcommand that creates a Restore for an instance via the Everest API. Required flags: `--instance` and `--namespace` to identify the target instance (`spec.instanceRef`), and `--backup` to reference the source Backup CR in the same namespace (`spec.dataSource.backup.backupRef`; `dataSource.type=Backup` is implied since it is the only type today). Only full-backup restores are in scope — point-in-time recovery is deliberately excluded and will be handled in a follow-up.

Optional flags:
- `--name` — explicit Restore name; when omitted the name is generated server-side (`metadata.generateName`) from the instance name.
- `--wait` / `--timeout` — block until the restore reaches a terminal state, reusing the shared waiter: `Succeeded` → exit 0, `Failed` → exit 1, timeout → exit 2; `Error` is documented as transient so the wait continues through it. Ctrl-C cancels only the wait — the restore keeps running server-side (stated in help text).

A 409 on an explicit `--name` is mapped to `restore %q already exists in namespace %q`. The command follows existing CLI conventions: `--json`/`--verbose` output modes, `--cluster`, and `--context` flags.

```
# Restore an instance from a backup
everestctl restore create --instance my-mongo --namespace everest \
  --backup pre-upgrade

# Wait for the restore to finish, bounded for CI
everestctl restore create --instance my-mongo --namespace everest \
  --backup nightly-20260727 --wait --timeout 30m
```

Applying a Restore CR with `kubectl apply` — bypasses Everest API authentication/RBAC and requires knowing the CRD schema.

### Additional Context

Mirrors `backup create` (#2578) and reuses its shared waiter machinery. This would introduce the `restore` command group, which also hosts `restore list` (separate issue). Out of scope for this issue, to be handled in follow-ups:
- PITR (`spec.dataSource.backup.pitr` with `latest`/`date`) — a future `--pitr` flag can be added without changing the flags introduced here.
- Restore-time parameters (`spec.parameters` validated against the BackupClass's `restoreParametersSchema`) via `--set`/`-f`.

VijetaPriya47:
Restores could already be triggered through the API and UI, but not from the terminal -- exactly the moment operators are most likely to need it, recovering from a shell instead of hand-crafting a Restore CR with kubectl or calling the API directly. This adds `everestctl restore create --instance --namespace --backup`, with an optional --name (falls back to metadata.generateName, same as backup create) and --wait/--timeout to block until the restore reaches a terminal state.

Mirrors backup create's CreateRunner/poll/409-conflict pattern and reuses the existing wait/waitcmd plumbing instead of growing a second copy. The one real structural difference is the payload: the source backup sits under spec.dataSource.backup.backupRef.name instead of a flat ref, since dataSource is shared with future PITR/seeding sources. PITR and restore-time parameters are deliberately left out -- #2620 calls both out of scope for a follow-up.

Built against #2652's typed ObjectMeta (restore list already migrated to it), and fixed the stale Restore CR example left in the backups-restore-architecture doc from the earlier *Ref.name refactor.

Closes #2620

### DIFF
--- a/pkg/cli/restore/create.go
+++ b/pkg/cli/restore/create.go
@@ -128,7 +128,7 @@ func (cr *CreateRunner) emitCreated(created *client.Restore, name string) error 
 
 // waitForRestore blocks until the restore reaches a terminal state, streaming
 // progress in pretty mode. JSON mode emits one final object on success.
-func (cr *CreateRunner) waitForRestore(ctx context.Context, c *client.ClientWithResponses, created *client.Restore, opts CreateOptions,name string) error {
+func (cr *CreateRunner) waitForRestore(ctx context.Context, c *client.ClientWithResponses, created *client.Restore, opts CreateOptions, name string) error {
 	var latest *client.Restore
 	basePoll := newRestorePoll(c, opts.Cluster, opts.Namespace, name)
 	poll := func(ctx context.Context) (*client.Restore, error) {
--- a/pkg/cli/restore/create.go
+++ b/pkg/cli/restore/create.go
@@ -1,0 +1,178 @@
+// everest
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package restore provides CLI business logic for restore management.
+package restore
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	"go.uber.org/zap"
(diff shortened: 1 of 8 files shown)


VijetaPriya47:
@recharte  could you review it?

pkg/cli/restore/create.go
```diff
@@ -0,0 +1,178 @@
...
+	// A 201 with an unparseable body would otherwise emit empty stdout.
+	if created == nil {
+		return fmt.Errorf("restore %q was created but the server returned an unreadable response body", name)
+	}
+	return writeRestoreJSON(created)
+}
+
+// waitForRestore blocks until the restore reaches a terminal state, streaming
+// progress in pretty mode. JSON mode emits one final object on success.
+func (cr *CreateRunner) waitForRestore(ctx context.Context, c *client.ClientWithResponses, created *client.Restore, opts CreateOptions,name string) error {
```
recharte: To make linter happy 😅 
```suggestion
func (cr *CreateRunner) waitForRestore(ctx context.Context, c *client.ClientWithResponses, created *client.Restore, opts CreateOptions, name string) error {
```

recharte (approved):

PR #2663, opened 2026-07-29
Fix install RenderTemplates error

### LINKED ISSUE #2662: Fix wrong error message when helm template rendering fails during install
### What is the issue?

During `everestctl install`, if Helm template rendering fails in the CatalogSource step, the CLI says it could not get the Everest CatalogSource namespace. That is not what failed. Rendering failed first.

The upgrade path already reports the right thing in `pkg/cli/upgrade/steps.go`. It says it could not render Helm templates. Install should do the same so the error points to the actual step that broke.

### Steps to Reproduce

1. Run `everestctl install` in a setup where Helm template rendering fails (bad chart values, broken Helm config, etc.).
2. Look at the output from the step called "Ensuring Everest CatalogSource is ready".
3. The error talks about the CatalogSource namespace even though template rendering failed before that lookup ran.

### Relevant Logs / Output

```bash
could not get Everest CatalogSource namespace: <underlying helm render error>
```

### OpenEverest Version

Latest version

### Installation Method

CLI (everestctl)

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

false200:
## Summary

Fixes a misleading error message during `everestctl install`. When Helm template rendering fails, the CLI now reports that rendering failed instead of saying it could not get the CatalogSource namespace.

Closes #2662

## What changed

In `pkg/cli/install/steps.go`, the error returned from `RenderTemplates` now matches the upgrade path in `pkg/cli/upgrade/steps.go`:

```go
return fmt.Errorf("could not render Helm templates: %w", err)
```

## Why

The old message pointed at the wrong step and made debugging harder. Upgrade already had the correct wording. This was a copy-paste bug in install.

## Test plan

- [ ] Run `everestctl install` in a setup where Helm template rendering fails
- [ ] Confirm the error says `could not render Helm templates` and not `could not get Everest CatalogSource namespace`


recharte (approved):
Thank you

PR #2667, opened 2026-07-30
[v2] Validate Backup/Restore resource references at the API layer on create

### LINKED ISSUE #2657: [v2] Validate Backup/Restore resource references at the API layer on create
> **Applies to v2 only** (`release-2.0` branch). The v1.x API on `main` has a different validation architecture and is not covered by this issue. All file references below are against `release-2.0`.

### Task Description

Extend the validation handler (`internal/server/handlers/validation/`) so that creating a `Backup` or `Restore` through the Everest API fails fast with a 400 when it references resources that don't exist or are in an unusable state, instead of returning 201 and surfacing the failure minutes later as a status condition.

- **`CreateRestore`**: `spec.dataSource.backup.backupRef.name` must reference an existing Backup in the namespace, and that Backup must be in the `Succeeded` state. Today the existence check runs only on the PITR path (`validateRestorePITR` bails out when `ds.Backup.PITR == nil`), producing an inconsistency: a PITR restore with a typo'd backup gets a crisp 400 ("backup 'X' does not exist"), while a plain restore with the same typo is accepted. Hoist the check out of the PITR-only path and add the state requirement. Note this deliberately rejects queueing a Restore against a still-`Running` backup — callers should wait for the backup to complete (e.g. `backup create --wait`) first.
- **`CreateBackup`**: `spec.instanceRef`, `spec.classRef`, and `spec.storageRef` must reference existing resources, and the BackupClass must list the instance's provider in `spec.supportedProviders` — moving the provider's async `BackupClassUnsupported` failure to request time. To avoid drift with the provider-side check, reuse (or extract) a shared helper from `provider-runtime/controller` rather than re-implementing the predicate — same approach `validateRestorePITR` already takes with `controller.ValidateRestorePITR`.

### Reasoning / Benefit

- Today, no synchronous layer catches a dangling reference: CRD CEL rules can't do cross-resource lookups by design, no admission webhook covers these CRs (only MonitoringConfig has one), the API validation handler skips the non-PITR path, and the CLI does no pre-flight. The first signal is an async condition (`SourceBackupNotFound` etc.).
- **Why the API layer and not a webhook:** the layered-validation doctrine — product API for synchronous UX, controller conditions for correctness, webhooks only for invariants where a bad object causes damage before reconciliation. A dangling ref is a UX problem, not a safety problem. A webhook also forces the `failurePolicy` dilemma: `Fail` adds an availability dependency to the disaster-recovery path (restores blocked because a webhook pod is down on an already-degraded cluster); `Ignore` makes the guarantee evaporate exactly when things are chaotic. Kubernetes core and mature operators (cert-manager, Prometheus operator, Gateway API) deliberately admit dangling refs and report via status — kubectl/GitOps users keep that contract via the existing conditions.
- Fixing it at the API layer covers UI, everestctl, and direct API consumers at once, runs within the requester's RBAC context, and follows the in-repo precedent (`validateRestorePITR` already does exactly this lookup pattern with `kubeConnector` and `ErrInvalidRequest` → 400).

### Additional Context

*(paths on `release-2.0`)*
- `internal/server/handlers/validation/restore.go` — `validateRestorePITR` (the pattern to generalize)
- `internal/server/handlers/validation/backup.go` — the `CreateBackup` passthrough (`// Add validation here if needed in the future`)
- Provider-side async conditions: `ReasonDataSourceSourceBackupNotFound`, `SourceBackupNotSucceeded`, `StorageMismatch` (`api/core/v1alpha1/instance_types.go`) — these remain the safety net for kubectl-applied CRs
- Came out of the #2656 review (`restore create` inherits the gap from `backup create` — neither pre-flights its references)

Adii-45:
## Summary

This PR implements the API-layer validation described in #2657 for the `release-2.0` branch

Previously, Backup and Restore creation requests relied on downstream processing for several resource reference checks, allowing invalid requests to pass through the API layer. This change adds fail-fast validation so invalid references return a `400 Bad Request` before the request proceeds further!

### Restore validation

- Generalized the existing restore validation so Backup existence and state validation are always performed
- Added validation that the referenced Backup exists
- Added validation that the referenced Backup is in the `Succeeded` state before allowing restore creation
- Preserved the existing PITR-specific validation flow while reusing the already-resolved Backup object instead of performing duplicate lookups

### Backup validation

Added validation during Backup creation to verify that:

- The referenced Instance exists
- The referenced BackupClass exists
- The referenced BackupStorage exists
- The selected BackupClass supports the referenced Instance's provider

These validations are performed before the request is forwarded to the next handler!

### Tests

Added unit tests covering the new validation logic for both Restore and Backup creation, including success and failure scenarios for:

- Missing referenced resources
- Invalid Backup state
- Unsupported provider compatibility
- Valid requests

## Result

The API now rejects invalid Backup and Restore creation requests earlier in the request lifecycle with appropriate validation errors instead of allowing them to progress further!

Closes #2657


Adii-45:
@recharte Hi! The PR is now ready for review..
I've implemented the API-layer validation described in the issue for the `release-2.0` branch and verified the changes locally! If you have any feedback or would like me to make any changes, I'd be happy to update the PR...
Thanks!

VijetaPriya47:
Hey @Adii-45
I believe those two shared helpers should be extracted into provider-runtime/controller/backup_validation.go (alongside the existing ValidateRestorePITR), and both the new API-layer validators and the existing inline checks in common.go should call them. 

Each caller keeps its own error/status wrapping (the API layer returns an error/400, ReconcileDataSource sets a DataSourceStatus with a reason string), the helper just centralizes the one thing that must never drift: what "succeeded" means and the comparison itself. :>
 
Update the branch also. 

@recharte WDYT?

Adii-45:
Thanks for the review @VijetaPriya47!

I've updated the implementation based on your feedback:

- extracted the shared validation helpers into `provider-runtime/controller/backup/validation.go`
- updated both API-layer validators and the existing controller logic in `common.go` to reuse them
- kept the caller-specific error/status handling while centralizing the shared validation logic
- updated the branch with the latest `release-2.0`

I'd appreciate any additional feedback or suggestions... Thanks!

provider-runtime/controller/backup/validation.go
```diff
@@ -0,0 +1,135 @@
...
+	// not exist.
+	ErrBackupClassNotFound = errors.New("backup class not found")
+	// ErrProviderUnsupported is returned when a BackupClass does not list a
+	// provider in its SupportedProviders.
+	ErrProviderUnsupported = errors.New("provider not supported by backup class")
+	// ErrRestorePITRUnsupported is returned by ValidateRestorePITR when a
+	// Restore requests point-in-time recovery but the resolved
+	// ProviderManaged BackupClass does not advertise supportsPITR.
+	ErrRestorePITRUnsupported = errors.New("point-in-time recovery is not supported")
+)
```
recharte: Small style thought on the messages. The Go convention is that each layer adds information no other layer has — the sentinel is the reason, the wrapper is the identity/context, and nothing repeats. Compare the stdlib:

```
open /tmp/x: no such file or directory       // *PathError adds op+path, ENOENT is the reason
strconv.Atoi: parsing "abc": invalid syntax  // *NumError adds func+input, ErrSyntax is the reason
```

Right now the sentinel and the detail say the same thing twice: `backup not found: backup 'x' does not exist`. If the interpolated part carries only identity, it reads better and gains information:

```go
return nil, fmt.Errorf("%w: %q in namespace %q", ErrBackupNotFound, name, ns)
// → backup not found: "daily-1" in namespace "team-a"

return backup, fmt.Errorf("%w: %q is in state %q, must be %q",
    ErrBackupNotSucceeded, name, backup.Status.State, backupv1alpha1.BackupStateSucceeded)
// → backup not succeeded: "daily-1" is in state "Running", must be "Succeeded"
```

Sentinel-first (`"%w: detail"`) is the right ordering here since the sentinel is the headline — no need for the stdlib's cause-last style.

internal/server/handlers/validation/backup.go
```diff
@@ -17,22 +17,66 @@ package validation
...
 	return h.next.GetBackup(ctx, cluster, namespace, name)
 }
 
-// CreateBackup proxies the request to the next handler.
+// CreateBackup validates the Backup's referenced resources before creating it.
 func (h *validateHandler) CreateBackup(ctx context.Context, cluster string, backup *v1alpha1.Backup) (*v1alpha1.Backup, error) {
-	// Add validation here if needed in the future
+	if err := h.validateBackupRefs(ctx, backup); err != nil {
+		return nil, errors.Join(ErrInvalidRequest, err)
+	}
```
recharte, line 43: Now that the helpers return sentinels, we can use them here: at the moment every failure gets joined with `ErrInvalidRequest`, so a transient API-server error (timeout, connection refused) comes back to the user as `400 Bad Request` rather than a 500. Something like:

```go
if err := h.validateBackupRefs(ctx, backup); err != nil {
    if errors.Is(err, backupvalidation.ErrBackupClassNotFound) ||
        errors.Is(err, backupvalidation.ErrProviderUnsupported) || /* ... */ {
        return nil, errors.Join(ErrInvalidRequest, err)
    }
    return nil, err
}
```

(or classify inside `validateBackupRefs` and only join there). Same applies to `CreateRestore` in restore.go.

provider-runtime/controller/backup/validation.go
```diff
@@ -0,0 +1,135 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package backup holds Backup/Restore reference-validation logic shared
+// between the API server's validation handlers and the provider runtime's
+// DataSource-seeding controller, so the existence/state/compatibility checks
+// are defined once and reused rather than re-implemented per caller.
+package backup
```
recharte: Two small things about the new package:

- The doc says the logic is shared with "the provider runtime's DataSource-seeding controller", but `Context.ReconcileDataSource` in `controller/common.go` still hand-rolls the same three checks (backup exists / Succeeded / class supports provider). Could you either wire that up too, or soften the doc so it matches reality?
- We already have `controller/backup_validation.go` holding closely-related helpers (`ValidateInstanceBackupAgainstClass`, `ValidateInstanceBackupPITRParameters`). Splitting backup validation across `controller` and `controller/backup` is a bit confusing — was there an import cycle forcing this? The API server already imports `provider-runtime/controller` (see `validation/secret.go`), so I don't think there is one. Keeping everything in `backup_validation.go` would be my preference.

provider-runtime/controller/backup/validation.go
```diff
@@ -0,0 +1,135 @@
...
+	ErrRestorePITRUnsupported = errors.New("point-in-time recovery is not supported")
+)
+
+// GetBackupFunc resolves a Backup by name. Callers close over their own
+// client/namespace so this package stays agnostic of how the lookup is
+// wired (a direct Kubernetes client vs. a reconciler Context).
+type GetBackupFunc func(ctx context.Context, name string) (*backupv1alpha1.Backup, error)
+
+// GetBackupClassFunc resolves a cluster-scoped BackupClass by name.
+type GetBackupClassFunc func(ctx context.Context, name string) (*backupv1alpha1.BackupClass, error)
```
recharte: Nice job pulling the shared rules out — one thought on the getter injection, though. Right now each call site has to wrap a fetch in a closure to hand it over:

```go
backup, err := backupvalidation.ResolveSucceededBackup(ctx, ds.Backup.BackupRef.Name,
    func(ctx context.Context, name string) (*backupv1alpha1.Backup, error) {
        return h.kubeConnector.GetBackup(ctx, ctrlclient.ObjectKey{Namespace: restore.GetNamespace(), Name: name})
    })
if err != nil {
    return err
}
```

I get why — the API server only has `kubeConnector` while the runtime has a `client.Client`, so a callback is the obvious way to bridge them. But what if we left the I/O with the caller and shared just the rule?

```go
backup, err := h.kubeConnector.GetBackup(ctx, ctrlclient.ObjectKey{Namespace: ns, Name: name})
if err != nil {
    if k8serrors.IsNotFound(err) {
        return fmt.Errorf("%w: %q in namespace %q", backupvalidation.ErrBackupNotFound, name, ns)
    }
    return fmt.Errorf("get backup %q: %w", name, err)
}
if err := backupvalidation.ValidateBackupSucceeded(backup); err != nil {
    return err
}
```

About the same number of lines, but everyone keeps their own client — which also means the message can finally say which namespace, something the helper can't do today since the namespace is hidden in the closure. And the shared package gets to shed `context.Context` and I/O entirely: just the sentinels plus two pure predicates, which makes the tests plain table rows instead of needing that `notFoundError` stand-in.

I think `ReconcileDataSource` would land in the same place when it adopts this — it needs a `DataSourceStatus{State, Reason, Message}` rather than an error, so the predicates are the half it can actually reuse.

What do you think?

provider-runtime/controller/backup/validation.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil, fmt.Errorf("%w: backup '%s' does not exist", ErrBackupNotFound, name)
+		}
+		return nil, fmt.Errorf("failed to get backup '%s': %w", name, err)
+	}
+	if backup.Status.State != backupv1alpha1.BackupStateSucceeded {
+		return backup, fmt.Errorf("%w: backup '%s' is in state '%s', must be '%s' to restore from it",
+			ErrBackupNotSucceeded, name, backup.Status.State, backupv1alpha1.BackupStateSucceeded)
+	}
```
recharte: Tiny one, but `Status.State` is `+optional` and stays empty until the controller's first reconcile, so anyone who does `backup create` followed quickly by `restore create` gets `backup 'daily-1' is in state '', must be 'Succeeded' to restore from it` — the empty quotes read like a bug. Since backup_controller.go sets `Pending` as soon as it gets there anyway, normalising is a two-liner:

```go
if backup.Status.State != backupv1alpha1.BackupStateSucceeded {
    state := backup.Status.State
    if state == "" {
        state = backupv1alpha1.BackupStatePending
    }
    return backup, fmt.Errorf("%w: %q is in state %q, must be %q to restore from it",
        ErrBackupNotSucceeded, name, state, backupv1alpha1.BackupStateSucceeded)
}
```

→ `backup not succeeded: "daily-1" is in state "Pending", must be "Succeeded" to restore from it`, which also nudges the user toward the right fix (wait, or use `backup create --wait`).

internal/server/handlers/validation/restore.go
```diff
@@ -33,41 +31,43 @@ func (h *validateHandler) GetRestore(ctx context.Context, namespace, name string
...
+// validateRestoreBackupRef rejects restores whose source Backup does not
+// exist or has not reached the Succeeded state. When the restore also
+// requests point-in-time recovery, it additionally rejects the request if
+// the Backup's BackupClass does not advertise PITR support.
+func (h *validateHandler) validateRestoreBackupRef(ctx context.Context, restore *backupv1alpha1.Restore) error {
 	ds := restore.Spec.DataSource
-	if ds.Backup == nil || ds.Backup.PITR == nil {
+	if ds.Backup == nil {
 		return nil
 	}
```
recharte: We validate `spec.instanceRef` on backup create but not on restore create — a restore into a non-existent Instance still gets a 201. Since `RestoreSpec.InstanceRef` is documented as "must already exist ... and use a provider listed in the BackupClass's SupportedProviders", it'd be nice to fetch the target Instance here and reuse `ValidateClassSupportsProvider` with the source backup's class.

internal/server/handlers/validation/backup_test.go
```diff
@@ -0,0 +1,172 @@
...
+				kubeConnector: kubeConnector,
+				next:          mockNext,
+			}
+
+			_, err := handler.CreateBackup(ctx, cluster, tt.backup)
+			if tt.err == "" {
+				require.NoError(t, err)
+				return
+			}
+			require.ErrorContains(t, err, tt.err)
```
recharte, line 213: Two or three additions that would make these tests guard the actual contract:

- `require.ErrorIs(t, err, ErrInvalidRequest)` — `ErrorContains` passes even if the join is dropped and the endpoint starts returning 500.
- `mockNext.AssertNotCalled(t, "CreateBackup", ...)` on the failure paths — right now nothing asserts the rejected request never reached the next handler.
- A case where the lookup fails with something other than NotFound (a fake-client interceptor does it in a couple of lines) — that's the path behind the 400-vs-500 point I raised on [this comment](https://github.com/openeverest/openeverest/pull/2667/changes#r3691459804).

Same goes for restore_test.go (lines 153–158).

provider-runtime/controller/backup/validation_test.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// notFoundError is a minimal apierrors-recognizable "not found" error for
+// exercising the k8serrors.IsNotFound branch without depending on a fake client.
+type notFoundError struct{}
+
+func (notFoundError) Error() string { return "not found" }
+func (notFoundError) Status() metav1.Status {
+	return metav1.Status{Reason: metav1.StatusReasonNotFound}
+}
```
recharte: Apimachinery can build one of these for you, so the stand-in type isn't needed —

```go
notFoundErr := k8serrors.NewNotFound(schema.GroupResource{Group: "backup.openeverest.io", Resource: "backups"}, "b")
```

Also gives you a realistic error string if a test ever fails and you're reading the output. (And if you end up going the pure-predicate route from my other comment, this helper disappears on its own anyway.)

recharte (changes requested):
Overall great job, the extracted package with typed sentinels is a much better shape than the inline checks, and routing `reconciler/restore.go` through it means the two sides can't drift.

Two comments carry most of the weight. The error classification is the main one: the sentinels are there now, but everything still gets tagged `ErrInvalidRequest`, so a flaky API server reads as a user error. The other is the injected getters — not a correctness issue, but it's the public API of a package providers will import, so it's a lot cheaper to settle now than later.

The rest — message wording, a couple of test assertions, one or two nits — is genuinely small stuff.

provider-runtime/controller/backup/validation.go
Adii-45: Thanks for the suggestion! Updated this in the latest changes..
The error messages now follow the sentinel-first pattern, with contextual details added by the caller, and the related tests have been updated accordingly!

internal/server/handlers/validation/backup.go
Adii-45: Validation errors are now classified before joining with `ErrInvalidRequest`, so only user-facing validation failures are wrapped while unexpected runtime errors are propagated unchanged!

provider-runtime/controller/backup/validation.go
Adii-45: I consolidated the closely related validation helpers into the existing `backup_validation.go` file and adjusted the package documentation to match the final shared validation flow! Changes had been made in the latest commit..

provider-runtime/controller/backup/validation.go
Adii-45: Thanks, I like this direction! The latest refactor now keeps the shared package focused on pure validation logic, with each caller responsible for its own lookup and context-specific error handling before calling the shared predicates! That also simplified the tests considerably..

provider-runtime/controller/backup/validation.go
Adii-45: `Status.State` is now normalized to `BackupStatePending` before constructing the error message, and I added a test covering the empty-state case!

internal/server/handlers/validation/restore.go
Adii-45: Restore validation now resolves the target `InstanceRef` and reuses `ValidateClassSupportsProvider` with the source backup's `BackupClass`, so provider compatibility is validated consistently during restore..!

internal/server/handlers/validation/backup_test.go
Adii-45: Added assertions for `ErrInvalidRequest`, verified the request doesn't reach the next handler on validation failures, and added coverage for non-`NotFound` lookup errors in both the backup and restore validation tests.

provider-runtime/controller/backup/validation_test.go
Adii-45: The custom stand-in is no longer used - the tests now exercise the real `IsNotFound` path, so there wasn't anything left to replace here..!

Adii-45:
Thanks for the detailed review! I've addressed all of the requested review comments in the latest commits, including the follow-up suggestions around validation flow, error handling, and test coverage.

If you notice anything else that should be adjusted, I'm happy to make further changes!

internal/server/handlers/validation/backup.go
recharte: This isn't done yet. The code still joins unconditionally:

```go
if err := h.validateBackupRefs(ctx, backup); err != nil {
    return nil, errors.Join(ErrInvalidRequest, err)
}
```

and `CreateRestore` at restore.go:37 is identical. So `failed to get instance 'x': server unavailable` is still an `ErrInvalidRequest`, and an unreachable API server is still reported to the user as `400 Bad Request` — exactly the case this thread is about.

Three changes needed:

1. Give the instance and storage not-found paths their own sentinels, next to `ErrInvalidRequest` in errors.go. Only the BackupClass path wraps one today (line 71); lines 54 and 64 return a bare `fmt.Errorf`, so there is nothing for `errors.Is` to match on:

   ```go
   return fmt.Errorf("%w: '%s' in namespace '%s'",
       ErrInstanceNotFound, backup.Spec.InstanceRef.Name, backup.GetNamespace())
   ```

2. Add a predicate covering the full set of user-error sentinels — the two new ones plus `controller.ErrBackupNotFound`, `ErrBackupClassNotFound`, `ErrBackupNotSucceeded`, `ErrProviderUnsupported` and `ErrRestorePITRUnsupported`.

3. Join only when that predicate holds, and return the error untouched otherwise:

   ```go
   if err := h.validateBackupRefs(ctx, backup); err != nil {
       if isInvalidRef(err) {
           return nil, errors.Join(ErrInvalidRequest, err)
       }
       return nil, err
   }
   ```

   The bare return is deliberate: everest.go's error handler unwraps k8s `StatusError`s to their real status code before it ever reaches the `ErrInvalidRequest` case, and anything else falls through to 500.

Same treatment in `CreateRestore`.

internal/server/handlers/validation/backup_test.go
recharte: The `ErrorIs` and `AssertNotCalled` additions are right, and the interceptor is the correct way to produce a non-NotFound failure.

The assertion on it is wrong, though. `errServerUnavailable` is an infrastructure failure, and line 202 asserts it *is* an `ErrInvalidRequest` — a 400. So this case pins the bug in place rather than catching it, which is the opposite of why it was added. restore_test.go:220 has the same problem.

What this needs:

1. Add `wantInvalidRequest bool` to the table, true for every case that exists today.
2. Set it false for the two server-error cases, and assert `require.NotErrorIs(t, err, ErrInvalidRequest)` there alongside the existing `require.ErrorIs(t, err, tt.getErr)`.

Both cases should fail against the handler as it stands today, and pass once the classification from the other thread is in.

internal/server/handlers/validation/restore.go
```diff
@@ -33,18 +33,21 @@ func (h *validateHandler) GetRestore(ctx context.Context, namespace, name string
...
-// when the BackupClass resolved via the source Backup does not advertise
-// PITR support. Restores without PITR options pass through untouched.
-func (h *validateHandler) validateRestorePITR(ctx context.Context, restore *backupv1alpha1.Restore) error {
+// validateRestoreBackupRef rejects restores whose source Backup does not
+// exist or has not reached the Succeeded state, whose target Instance does
+// not exist, or whose Backup's BackupClass does not support the target
+// Instance's provider. When the restore also requests point-in-time
+// recovery, it additionally rejects the request if the BackupClass does not
+// advertise PITR support.
+func (h *validateHandler) validateRestoreBackupRef(ctx context.Context, restore *backupv1alpha1.Restore) error {
```
recharte: Small one: this now validates `spec.instanceRef` and the class/provider pairing too, so `validateRestoreBackupRef` undersells it — and its opposite number on the backup side is `validateBackupRefs`. `validateRestoreRefs` would line the two up.

internal/server/handlers/validation/restore.go
```diff
@@ -54,18 +57,39 @@ func (h *validateHandler) validateRestorePITR(ctx context.Context, restore *back
...
-			return fmt.Errorf("backup class '%s' does not exist", backup.Spec.ClassRef.Name)
+			return fmt.Errorf("%w: '%s'", controller.ErrBackupClassNotFound, backup.Spec.ClassRef.Name)
 		}
 		return fmt.Errorf("failed to get backup class '%s': %w", backup.Spec.ClassRef.Name, err)
 	}
+	if err := controller.ValidateClassSupportsProvider(bc, instance.Spec.ProviderRef.Name); err != nil {
+		return err
+	}
+
+	if ds.Backup.PITR == nil {
```
recharte: `ValidateRestorePITR` already returns nil when `DataSource.Backup.PITR` is nil — it's the first thing it checks — so this guard is a no-op. Dropping it leaves the tail as a single `return controller.ValidateRestorePITR(restore, bc)`.

internal/server/handlers/validation/restore.go
Adii-45: Renamed the helper to `validateRestoreRefs` to better reflect its broader responsibility and align it with `validateBackupRefs`

internal/server/handlers/validation/restore.go
Adii-45: Removed the redundant `ds.Backup.PITR == nil` guard since `ValidateRestorePITR` already handles that case internally. The function now ends with a single return as suggested

internal/server/handlers/validation/backup.go
recharte: @Adii-45 I saw you pushed some new changes but haven't addressed this one yet, tagging you here just to make sure this didn't get lost but take your time, there's no hurry.

Adii-45: Oops, I mistakenly missed this thread while addressing the other review comments. I've addressed it now..

Implemented the requested error classification by introducing dedicated sentinels for the instance/storage NotFound paths, added predicate coverage for all validation sentinels, and updated both CreateBackup and CreateRestore to wrap only validation errors with `ErrInvalidRequest` while propagating infrastructure/runtime errors unchanged. I also updated the corresponding regression tests to cover these cases explicitly!

internal/server/handlers/validation/backup_test.go
Adii-45: Oops, I mistakenly missed this thread too while working through the other review comments!

I've addressed this now by making the test expectations explicit for the server-error cases. The infrastructure failure scenarios now assert `NotErrorIs(..., ErrInvalidRequest)`, while validation failures continue to assert `ErrorIs(..., ErrInvalidRequest)`, so the tests now verify the intended 400 vs. 500 classification behavior..

provider-runtime/controller/backup_validation.go
```diff
@@ -43,6 +43,34 @@ var ErrBackupClassLimitsExceeded = errors.New("backup class limits exceeded")
...
+// lookup for a referenced BackupClass returns not-found.
+var ErrBackupClassNotFound = errors.New("backup class not found")
+
+// ErrInstanceNotFound is a sentinel callers may wrap when their own lookup
+// for a referenced Instance returns not-found.
+var ErrInstanceNotFound = errors.New("instance not found")
+
+// ErrBackupStorageNotFound is a sentinel callers may wrap when their own
+// lookup for a referenced BackupStorage returns not-found.
+var ErrBackupStorageNotFound = errors.New("backup storage not found")
```
recharte: Apologies for adding to this one: we've been round several iterations already, and here I am asking for something more, in a file you'd otherwise be finished with. I do think this is the right moment rather than a later one — you're already editing this block, and splitting it means a second pass over the same exported surface. But if you'd rather land a one-line fix and leave the rest to a follow-up, say so and I'll open the issue. That's a perfectly good outcome and I won't hold the PR for it.

The bug, in this file: `ValidateRestorePITR` has two error returns and only one carries a sentinel. The date branch —

```go
if pitr.Type == backupv1alpha1.PITRTypeDate && pitr.Date == nil {
    return fmt.Errorf(
        "spec.dataSource.backup.pitr.date must be set when type is %q",
        backupv1alpha1.PITRTypeDate,
    )
}
```

— returns a bare error, so `isValidationError` says false, `CreateRestore` propagates it unwrapped, and a restore with `pitr: {type: date}` and no date now returns 500. Before this commit it was joined unconditionally and returned 400, so it's a regression. It is reachable: the CRD's CEL rule catches this at admission, but the handler runs before the object reaches the API server — the "paths that bypass admission" case this function's own doc comment cites.

Rather than add one more line to the list, I'd fix the shape that let it through. `isValidationError` lives in `internal/server/handlers/validation` and enumerates sentinels declared here, in a different module path, and `TestIsValidationError` re-lists the same seven by hand — so the test mirrors the omission instead of catching it.

**`provider-runtime/controller/backup_validation.go`** — replace the sentinel block with an umbrella the others report membership in:

```go
// ErrInvalidReference is the umbrella for every Backup/Restore
// reference-validation failure below: a caller-supplied reference that does
// not resolve, or resolves to an object that cannot be used for the
// requested operation. Callers that need to tell "the user named something
// wrong" from "the cluster is unreachable" match on this, so adding a
// sentinel never requires editing a list somewhere else.
//
// Instance-level backup config errors (ErrBackupClassLimitsExceeded,
// ErrPITRConfigInvalid) deliberately stay outside it — those surface on
// conditions rather than rejecting a request.
var ErrInvalidReference = errors.New("invalid reference")

// referenceError is a sentinel that also satisfies
// errors.Is(err, ErrInvalidReference). Declare new reference-validation
// sentinels with this type rather than errors.New.
type referenceError string

func (e referenceError) Error() string { return string(e) }
func (e referenceError) Is(target error) bool { return target == ErrInvalidReference }

const (
    ErrBackupNotFound          = referenceError("backup not found")
    ErrBackupClassNotFound     = referenceError("backup class not found")
    ErrInstanceNotFound        = referenceError("instance not found")
    ErrBackupStorageNotFound   = referenceError("backup storage not found")
    ErrBackupNotSucceeded      = referenceError("backup not succeeded")
    ErrProviderUnsupported     = referenceError("provider not supported by backup class")
    ErrRestorePITRUnsupported  = referenceError("point-in-time recovery is not supported")
    ErrRestorePITRDateRequired = referenceError("point-in-time recovery date is required")
)
```

**Same file**, in `ValidateRestorePITR`, wrap the date branch:

```go
return fmt.Errorf("%w: type is %q", ErrRestorePITRDateRequired, backupv1alpha1.PITRTypeDate)
```

**`internal/server/handlers/validation/backup_restore_errors.go`** — the predicate collapses to a single match:

```go
func isValidationError(err error) bool {
    return errors.Is(err, controller.ErrInvalidReference)
}
```

**`internal/server/handlers/validation/backup_restore_errors_test.go`** — `TestIsValidationError` becomes two cases that stay true as sentinels get added: a wrapped `referenceError` is one, `fmt.Errorf("...: %w", errors.New("etcd unavailable"))` isn't.

I checked every in-tree caller before suggesting this — all of them either wrap with `%w` or match with `errors.Is`, both unchanged by typed constants, and nothing takes an address or reassigns one. So it's source-compatible.

Concretely:

1. `provider-runtime/controller/backup_validation.go` — add `ErrInvalidReference` and `referenceError`, convert the eight sentinels to consts;
2. same file — wrap the date branch in `ValidateRestorePITR` with `ErrRestorePITRDateRequired`;
3. `internal/server/handlers/validation/backup_restore_errors.go` — reduce `isValidationError` to the single `errors.Is`;
4. `internal/server/handlers/validation/backup_restore_errors_test.go` — trim `TestIsValidationError` to the two structural cases;
5. `internal/server/handlers/validation/restore_test.go` — add a case for `type: date` with no date, asserting `wantInvalidRequest: true`.

internal/server/handlers/validation/backup.go
recharte: That's it — the gate plus sentinels on the instance and storage paths is what was missing. One case still slips past it and I'd like to fix the shape rather than the symptom; details on the sentinel block in `provider-runtime/controller/backup_validation.go`.

provider-runtime/controller/backup_validation.go
Adii-45: Thanks again for the detailed review! I've now addressed the requested changes:

- Introduced the umbrella `ErrInvalidReference` sentinel and updated the existing reference validation sentinels to report membership through it
- Wrapped the missing PITR date validation with `ErrRestorePITRDateRequired`
- Simplified `isValidationError()` to use the umbrella sentinel
- Updated the validation error tests and added the requested restore regression test for the missing PITR date case

I've also run the relevant validation tests, controller tests, and the full repository test suite locally, and everything is passing. Let me know if you notice anything else that should be adjusted...!

recharte (approved):
Nice work on this — thanks for your patience through the review.

Issue #2659, opened 2026-07-29
Add `everestctl backupstorage create` subcommand


recharte:
### Is your feature request related to a problem?

The CLI backup workflow is not usable end-to-end: `backup create --storage my-s3` presumes a BackupStorage that can only be created through the UI or by hand-crafting the CR plus its credentials Secret with kubectl. Operators bootstrapping a cluster from a shell hit a wall at the very first step.

### Describe the solution you'd like

Add a `backupstorage create` subcommand to the existing command group. Required flags: `--name`, `--namespace`, `--type` (currently `s3`), `--bucket`, `--region`, `--endpoint-url` (required by the schema, with server-side URL validation); optional: `--verify-tls` (default true), `--force-path-style` (default false). Credentials, two modes (mutually exclusive, exactly one required):

- `--credentials-secret` — reference an existing Secret (`spec.credentialsSecretRef`), for users managing secrets externally;
- `--access-key-id` with the secret key supplied via environment variable or interactive prompt (never a plain flag, to keep it out of shell history/process lists) — the CLI sets the spec's write-only `accessKeyId`/`secretAccessKey` fields and defaults `credentialsSecretRef` to `<name>-credentials`; the server-side machinery persists them into that Secret and clears the fields. No client-side Secret management needed.

The command follows existing CLI conventions: `--json`/`--verbose` output modes, `--cluster`, and `--context` flags. A 409 on `--name` is mapped to `backup storage %q already exists in namespace %q`.

### How would users interact with this feature?

```
# Create an S3 backup storage, secret key prompted interactively
everestctl backupstorage create --name my-s3 --namespace everest \
  --type s3 --bucket backups --region eu-west-1 \
  --endpoint-url https://s3.eu-west-1.amazonaws.com --access-key-id AKIA...

# Reference an existing credentials secret
everestctl backupstorage create --name my-minio --namespace everest \
  --type s3 --bucket backups --region us-east-1 \
  --endpoint-url https://minio.local:9000 --credentials-secret minio-credentials
```

### Alternative solutions considered

`kubectl apply` of the BackupStorage CR + Secret — bypasses Everest API authentication/RBAC and requires knowing both schemas and their coupling (secret key names) by heart.

### Additional Context

Applies to v2 (`release-2.0`). Completes the loop for the backup command family (#2578, #2620): with this, an operator can go from empty namespace to restored instance entirely from the CLI. `backupstorage delete` (separate issue) and `update`/`patch` are follow-ups.

[...]

PR #2670, opened 2026-07-30
Add default response to backup storage 

### LINKED ISSUE #2669: [v2] Add `default` response to `createBackupStorage` and `listBackupStorages`
> **Applies to v2 only** (`release-2.0` branch). File references are against `release-2.0`.

### Task Description

Add a `default` response (schema: `Error`) to the `createBackupStorage` and `listBackupStorages` operations in `api/openapi/http-api.yaml`, then re-run `make gen`.

The change is **purely additive** — keep the existing `'400'`/`'500'` declarations and add `default:` alongside them, exactly as `patchBackupStorage` and `updateBackupStorage` already do. The generated response structs gain a `JSONDefault` field without losing `JSON400`/`JSON500`, so no existing call site breaks.

Also update `pkg/cli/backupstorage/` to use `JSONDefault` for the generic server-error fallback once it exists, so the error handling matches `backup create`/`restore create`.

### Reasoning / Benefit

The BackupStorage endpoints are currently inconsistent with each other:

| Operation | Declared responses |
|---|---|
| `getBackupStorage` | 200, 404, **default** |
| `patchBackupStorage` | 200, 400, 404, **default** |
| `updateBackupStorage` | 200, 400, 404, **default** |
| `deleteBackupStorage` | 204, 400, **default** |
| `createBackupStorage` | 201, 400, 500 |
| `listBackupStorages` | 200, 400, 500 |

Without a `default` arm, the generated client's response parser has no catch-all: any status the spec doesn't list explicitly leaves **every** typed field `nil`. That's not hypothetical — the server returns 409 on `AlreadyExists` (`internal/server/everest.go`, the `IsAlreadyExists` branch) and 401/403 from the auth and RBAC layers, none of which are declared. Consumers are left inspecting raw `Body` bytes or working around it with hand-rolled unmarshalling.

Since the server can surface arbitrary Kubernetes errors on these endpoints, `default` is the appropriate catch-all, and it's already the convention for the newer v2 resources.

### Additional Context

- `api/openapi/http-api.yaml` — the two operations to change; `patchBackupStorage` is the pattern to copy
- `client/everest-client.gen.go` — `ParseCreateBackupStorageResponse` (no catch-all arm) vs `ParseCreateBackupResponse` (`case ... && true` catch-all)
- Came out of the #2659 review, where the missing `JSONDefault` prompted a workaround in the CLI
- **Not blocked by #2659**, and doesn't block it — the spec change is additive (`JSON400`/`JSON500` are kept), so both orderings compile. If #2659 is still open when this is picked up, just coordinate on `pkg/cli/backupstorage/create.go` to avoid a merge conflict; whoever lands second adapts the error handling to use `JSONDefault`.
- **Out of scope:** the same explicit-codes style appears on the older endpoint families (monitoringconfigs, configmaps, secrets, auth/session). Normalizing those is a broader question — worth a separate discussion, since some of them may genuinely have closed response sets.

false200:
## Summary

`createBackupStorage` and `listBackupStorages` were missing a `default` error response in the OpenAPI spec. Every other BackupStorage endpoint already had one. This adds it to both, regenerates the clients, and wires up the list command to use the new catch-all field.

Part of  #2669

## Changes

Added `default: Error` alongside the existing 400 and 500 responses in `api/openapi/http-api.yaml`, matching `patchBackupStorage` and `updateBackupStorage`. Ran codegen for the Go client, server, UI types, and api-tests types. Updated `pkg/cli/backupstorage/list.go` to read `JSONDefault` on failure, same pattern as backup and restore create.

## Testing

- [x] `go test ./pkg/cli/backupstorage/...` passes
- [x] `CreateBackupStorageResponse` and `ListBackupStoragesResponse` include `JSONDefault` in `client/everest-client.gen.go`
- [x] `ui/api/http-api.types.ts` lists `default` on both operations


pkg/cli/backupstorage/list.go
```diff
@@ -121,6 +121,9 @@ func (sl *ListRunner) listBackupStorages(
 		return nil, fmt.Errorf("failed to list backup storages in namespace %q: %w", namespace, err)
 	}
 	if resp.StatusCode() != http.StatusOK || resp.JSON200 == nil {
+		if resp.JSONDefault != nil && resp.JSONDefault.Message != nil {
+			return nil, fmt.Errorf("server error: %s", *resp.JSONDefault.Message)
+		}
```
recharte: @VijetaPriya47 how do you think we should handle the error cases? Should we have specific cases for 409, 400 and 500 and a catch all for everything else as initially suggested [here](https://github.com/openeverest/openeverest/issues/2659#issuecomment-5129086792) or do you suggest something else?

VijetaPriya47: @recharte  Yeah, I think we should mirror what you suggested for `create`, explicit branches for what's declared, `JSONDefault` as the catch-all for what isn't:

@false200 

```go
 if resp.JSON400 != nil && resp.JSON400.Message != nil {
     return nil, fmt.Errorf("server error: %s", *resp.JSON400.Message)
 }
 if resp.JSON500 != nil && resp.JSON500.Message != nil {
     return nil, fmt.Errorf("server error: %s", *resp.JSON500.Message)
 }
 if resp.JSONDefault != nil && resp.JSONDefault.Message != nil {
     return nil, fmt.Errorf("server error: %s", *resp.JSONDefault.Message)
 }
 return nil, fmt.Errorf("unexpected response listing backup storages in namespace %q: %s", namespace, resp.Status())
 ```

No 409 branch needed here, that was specific to `create`'s "already exists" case, `list` doesn't have an equivalent conflict scenario.
The generated `ParseListBackupStoragesResponse` is a switch matching `StatusCode == 400`/`== 500` *before* falling through to the `JSONDefault` catch-all case, and Go switches stop at the first match. So `JSONDefault` never actually fires for a real 400 or 500, only for statuses the spec doesn't declare at all (401/403/etc.). The current code (in this PR) silently drops the message on exactly the two response shapes the spec calls out explicitly, and only helps on the undeclared ones. 

TLDR: Check `JSON400`/`JSON500` explicitly first.

**Comment by VijetaPriya47** [comment]
@false200 Also, the issue asks to update pkg/cli/backupstorage/ , and this PR's own title says "create", but only list.go was touched. You don't need to add create.go here, that's already being worked on separately #2659, so drop the auto-close keyword ( Part of #2669) and I would add create.go as its own PR.

File: pkg/cli/backupstorage/list.go (hunk 1 above)
Comment (VijetaPriya47) [reply pkg/cli/backupstorage/list.go]: @recharte  we could do this also, WDYT?
A tiny helper in pkg/cli/clienterr (same shape as pkg/cli/clientmeta)

```go
 // Message returns the first message found among the given candidates
 // (typically a response's JSON400/JSON500/JSONDefault fields, checked in that
 // order), or false if none carry one.
 func Message(candidates ...*client.Error) (string, bool) {
     for _, c := range candidates {
         if c != nil && c.Message != nil {
             return *c.Message, true
         }
     }
     return "", false
}

Usage collapses to one call instead of three ifs:
if msg, ok := serverErrorMessage(resp.JSON400, resp.JSON500, resp.JSONDefault); ok {
    return nil, fmt.Errorf("server error: %s", msg)
}
return nil, fmt.Errorf("unexpected response listing backup storages in namespace %q: %s", namespace, resp.Status())
```

false200:
> @false200 Also, the issue asks to update pkg/cli/backupstorage/ , and this PR's own title says "create", but only list.go was touched. You don't need to add create.go here, that's already being worked on separately #2659, so drop the auto-close keyword ( Part of #2669) and I would add create.go as its own PR.

Sure.

pkg/cli/backupstorage/list.go
false200: > @recharte Yeah, I think we should mirror what you suggested for `create`, explicit branches for what's declared, `JSONDefault` as the catch-all for what isn't:
> 
> @false200
> 
> ```go
>  if resp.JSON400 != nil && resp.JSON400.Message != nil {
>      return nil, fmt.Errorf("server error: %s", *resp.JSON400.Message)
>  }
>  if resp.JSON500 != nil && resp.JSON500.Message != nil {
>      return nil, fmt.Errorf("server error: %s", *resp.JSON500.Message)
>  }
>  if resp.JSONDefault != nil && resp.JSONDefault.Message != nil {
>      return nil, fmt.Errorf("server error: %s", *resp.JSONDefault.Message)
>  }
>  return nil, fmt.Errorf("unexpected response listing backup storages in namespace %q: %s", namespace, resp.Status())
> ```
> 
> No 409 branch needed here, that was specific to `create`'s "already exists" case, `list` doesn't have an equivalent conflict scenario. The generated `ParseListBackupStoragesResponse` is a switch matching `StatusCode == 400`/`== 500` _before_ falling through to the `JSONDefault` catch-all case, and Go switches stop at the first match. So `JSONDefault` never actually fires for a real 400 or 500, only for statuses the spec doesn't declare at all (401/403/etc.). The current code (in this PR) silently drops the message on exactly the two response shapes the spec calls out explicitly, and only helps on the undeclared ones.
> 
> TLDR: Check `JSON400`/`JSON500` explicitly first.

Oh yeahh, makes sense.

false200: The second option should be good, as it reduces noise from multiple if statements.

false200:
For the fix, waiting for recharte to confirm among both the options available.

VijetaPriya47:
@false200 or better you can raise another PR for create once it gets merged. Will ping you then :>

false200:
Sure, you mean once this is merged right? Then to work on create.

VijetaPriya47:
yup, if the shared helper gets approved and after this PR/

pkg/cli/backupstorage/list.go
recharte: Hey folks, forry for the delay and the back-and-forth I'm creating here — I pushed back on the manual `json.Unmarshal(resp.Body, ...)` over in #2659 and I think I gave you the wrong steer. Let me try again properly.

What I objected to was real, but I aimed at the wrong target. The problem with the hand-rolled unmarshal wasn't the unmarshal, it was that it lived at the call site — which meant every new subcommand would grow its own copy. But the three-`if` version has exactly the same disease: `JSON400`/`JSON500`/`JSONDefault` are artifacts of **oapi-codegen**, not domain concepts, and we've let a code generator's internal shape become our error-handling API across ~15 call sites. That's why this thread keeps going in circles — every option on the table is a different way of arranging deck chairs at the call site.

@VijetaPriya47's `clienterr.Message(...)` helper is the best of the three (definitely better than three `if`s), and if we want to unblock this PR today I'm fine with it. But I don't think it's where we should end up.

The thing I'd actually build: convert HTTP failures into typed Go errors once, at the transport boundary, and never look at `JSONxxx` again. That's what `client-go` does (`*k8serrors.StatusError`), what AWS SDK v2 does (`smithy.APIError`), what `googleapi.CheckResponse` does. And we already have the seam — `NewAPIClient` is the single construction site and `client.WithHTTPClient` takes any `HttpRequestDoer`, and I checked that errors from the Doer propagate untouched through every generated `*WithResponse` method:

```go
type Doer struct{ Base client.HttpRequestDoer }

func (d *Doer) Do(req *http.Request) (*http.Response, error) {
	resp, err := d.Base.Do(req)
	if err != nil || resp.StatusCode < 400 {
		return resp, err
	}
	defer resp.Body.Close() //nolint:errcheck

	body, err := io.ReadAll(io.LimitReader(resp.Body, maxErrorBody))
	if err != nil {
		return nil, fmt.Errorf("read error response: %w", err)
	}
	e := &Error{StatusCode: resp.StatusCode, Status: resp.Status,
		Method: req.Method, Path: req.URL.Path, Body: body}
	var payload client.Error
	if json.Unmarshal(body, &payload) == nil && payload.Message != nil {
		e.Message = *payload.Message
	}
	return nil, e
}
```

plus `IsNotFound` / `IsAlreadyExists` / `IsForbidden` predicates over `errors.As`, mirroring the `k8serrors` idiom we all already know — which is fitting, since `everest.go` derives these status codes from those very errors on the server side.

Call sites then collapse to:

```go
resp, err := c.ListBackupStoragesWithResponse(ctx, cluster, ns)
if err != nil {
	return nil, fmt.Errorf("list backup storages in namespace %q: %w", ns, err)
}
```

and for create:

```go
if err != nil {
	if apierr.IsAlreadyExists(err) {
		return fmt.Errorf("backup storage %q already exists in namespace %q", opts.Name, ns)
	}
	return fmt.Errorf("create backup storage %q: %w", opts.Name, err)
}
```

It also composes fine with the existing `authTransport` — that stays below as a `RoundTripper` doing token refresh, so a 401 that survives a refresh just surfaces as a normal typed error.

To be upfront about the thing I got wrong: yes, this decodes the body directly, which is what I told you not to do. The difference I should have articulated the first time is one reviewed function at the boundary vs. a pattern that spreads. My "it'll become the thing everyone copies" worry was about the latter, and I incorrectly generalized it into "never touch the body". The other trade-off worth naming: we'd give up per-endpoint typed error payloads. Costs us nothing today since every error response `$ref`s the same `Error` schema, and keeping the raw `Body` on the error type means a future richer schema is still decodable at the one call site that needs it.

So concretely, what I'd suggest:

1. Land this PR with @VijetaPriya47's helper — it's a strict improvement and I don't want to block the backup-storage work on a refactor.
2. I'll file the `apierr` issue and we do the sweep separately.
3. Either way, please add a test asserting the message survives a real `400` — list_test.go currently only asserts on `"unexpected response"`, which is precisely the assertion that let this slip through.

@false200 sorry for the extra laps on your first PR here — the churn is on the design not on your patch, which was a perfectly reasonable read of the issue as written.

false200: Sure, I'll update the PR to use clienterr.Message for now along with the unit tests.

false200: Do you want me to add uni tests separately for clienterr.go which is the file I created to have the Message function in? For now I have added tests in list_test.go to test the list.go which has the function (clienterr.Message(...)) imported and used.

recharte: Well, `clienterr.Message()` doesn't have a lot of logic so it's not like we need extensive tests, I'll leave it up to you do decide if it's worth adding just a couple of simple cases.

pkg/kubernetes/resources.go
```diff
@@ -120,7 +120,7 @@ func (k *Kubernetes) getEKSVolumeCount(node corev1.Node) (uint64, error) {
 	}
 	// ... however, if the node type is one of M5, C5, R5, T3, Z1D it's 25.
 	limitedVolumesSet := map[string]struct{}{
-		"m5": {}, "c5": {}, "r5": {}, "t3": {}, "z1d": {},
+		"m5": {}, "c5": {}, "r5": {}, "t3": {}, "t1d": {},
```
recharte: Please make sure to rebase your branch properly as this is reverting commit e55e4e1fccd0770ffb5a4567f65180111a05c58b that we cherry-picked a few hours ago.

pkg/cli/backupstorage/list.go
false200: Sure.

pkg/kubernetes/resources.go
false200: Sorry about that, I hadn't synced my branch with the latest release 2.0, so the old t1d change slipped back in. I'll fix it and push an update. Thanks for pointing it out.

pkg/cli/clienterr/clienterr.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Package clienterr provides helpers for reading error bodies from generated
+// OpenAPI client responses.
+package clienterr
+
+import "github.com/openeverest/openeverest/v2/client"
+
+// Message returns the first message found among the given candidates
+// (typically resp.JSON400, resp.JSON500, resp.JSONDefault), or false if none do.
+func Message(candidates ...*client.Error) (string, bool) {
```
recharte: Could we make the drift hazard explicit here? Something like:
```suggestion
// Message returns the first message found among the given candidates, or false
// if none carry one. Callers pass the error fields their operation declares in
// the OpenAPI spec (typically resp.JSON400, resp.JSON500, resp.JSONDefault) —
// when an operation gains a declared status, its call sites must be updated to
// pass the new field, or the message will be silently dropped again.
func Message(candidates ...*client.Error) (string, bool) {
```
The current wording reads like the three-field list is universal, and the next person will copy it onto an endpoint that also declares a 409 — which is exactly the bug this PR is fixing. 🙂

pkg/cli/clienterr/clienterr_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestMessage(t *testing.T) {
+	t.Parallel()
+	badRequest := "bad request"
+	msg, ok := clienterr.Message(&client.Error{Message: &badRequest})
+	require.True(t, ok)
+	assert.Equal(t, badRequest, msg)
+
+	_, ok = clienterr.Message(nil, &client.Error{})
+	assert.False(t, ok)
+}
```
recharte, line 42: Small one: the doc says "returns the first message found", but we never pass two populated candidates, so the ordering behaviour isn't actually covered. Since that ordering is the whole point of the helper, could we add something like:

```suggestion

	first, second := "first", "second"
	msg, ok = clienterr.Message(nil, &client.Error{Message: &first}, &client.Error{Message: &second})
	require.True(t, ok)
	assert.Equal(t, first, msg)
}
```

pkg/cli/backupstorage/list_test.go
```diff
@@ -189,6 +190,26 @@ func TestBackupStorageList_ServerError(t *testing.T) {
 	assert.Contains(t, err.Error(), "unexpected response")
 }
 
+func TestBackupStorageList_ErrorResponses(t *testing.T) {
```
recharte, line 175: Nice to see a real message assertion here. One gap though: `TestBackupStorageList_ServerError` above writes a bare 500 with no body, so it exercises the fallback path rather than the message path — which means **400 is the only status whose message we actually assert**, and a real 500 from the API does carry a JSON body.

Could we make this table-driven over 400 / 500 / an undeclared status like 403? That covers all three fields we pass to the helper, and it fixes the plural test name currently describing a single case.

recharte (dismissed):
Looks good 💪 

Left three small nits (a doc-comment clarification, one missing test assertion, one test-coverage gap) — none of them blocking but it would be great if you could fold them in before merge.

false200:
Sure, I'll have them enhanced.

pkg/cli/clienterr/clienterr_test.go
false200: Sure.

pkg/cli/backupstorage/list_test.go
false200: Alright, done.

recharte (approved):
Looks great @false200 🚀  thank you for the quick turnaround addressing the review comments.

false200:
Glad to have it fixed.

PR #2671, opened 2026-07-30
everestctl: add backup-storage create subcommand

### LINKED ISSUE #2659: Add `everestctl backupstorage create` subcommand

VijetaPriya47:
Operators bootstrapping a cluster from a shell had no way to create a BackupStorage without either using the UI or hand-crafting the CR plus its credentials Secret with kubectl. This adds `everestctl backup-storage create`, with credentials provided either by pointing at an existing Secret (--credentials-secret) or by an access key id plus a secret access key that's never accepted as a plain flag -- supplied via an environment variable, or an interactive masked prompt when running on a real terminal.

The prompt is gated on whether stdin is actually a terminal, not on whether --json/--verbose was passed: --verbose alone no longer blocks an interactive session, and piped or redirected stdin (a CI job that forgot --json, for instance) now fails loudly with an actionable message instead of silently hanging on a prompt nobody can answer.

Error handling on create keeps the explicit conflict-on-name-reuse message and reads the structured message the server provides for 400/500 responses, falling back to the plain status line for anything else -- deliberately not parsing the raw response body for unmapped statuses.
Closes #2659


VijetaPriya47:
@recharte check it out once

pkg/cli/backupstorage/create.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		switch {
+		case opts.CredentialsSecret != "":
+			bs.Spec.S3.CredentialsSecretRef.Name = opts.CredentialsSecret
+		case opts.AccessKeyID != "":
+			bs.Spec.S3.AccessKeyId = &opts.AccessKeyID
+			bs.Spec.S3.SecretAccessKey = &opts.SecretAccessKey
+			// CredentialsSecretRef.Name is left empty; the server defaults it
+			// to "<name>-credentials" (internal/server/handlers/k8s/backup_storage.go).
```
recharte: Leaving `credentialsSecretRef.name` empty means this mode never gets past request validation: the field is `required` with `minLength: 1` on the CRD schema, the create body `$ref`s that schema, and `apiGroup.Use(middleware.OapiRequestValidatorWithOptions(...))` (`internal/server/everest.go:254`) runs before any handler. So:

```
$ everestctl backup-storage create --name my-s3 --access-key-id AKIA... ...
Error: server error: request body has an error: doesn't match schema:
  Error at "/spec/s3/credentialsSecretRef/name": minimum string length is 1
```

And the comment isn't right either — there's no defaulting in `internal/server/handlers/k8s/backup_storage.go`. `applyS3Credentials` uses `s3.CredentialsSecretRef.Name` verbatim as the Secret's name, so even if the request got through it would try to create a Secret with an empty name.

That's my fault: the issue told you the server would do it, and it doesn't. Sorry for the wasted detour.

The issue is also wrong about the *name*. It says `<name>-credentials`, but the UI has been deriving this client-side for a while, with a different string:

```ts
// ui/apps/everest/src/api/backupStorage.ts:34
credentialsSecretRef: {
  name: `backup-storage-${formValues.name}-credentials`,
},
```

Since the Secret is user-visible in their namespace, having the CLI and the UI disagree is the worse outcome, and the name is expensive to change once shipped. So let's follow the UI:

```suggestion
		case opts.AccessKeyID != "":
			bs.Spec.S3.AccessKeyId = &opts.AccessKeyID
			bs.Spec.S3.SecretAccessKey = &opts.SecretAccessKey
			// Matches the UI's convention (ui/apps/everest/src/api/backupStorage.ts)
			// so both clients name the credentials Secret identically.
			bs.Spec.S3.CredentialsSecretRef.Name = "backup-storage-" + opts.Name + "-credentials"
```

`TestRun_HappyPath_AccessKeyID` needs flipping too — `assert.Empty(t, gotBody.Spec.S3.CredentialsSecretRef.Name)` currently locks the broken behaviour in. `assert.Equal(t, "backup-storage-my-s3-credentials", ...)` is the assertion I'd want.

Longer term we need to move the defaulting to somewhere on the backend but that's a separate issue, don't let it hold this PR up.

commands/backupstorage/create.go
```diff
@@ -0,0 +1,160 @@
...
+
+  - --credentials-secret: reference an existing Secret (in the same
+    namespace) that already holds the storage credentials.
+  - --access-key-id plus a secret access key supplied via the
+    EVEREST_BACKUP_STORAGE_SECRET_ACCESS_KEY environment variable, or an
+    interactive prompt if the env var is unset and stdin is a terminal
+    (skipped for --json, or when stdin is piped/redirected, e.g. in CI).
+    The secret access key is never accepted as a plain flag. The server
+    stores it in a Secret named "<name>-credentials" and never persists
+    it on the BackupStorage object.
```
recharte: > The server stores it in a Secret named "&lt;name&gt;-credentials" and never persists it on the BackupStorage object.

Second half is true — `applyS3Credentials` clears both write-only fields before the CR is persisted. The naming isn't: the server uses whatever `credentialsSecretRef.name` says, and has no default. Again, that came from the issue, which was wrong; sorry.

Once the CLI derives the name (see the comment on `pkg/cli/backupstorage/create.go`), this becomes accurate with the UI's string:

```suggestion
    The secret access key is never accepted as a plain flag. The CLI
    points credentialsSecretRef at "backup-storage-<name>-credentials"
    (the same name the UI uses); the server writes the credentials into
    that Secret and never persists them on the BackupStorage object.
```

pkg/cli/backupstorage/create.go
```diff
@@ -0,0 +1,171 @@
...
+// createStorageError maps a non-201 response to an error.
+func createStorageError(resp *client.CreateBackupStorageResponse) error {
+	if resp.JSON400 != nil && resp.JSON400.Message != nil {
+		return fmt.Errorf("server error: %s", *resp.JSON400.Message)
+	}
+	if resp.JSON500 != nil && resp.JSON500.Message != nil {
+		return fmt.Errorf("server error: %s", *resp.JSON500.Message)
+	}
+	return fmt.Errorf("unexpected response creating backup storage: %s", resp.Status())
+}
```
recharte, line 160: Let's switch to using `clienterr.Message()` just like `list.go` does:

```suggestion
// createStorageError maps a non-201 response to an error.
func createStorageError(resp *client.CreateBackupStorageResponse) error {
	if msg, ok := clienterr.Message(resp.JSON400, resp.JSON500); ok {
		return fmt.Errorf("server error: %s", msg)
	}
	return fmt.Errorf("unexpected response creating backup storage: %s", resp.Status())
}
```

commands/backupstorage/create.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func createPreRun(cmd *cobra.Command, _ []string) {
+	createCfg.Pretty = !cmd.Flag(cli.FlagVerbose).Changed && !cmd.Flag(cli.FlagJSON).Changed
+}
+
+// createRun resolves the secret access key (env var, else interactive prompt) before
+// building the request.
+func createRun(cmd *cobra.Command, _ []string) {
+	if createOpts.AccessKeyID != "" {
```
recharte, line 131: The secret key gets prompted before `--type` is validated: `resolveSecretAccessKey` runs here, then the API client is built, and only then does `newBackupStorage` reach the `switch opts.Type`. So a typo'd type takes the credential first and rejects afterwards:

```
$ everestctl backup-storage create --type gcs --access-key-id AKIA... ...
❓ Secret access key: ********
Error: unsupported storage type "gcs" (only "s3" is supported today)
```

Asking someone to hand over a credential before telling them the command was never going to work is a poor trade. Cheapest fix is to reject the type in `createPreRun`, or lift the `switch` out of `newBackupStorage` into a small `validate()` called before `resolveSecretAccessKey`.

commands/backupstorage/create.go
```diff
@@ -0,0 +1,160 @@
...
+// resolveSecretAccessKey returns the secret access key from the environment, falling
+// back to an interactive prompt when running on a real terminal.
+func resolveSecretAccessKey(cmd *cobra.Command) (string, error) {
+	if v := os.Getenv(envSecretAccessKey); v != "" {
+		return v, nil
+	}
+	if cmd.Flag(cli.FlagJSON).Changed || !term.IsTerminal(int(os.Stdin.Fd())) {
+		return "", fmt.Errorf("%s must be set: not running interactively", envSecretAccessKey)
+	}
+	return tui.NewInputPassword(cmd.Context(), "Secret access key").Run()
```
recharte: Pressing Enter at the prompt yields `""`, which still goes on the wire as `"secretAccessKey": ""` — `omitempty` on a `*string` only skips a nil pointer, and this one points at an empty string. The server then hits:

```go
// internal/server/handlers/k8s/backup_storage.go
case s3.AccessKeyID != "" && s3.SecretAccessKey == "":
	return errors.New("secretAccessKey is not provided")
```

which surfaces a round-trip later as a wrapped `failed to apply S3 credentials`. `tui` already has the knob for catching it locally — `pkg/accounts/cli/utils.go` uses it:

```go
return tui.NewInputPassword(cmd.Context(), "Secret access key",
	tui.WithPasswordValidation(func(s string) error {
		if s == "" {
			return errors.New("secret access key cannot be empty")
		}
		return nil
	}),
).Run()
```

recharte (changes requested):

pkg/cli/backupstorage/create.go
```diff
@@ -0,0 +1,171 @@
...
+// created backup storage in JSON mode (so `create --json` is parseable either way).
+func (cr *CreateRunner) emitCreated(created *client.BackupStorage, name string) error {
+	if cr.config.Pretty {
+		_, _ = fmt.Fprint(os.Stdout, output.Success("Backup storage %q created", name))
+		return nil
+	}
+	// A 201 with an unparseable body would otherwise emit empty stdout.
+	if created == nil {
+		return fmt.Errorf("backup storage %q was created but the server returned an unreadable response body", name)
+	}
```
recharte: One more I missed on the first pass. This branch can't be reached: `Run` already returns `createStorageError(resp)` when `resp.JSON201 == nil` (line 87), so `emitCreated` never sees a nil `created`. `TestRun_JSONErrorsOnEmptyCreateBody` confirms it — it asserts on `"unexpected response creating backup storage"`, which is the *upstream* message, not this one. Same story for the `if bs == nil` guard in `writeBackupStorageJSON` below.

The shape is copied from `pkg/cli/backup/create.go` where it's equally dead, so I'd appreciate if you could handle it there as well.

VijetaPriya47:
@recharte resolved all the issues, and tested with kind cluster. could you check again.

recharte:
Thanks, everything looks good from the backup storage side.

One thing before I do the final pass: 9d79688 ("feat(instance): add instance delete command") looks like it landed on this branch by accident — 1,026 lines across 10 files, none of it related to #2659. Could you move it onto its own branch and open a separate PR with its own issue? I'll pick that one up next.

VijetaPriya47:
yeah, I saw. Was about to do the same 😅. I was testing on this branch that's why.

pkg/cli/backupstorage/create.go
```diff
@@ -0,0 +1,171 @@
...
+		case opts.CredentialsSecret != "":
+			bs.Spec.S3.CredentialsSecretRef.Name = opts.CredentialsSecret
+		case opts.AccessKeyID != "":
+			bs.Spec.S3.AccessKeyId = &opts.AccessKeyID
+			bs.Spec.S3.SecretAccessKey = &opts.SecretAccessKey
+			bs.Spec.S3.CredentialsSecretRef.Name = "backup-storage-" + opts.Name + "-credentials"
+		}
+	default:
+		return nil, ValidateType(opts.Type)
+	}
```
recharte, line 149: `default: return nil, ValidateType(opts.Type)` is only correct today because `ValidateType` rejects everything that isn't `"s3"`. Two things worth doing here.

**Use the generated enum for the check.** The client already ships `client.BackupStorageSpecTypeS3` and a `Valid()` method on `BackupStorageSpecType`, both regenerated from the spec:

```go
// client/crds.gen.go
func (e BackupStorageSpecType) Valid() bool {
	switch e {
	case BackupStorageSpecTypeS3:
		return true
	default:
		return false
	}
}
```

So `ValidateType` doesn't need to hold its own copy of the valid set — it can just ask:

```go
func ValidateType(t string) error {
	if !client.BackupStorageSpecType(t).Valid() {
		return fmt.Errorf("unsupported storage type %q", t)
	}
	return nil
}
```

The hand-written `t != "s3"` has to be found and updated by hand the day the spec gains a second type; this version updates itself with `make gen`. The error message has to lose the `(only "s3" is supported today)` parenthetical for the same reason — it would go stale too.

**Then let the `default` arm fail on its own terms.** Once `ValidateType` accepts more than one type, a `switch` here that hasn't grown the matching `case` returns `(nil, nil)`, and `Run` dereferences it a few lines later:

```go
bs, err := newBackupStorage(opts)   // (nil, nil)
if err != nil { return err }        // not taken
resp, err := c.CreateBackupStorageWithResponse(ctx, opts.Cluster, opts.Namespace, *bs)  // panic
```

This arm means "this switch has no branch for that type", which is an error whatever `ValidateType` thinks:

```suggestion
	default:
		return nil, fmt.Errorf("no request builder for storage type %q", opts.Type)
	}
```

recharte (dismissed):
Sorry — I said this was ready to go and then found one more thing worth addressing. It's small and not a bug today, just a trap left for whoever adds the second storage type. Everything else from the last round checks out.

pkg/cli/backupstorage/create.go
```diff
@@ -0,0 +1,171 @@
...
+func ValidateType(t string) error {
+	if t != "s3" {
+		return fmt.Errorf("unsupported storage type %q (only \"s3\" is supported today)", t)
+	}
+	return nil
+}
+
+// newBackupStorage builds the request body for a create call.
+func newBackupStorage(opts CreateOptions) (*client.BackupStorage, error) {
+	bs := client.BackupStorage{
```
recharte, line 113: @VijetaPriya47 the unit tests CI is failling. My suggestion changed the message here, but we didn't check that `TestRun_UnsupportedType` asserts on the old one, so applying it turned CI red.

There's something real hiding underneath it, though. That `default` arm was quietly doing two jobs: guarding the switch and validating the type. Now that it only does the first, nothing validates `opts.Type` on this path — `createRun` does it, but anything calling `Run` directly, like the unit test, walks straight past.

Easiest fix is to validate at the top of the builder:

```suggestion
func newBackupStorage(opts CreateOptions) (*client.BackupStorage, error) {
	if err := ValidateType(opts.Type); err != nil {
		return nil, err
	}

	bs := client.BackupStorage{
```

That gets the good error message back, the test passes unchanged (I ran it locally), and the `default` arm goes back to being a pure "shouldn't happen" guard.

VijetaPriya47: okay, let me apply the changes, and test it on my local too. In the meantime could you check upon the instance delete? @recharte

VijetaPriya47: okay could you check it out now

recharte (dismissed):
Everything looks good now, thank you 🤗

VijetaPriya47:
okay now it's good to go @recharte

recharte (approved):

PR #2719, opened 2026-07-31
Fix init container resource counting in GetConsumedCPUAndMemory

### LINKED ISSUE #2720: GetConsumedCPUAndMemory undercounts resources when some init containers are still running
### What is the issue?

`GetConsumedCPUAndMemory` skips init container CPU/memory requests when **any** init container has terminated, instead of checking each init container individually.

When a pod has multiple init containers and some have finished while others are still running, the running init containers' resource requests are **not included** in cluster resource calculations.

**Expected:** Count running init containers + main containers  
**Actual:** Once the first init container finishes, **all** init container requests are ignored

### Steps to Reproduce

1. Create a running pod with:
   - `init-a` - terminated, CPU request `100m`
   - `init-b` - still running, CPU request `200m`
   - `app` - main container, CPU request `500m`
2. Call `GetConsumedCPUAndMemory` for the pod's namespace.

**Expected:** `700m` CPU (200m + 500m)  
**Actual:** `500m` CPU - running `init-b` is ignored

3. Root cause in `pkg/kubernetes/resources.go`:

```go
if !IsContainerInState(ppod.Status.InitContainerStatuses, ContainerStateTerminated) {
```

`IsContainerInState` returns true if **any** init container terminated, not the specific one being checked.

### Relevant Logs / Output

Verified with unit test `TestGetConsumedCPUAndMemoryCountsRunningInitContainers`:

```bash
go test -race ./pkg/kubernetes/... -run TestGetConsumedCPUAndMemoryCountsRunningInitContainers -v
# Before fix: 500m (bug)
# After fix:  700m (correct)
```

### OpenEverest Version

main (2026-07-31)

### Installation Method

Helm

### Kubernetes Distribution

Other (please specify below)

### Environment Details

Kubernetes Distribution: Any (bug is in Go resource aggregation logic, not K8s-specific)
Local test cluster: k3d v1.36.x
Affected file: pkg/kubernetes/resources.go
Affected function: GetConsumedCPUAndMemory
Host OS: Ubuntu Linux
Go Version: 1.26.3

### Possible Solution

Check termination state per init container using `IsNamedContainerInState(name, statuses, state)` instead of `IsContainerInState` on the full status list.

Fix submitted in PR #2719.

### Additional Context

- No existing GitHub issue found for this bug
- PR: https://github.com/openeverest/openeverest/pull/2719

Jay2006sawant:
## Summary

Fixes incorrect cluster resource calculation when a Running pod has a terminated regular init container together with a still-running native sidecar (`initContainers[].restartPolicy: Always`).

`GetConsumedCPUAndMemory` used `IsContainerInState` on the full init container status list, which returned true if **any** init container had terminated. That dropped every init container request once the first regular init finished, so the sidecar's CPU/memory requests were missing from consumed resources. Because `available = capacity - consumed`, that undercount inflated reported cluster headroom.

## Changes

* Update `IsContainerInState` to check a named container instead of any container in the list
* Replace the JSON state round trip with direct `Waiting` / `Terminated` nil checks
* Count non-terminated init containers (including native sidecars) in `GetConsumedCPUAndMemory`
* Add unit tests for the helper and a sidecar regression fixture

## Test Plan

* [x] `go test -race ./pkg/kubernetes/... -run 'TestIsContainerInState|TestGetConsumedCPUAndMemory' -v`
* [x] `go test -race ./pkg/kubernetes/...`
* [x] `make copyright-check` on changed files

Fixes #2720

### DIFF
--- a/pkg/kubernetes/container_test.go
+++ b/pkg/kubernetes/container_test.go
@@ -1,0 +1,53 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kubernetes
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestIsContainerInState(t *testing.T) {
+	t.Parallel()
+
+	statuses := []corev1.ContainerStatus{
+		{
+			Name: "init-a",
+			State: corev1.ContainerState{
+				Terminated: &corev1.ContainerStateTerminated{},
+			},
+		},
+		{
+			Name: "init-b",
+			State: corev1.ContainerState{
+				Running: &corev1.ContainerStateRunning{},
+			},
+		},
+		{
+			Name: "init-c",
+			State: corev1.ContainerState{
+				Waiting: &corev1.ContainerStateWaiting{},
+			},
+		},
+	}
+
+	assert.True(t, IsContainerInState(statuses, "init-a", ContainerStateTerminated))
+	assert.False(t, IsContainerInState(statuses, "init-b", ContainerStateTerminated))
+	assert.False(t, IsContainerInState(statuses, "init-missing", ContainerStateTerminated))
+	assert.True(t, IsContainerInState(statuses, "init-c", ContainerStateWaiting))
+	assert.False(t, IsContainerInState(statuses, "init-a", ContainerStateWaiting))
+}
--- a/pkg/kubernetes/resources_test.go
+++ b/pkg/kubernetes/resources_test.go
@@ -17,12 +17,17 @@
 package kubernetes
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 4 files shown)


pkg/kubernetes/resources_test.go
```diff
@@ -89,6 +93,54 @@ func TestGetEKSVolumeCount(t *testing.T) {
...
+
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "db-pod",
+			Namespace: "everest",
+		},
+		Spec: corev1.PodSpec{
+			InitContainers: []corev1.Container{
+				containerWithCPURequest("init-a", "100m"),
+				containerWithCPURequest("init-b", "200m"),
```
recharte: The regression fixture can't occur in a real cluster: `init-b` is a plain init container, and the kubelet holds a pod in `Pending` until every non-restartable init container has terminated, so a `Running` pod with a plain init container still running never exists. The fake client doesn't validate it, so the test passes either way — but the case this fix actually addresses is a native sidecar, `initContainers[].restartPolicy: Always`, which stays Running for the pod's whole life.

Marking `init-b` as a sidecar costs one line and makes the fixture describe the state that reaches this code path:

```go
initB := containerWithCPURequest("init-b", "200m")
initB.RestartPolicy = ptr.To(corev1.ContainerRestartPolicyAlways)
```

pkg/kubernetes/container_test.go
```diff
@@ -0,0 +1,55 @@
+// everest
+// Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
```
recharte: `scripts/add_copyright.py` prepends a header to a brand-new file with the OpenEverest line only — the `// everest` marker and the Percona line exist in `resources_test.go` because it pre-dates the fork and the script inserted the OpenEverest line beneath what was already there. `make copyright-check` passes either way since it only looks for the OpenEverest line.

```suggestion
// Copyright (C) 2026 The OpenEverest Contributors
```

recharte (commented):
The description still describes the first approach — `IsNamedContainerInState` plus a `containerStatusHasState` helper — and none of that is in the diff any more. Worth updating before merge, since it lands as the squash commit message.

Jay2006sawant:
Hey @recharte, thanks for the review!!
 have updated the PR description to match the current approach, marked init-b as a native sidecar in the test fixture, and fixed the copyright header on the new test file. Ready for another look when you have a minute.

recharte (approved):
Both points addressed, thanks — and the description reads much better now.

Issue #2660, opened 2026-07-29
Add `everestctl backupstorage delete` subcommand


recharte:
### Is your feature request related to a problem?

BackupStorages can be listed from the terminal but not removed — decommissioning a storage (rotated credentials, migrated bucket) still requires the UI or `kubectl delete`, bypassing Everest API authentication/RBAC.

### Describe the solution you'd like

Add a `backupstorage delete` subcommand. Required flags: `--name`, `--namespace`. Destructive, so it follows the shared confirmation pattern established by `instance delete` (#2658), at the resource-removal tier:

- **Interactive (stdin is a TTY, no `--yes`):** a y/N prompt, default **No**. Declined exits non-zero without doing anything.
- **`--yes`/`-y`** skips the confirmation.
- **Non-interactive:** if stdin is not a TTY *or* `--json` is set, and `--yes` wasn't given, fail immediately (exit 1) with *"confirmation required in non-interactive mode; re-run with --yes"* — never block on a prompt nobody can answer.
- `--force` is deliberately not introduced; the name stays reserved for overriding server-side refusals (e.g. a storage still referenced), never as a `--yes` synonym.

A 404 is mapped to `backup storage %q not found in namespace %q`; deleting a storage still referenced by instances or backups surfaces the server's error as-is. Follows existing CLI conventions: `--json`/`--verbose`, `--cluster`, `--context`.

### How would users interact with this feature?

```
# Delete a backup storage (prompts y/N)
everestctl backupstorage delete --name my-s3 --namespace everest

# Scripted
everestctl backupstorage delete --name my-s3 --namespace everest -y
```

### Alternative solutions considered

`kubectl delete backupstorage` — bypasses Everest API authentication/RBAC.

### Additional Context

Applies to v2 (`release-2.0`). Reuses the shared confirmation helper introduced with `instance delete` (#2658); whichever lands first establishes it. The credentials Secret's lifecycle needs a decision: delete it when it follows the CLI-managed `<name>-credentials` convention from `backupstorage create` (#2659), leave it when it was externally referenced.

PR #2762, opened 2026-08-02
feat(instance): add instance delete command

### LINKED ISSUE #2658: Add `everestctl instance delete` subcommand

Instances can be created, listed, and watched from the terminal, but not deleted — every scripted lifecycle (CI environments, dev clusters, demo teardowns) still ends in the UI or a hand-crafted `kubectl delete`, bypassing Everest API authentication/RBAC.

Add an `everestctl instance delete` subcommand that deletes an Instance via the Everest API. Required flags: `--name` and `--namespace`. What happens to dependent Backup/Restore CRs is governed by the Instance's `spec.deletionPolicy` (Cascade/Orphan) by default, but the API supports overriding it at delete time (`deleteInstance`'s `deletionPolicy` query parameter) — expose that as an optional `--deletion-policy Cascade|Orphan` flag. When omitted, the policy declared on the Instance applies; the help text should explain the semantics (Cascade deletes all Backup/Restore CRs referencing the instance, Orphan leaves them behind).

This is the first destructive data operation in the instance/backup/restore command family, so it establishes the confirmation pattern the future delete subcommands will reuse:

- **Interactive (stdin is a TTY, no `--yes`):** since this is irreversible data loss, use a type-the-name confirmation (gh/terraform style) rather than y/N. The prompt states the blast radius first, including the *effective* deletion policy (from `--deletion-policy` if given, else the Instance's spec), e.g. *"This permanently deletes instance \"my-mongo\" and, per policy Cascade, all its Backup and Restore objects. Type the instance name to confirm."* A declined/mismatched confirmation exits non-zero without doing anything.
- **Non-interactive:** if stdin is not a TTY *or* `--json` is set, and `--yes` wasn't given, fail immediately (exit 1) with *"confirmation required in non-interactive mode; re-run with --yes"* — never block on a prompt nobody can answer.
- `--force` is deliberately **not** introduced: that name stays reserved for overriding server-side refusals in the future, never as a `--yes` synonym.

Optional flags:
- `--wait` / `--timeout` — block until the instance is actually gone, reusing the shared waiter: poll until the GET returns 404 → exit 0, timeout → exit 124. Exit 1 is reserved for command errors (API, auth, bad flags): there is deliberately **no** "deletion failed" outcome, because the runtime leaves a deleting Instance in `Terminating` with no message and no condition even when cleanup is stuck, so such a branch could never fire. Revisit once the runtime publishes deletion progress. The timeout message must say the deletion is still running server-side and point at `everestctl backup list` / `restore list`, since a stuck cascade is almost always a child that won't finalize. Ctrl-C cancels only the wait — deletion continues server-side (stated in help text).
- `--ignore-not-found` — kubectl/helm-style idempotence for scripted teardown: an already-absent instance exits 0 instead of erroring, short-circuiting both the confirmation prompt and `--wait`. Without the flag a 404 stays exit 1 with `instance %q not found in namespace %q`. The same flag name and semantics apply to `backupstorage delete` (#2660) and future delete subcommands.

A 404 on the initial delete is mapped to `instance %q not found in namespace %q`. The command follows existing CLI conventions: `--json`/`--verbose` output modes, `--cluster`, and `--context` flags.

```
# Delete an instance (states blast radius, asks to type the name)
everestctl instance delete --name my-mongo --namespace everest

# Keep backups around despite the instance's Cascade policy
everestctl instance delete --name my-mongo --namespace everest --deletion-policy Orphan

# Scripted teardown, wait until fully removed
everestctl instance delete --name ci-run-421 --namespace ci -y --wait --timeout 10m
```

`kubectl delete instance` — works but bypasses Everest API authentication/RBAC and skips any policy enforcement at the API layer.

### Additional Context

Applies to v2 (`release-2.0`). Completes the create/list/status/delete lifecycle for the `instance` command group.

The confirmation behavior should live in a shared helper (e.g. `pkg/cli/confirm`, building on the existing `pkg/cli/tui` inputs) with two tiers — type-the-name for data-loss operations, y/N for resource removals — so `backupstorage delete` and future `backup delete`/`restore delete` reuse it instead of growing copies (same rationale as `pkg/cli/wait`/`waitcmd`). As part of establishing `--yes` as the uniform flag, `uninstall` should gain `--yes` as the documented form with `--assume-yes` kept as a compatibility alias.

The `--wait` machinery reuses `pkg/cli/wait` + `pkg/cli/waitcmd`; "wait until 404" is a new condition shape worth adding to the shared waiter rather than a per-command copy.

VijetaPriya47:
Adds `everestctl instance delete`, closing out the create/list/status/delete lifecycle for the instance command group. Deleting through the CLI now goes through the Everest API (auth + RBAC included) instead of falling back to kubectl.

- `--deletion-policy` overrides the instance's own Cascade/Orphan policy for this delete only.
- Since this is the first destructive command in the instance/backup/restore family, the confirmation flow lives in a new shared pkg/cli/confirm package rather than being one-off: type-the-name confirmation in a terminal, --yes to skip it, and a fast, clear failure in non-interactive/--json contexts instead of hanging on a prompt nobody can answer.
- `--ignore-not-found` makes repeated/scripted teardown calls idempotent.
- `--wait`/`--timeout` block until the instance is actually gone. Note: a stuck cascade delete just times out for now (no distinct failure state) - provider-runtime has no way to report "this got stuck" yet, so that's left for a follow-up alongside the corresponding backend change.
- uninstall now takes --yes as its primary flag, with --assume-yes kept around as a deprecated alias so existing scripts don't break.

Tested against a real kind cluster (login, create, delete, --ignore-not-found, --deletion-policy, --wait, and the interactive type-the-name prompt including a wrong-name rejection) in addition to the unit tests.
Closes #2658


VijetaPriya47:

VijetaPriya47:
@recharte could u check it out?

commands/instance/delete.go
```diff
@@ -0,0 +1,132 @@
...
+		output.PrintError(err, logger.GetLogger(), deleteCfg.Pretty)
+		os.Exit(waitcmd.ExitFailed)
+	}
+
+	ctx := cmd.Context()
+	stop := func() {}
+	if deleteOpts.Wait {
+		// Ctrl-C cancels only the wait; deletion continues server-side.
+		var cancel context.CancelFunc
+		ctx, cancel = signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
```
recharte: With `--wait`, the signal-notified context is installed *before* `Run` — so it also wraps the confirmation prompt. `tui.NewInput` passes that ctx to bubbletea via `tea.WithContext`, and bubbletea's `Program.Run` checks `externalCtx.Err()` before the model's own `interrupt` flag:

```go
killed := p.externalCtx.Err() != nil || p.ctx.Err() != nil || err != nil
if killed {
    if err == nil && p.externalCtx.Err() != nil {
        err = fmt.Errorf("%w: %w", ErrProgramKilled, p.externalCtx.Err())
```

So hitting Ctrl-C at the "type the instance name" prompt never yields `tui.ErrUserInterrupted` — `confirm.Name` returns `ErrProgramKilled: context canceled`, `waitcmd.ExitCode` matches `errors.Is(err, context.Canceled)`, and the user gets:

```
wait cancelled; instance "my-mongo" deletion continues in the background — check with 'everestctl instance status'
```

…exit 130, when in fact no DELETE was ever issued and the instance is untouched. On a destructive command that's the worst possible direction for the message to be wrong in.

Two ways out:

1. Install the signal handler only around the wait — move `signal.NotifyContext` into `Deleter.waitForDeletion`, so confirmation runs on the plain command context.
2. In `confirm.Name`/`YesNo`, treat a cancelled context the same as `tui.ErrUserInterrupted` and return `ErrDeclined` — an aborted prompt is a decline, never an aborted wait.

I'd go with the first: nothing before the DELETE has any business being cancellable-but-continuing.

pkg/cli/instance/delete.go
```diff
@@ -0,0 +1,233 @@
...
+	c, err := authcli.NewAPIClient(authcli.Config{Pretty: id.config.Pretty}, id.l.Desugar().Sugar(), cfgPath, opts.Context)
+	if err != nil {
+		return err
+	}
+
+	if opts.IgnoreNotFound && id.instanceAlreadyGone(ctx, c, opts) {
+		return id.emitDeleted(opts)
+	}
+
+	confirmOpts := confirm.Options{Yes: opts.Yes, JSON: !id.config.Pretty, IsTerminal: opts.IsTerminal}
```
recharte: `Config.Pretty` is `!verbose && !json`, so `JSON: !id.config.Pretty` is also true under `--verbose`. That makes

```
everestctl instance delete --name my-mongo --namespace everest --verbose
```

fail on a normal terminal with *"confirmation required in non-interactive mode; re-run with --yes"* — which pushes anyone debugging with `-v` into `--yes`, exactly the habit we don't want around a destructive command.

`--verbose` only raises the log level (`MustInitVerboseLogger`, development config, stderr); it says nothing about stdin. And `backupstorage create` — the one existing prompt in this command family — already gates on the real flag:

```go
// commands/backupstorage/create.go:159
if cmd.Flag(cli.FlagJSON).Changed || !term.IsTerminal(int(os.Stdin.Fd())) {
```

So as it stands `backupstorage create -v` prompts and `instance delete -v` refuses, which is the kind of divergence that's much cheaper to fix now than once three more delete subcommands have copied it.

Could we carry the flag instead of deriving it — a `JSON bool` on `DeleteOptions` set from `cmd.Flag(cli.FlagJSON).Changed`, passed into `confirm.Options`? Output format can keep using `Pretty` as it does today; it's only the interactivity gate that should follow #2658's "not a TTY *or* `--json`".

pkg/cli/instance/delete.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// confirmDeletion shows the blast radius and asks the user to type the name back.
+func (id *Deleter) confirmDeletion(ctx context.Context, c *client.ClientWithResponses, confirmOpts confirm.Options, opts DeleteOptions) error {
+	policy := opts.DeletionPolicy
+	if policy == "" && confirm.WillPrompt(confirmOpts) {
+		policy = id.fetchEffectivePolicyForPrompt(ctx, c, opts)
+	}
+	if policy == "" {
+		policy = string(client.DeleteInstanceParamsDeletionPolicyCascade)
```
recharte, line 178: When `fetchEffectivePolicyForPrompt` can't read the instance (network error, non-200, or just RBAC — delete permission without get permission is a perfectly reachable policy), `policy` falls back to `Cascade` and the prompt then states as fact:

```
... and, per policy Cascade, deletes all its Backup and Restore objects.
```

The blast-radius line is the one thing the user reads before typing the name, so it shouldn't name a policy we didn't actually observe. Falling back to the scarier of the two is the right direction, but "we couldn't check" and "it's Cascade" are different statements.

Could we keep the fallback behaviour but make the wording honest when the lookup failed? Something like *"... and, per the instance's own deletion policy, may delete all its Backup and Restore objects"*, so only a verified policy — or an explicit `--deletion-policy` — is ever named.

pkg/cli/instance/delete.go
```diff
@@ -0,0 +1,233 @@
...
+		return fmt.Errorf("invalid --deletion-policy %q: must be Cascade or Orphan", opts.DeletionPolicy)
+	}
+
+	c, err := authcli.NewAPIClient(authcli.Config{Pretty: id.config.Pretty}, id.l.Desugar().Sugar(), cfgPath, opts.Context)
+	if err != nil {
+		return err
+	}
+
+	if opts.IgnoreNotFound && id.instanceAlreadyGone(ctx, c, opts) {
+		return id.emitDeleted(opts)
```
recharte: The short-circuit itself is right, and matches what we agreed on the issue. But it routes into `emitDeleted`, which prints `Instance "my-db" deleted` and, in JSON mode, `{"name":...,"deleted":true}` — for an instance that this run never deleted, because it wasn't there.

`kubectl delete --ignore-not-found` deliberately says nothing at all in this case, precisely so "deleted" keeps meaning "I deleted it". Since this command sets the pattern for `backupstorage delete` (#2660) and the rest of the family, worth getting it right now: something like `Instance "my-db" not found in namespace "everest"; nothing to delete` in pretty mode, and `"deleted": false` (or a distinct `"notFound": true`) in JSON. Exit code stays 0 either way.

pkg/cli/instance/delete.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	msg := blastRadiusMessage(opts.Name, opts.Namespace, policy)
+	return confirm.Name(ctx, confirmOpts, msg, opts.Name)
+}
+
+// fetchEffectivePolicyForPrompt looks up the instance's policy just to show
+// it in the prompt. On failure it falls back to Cascade — the delete call
+// below is what actually checks the instance exists.
+func (id *Deleter) fetchEffectivePolicyForPrompt(ctx context.Context, c *client.ClientWithResponses, opts DeleteOptions) string {
```
recharte: Small one: with `--ignore-not-found` and no `--yes` on a TTY, this is the second GET of the same object — `instanceAlreadyGone` already fetched it a few lines earlier.

Having `instanceAlreadyGone` return the fetched `*client.Instance` (nil when 404) would let `confirmDeletion` read the policy straight off it. Halves the round-trips, and removes the window where the policy shown in the prompt came from a different read than the existence check.

pkg/cli/instance/delete.go
```diff
@@ -0,0 +1,233 @@
...
+	inst := resp.JSON200
+	// DeletionPolicy is generated as interface{}, so read it as a plain string.
+	policy, ok := inst.Spec.DeletionPolicy.(string)
+	if !ok {
+		return ""
+	}
+	return policy
+}
+
+func blastRadiusMessage(name, namespace, policy string) string {
```
recharte: Could we get a test on this one? It's the single piece of output a user acts on irreversibly, and the Cascade/Orphan branch is exactly the kind of thing that gets inverted in a later refactor without anyone noticing — the tests would all still pass, and the prompt would quietly tell people their backups are safe while they're not.

It's a pure function, so a three-case table (explicit `Orphan`, explicit `Cascade`, and the unresolved fallback) is cheap insurance.

recharte (changes requested):
The delete semantics are right — the exit-code contract, the --wait scoping and the timeout message all match what we settled on #2658, and pulling confirmation into pkg/cli/confirm was the right call.

One blocker: Ctrl-C at the confirmation prompt currently tells the user their instance is being deleted in the background when nothing was deleted at all. The rest is small.

commands/instance/delete.go
VijetaPriya47: gone with the first

VijetaPriya47:

recharte:
> @recharte could you check it out?

@VijetaPriya47 did you forget to push something? I only see a merge commit after my review...

recharte:
@VijetaPriya47 It looks like you pushed all changes to #2765, please bring to this branch the ones that relate to the instance command

VijetaPriya47:
Yup, I cherry picked it here. could you check out now ?

pkg/cli/instance/delete.go
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil {
+		return err
+	}
+
+	if !alreadyGone {
+		id.l.Infof("deleted instance %q in namespace %q", opts.Name, opts.Namespace)
+	}
+
+	if !opts.Wait || alreadyGone {
+		return id.emitDeleted(opts)
```
recharte: The `emitAlreadyGone` split is right, but it only covers the pre-check. When `checkDeleteResponse` returns `alreadyGone` — `--ignore-not-found` and the DELETE itself came back 404 — control still falls into `emitDeleted`, so the user gets `Instance "my-db" deleted` and `{"deleted": true}` for something this run didn't delete. Same statement as the thread above, just on the narrower path.

It's reachable two ways: a genuine race (something else removed the instance between the pre-check and the DELETE), and the case where `checkInstanceExists` returned `(nil, false)` because the GET was a 403 or a transient 500 rather than a 404 — which is exactly the RBAC shape we discussed on the policy lookup, delete permission without read permission.

```go
if alreadyGone {
    return id.emitAlreadyGone(opts)
}
if !opts.Wait {
    return id.emitDeleted(opts)
}
```

That also drops the `|| alreadyGone` from the `--wait` guard, which is only there to stop us waiting for something that was never deleted — `emitAlreadyGone` returning early does that on its own.

pkg/cli/instance/delete_test.go
```diff
@@ -0,0 +1,407 @@
...
+
+	id := NewDeleter(Config{Pretty: false}, zap.NewNop().Sugar())
+	err := id.Run(context.Background(), opts, cfgPath)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "confirmation required in non-interactive mode")
+}
+
+// TestDelete_VerboseAloneDoesNotForceNonInteractive proves --verbose alone
+// does not force non-interactive mode, only opts.JSON should.
+func TestDelete_VerboseAloneDoesNotForceNonInteractive(t *testing.T) {
```
recharte, line 272: This builds its own `confirm.Options` and asserts `WillPrompt` on it — which restates the line from `Run` rather than exercising it. Revert `Run` to `JSON: !id.config.Pretty` and this test still passes, as does every other test in the file: the ones that reach the gate either set `Yes: true`, or set `IsTerminal` false, or pass `JSON: true` explicitly, so none of them discriminate.

So the fix is real but nothing is holding it in place — same failure mode as the blast-radius message, and that one you covered nicely.

The discriminating case is `Config{Pretty: false}` (i.e. `--verbose`), `JSON: false`, `IsTerminal` true, no `--yes`. Under the old behaviour that's `ErrNonInteractive`; under the new one it goes on to attempt a prompt, which fails fast with a bubbletea TTY error in a test process. So:

```go
err := id.Run(context.Background(), opts, cfgPath)
assert.NotErrorIs(t, err, confirm.ErrNonInteractive)
```

Asserting on the *absence* of the sentinel rather than on success keeps it robust — it doesn't care how bubbletea fails without a terminal, only that we got past the gate.

recharte (commented):
Thanks — the Ctrl-C scoping is exactly right, and the blast-radius table is the kind of test that'll still be earning its keep in a year.

Two small things left, both follow-ons from the last round rather than anything new.

recharte (approved):
Good job 🚀

PR #2764, opened 2026-08-02
ci(triage): introduce a triage and self-serve assignment process

recharte:
## Why

Between 6 June and 24 July, 26 issues were opened here by people outside the maintainer team. **Seventeen of them — 65% — never received a reply from a maintainer.** Of the nine that did, the median wait was 3.6 days and the longest was 38.

Meanwhile `CONTRIBUTING.md` said `good first issue` items could be picked up *without* assignment, which is a race by construction, while everything else required waiting for a maintainer to assign — a queue that, per the numbers above, mostly was not serviced. Contributors responded rationally: they stopped waiting. We have had duplicated work, a first-time contributor ghosted for 24 days, and pull requests opened against issues we had never agreed to fix.

**The problem is not that contributors ignore our process. It is that our process contains a step we do not reliably perform.** This PR's central idea is therefore to *take maintainers off the critical path* for the common case, while keeping the decisions that actually need judgement in human hands.

---

## The issue lifecycle

```mermaid
stateDiagram-v2
    [*] --> NeedsTriage: issue opened
    NeedsTriage --> NeedsInfo: needs more detail
    NeedsInfo --> NeedsTriage: reporter responds
    NeedsTriage --> Closed: not a bug, won't fix, duplicate
    NeedsTriage --> Accepted: maintainer accepts (target: 5 business days)
    Accepted --> HelpWanted: we would welcome community help
    HelpWanted --> Assigned: contributor comments /assign (instant, no maintainer)
    Accepted --> Assigned: maintainer assigns someone
    Assigned --> HelpWanted: 3 weeks with no progress
    Assigned --> Closed: pull request merged
```

Labels are the state machine. Anything without `triage/accepted` has not been agreed yet; only `help wanted` is open for self-service.

| Label | Meaning |
| --- | --- |
| `needs-triage` | Not reviewed yet. Applied automatically to every new issue. |
| `triage/needs-information` | Blocked on the reporter. |
| `triage/accepted` | We agree this should be done. |
| `help wanted` | Accepted **and** open for community help. Self-serve claimable. |
| `good first issue` | A `help wanted` issue needing no deep background. Signposted for first-timers. |
| `lifecycle/frozen` | Exempt from automatic unassignment. |

---

## What a contributor does

1. **Find something.** Browse [`help wanted`](https://github.com/openeverest/openeverest/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+no%3Aassignee). Anything else has not been agreed yet.
2. **Claim it.** Comment `/assign`. A bot assigns you within seconds — no maintainer involved. `/unassign` releases it at any time.
3. **If it is not `help wanted`**, comment to say you would like it and wait for triage. **If nobody replies within 5 business days, start anyway** and say so in the thread. You should never be blocked by our silence.
4. **Open a PR** referencing `Fixes #123`.

**Rules that apply to them:**

- **Opening an issue does not reserve it.** It has not been triaged; we may not agree it is a bug, or may want it fixed differently.
- **At most two open assigned issues** at a time.
- **If you already have a merged PR here, consider leaving `good first issue` items for people who don't.** A courtesy, not a rule — the bot won't stop you.
- **Do not open a PR against someone else's assigned issue.** It will be closed.
- **Never present output you did not produce.** Reasoning from reading the code is fine — say that is what you did. Invented logs and stack traces are worse than no evidence, because they manufacture confidence that something was reproduced.

---

## What a maintainer does

1. **Work the triage queue** — [`needs-triage`, oldest first](https://github.com/openeverest/openeverest/issues?q=is%3Aissue+is%3Aopen+label%3Aneeds-triage+sort%3Acreated-asc). Target: **5 business days**, which is the one new obligation this PR puts on us.
2. **Decide**, and apply one of: `triage/accepted`, `triage/needs-information`, or close with an explanation.
3. **If accepted and we would welcome help**, add `help wanted`. From that moment the issue is self-serve and needs nothing further from us.
4. **Add `good first issue` only if it clears the bar**: the solution is described, the relevant files are named, and the tests to change are identified. *If you cannot write that down, it is not a good first issue* — that is what turns a `good first issue` queue into a trap.

Maintainers bypass every bot gate, and can use `/assign @someone` to assign others.

---

## What the automation does

Three workflows, none of which make judgement calls:

| Workflow | Trigger | Behaviour |
| --- | --- | --- |
| `label-sync` | `labels.yml` changes on `main` | Creates/updates the labels above. Additive — never deletes labels not listed. |
| `assign` | `/assign`, `/unassign`, `/take` comment | Assigns the commenter if the issue is `help wanted`, unassigned, and they are under the cap. Otherwise explains why not. |
| `assignment-reaper` | Daily | After 14 days with no activity, asks the assignee if they are still on it. After 21, unassigns and says how to take it back. Skips maintainers, anyone with an open PR, and `lifecycle/frozen`. |

`/assign` matches Prow's syntax, so anyone who has contributed to a Kubernetes project already knows it, and the label names follow Kubernetes conventions so we can grow into `priority/*` and `area/*` later without renaming.

---

## Drawbacks, honestly

- **More process for a five-maintainer project.** Every rule here is one somebody enforces until it is automated.
- **The 5-business-day commitment is the hard part**, and the most likely thing to fail. We are currently at 23%.
- **There are zero claimable issues right now.** This advertises a self-serve shelf we have not stocked. Grooming ~10–15 `help wanted` items is the real prerequisite, and no amount of automation substitutes for it.
- **The `good first issue` courtesy is unenforced**, so it relies on goodwill. The two-issue cap stops anyone hoarding them concurrently, but not sequentially.

## Alternatives considered

- **Just reply faster.** That is the actual fix, and the data says we have not managed it. The automation exists so the process survives us being busy.
- **Prow**, as Kubernetes uses. Rejected: it means operating a 24/7 service, which is the wrong answer to a maintainer-capacity problem. Staying Prow-*compatible* keeps that door open for free.
- **`takanome-dev/assign-issue-action`** (used by JabRef, OpenSCAD, OrcaSlicer). Rejected after reading its source: its command parser is substring-based, which makes Prow's `/assign` unusable and fires on comments that merely mention the command.

## Prior art

[Kubernetes issue triage guidelines](https://kubernetes.dev/docs/guide/issue-triage/) and [`help wanted` / `good first issue` criteria](https://kubernetes.dev/docs/guide/help-wanted/).

---

## Questions for reviewers

1. **Is 5 business days a commitment we can keep?** Would a weekly rotation help, or is ad-hoc fine?
2. **Is the `good first issue` courtesy strong enough unenforced?** It started as a hard gate; we softened it because refusing a willing contributor is the worse failure, and the two-issue cap already prevents hoarding. Say if you'd rather it were enforced.
3. **Who stocks the `help wanted` pool, and by when?** This is the critical path.
4. **Is the AI-assisted contributions section pitched right?** It is deliberately not a tool ban, but it does commit us to closing unverified reports.

## Not in this PR

A PR guard blocking pull requests against untriaged issues (needs a branch-protection change), the maintainer triage playbook in `docs/process/`, and the CNCF-style contributor ladder with Triager/Reviewer tiers (needs a governance vote).

## Before merging

`/assign` needs testing from an account that is not a repository collaborator — GitHub silently drops assignees it considers ineligible. The workflow reads the assignment back and reports failure rather than failing silently, but the happy path is unverified.

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -1,0 +1,138 @@
+name: Issue assignment
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+
+env:
+  MAX_OPEN_ISSUES: 2
+
+jobs:
+  assign:
+    # Issues only; PR comments arrive on the same event.
+    if: >-
+      github.event.issue.pull_request == null &&
+      (contains(github.event.comment.body, '/assign') ||
+       contains(github.event.comment.body, '/unassign') ||
+       contains(github.event.comment.body, '/take'))
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const issue = context.payload.issue;
+            const issue_number = issue.number;
+            const commenter = context.payload.comment.user.login;
+
+            // Commands must start a line, matching the Prow convention.
+            const line = context.payload.comment.body
+              .split(/\r?\n/)
+              .map((l) => l.trim())
+              .find((l) => /^\/(assign|unassign|take)\b/.test(l));
+            if (!line) return;
+
+            const parsed = line.match(/^\/(assign|unassign|take)(?:\s+@([A-Za-z0-9-]+))?/);
+            const verb = parsed[1] === 'take' ? 'assign' : parsed[1];
+            const target = parsed[2] || commenter;
+
+            const reply = (body) =>
+              github.rest.issues.createComment({ owner, repo, issue_number, body });
+
+            const isMaintainer = async (login) => {
+              try {
+                const { data } = await github.rest.repos.getCollaboratorPermissionLevel({
+                  owner, repo, username: login,
+                });
+                return ['admin', 'write', 'maintain'].includes(data.permission);
+              } catch {
+                return false;
+              }
+            };
+
+            const commenterIsMaintainer = await isMaintainer(commenter);
+
+            // Assigning somebody else is a maintainer action; otherwise anyone
+            // could hand their work to a stranger.
+            if (target !== commenter && !commenterIsMaintainer) {
+              await reply(`@${commenter} only maintainers can assign an issue to someone else. Comment \`/${verb}\` on its own to ${verb === 'assign' ? 'take' : 'release'} it yourself.`);
+              return;
+            }
+
+            if (verb === 'unassign') {
+              await github.rest.issues.removeAssignees({
+                owner, repo, issue_number, assignees: [target],
+              });
+              await reply(`Unassigned @${target}. This issue is open for anyone to pick up again.`);
+              return;
+            }
+
+            if (issue.state !== 'open') {
+              await reply(`@${commenter} this issue is closed.`);
+              return;
+            }
+
+            const assignees = (issue.assignees || []).map((a) => a.login);
+            if (assignees.length > 0) {
+              if (assignees.includes(target)) return; // already theirs, nothing to do
+              await reply(`@${commenter} this issue is already assigned to @${assignees[0]}. Please don't open a competing pull request — if it looks abandoned, ask about its status in this thread.`);
+              return;
+            }
+
+            const labels = (issue.labels || []).map((l) => l.name);
+
+            // Maintainers bypass the gates below.
+            if (!commenterIsMaintainer) {
+              if (!labels.includes('help wanted')) {
+                await reply(`@${commenter} this issue isn't open for self-assignment yet — it needs the \`help wanted\` label, which a maintainer adds once the issue has been triaged and we'd welcome help with it.\n\nComment to say you'd like to take it and a maintainer will get back to you. See [CONTRIBUTING.md](https://github.com/${owner}/${repo}/blob/main/CONTRIBUTING.md#working-on-issues).`);
+                return;
+              }
+
+              const open = await github.rest.search.issuesAndPullRequests({
+                q: `repo:${owner}/${repo} is:issue is:open assignee:${commenter}`,
+                advanced_search: 'true',
+                per_page: 1,
+              });
+              const cap = Number(process.env.MAX_OPEN_ISSUES);
+              if (open.data.total_count >= cap) {
+                await reply(`@${commenter} you already have ${open.data.total_count} open assigned issue(s), and the limit is ${cap}. Please finish or \`/unassign\` one first — holding several at once blocks work others could be doing.`);
+                return;
+              }
+
+              if (labels.includes('good first issue')) {
+                const merged = await github.rest.search.issuesAndPullRequests({
+                  q: `repo:${owner}/${repo} is:pr is:merged author:${commenter}`,
+                  advanced_search: 'true',
+                  per_page: 1,
+                });
+                if (merged.data.total_count > 0) {
+                  await reply(`@${commenter} \`good first issue\` items are reserved for people who haven't had a pull request merged here yet, and you already have ${merged.data.total_count}. Please leave this one for a newcomer and pick up a [\`help wanted\`](https://github.com/${owner}/${repo}/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+no%3Aassignee) issue instead — thanks for understanding.`);
+                  return;
+                }
+              }
+            }
+
+            await github.rest.issues.addAssignees({
+              owner, repo, issue_number, assignees: [target],
+            });
+
+            // GitHub silently drops assignees it considers ineligible, so
+            // confirm rather than trusting the call to have taken effect.
+            const { data: after } = await github.rest.issues.get({ owner, repo, issue_number });
+            if (!(after.assignees || []).some((a) => a.login === target)) {
+              await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              return;
+            }
+
+            await reply([
+              `Assigned to @${target}. Thanks for picking this up!`,
+              '',
+              'A few things worth knowing:',
+              '',
+              '- Open your pull request with `Fixes #' + issue_number + '` in the description.',
+              '- If you get stuck or change your mind, comment `/unassign` — no hard feelings.',
+              '- If there is no visible progress for two weeks we will check in, and unassign a week after that so somebody else can pick it up.',
+            ].join('\n'));
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -86,6 +86,13 @@ jobs:
 
             // Maintainers bypass the gates below.
             if (!commenterIsMaintainer) {
+              // Filing an issue does not reserve it: it has not been triaged,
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -38,15 +38,49 @@ If you are trying to report a vulnerability, please refer to our [Security Polic
...
+Labels tell you where an issue stands:
 
-**Exception: `good first issue`.** Issues labeled [`good first issue`](https://github.com/openeverest/openeverest/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) are explicitly reserved for new contributors and can be picked up without prior assignment. Feel free to start working and open a PR when ready.
+| Label | What it means for you |
+| --- | --- |
+| `needs-triage` | Not reviewed yet. Please don't start work on it. |
+| `triage/needs-information` | We're waiting on the reporter before deciding. |
+| `triage/accepted` | We agree this should be done. |
+| [`help wanted`](https://github.com/openeverest/openeverest/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+no%3Aassignee) | Accepted, and we'd welcome community help. **Claim these yourself with `/assign`.** |
+| [`good first issue`](https://github.com/openeverest/openeverest/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22+no%3Aassignee) | A `help wanted` issue that needs no deep background. Reserved for first-time contributors. |
```
chilagrow: I don't think it's strictly necessary to reserve it for the first time contributors. A contributor can take up at most 2 issues, which limits parallel work. 😉

.github/workflows/assign.yml
```diff
@@ -0,0 +1,145 @@
...
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+
+env:
+  MAX_OPEN_ISSUES: 2
```
chilagrow, line 11: Any reason it's 2 issues at a time? To keep focus, I feel 1 issue at a time seems sensible too 🤗

chilagrow (commented):
Thanks for this 🤗 a couple minor questions

.github/workflows/assign.yml
recharte: The only reason is that PR reviews sometimes take a bit longer so I thought it would be reasonable that a contributor could take a second issue while the first one is being reviewed

CONTRIBUTING.md
recharte: Agree, especially at our scale. Softened the rule.

.github/workflows/assign.yml
chilagrow: Thanks, yes review can be a bottleneck at the moment, allowing 2 issues for this reason is sensible 👍

chilagrow (dismissed):
Thank you for this 🙏

CONTRIBUTING.md
```diff
@@ -38,15 +38,51 @@ If you are trying to report a vulnerability, please refer to our [Security Polic
...
+
+**For anything else, comment to say you'd like to take it** — for example, *"I'd like to work on this"* — and wait for a maintainer to accept the issue and assign it. Waiting stops two people solving the same problem in parallel, and stops you spending time on a report that turns out not to be a bug or that we have decided not to fix.
+
+**If nobody has replied within 5 business days, go ahead and start**, and leave a second comment saying so. You should not be blocked by our silence. If someone else has already said they started, coordinate with them in the thread rather than opening a competing pull request.
+
+### While you are working
+
+Please do not open a pull request for an issue assigned to someone else — it will be closed. Reference the issue with `Fixes #123` in your pull request description.
+
+If an assigned issue shows no visible progress for two weeks, a bot will ask whether you are still on it, and will unassign it about a week later if there is no reply. No hard feelings — comment `/assign` and it is yours again. If you know a change will take longer than that, say so in the thread and we will add `lifecycle/frozen`, which exempts it.
```
solovevayaroslavna: UI issues are frequently gated on design/UX review or screenshots, so they can legitimately sit >2 weeks with no thread activity while not being abandoned. lifecycle/frozen covers this but is manual and easy to forget, and semantically means "frozen indefinitely" rather than "temporarily blocked". Could the reaper also skip a status/blocked (e.g. blocked/needs-design) label?

solovevayaroslavna (dismissed):

CONTRIBUTING.md
recharte: Good catch, and I've pushed a change for it — but I want to lay out the reasoning, because it changed how I think about part of this.

**What I did:** added `status/blocked` ("waiting on something we owe — a design review, a decision, an upstream fix"), which the reaper now skips alongside `lifecycle/frozen`. I kept them as two labels rather than overloading one, exactly for the semantic reason you gave: `lifecycle/frozen` means *"this is a long piece of work"*, `status/blocked` means *"this is stuck on us"*. Different states, different resolutions — and only one of them is our debt to clear.

**Where I didn't follow you:** I went generic rather than `blocked/needs-design`. The same situation arises blocked on an API decision, an upstream operator fix, or a review, and sub-typing before we have the volume to justify it is how a label taxonomy rots. If design review needs its own queue, `label:ui label:status/blocked` gives you one. Rust, who have far more volume than us, deliberately keep a single `S-blocked` for this reason.

**On "manual and easy to forget":** a new label doesn't actually fix that — it's equally manual. So I also reworded the day-14 reminder to name the remedy: *"if you're waiting on us — a design review, a decision, an upstream fix — say so and we'll add `status/blocked`."* Forgetting now self-heals, because the prompt arrives exactly when it's relevant. I suspect that does more work than the label does.

**Two things that may already cover more of your case than it looks:** the reaper skips anyone with an open PR in the repo, including drafts — so "code written, waiting on design sign-off" was never at risk. And unassignment only happens *after* a warning, so replying "still blocked on design" resets the clock. The genuinely uncovered case was narrow: blocked on design before any code exists, and the contributor doesn't see the nag.

**The wider point your comment raised**, which I'd like your view on: if something needs a design spec before anyone can implement it, arguably it shouldn't be `help wanted` yet at all. Kubernetes defines `triage/accepted` as *"ready to be actively worked on"* — if a design still has to be written, it isn't. We already have the [`docs/process/`](https://github.com/openeverest/openeverest/blob/main/docs/process/README.md) proposal templates and the `specs` repo for that. So "needs design" is often a signal we marked something ready too early, rather than a blocked assignment. `status/blocked` then covers the narrower case: someone is mid-implementation and hits a decision only we can make.

One caveat worth naming: `status/blocked` can itself go stale — an issue waiting on a design review nobody ever does becomes invisible. I don't think that needs automation, but `label:status/blocked sort:updated-asc` is worth a look during triage, and I'll note it in the triage playbook when that lands.

solovevayaroslavna: Yes! I completely agree with you. And you're right that if a task requires design, it means that it wasn't sufficiently developed at the start, but this happens when we're waiting for something in the process, and it's more appropriate to separate these situations. I don't see a way to automate this yet, so I think the key is to make the task visible.

recharte: Agreed, I think we need to discuss this further in the scope of a triage playbook.
I'm thinking about two things that should be there: `label:status/blocked` `sort:updated-asc` as a standing query to walk during triage, and a note that clearing that list is our debt rather than the contributor's.
If you'd like the design-gated subset surfaced on its own — a board column, or just a saved `label:ui` `label:status/blocked` view. We can figure something out.

PR #2765, opened 2026-08-02
Ctl everestctl backup storage delete

### LINKED ISSUE #2660: Add `everestctl backupstorage delete` subcommand

VijetaPriya47:
Deletes a BackupStorage through the Everest API, following the resource-removal confirmation flow from the shared pkg/cli/confirm helper (--yes skips the y/N prompt and the command fails fast in non-interactive contexts). An optional --wait/--timeout polls until the BackupStorage is actually removed, since deleting a storage that is still referenced by an Instance or Backup returns 204 No Content but leaves the resource in the Terminating state rather than failing synchronously. The credentials Secret is not deleted explicitly—the BackupStorage controller sets an owner reference on it, allowing Kubernetes garbage collection to remove the Secret automatically once the BackupStorage has been deleted.
Closes #2660 
Branched on issue #2658

### DIFF
--- a/commands/backupstorage/delete.go
+++ b/commands/backupstorage/delete.go
@@ -76,7 +76,7 @@ list'/'everestctl backup list' for the namespace.`,
 
 func init() {
 	deleteCmd.Flags().StringVar(&deleteOpts.Name, cli.FlagBackupStorageName, "", "Backup storage name (required)")
-	deleteCmd.Flags().StringVar(&deleteOpts.Namespace, cli.FlagBackupStorageNamespace, "", "Namespace the backup storage is in (required)")
+	deleteCmd.Flags().StringVarP(&deleteOpts.Namespace, cli.FlagBackupStorageNamespace, "n", "", "Namespace the backup storage is in (required)")
 	deleteCmd.Flags().StringVar(&deleteOpts.Cluster, cli.FlagBackupStorageCluster, "main", "Cluster name")
 	deleteCmd.Flags().StringVar(&deleteOpts.Context, cli.FlagBackupStorageContext, "", "Context to use (default: current context)")
 	deleteCmd.Flags().BoolVarP(&deleteOpts.Yes, cli.FlagYes, "y", false, "Skip the confirmation prompt")
@@ -110,7 +110,7 @@ func deleteRun(cmd *cobra.Command, _ []string) {
 	d := backupstoragecli.NewDeleter(*deleteCfg, logger.GetLogger())
 	runErr := d.Run(cmd.Context(), *deleteOpts, cfgPath)
 	os.Exit(waitcmd.ExitCode(runErr, logger.GetLogger(), deleteCfg.Pretty,
-		fmt.Sprintf("wait cancelled; backup storage %q deletion continues in the background", deleteOpts.Name),
+		fmt.Sprintf("wait cancelled; backup storage %q deletion continues in the background — check with 'everestctl backup-storage list --namespace %s'", deleteOpts.Name, deleteOpts.Namespace),
 		fmt.Sprintf("timed out after %s waiting for backup storage %q to be deleted; deletion is still running server-side — a stuck delete is almost always an Instance or Backup that still references the storage, check 'everestctl instance list --namespace %s' / 'everestctl backup list --namespace %s'",
 			deleteOpts.Timeout, deleteOpts.Name, deleteOpts.Namespace, deleteOpts.Namespace),
 	))
--- a/commands/backupstorage/delete.go
+++ b/commands/backupstorage/delete.go
@@ -48,7 +48,14 @@ unless it was externally referenced before being adopted.
 Interactively (a terminal, no --yes), you'll be asked to confirm with y/N.
 Pass --yes/-y to skip the prompt; in a non-interactive context (no terminal,
 or --json) omitting --yes fails immediately instead of hanging on a prompt
-nobody can answer.`,
+nobody can answer.
+
+--ignore-not-found treats an already-absent backup storage as success and
+skips both the confirmation prompt and --wait entirely, since there's
+nothing to confirm or wait for. If --wait times out, the deletion is still
+running server-side, and a stuck delete is almost always an Instance or
+Backup that still references the storage, so point at 'everestctl instance
+list'/'everestctl backup list' for the namespace.`,
 		Example: `  # Delete a backup storage (prompts y/N)
   everestctl backup-storage delete --name my-s3 --namespace everest
 
@@ -56,7 +63,10 @@ nobody can answer.`,
   everestctl backup-storage delete --name my-s3 --namespace everest -y
 
   # Scripted teardown, wait until fully removed
-  everestctl backup-storage delete --name my-s3 --namespace everest --yes --wait --timeout 5m`,
+  everestctl backup-storage delete --name my-s3 --namespace everest --yes --wait --timeout 5m
+
+  # Idempotent teardown: treat "already gone" as success
+  everestctl backup-storage delete --name my-s3 --namespace everest --yes --ignore-not-found`,
 		PreRun: deletePreRun,
 		Run:    deleteRun,
 	}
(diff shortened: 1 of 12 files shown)


VijetaPriya47:
@recharte could you check this out?

recharte:
@VijetaPriya47 please resolve the conflicts here before I jump onto the review.

commands/backupstorage/delete.go
```diff
@@ -0,0 +1,119 @@
...
+	deleteCmd = &cobra.Command{
+		Use:   "delete [flags]",
+		Args:  cobra.NoArgs,
+		Short: "Delete a backup storage",
+		Long: `Delete a BackupStorage through the Everest API.
+
+The name and namespace flags are required. The credentials Secret is not
+deleted by this command directly — it's owned by the BackupStorage and is
+garbage-collected by Kubernetes once the BackupStorage is actually gone,
+unless it was externally referenced before being adopted.
```
vickysharma-prog: This reads as a guarantee that a Secret passed via `--credentials-secret` survives the delete. Per the adoption rule in `backupstorage_controller.go` it does not: any Secret without an existing controller is adopted, user-created ones included, and GC then removes it. Worth rewording however the lifecycle question in #2660 gets settled.

vickysharma-prog (commented):
Read this against #2660 and against `instance delete`, since that is the pattern it is meant to follow.

The mechanics line up. Confirmation reuses `pkg/cli/confirm`, so the non-interactive failure is the exact `confirmation required in non-interactive mode; re-run with --yes` the issue specified. `--json` forces non-interactive while `--verbose` alone does not, and there is a test pinning that difference. 404 maps to the wording from the issue, other statuses surface the server message, and `--force` is absent as intended. The option struct matches `pkg/cli/instance/delete.go` field for field.

I checked the claim about Terminating rather than a synchronous failure. `handleFinalizer` in `internal/controller/backup/backupstorage_controller.go` keeps the finalizer while any Instance or Backup still references the storage, so the description is accurate, and `deleteCondition` telling the user it is "likely referenced by an Instance or Backup" is the right thing to surface.

One thing worth settling before merge: the credentials Secret. @recharte this is your call rather than something the PR can decide on its own.

#2660 asked for a decision delete it when it follows the CLI-managed `<name>-credentials` convention, leave it when it was externally referenced. This PR answers with garbage collection, and the controller's adoption is wider than that split. The condition there is only:

```go
// Adopt the secret if it has no controller yet.
if metav1.GetControllerOf(secret) == nil {
```

A Secret the user created and passed via `--credentials-secret` has no controller, so it is adopted too and GC removes it on delete:

1. User creates `shared-s3-creds` by hand.
2. `backupstorage create --name a --credentials-secret shared-s3-creds`
3. `backupstorage create --name b --credentials-secret shared-s3-creds`
4. A reconciles first and adopts the Secret. B sees a controller already set and skips.
5. `backupstorage delete --name a` takes the Secret with it, and B is left pointing at nothing.

That is the "leave it when it was externally referenced" case, and it is not left.

The cause sits in the controller rather than this command, so it may well not be yours to fix here. The help text is, though, and I have left a note on it inline.

commands/backupstorage/delete.go
VijetaPriya47: You're right, I cheked the controller and  the check is purely GetControllerOf(secret) == nil, nothing in it looks at how the storage came to reference the Secret, so a freshly-created one handed in via --credentials-secret gets adopted exactly like a CLI-generated one. I have corrected the wording to be accurate.

recharte: Thanks folks, this will be addressed in a separate issue.

pkg/cli/backupstorage/delete.go
```diff
@@ -0,0 +1,243 @@
...
+		return err
+	}
+
+	if !alreadyGone {
+		bd.l.Infof("deleted backup storage %q in namespace %q", opts.Name, opts.Namespace)
+	}
+
+	if !opts.Wait || alreadyGone {
+		return bd.emitDeleted(opts)
+	}
```
recharte: `alreadyGone` from `checkDeleteResponse` falls through to `emitDeleted`, so this path prints `Backup storage "my-s3" deleted` / `{"deleted":true}` for a storage this run never deleted — the exact thing `emitAlreadyGone`'s own comment says must not happen ("deleted must stay false, it should only mean this run actually deleted something"). `instance delete` returns `emitAlreadyGone` here.

The pre-check short-circuit above hides it most of the time, but not always: it only fires on a clean 404 from the GET, so a transient GET failure, a 403 from read-restricted RBAC, or a storage that disappears between the GET and the DELETE all land here.

```suggestion
	if alreadyGone {
		return bd.emitAlreadyGone(opts)
	}
	bd.l.Infof("deleted backup storage %q in namespace %q", opts.Name, opts.Namespace)

	if !opts.Wait {
		return bd.emitDeleted(opts)
	}
```

Worth a test too — `TestDelete_NotFound_WithIgnoreNotFound_Succeeds` looks like it covers this, but its GET handler is nil (404), so the short-circuit fires and the DELETE is never issued; the branch has no coverage and asserts nothing about output. A variant with a GET that 200s and a DELETE that 404s would pin it.

pkg/cli/backupstorage/delete.go
recharte, line 1: `checkDeleteResponse`, `emitDeleted`, `emitAlreadyGone`, `writeDeleteResultJSON`, `waitForDeletion`, the delete poll and `deleteCondition` are all line-for-line copies of their `pkg/cli/instance` counterparts, differing only in the client method and the noun in the strings.

#2658 called this out in advance: *"'wait until 404' is a new condition shape worth adding to the shared waiter rather than a per-command copy"* — the same rationale that produced `pkg/cli/wait`/`waitcmd`/`confirm`. It didn't happen for `instance delete`, and this is the second copy, with `backup delete` and `restore delete` still to come. The other comment on this file is precisely the drift this shape produces: one branch was reordered in `instance delete` and the copy here didn't follow.

The generic parts are small and mechanical — a `Gone`-style `wait.Condition`, and a shared delete-result emitter (name/namespace/deleted JSON + the two pretty lines) alongside `waitcmd`. The poll is the only genuinely resource-specific piece, and even that is "GET → 404 means gone" over a different client method.

commands/backupstorage/delete.go
```diff
@@ -0,0 +1,122 @@
...
+
+  # Idempotent teardown: treat "already gone" as success
+  everestctl backup-storage delete --name my-s3 --namespace everest --yes --ignore-not-found`,
+		PreRun: deletePreRun,
+		Run:    deleteRun,
+	}
+	deleteCfg  = &backupstoragecli.Config{}
+	deleteOpts = &backupstoragecli.DeleteOptions{}
+)
+
```
recharte, line 76: `backup-storage list` and `backup-storage create` both register `--namespace` with the `-n` shorthand, so `everestctl backup-storage delete -n everest --name my-s3` fails with `unknown shorthand flag: 'n'` in the middle of a group where it works everywhere else.

```suggestion
	deleteCmd.Flags().StringVarP(&deleteOpts.Namespace, cli.FlagBackupStorageNamespace, "n", "", "Namespace the backup storage is in (required)")
```

commands/backupstorage/delete.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if err := waitcmd.ValidateWaitFlags(deleteOpts.Wait, cmd.Flags().Changed(cli.FlagBackupStorageTimeout), deleteOpts.Timeout); err != nil {
+		output.PrintError(err, logger.GetLogger(), deleteCfg.Pretty)
+		os.Exit(waitcmd.ExitFailed)
+	}
+
+	// Ctrl-C only cancels --wait; the cancellable context is set up inside
+	// Deleter.waitForDeletion, not here.
+	d := backupstoragecli.NewDeleter(*deleteCfg, logger.GetLogger())
```
recharte, line 110: `instance delete`'s cancel message ends with "— check with 'everestctl instance status'". This one stops at "continues in the background", leaving the user with no next command, even though the timeout message a couple of lines below does point somewhere. `everestctl backup-storage list --namespace %s` would close it out.

VijetaPriya47:
Hey, @recharte I have made changes in other commands also for clienterr.Message fix. Could you check it out now? Before opening the issue, wanted to reverify it once?

recharte (approved):
Thanks @VijetaPriya47. Nice work.

PR #2769, opened 2026-08-03
fix(auth): backport token blocklist update error handling

### LINKED ISSUE #2766: Blocklist update failures are silently swallowed on release-2.0 (missing backport of #2086)
### What is the issue?

`tokenStore.Add` in `pkg/session/token_store.go` declares `updateErr` for the `UpdateSecret` call but returns `err` from inside the `if updateErr != nil` branch. At that point `err` is guaranteed nil, because the earlier `GetSecret` already returned on failure — so a failed write to the blocklist Secret is reported as success.

The consequence is that server-side logout stops being reliable under exactly the class of transient error it was built to tolerate. `blocklist.Block` wraps `Add` in a `backoff.Retry` with 10 attempts, but a nil return makes the retry loop treat attempt 1 as successful, `POST /v1/auth/revoke` answers 204, and the access token is never blocklisted — it stays valid until it expires on its own.

Impact is bounded: the refresh token is revoked correctly through the separate `tokenRegistry.Revoke` path, so no new access tokens can be minted. The exposure window is the remaining `AccessTokenTTL` of the already-issued access token.

This is the same defect as #2085, fixed on `main` by #2086 and shipped in v1.16.0. It is present on `release-2.0` because that fix was never backported.

### Steps to Reproduce

Deterministically, as a unit test in `pkg/session`:

1. Build a `fakeclient` with an `interceptor.Funcs{Update: ...}` that returns a conflict error (`k8serrors.NewConflict`) for Secret writes.
2. Construct a `tokenStore` against it and call `Add(ctx, token)`.
3. Observe that `Add` returns nil, and that a following `Exists(ctx, token)` reports the token is not blocklisted.

At runtime:

1. Cause `UpdateSecret` on the blocklist Secret to fail transiently — a 409 from two concurrent logouts, or controller-runtime cache staleness.
2. Log out.
3. Replay the same access token against a protected endpoint; it is still accepted.

### Relevant Logs / Output

```go
secret = addDataToSecret(ts.l, secret, shortenedToken, time.Now().UTC())
_, updateErr := ts.client.UpdateSecret(ctx, secret)
if updateErr != nil {
	return err // always nil here
}
return nil
```

The log line does fire, so the failure is visible in logs — it just never reaches the caller.

### OpenEverest Version

`release-2.0` (the v2 development branch) @ efd04062, and the `v2.0.0-dev.1` developer preview tag cut from it.

### Installation Method

Helm

### Kubernetes Distribution

Other (please specify below)

### Environment Details

Not environment-specific. The defect is in error propagation and reproduces on any distribution and any installation method.

No shipped release is affected. `main` and every v1.16.x tag carry the fix; the bug exists only on the v2 development line, where the sole tagged artifact is the `v2.0.0-dev.1` preview. Worth fixing now precisely so it never reaches a v2 release.

### Possible Solution

`return updateErr` instead of `return err`, matching what #2086 already did on `main`.

Note that `git cherry-pick 6727c51e` will not apply cleanly: the branches have diverged around this code (`release-2.0` has a 4-arg `newTokenStore` and `Add` reads `ts.namespace`, `main` has 3 args and hardcodes `common.SystemNamespace`), and the commit also carries a copyright-header hunk that is already present. The edit is easier applied by hand.

Two PRs are planned:

1. `release-2.0` — the fix plus a regression test. Going first, so the test is reviewed on the branch where it visibly goes red then green.
2. `main` — the same test, ported. The fix is already there, so this one closes the coverage gap only.

The test should assert the invariant rather than the return value: `Add` reports success only when the token is genuinely on the blocklist, verified by reading it back. A test that only asserts "an error came back" would still pass against an `Add` that returns the right error but writes nothing.

### Additional Context

`pkg/session/blocklist_test.go` has no case where `UpdateSecret` fails — the fake client's writes always succeed. That gap is why the original bug shipped, and why the missing backport went unnoticed: the suite on `release-2.0` is green.

Reported privately via the process in `SECURITY.md`. Handling it in the open rather than as a Security Advisory: the fix and its reasoning have been in a public commit since May, so an advisory protects nobody, and there is no v2 release behind this branch.

The broader concern is the backport gap itself — a fix that landed on `main` in May is still absent from the v2 development line nearly three months later, with no signal that it was missing.

Related: #2085 (original report), #2086 (fix on `main`).

Anuragkumar-687:
## What

Backports the fix from #2086 to `release-2.0`.

Part of #2766

`tokenStore.Add` incorrectly returned the `GetSecret` error instead of the `UpdateSecret` error, causing blocklist update failures to be reported as success and preventing the retry logic from running.

## Changes

- Return the `UpdateSecret` error from `tokenStore.Add`, matching the existing fix in #2086.
- Add a regression test covering `UpdateSecret` failures.

## Testing

Added `TestTokenStore_Add_UpdateFailure`, which verifies that:

- `Add` returns an error when the Secret update fails.
- A read-back using `Exists` confirms the token was not blocklisted.

Verified locally:

- `go test ./pkg/session/...`
- `go test -race ./pkg/session/...`

## Related

- Backports #2086

### DIFF
--- a/pkg/session/blocklist_test.go
+++ b/pkg/session/blocklist_test.go
@@ -28,8 +28,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 
 	"github.com/openeverest/openeverest/v2/pkg/common"
 	"github.com/openeverest/openeverest/v2/pkg/kubernetes"
@@ -435,6 +437,71 @@ func TestBlocklist_IsBlocked(t *testing.T) {
 			assert.Equal(t, tc.blocked, blocked)
 		})
 	}
+}
+
+// TestTokenStore_Add_UpdateFailure asserts the invariant that Add reports
+// success only when the token is genuinely blocklisted. When the underlying
+// Secret update fails, Add must return the error rather than swallow it, and
+// the token must not be persisted to the blocklist Secret.
+//
+// This is a regression test for the missing backport of #2086 (see #2766): on
+// the unfixed code Add returns the nil `err` from the earlier GetSecret instead
+// of the update error, so a failed write is reported as success and the token
+// is never blocklisted.
+func TestTokenStore_Add_UpdateFailure(t *testing.T) {
+	t.Parallel()
+
+	l := zap.NewNop().Sugar()
+	ctx := context.Background()
+
+	//nolint:gosec // test token ID, not a credential
+	const shortenedToken = "9d1c1f98-a479-41e3-8939-c7cb3e049a331743679478"
+
+	// Fail every Secret update to simulate a transient write error (e.g. a 409
+	// conflict from concurrent logouts). Reads and the init-time create still
+	// succeed, so Add reaches the update and fails there.
+	mockClient := fakeclient.NewClientBuilder().
+		WithScheme(kubernetes.CreateScheme()).
+		WithInterceptorFuncs(interceptor.Funcs{
+			Update: func(
+				ctx context.Context,
+				c ctrlclient.WithWatch,
+				obj ctrlclient.Object,
+				opts ...ctrlclient.UpdateOption,
+			) error {
+				if _, ok := obj.(*corev1.Secret); ok {
+					return k8serrors.NewConflict(
+						schema.GroupResource{Resource: "secrets"},
+						obj.GetName(),
+						errors.New("simulated conflict"),
+					)
+				}
+				return c.Update(ctx, obj, opts...)
+			},
+		})
+	k := kubernetes.NewEmpty(l, "test-ns").WithKubernetesClient(mockClient.Build())
+
+	store, err := newTokenStore(ctx, k, l, "test-ns")
+	require.NoError(t, err)
+
+	// The Secret write fails, so Add must surface the error and not report
+	// success. The defect returned the wrong error variable, so pin that the
+	// surfaced error is the UpdateSecret conflict rather than merely "an error":
+	// kubernetes.UpdateSecret returns the client error unwrapped.
+	err = store.Add(ctx, shortenedToken)
+	require.Error(t, err)
+	assert.True(t, k8serrors.IsConflict(err), "Add must surface the UpdateSecret error, got %v", err)
+
+	// Read-back invariant: because the write failed, the token must not have
+	// reached the Secret. As in TestBlocklist_Block, the fake client does not do
+	// the StringData -> Data transformation the real API server does, so the
+	// assertion has to be on StringData.
+	secret, err := k.GetSecret(ctx, ctrlclient.ObjectKey{
+		Name:      common.EverestBlocklistSecretName,
+		Namespace: "test-ns",
+	})
+	require.NoError(t, err)
+	assert.NotContains(t, secret.StringData[dataKey], shortenedToken)
 }
 
 func mockNewBlocklist(ctx context.Context, logger *zap.SugaredLogger, mockClient TokenStoreClient) (Blocklist, error) {
--- a/pkg/session/token_store.go
+++ b/pkg/session/token_store.go
@@ -76,9 +76,9 @@ func (ts *tokenStore) Add(ctx context.Context, shortenedToken string) error {
 	}
 
 	secret = addDataToSecret(ts.l, secret, shortenedToken, time.Now().UTC())
-	_, updateErr := ts.client.UpdateSecret(ctx, secret)
-	if updateErr != nil {
-		ts.l.Errorf("failed to update %s secret in the %s namespace with the %s shortened token, retrying: %v", secret.Name, secret.Namespace, shortenedToken, updateErr)
+	_, err = ts.client.UpdateSecret(ctx, secret)
+	if err != nil {
+		ts.l.Errorf("failed to update %s secret in the %s namespace with the %s shortened token, retrying: %v", secret.Name, secret.Namespace, shortenedToken, err)
 		return err
 	}
 	return nil


vickysharma-prog:
@Anuragkumar-687 the backport against #2086 and against `release-2.0`.

The fix itself is right. The hunk in `token_store.go` is byte-identical to the one merged in #2086, which is what a backport should look like. #2766 suggests `return updateErr`; reassigning `err` to match the original commit is the better call for a backport, since the two branches now read the same.

The bug reproduces as described. When the update fails, `err` still holds the nil result of the earlier `GetSecret`, so `Add` returns nil on a failed write, and `blocklist.Block` wraps `Add` in `backoff.Retry`, so the nil ends the retry loop instead of driving it.

The test does the thing #2766 specifically asked for: it asserts the invariant by reading back through `Exists`, not just that an error came back. That distinction matters, since a variant returning the right error while writing nothing would pass the weaker check. It is also a real regression test on the unfixed code `Add` returns nil and `require.Error` fails.

The description says `Fixes #2766`, which will auto-close that issue on merge. #2766 plans two PRs: this one on `release-2.0`, and a second porting the same test to `main`, where the fix already exists but the coverage gap does not. Closing the issue here would lose the second one.

`Part of #2766` instead. Same ask @recharte made on #2617 for the same reason.

pkg/session/blocklist_test.go
```diff
@@ -437,6 +439,63 @@ func TestBlocklist_IsBlocked(t *testing.T) {
...
+// TestTokenStore_Add_UpdateFailure asserts the invariant that Add reports
+// success only when the token is genuinely blocklisted. When the underlying
+// Secret update fails, Add must return the error rather than swallow it, and
+// the token must not be readable back via Exists.
+//
+// This is a regression test for the missing backport of #2086 (see #2766): on
+// the unfixed code Add returns the nil `err` from the earlier GetSecret instead
+// of the update error, so a failed write is reported as success and the token
+// is never blocklisted.
+func TestTokenStore_Add_UpdateFailure(t *testing.T) {
```
vickysharma-prog, line 451: This targets `tokenStore.Add`, and the split in this package puts those in `token_store_test.go` (`TestCleanupOld`, `TestAddDataToSecret`) while `blocklist_test.go` holds the `TestBlocklist_*` cases. The name you've given it already says `TokenStore`, so it reads like it belongs in the other file. Non-blocking.

pkg/session/blocklist_test.go
```diff
@@ -437,6 +439,63 @@ func TestBlocklist_IsBlocked(t *testing.T) {
...
+			},
+		})
+	k := kubernetes.NewEmpty(l, "test-ns").WithKubernetesClient(mockClient.Build())
+
+	store, err := newTokenStore(ctx, k, l, "test-ns")
+	require.NoError(t, err)
+
+	// The Secret write fails, so Add must surface the error and not report success.
+	err = store.Add(ctx, shortenedToken)
+	require.Error(t, err)
```
vickysharma-prog, line 492: This pins that the error is returned, and the `Exists` read-back below covers the invariant, which is the part that matters. What is not covered is the consequence #2766 leads with, that the retry loop now actually runs. A case through `blocklist.Block` counting `tokenStore.Add` attempts on a persistent conflict would close that, though with `maxRetries = 10` at a 500ms constant backoff it would add about five seconds to the suite unless the interval is made injectable. Note rather than a request.

vickysharma-prog (changes requested):
Two non-blocking notes inline. CI green, DCO signed.

Anuragkumar-687:
Hi @recharte! The release-2.0 backport is ready for review whenever you have a chance. It includes the backported fix from #2086 along with the regression test we discussed in #2766. As you suggested, once this PR is merged, I'll open the follow-up PR against main to port the same regression test and close the coverage gap there. I'd really appreciate your feedback whenever you have time. Thanks!

pkg/session/blocklist_test.go
```diff
@@ -437,6 +439,63 @@ func TestBlocklist_IsBlocked(t *testing.T) {
...
+
+	// Read-back invariant: because the write failed, the token must not be
+	// blocklisted. Asserting the return value alone would not catch a variant
+	// that reports success while writing nothing.
+	blocked, err := store.Exists(ctx, shortenedToken)
+	require.NoError(t, err)
+	assert.False(t, blocked)
```
recharte: `Exists` reads `secret.Data`, and the fake client never populates it — it stores the `StringData` you hand it without doing the API server's StringData → Data conversion. So `blocked` is `false` here whether the write failed or succeeded, and this assertion holds unconditionally. Dropping the interceptor so `Add` fully succeeds: the persisted Secret comes back `Data=map[] StringData=map[list:9d1c1f98-…]` and `Exists` still reports `false`.

`TestBlocklist_Block` hit the same thing 240 lines up and says so — *"the mocked client does not do this StringData -> Data transformation in Secrets which the actual k8a API do, so we only check the StringData field"*. Same idiom here:

```suggestion
	// Read-back invariant: because the write failed, the token must not have
	// reached the Secret. As in TestBlocklist_Block, the fake client does not do
	// the StringData -> Data transformation the real API server does, so the
	// assertion has to be on StringData.
	secret, err := k.GetSecret(ctx, ctrlclient.ObjectKey{
		Name:      common.EverestBlocklistSecretName,
		Namespace: "test-ns",
	})
	require.NoError(t, err)
	assert.NotContains(t, secret.StringData[dataKey], shortenedToken)
```

pkg/session/blocklist_test.go
recharte, line 492: The defect was returning the wrong error variable, so "an error came back" is a notch weaker than the contract this is guarding — `return errors.New("boom")` satisfies it too. `kubernetes.UpdateSecret` returns the client error unwrapped, so the conflict can be pinned directly:

```suggestion
	err = store.Add(ctx, shortenedToken)
	require.Error(t, err)
	assert.True(t, k8serrors.IsConflict(err), "Add must surface the UpdateSecret error, got %v", err)
```

pkg/session/blocklist_test.go
recharte: Placement is fine as it is: `token_store_test.go` covers the pure helpers (`cleanupOld`, `addDataToSecret`) and never builds a client, whereas every fake-client-backed case and `mockNewBlocklist` live in this file. Moving it would drag the harness across for one test.

pkg/session/blocklist_test.go
recharte: `TestBlocklist_Block` already covers this: it deletes the Secret, calls `Block` again and asserts the `NotFound` surfaces, which only happens after `backoff.Retry` exhausts all 10 attempts. That case runs in 5.15s today. The only gap left is the same loop driven by an `UpdateSecret` failure rather than a `GetSecret` one, and that would cost a second 5s to pin a path the existing case already exercises. Not adding a test for it — no change needed here.

recharte (changes requested):
One thing that isn't on a line: the description says `Fixes #2766`, so merging auto-closes it — and #2766 covers two PRs, this one and the port of the test to `main`. Could you change it to `Part of #2766`?

Anuragkumar-687:
Thanks @recharte, addressed the review comments.

* Replaced the `Exists` read-back with a direct `Secret.StringData` check, since the fake client does not perform the `StringData -> Data` conversion.
* Strengthened the error assertion to verify that `Add` surfaces the `UpdateSecret` Conflict error, rather than just returning any error.
* Kept the test in `blocklist_test.go` and did not add another retry-loop test, as discussed.
* Updated the issue reference from `Fixes #2766` to `Part of #2766`.

`go test ./pkg/session/...` and `go test -race ./pkg/session/...` both pass.

recharte (approved):
Thanks @Anuragkumar-687 well done

PR #2456, opened 2026-06-26
fix: use a dedicated HTTP client for S3 validation

ashnaaseth2325-oss:
# SUMMARY

This PR prevents backup storage validation from modifying the shared `http.DefaultClient` by using a dedicated HTTP client for S3 connectivity checks. The change keeps the validation behavior the same while ensuring HTTP settings remain isolated to the current request.

Primarily affected module:

* `internal/server/handlers/validation/backup_storage.go`

# FIX

```go
// Before
c := http.DefaultClient
c.Timeout = timeoutS3AccessSec * time.Second

// After
c := &http.Client{
    Timeout: timeoutS3AccessSec * time.Second,
}
```

PR #2781, opened 2026-08-04
fix: avoid mutating http.DefaultClient in s3Access

### LINKED ISSUE #2780: Data race: s3Access() mutates the global http.DefaultClient instead of creating a new client
## Description

`s3Access()` in `internal/server/handlers/validation/backup_storage.go` (lines 225–229) copies `http.DefaultClient` by assignment, which only copies the pointer —not the underlying struct. Subsequent field mutations (`Timeout`, `Transport`) end up mutating the shared global `http.DefaultClient` instance.

## Current Code

```go
// Lines 225-229
c := http.DefaultClient          // copies the POINTER, not the struct
c.Timeout = timeoutS3AccessSec * time.Second
c.Transport = &http.Transport{
    TLSClientConfig: &tls.Config{InsecureSkipVerify: !verifyTLS},
}
```

## Why This Is a Problem

1. **Data race / concurrency bug**  Everest handles concurrent requests. Two simultaneous calls to `s3Access()` (e.g., two users adding backup storages at the same time) will race over writing to `http.DefaultClient.Timeout` and `http.DefaultClient.Transport`. This is a classic Go data race.

2. **Global state pollution**  After the first call, `http.DefaultClient` is left with a modified `Timeout` and `Transport`. Other code paths that use `http.DefaultClient.Do(req)` (e.g. `telemetry.go`, `oidc.go`, `version_service/client.go`) will unknowingly inherit the modified client.

3. **TLS verification bypass leak** If a user creates a backup storage with `verifyTLS: false`, `InsecureSkipVerify: true` leaks globally into `http.DefaultClient`, disabling TLS verification for **all** subsequent HTTP calls made anywhere in the process.

## Proposed Fix

Create a new `http.Client` instead of copying the pointer:

```go
// AFTER (correct)
c := &http.Client{
    Timeout: timeoutS3AccessSec * time.Second,
    Transport: &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: !verifyTLS}, //nolint:gosec
    },
}
```

## Affected File
`internal/server/handlers/validation/backup_storage.go`, lines 225–229

### A. Global state pollution (deterministic)

1. Call the backup storage validation endpoint (which triggers `s3Access()`) with `verifyTLS: false` for one storage  e.g. `POST /v1/backup-storages` (or equivalent validation call) with a payload where `verifyTLS` is `false`.
2. After that call completes, trigger any other code path in the same process that uses `http.DefaultClient` directly (e.g. `telemetry.go`, `oidc.go`, or `version_service/client.go`), pointed at a server with an invalid/self-signed TLS certificate.
3. Observe that the second call succeeds despite the invalid certificate  proving `InsecureSkipVerify: true` leaked into the shared `http.DefaultClient`, even though that call never opted out of TLS verification.

### B. Data race (requires Go race detector)

1. Build/run the server with the race detector enabled:
```bash
   go run -race ./cmd/everest-server
```
   (or `go test -race ./...` if there's a test that exercises `s3Access()`)
2. Fire two or more concurrent requests that both call `s3Access()` e.g. two simultaneous "add backup storage" calls with different `verifyTLS` values:
```bash
   curl -X POST http://localhost:8080/v1/backup-storages -d '{"verifyTLS": true, ...}' &
   curl -X POST http://localhost:8080/v1/backup-storages -d '{"verifyTLS": false, ...}' &
   wait
```
3. The Go race detector reports a data race on `http.DefaultClient.Timeout` / `http.DefaultClient.Transport`, since both goroutines write to the same shared pointer's fields concurrently.

### Expected Result
Each call to `s3Access()` should use its own isolated `*http.Client`, with no cross-request interference and no race detected.

### Actual Result
- All calls share and mutate the same global `http.DefaultClient`.
- `-race` flags concurrent writes to it.
- A `verifyTLS: false` request can silently disable TLS verification for unrelated requests elsewhere in the process.

### Relevant Logs / Output

_No response_

### OpenEverest Version

OpenEverest v1.13.0 (latest as of writing)

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

Bhupesh-081:
## Summary
Fix a data race and global state pollution caused by mutating `http.DefaultClient` in `s3Access()`.

## What changed
`internal/server/handlers/validation/backup_storage.go`

```diff
-	c := http.DefaultClient
-	c.Timeout = timeoutS3AccessSec * time.Second
-	c.Transport = &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: !verifyTLS}, //nolint:gosec
-	}
+	c := &http.Client{
+		Timeout: timeoutS3AccessSec * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: !verifyTLS}, //nolint:gosec
+		},
+	}
```

## Root Cause
`http.DefaultClient` is a `*http.Client` (a pointer). Assigning it to a local variable `c` copies only the pointer value, not the underlying struct. All subsequent field assignments (`c.Timeout`, `c.Transport`) therefore directly mutate the shared global `http.DefaultClient`.

## Impact
1. **Data race** Under concurrent requests (e.g., two users adding a backup storage simultaneously), both goroutines race to write `http.DefaultClient.Timeout` and `http.DefaultClient.Transport`. This is a classic Go data race detectable with `-race`.
2. **Global TLS config leak**  After the first call with `verifyTLS: false`, `http.DefaultClient.Transport` is permanently set to `InsecureSkipVerify: true`. All subsequent code in the process that calls `http.DefaultClient.Do()` including `internal/server/telemetry.go`, `pkg/oidc/oidc.go`, and `pkg/version_service/client.go`  will silently skip TLS verification.
3. **Timeout leak**  The 2-second timeout is permanently stamped on `http.DefaultClient`, affecting all other outbound HTTP calls for the lifetime of the process.

## Fix
Construct a fresh `*http.Client` literal per invocation. Each call to `s3Access()` now owns its own isolated, short-lived HTTP client with the correct timeout and TLS configuration, leaving `http.DefaultClient` untouched.

## Testing
- `go build ./internal/server/handlers/validation/...` ✅

Fixes #2780


internal/server/handlers/validation/backup_storage.go
```diff
@@ -222,10 +222,11 @@ func s3Access(
 		endpoint = nil
 	}
 
-	c := http.DefaultClient
-	c.Timeout = timeoutS3AccessSec * time.Second
-	c.Transport = &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: !verifyTLS}, //nolint:gosec
+	c := &http.Client{
```
vickysharma-prog, line 225: Non-blocking, and not a regression from this PR: a fresh `http.Transport` per call also means a fresh connection pool per call, and those idle connections stay around until GC rather than being reused. The stdlib docs are explicit that transports are meant to be reused. For a validation endpoint that a user can trigger repeatedly it may be worth either a package-level pair of transports keyed on `verifyTLS`, or a `defer c.CloseIdleConnections()` here. Fine to leave for a follow-up.

vickysharma-prog (changes requested):
Confirmed the bug and the fix. `http.DefaultClient` is a `*http.Client`, so `c := http.DefaultClient` copies the pointer and both field writes land on the global. Building a fresh client is the right fix.

I checked the impact claims rather than take them on trust, and they hold. `internal/server/telemetry.go:65`, `pkg/oidc/oidc.go:66` and `pkg/version_service/client.go:89,150` all call `http.DefaultClient.Do`, so one backup storage added with `verifyTLS: false` really did leave `InsecureSkipVerify: true` on the client those three use for the rest of the process. That is the part worth leading with; the timeout leak is the milder half.

I also grepped for other mutation sites. `DefaultClient` appears in exactly those four files and the other three only read it, so this was the only one.

One ask before merge: a regression test. Testing lists `go build` only, and nothing in the suite would notice if this came back. It is cheap here `config.Debug` is `false` in default builds (`//go:build !debug`), so `s3Access` runs past its early return under `go test`, and `backup_storage_test.go` already exists in the package. Calling `s3Access` with an unreachable endpoint and then asserting `http.DefaultClient.Timeout == 0` and `http.DefaultClient.Transport == nil` goes red on the old code and green on yours, with no network and no `-race` needed.

Bhupesh-081:
@vickysharma-prog  Added the regression test as requested (TestS3Access_DoesNotMutateDefaultClient in backup_storage_test.go). It asserts http.DefaultClient.Timeout and .Transport are left untouched after calling s3Access() against an unreachable endpoint.

Verified it actually catches the regression before pushing:

1. Against the old buggy code (c := http.DefaultClient) FAILS, showing the exact mutation: Timeout gets stamped to 2s and Transport gets set globally.
2. Against the current fix (c := &http.Client{...}) PASSES cleanly, no network, no -race needed.

internal/server/handlers/validation/backup_storage_test.go
```diff
@@ -290,6 +292,23 @@ func TestValidateBucketName(t *testing.T) {
 	}
 }
 
+func TestS3Access_DoesNotMutateDefaultClient(t *testing.T) {
```
shivansh-source, line 299: one case is missing from the test for parellels

internal/server/handlers/validation/backup_storage.go
```diff
@@ -222,10 +222,11 @@ func s3Access(
...
+		Timeout: timeoutS3AccessSec * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: !verifyTLS}, //nolint:gosec
```
chilagrow: Do you `gosec` linter? Why?

go1.26.3.linux-amd64.tar.gz
chilagrow, line 1: What is this file, please do not add irrelevant files please.

internal/server/handlers/validation/backup_storage_test.go
```diff
@@ -290,6 +292,23 @@ func TestValidateBucketName(t *testing.T) {
...
+	// Sanity check- http.DefaultClient should start in its zero-value state.
```
chilagrow: Please check the linter warnings

internal/server/handlers/validation/backup_storage_test.go
```diff
@@ -290,6 +292,23 @@ func TestValidateBucketName(t *testing.T) {
...
+	require.Equal(t, time.Duration(0), http.DefaultClient.Timeout)
+	require.Nil(t, http.DefaultClient.Transport)
+
+	l := zap.NewNop().Sugar()
+	endpoint := "http://127.0.0.1:1" // unreachable; connection refused immediately
+
+	// The result of s3Access is irrelevant here we only care that  http.DefaultClient is left untouched.
```
chilagrow: But ignoring error is not a good practice. Error should be checked

chilagrow (changes requested):
Please check linter, unnecessary files before submitting PRs for review please.

Also as issue pointed out, DefaultClient is used in telemetry, oidc and other places. What are the reason for using DefaultClient there?

go1.26.3.linux-amd64.tar.gz
Bhupesh-081: I apologize for adding irrelevant files, they were add by mistake during an installation process, I have removed them locally, Changes will be reflected in commit.

internal/server/handlers/validation/backup_storage.go
Bhupesh-081: Thanks for confirming the fix and for double-checking the impact claims appreciate the thoroughness on that.

On the connection pooling note: agreed, and noted. I'll leave the fresh Transport-per-call as is for this PR since it's scoped to fixing the mutation bug, but happy to open a follow-up issue for the package-level-transport-pair-keyed on-verifyTLS approach if that's useful for the team to track separately. Let me know if you'd like me to file it.

internal/server/handlers/validation/backup_storage_test.go
Bhupesh-081: t.Parallel() , this test asserts on the shared http.DefaultClient global, so running it in parallel with other tests touching that same global could produce flaky results. Please share your Opinion about this I would like to work on further review

internal/server/handlers/validation/backup_storage.go
Bhupesh-081: It's user-controlled, not hardcoded,verifyTLS comes from the API request (params.VerifyTLS), letting users opt out of cert verification when connecting to self-hosted S3-compatible endpoints (MinIO, Ceph, etc.) that use self-signed certs. Added an inline comment explaining this instead of leaving the suppression bare. If any issues please give feedback, I would like to follow up.

internal/server/handlers/validation/backup_storage_test.go
Bhupesh-081: since this test asserts on process-global state and shouldn't run concurrently with other tests touching http.DefaultClient. Thats why I didn't included it please share insights regarding this change if any.

internal/server/handlers/validation/backup_storage_test.go
Bhupesh-081: Fixed. the test now does:

err := s3Access(l, &endpoint, ...)
require.Error(t, err)

before asserting on http.DefaultClient state, so the error is checked rather than discarded.

Bhupesh-081:
apologies for the noise, and thanks for flagging both. On linter/unnecessary files: fixed. The go1.26.3.linux-amd64.tar.gz was leftover from a local Go toolchain upgrade and got committed by mistake removed.

 since the test asserts on process-global state and shouldn't run concurrently with other tests touching http.DefaultClient).

On DefaultClient in telemetry.go, oidc.go, and version_service/client.go: those three only *read* it via .Do() — they don't mutate its fields the way s3Access() did. So they weren't affected by the data race or the TLS/timeout leak this PR fixes; s3Access() was the only mutation site (confirmed by grepping DefaultClient usage across the codebase).

That said, if you'd prefer those three also move off the shared globalclient entirely  for consistency/isolation, even though they're safe as read-only callers I'm happy to open a follow-up issue to track that separately, since it's a broader change than this PR's scope. Let me know if you'd like me to file it.

recharte:
@Bhupesh-081 thanks for this, and for turning the review around so quickly.

Two things. We missed that #2456 has been open since June with the same fix, and it's merged now — our filing failure, not yours. And on the regression test: I'd rather not take it, despite it being asked for. It asserts `http.DefaultClient` is untouched, which any implementation that doesn't reach for the global passes for free, so it only catches someone deliberately writing that line again. It also asserts on process-global state as a precondition, which couples it to every other test in the package and rules out running it in parallel.

@chilagrow's question on `http.DefaultClient` elsewhere: telemetry, OIDC and the version service only read it, never mutate it, so there's nothing further to fix there.

Bhupesh-081:
No problem  @recharte at all, and thanks for the clarification! I appreciate the update on the contribution workflow the new `/assign` capability definitely makes it easier for new contributors like me to get involved.

I'm currently exploring the codebase and learning the project's architecture, with the goal of contributing more actively over time. I'll also make sure to check for existing issues and PRs before reporting or working on something to avoid duplicates.

Since my recent PR is now closed, I'm free to pick up another task. If there's any issue you think would be a good fit, Please be free to assign it right away I'd be happy to work on it and use it as an opportunity to learn more about the project.

Thank you @chilagrow @shivansh-source @vickysharma-prog for reviewing  my  written code it meants a lot for me.

PR #2791, opened 2026-08-04
fix(rbac): avoid panic on invalid policy refresh

### LINKED ISSUE #2786: Unhandled panic in RBAC ConfigMap background informer causes API server process crash
### What is the issue?

In 
pkg/rbac/rbac.go
, the background informer watches for updates to the everest-rbac ConfigMap via refreshEnforcerInBackground.

When an update occurs, the callback attempts to reload and validate the Casbin policy. If the updated ConfigMap contains invalid policy syntax or a corrupted rule, the callback triggers a panic(...):

if err := enforcer.LoadPolicy(); err != nil {
    panic("invalid policy detected - " + err.Error())
}
if err := validatePolicy(enforcer); err != nil {
    panic("invalid policy detected - " + err.Error())
}
if err := loadAdminPolicy(enforcer); err != nil {
    panic("failed to load admin policy - " + err.Error())
}
Because this informer callback runs in a background goroutine, an unhandled panic brings down the entire everest API server process (resulting in a container crash-loop / Denial of Service).

### Steps to Reproduce

Start OpenEverest locally (make dev-up).
Edit the everest-rbac ConfigMap in Kubernetes to introduce invalid syntax (e.g. malformed role policy string).
Observe the background informer triggering refreshEnforcerInBackground.
The everest API server process panics and crashes immediately.

### Relevant Logs / Output

Expected Behavior: If an invalid RBAC policy ConfigMap update occurs, the API server should log an error (l.Errorf), reject/skip the bad update, and retain the previously loaded valid policy in memory—allowing the server to remain online and operational.

Actual Behavior: The background goroutine panics and crashes the API server process.

### OpenEverest Version

v1.4.0 (or main / latest)

### Installation Method

CLI (everestctl)

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Update refreshEnforcerInBackground in 
pkg/rbac/rbac.go
Instead of mutating the active enforcer and calling panic(...) on errors during ConfigMap updates, load and validate the updated policy using a candidate/temporary enforcer instance.
If validation succeeds, apply the new policy to the main enforcer.
If validation fails, log an error (l.Errorf("invalid RBAC policy update: %v", err)) and retain the existing valid policy in memory so the API server continues operating without crashing.

### Additional Context

_No response_

RishiSingh13:
## Description

This PR fixes the RBAC policy refresh behavior when an invalid policy is detected in the `everest-rbac` ConfigMap.

Previously, errors encountered while reloading, validating, or restoring the RBAC policy triggered a panic in the background informer callback, which could terminate the API server process.

This change replaces those panics with error logging and an early return, allowing the API server to remain running when an invalid RBAC policy update is detected.

## Changes

* Replaced `panic(...)` calls in `refreshEnforcerInBackground` with error logging.
* Return early when RBAC policy loading, validation, or admin policy loading fails.
* Prevent invalid RBAC ConfigMap updates from crashing the API server.

## Testing

* Ran the relevant RBAC package tests.
* Verified the code is formatted with `gofmt`.

## Related Issue

Closes #2786


shivansh-source:
hello @RishiSingh13 can u rebase ur pr it is out of date . U will not get any merge  conflicts as the change is really small still if u get one try to solve it and put the pr . ping if u need any help :)

pkg/rbac/rbac.go
```diff
@@ -122,19 +122,22 @@ func refreshEnforcerInBackground(
 		if !ok || cm.GetName() != common.EverestRBACConfigMapName {
 			return
 		}
-		if err := enforcer.LoadPolicy(); err != nil {
-			panic("invalid policy detected - " + err.Error())
+			if err := enforcer.LoadPolicy(); err != nil {
```
shivansh-source: here indenting is wrong . the change is of one line but because of indenting the whole code section is shown as changed

amh1k:
@RishiSingh13 I think one of your commits is unsigned that's why DCO is failing

RishiSingh13:
> @RishiSingh13 I think one of your commits is unsigned that's why DCO is failing

Thanks for pointing that out! I've added the missing DCO sign-off and updated the PR.

RishiSingh13:
Thanks for the review! @shivansh-source  
 I've rebased the branch, fixed the indentation issue, and updated the PR.

spron-in (dismissed):
Lgtm.
I would be good to cherry pick this into release-2.0 branch

spron-in:
@RishiSingh13 it would be awesome if you cherry pick this fix into release-2.0 branch as well

spron-in:
Also DCO is still failing.

RishiSingh13:
Hi @spron-in, just a quick update on this PR. The DCO is now passing and the changes are approved. The PR is currently waiting on the Merge Gatekeeper workflow/maintainer approval.

I haven't made any further changes since the last update. Please let me know if anything else is needed from my side, including the cherry-pick into `release-2.0`.

pkg/rbac/rbac.go
```diff
@@ -117,24 +117,30 @@ func refreshEnforcerInBackground(
...
-			panic("invalid policy detected - " + err.Error())
+			l.Errorf("Invalid RBAC policy detected: %s", err)
+			return
 		}
 		// Calling LoadPolicy() re-writes the entire model, so we need to add back the admin role.
 		if err := loadAdminPolicy(enforcer); err != nil {
-			panic("failed to load admin policy - " + err.Error())
+			l.Errorf("Failed to load admin policy: %s", err)
+			return
 		}
```
recharte: `LoadPolicy()` is atomic in casbin v2.135 — it builds a new model and installs it only on success — so by the time `validatePolicy` runs, the enforcer is already serving the new policy. Returning here leaves the policy we just rejected live, and without any `everest-admin` grants: `LoadPolicy()` wiped the in-memory ones and `loadAdminPolicy` below never runs, so every admin is denied everything until the process restarts. The `loadAdminPolicy` branch has the same problem. #2786 asks for the opposite — retain the previously loaded valid policy.

Validating a candidate first keeps the live enforcer untouched, and `newEnforcer` already does load + `loadAdminPolicy` + `validatePolicy` and returns the error:

```suggestion
		// Validate the incoming policy on a throwaway enforcer, so that an invalid
		// update never reaches the live one.
		if _, err := newEnforcer(enforcer.GetAdapter(), false); err != nil {
			l.Errorf("Invalid RBAC policy detected, keeping the previous policy: %s", err)
			return
		}
		if err := enforcer.LoadPolicy(); err != nil {
			l.Errorf("Failed to load RBAC policy: %s", err)
			return
		}
		// Calling LoadPolicy() re-writes the entire model, so we need to add back the admin role.
		if err := loadAdminPolicy(enforcer); err != nil {
			l.Errorf("Failed to load admin policy: %s", err)
			return
		}
```

pkg/rbac/rbac.go
```diff
@@ -117,24 +117,30 @@ func refreshEnforcerInBackground(
 		informer.WithLogger(l),
 		informer.Watches(&corev1.ConfigMap{}, common.SystemNamespace),
 	)
+	if err != nil {
```
recharte, line 120: Good catch, and worth keeping even though it's outside #2786: `informer.New` returns `nil, err`, so the old code would have nil-dereferenced on `inf.OnUpdate`, and the `Start` error was being dropped in favour of the constructor's (nil) one.

recharte (changes requested):
Thanks for picking this up — the panic removal is right, but the ordering needs a look before it can land.

RishiSingh13:
Thanks for the detailed review, @recharte. I’ve updated the refresh ordering so the incoming policy is validated using a throwaway enforcer before modifying the live enforcer. I also fixed the informer creation error handling as suggested.

The changes have been pushed. Please take another look when you get a chance.

pkg/rbac/rbac.go
recharte: @RishiSingh13 sorry, I meant this as praise for the `err` check itself, not a request to change the wrapping — `errors.Join(err, errors.New("failed to create informer"))` was right, and it matches how `inf.Start` is wrapped a few lines down. Mind putting it back?

recharte:
@RishiSingh13 can you please fix the conflicts?

RishiSingh13:
> @RishiSingh13 can you please fix the conflicts?

Fixed the merge conflicts by merging the latest `upstream/main` into the branch and pushed the changes. The PR should now be conflict-free.

recharte (approved):
Thanks @RishiSingh13 🚀

PR #2800, opened 2026-08-05
fix: remove incorrect IncreaseTimeout call from DeleteSession

### LINKED ISSUE #2908: bug: DeleteSession calls IncreaseTimeout on every logout causing self-inflicted auth DoS on shared IPs
### What is the issue?

In internal/server/session.go:71–72, DeleteSession unconditionally
calls e.attemptsStore.IncreaseTimeout(ctx.RealIP()) on every
logout — the exact same call CreateSession makes only on a failed
login. Both share one IP-keyed rate limiter whose timeout doubles
each call, starting at 20ms and growing toward a 6-hour expiry.
On a shared IP (NAT, corporate proxy, multiple tabs), roughly
25–30 logouts push the timeout past the lockout window and block
all logins from that IP for hours with zero failed attempts.

### Steps to Reproduce

1. Deploy Everest behind a shared NAT or corporate proxy
2. Have 25–30 users/tabs log out normally from the same IP
3. Attempt any login from that IP

### Relevant Logs / Output

Login succeeds normally regardless of how many logouts occurred

Actual Output:
Login attempt returns:
  HTTP 429 Too Many Requests
  {
    "message": "Too many requests, please try again later",
    "retryAfter": 21600
  }
Log shows:
  WARN  session: IP 203.0.113.5 rate limited,
        timeout=21600s (6h), reason=logout

### OpenEverest Version

v1.0.0

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Remove IncreaseTimeout call from DeleteSession entirely.
Only call it on a failed login attempt in CreateSession.

internal/server/session.go:71–72

### Additional Context
This is a server-side request forgery (SSRF) primitive. Even
though the caller gets a 403, the damage is already done —
the probe has fired. In cloud/Kubernetes environments this is
particularly dangerous because the metadata endpoint
(169.254.169.254 on AWS/GCP/Azure) is reachable from the pod
network. An attacker can enumerate internal services, probe
firewall rules, or extract cloud instance credentials via
IMDSv1 using nothing more than a valid Everest login.
The fix is a one-line reorder of handler registration —
zero functional impact on the happy path since RBAC-authorized
users still reach the validation handler as before.
This does NOT require any change to the validation logic itself.
_No response_

ayushgupta-h:
`DeleteSession` (logout) was calling `attemptsStore.IncreaseTimeout(ip)`, the brute-force penalty meant exclusively for failed login attempts. Logout is a successful authenticated operation — `DELETE /session` inherits the global `BearerAuth` requirement — so it must not be charged a failure penalty.

In the normal flow the effect is small: a successful `CreateSession` calls `attemptsStore.CleanupVisitor(ip)`, which deletes the visitor entry outright, so a following logout starts again at `initialTimeout` (20ms) and doubles it once, to 40ms. The timeout only compounds across consecutive logouts with no successful login from the same IP in between (e.g. many users behind a shared NAT/proxy logging out), where it can grow until the `/v1/session` rate limiter rejects logins from that IP.

Remove the call; the penalty remains applied only on failed logins in `CreateSession`.

Fixes #2908


.gitignore
```diff
@@ -2,6 +2,7 @@
 /.idea/
 
 /bin/
+everest.exe
```
recharte: `everest.exe` isn't produced by anything in this repo: `make build` writes the server to `bin/everest`, already covered by the existing `/bin/` entry, and the only Windows artifact any target produces is `dist/everestctl.exe` from `make release-cli`. That makes this a local artifact of your own `go build -o everest.exe ./cmd`, which belongs in `.git/info/exclude` or your global gitignore rather than the shared one.

Could you drop 89bd1f2 and keep the PR to the one-line fix?

recharte (changes requested):
The fix itself is right: `IncreaseTimeout` is the failed-login penalty, and logout is an authenticated operation — `DELETE /session` inherits the global `BearerAuth` requirement, only `POST /session` opts out with `security: []` — so it should never have been charged one.

The description overstates the effect, and since this repo squashes with `COMMIT_MESSAGES` the wording in 77c840c goes into `main`'s history as-is. The backoff doesn't compound in the normal flow: a successful `CreateSession` calls `attemptsStore.CleanupVisitor(ip)`, which deletes the visitor entry outright, so the following logout starts again at `initialTimeout` (20ms) and doubles once, to 40ms. It only compounds across consecutive logouts with no successful login in between. Could you reword "progressively blocking legitimate subsequent logins from the same address" to match?

recharte:
@ayushgupta-h thanks for the fix — since only the wording and the unrelated .gitignore commit were outstanding, I've used maintainer edits to revert that commit and will adjust the commit message at merge so this can land. The same change will be ported to release-2.0 separately.

recharte (approved):

Issue #2793, opened 2026-08-05
Incorrect time truncation in database_cluster.go causes inconsistent timestamps


Harkirat1309:
### What is the issue?

The time truncation logic in `database_cluster.go` is incorrect. The current code attempts to remove nanoseconds by calling `now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)`, but this does not truncate the timestamp to whole-second precision. Instead, it truncates the time to a multiple of its own current nanosecond component, which can lead to inconsistent and unintended timestamp values.

This creates a logic bug because the resulting time depends on the current nanosecond value rather than using a fixed unit of truncation. The expected behavior is to consistently strip sub-second precision by truncating with `time.Second`, i.e. `now.Truncate(time.Second)`.

### Steps to Reproduce

1. Open `database_cluster.go` and locate the code path where the current time is truncated using `now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)`.
2. Run the logic with different `time.Now()` values that have different nanosecond components.
3. Compare the output of `now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)` with `now.Truncate(time.Second)`.
4. Observe that the current implementation does not consistently truncate to whole seconds and may return unexpected values depending on the nanosecond portion of the timestamp.

Minimal example:
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	now := time.Now()

	wrong := now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)
	correct := now.Truncate(time.Second)

	fmt.Println("original:", now)
	fmt.Println("wrong   :", wrong)
	fmt.Println("correct :", correct)
}
``

### Relevant Logs / Output

There may not be a runtime error or explicit log message for this issue, since it is a logic bug rather than a crash. The problem is visible in the produced timestamp values.

Example output from the minimal reproduction:

```text
original: 2026-08-05 14:23:45.987654321 +0000 UTC
wrong   : 2026-08-05 14:23:45.975308637 +0000 UTC
correct : 2026-08-05 14:23:45 +0000 UTC

### OpenEverest Version

Unknown at the moment. I identified this issue by reviewing the source code in `database_cluster.go`, rather than from a specific deployed OpenEverest version.  If needed, maintainers can map it to the affected release(s) based on when this logic was introduced.

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

[...]

Issue #2801, opened 2026-08-05
fix: latestRestorableDate truncates nanoseconds incorrectly, causing PITR restore failures


ayushgupta-h:
## Bug: `latestRestorableDate` — Nanosecond Truncation is a No-Op, Sub-Second Precision Leaks into PITR Date

**Component:** `internal/server/handlers/k8s/database_cluster.go`
**Severity:** Medium

### I tried to do
Call `GET /v1/namespaces/{ns}/database-clusters/{name}/pitr` on a database cluster with PITR enabled and a successful backup older than the upload interval, then use the returned `latestDate` to submit a PITR restore request.

### I expected
The `latestDate` returned by `latestRestorableDate` to have all sub-second precision stripped (per the code comment `// delete nanoseconds since they're not accepted by restoration`), so the timestamp is accepted by the restore operator.

### But instead
`latestDate` still carries a non-zero nanosecond component, because line 389 truncates using the wrong duration:

```go
now = now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)
```

`time.Time.Truncate(d)` rounds down to a multiple of `d` since the zero time. Using `now.Nanosecond()` (the *current* nanosecond value, an `int`) as the duration produces an arbitrary sub-second truncation boundary — not a true "strip to whole seconds" operation. It's effectively a no-op for its stated purpose.

The restore operator rejects timestamps with nanosecond precision, so PITR restore requests using this timestamp **silently fail** downstream, even though the API response looks valid.

**Affected engines:** old PXC (< 1.14.0), all PSMDB, all PostgreSQL clusters (anything going through the heuristics path).

---

### Root Cause
Confusion between:
- `time.Nanosecond` — the duration constant (`1ns`)
- `now.Nanosecond()` — method returning the nanosecond-within-second component

### Fix
```diff
 func latestRestorableDate(now, latestBackupTime time.Time, heuristicsInterval int) *time.Time {
     if heuristicsInterval == 0 {
         return nil
     }
     // delete nanoseconds since they're not accepted by restoration
-    now = now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)
+    now = now.Truncate(time.Second)
     date := now.Add(-time.Duration(heuristicsInterval) * time.Second).UTC()
```

### Steps to Reproduce
1. Have a database cluster with PITR enabled and a successful backup older than the upload interval.
2. Call `GET /v1/namespaces/{ns}/database-clusters/{name}/pitr`.
[...]

PR #2802, opened 2026-08-05
fix: correct nanosecond truncation in latestRestorableDate

ayushgupta-h:
What's the problem?

In `internal/server/handlers/k8s/database_cluster.go`, the `latestRestorableDate` function tries to strip sub-second precision from the current time before calculating the latest PITR date.

However, the truncation logic was written as:

```go
// delete nanoseconds since they're not accepted by restoration
now = now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)
```

`time.Truncate(d)` rounds a `time.Time` down to the nearest multiple of `d`. Using `now.Nanosecond() * time.Nanosecond` passes a dynamic sub-second duration (e.g. 482,109,234 ns) into `Truncate`, rather than truncating to the nearest full second.

As a result, sub-second precision was never properly stripped from `now`, leaving nanosecond residue in the calculated `latestDate` timestamp. When a user tries to perform a Point-in-Time Recovery (PITR) restore using this timestamp, the database operator rejects the request due to the sub-second precision.

### Why wasn't this caught by unit tests?

The existing test case in `TestLatestRestorableDate` initialized `now` using `time.Date(2024, 3, 12, 12, 0, 0, 0, time.UTC)` where the nanosecond field was explicitly set to `0`. Because `now.Nanosecond()` evaluated to `0`, the broken `Truncate` call returned the time unmodified, which happened to have zero nanoseconds, completely masking the bug.

### What changed?

1. Replaced the self-referential `Truncate` call with `now = now.Truncate(time.Second)`, which cleanly strips all sub-second precision (milliseconds, microseconds, nanoseconds) as intended.
2. Added a test case to `TestLatestRestorableDate` where `now` has `999999999` nanoseconds to ensure sub-second truncation is explicitly verified and won't regress.

### Code Diff

```diff
diff --git a/internal/server/handlers/k8s/database_cluster.go b/internal/server/handlers/k8s/database_cluster.go
index 067ef114..15bf16c8 100644
--- a/internal/server/handlers/k8s/database_cluster.go
+++ b/internal/server/handlers/k8s/database_cluster.go
@@ -386,7 +386,7 @@ func latestRestorableDate(now, latestBackupTime time.Time, heuristicsInterval in
 		return nil
 	}
 	// delete nanoseconds since they're not accepted by restoration
-	now = now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)
+	now = now.Truncate(time.Second)
 	// heuristic: latest restorable date is now minus uploadInterval
 	date := now.Add(-time.Duration(heuristicsInterval) * time.Second).UTC()
 	// not able to restore if after the latest backup passed less than uploadInterval time,
```

### DIFF
--- a/internal/server/handlers/k8s/database_cluster.go
+++ b/internal/server/handlers/k8s/database_cluster.go
@@ -386,7 +386,7 @@ func latestRestorableDate(now, latestBackupTime time.Time, heuristicsInterval in
 		return nil
 	}
 	// delete nanoseconds since they're not accepted by restoration
-	now = now.Truncate(time.Duration(now.Nanosecond()) * time.Nanosecond)
+	now = now.Truncate(time.Second)
 	// heuristic: latest restorable date is now minus uploadInterval
 	date := now.Add(-time.Duration(heuristicsInterval) * time.Second).UTC()
 	// not able to restore if after the latest backup passed less than uploadInterval time,
--- a/internal/server/handlers/k8s/database_cluster_test.go
+++ b/internal/server/handlers/k8s/database_cluster_test.go
@@ -42,6 +42,8 @@ func TestLatestRestorableDate(t *testing.T) {
 
 	now := time.Date(2024, 3, 12, 12, 0, 0, 0, time.UTC)
 	restorable := now.Add(-600 * time.Second)
+	nowWithNanos := time.Date(2024, 3, 12, 12, 0, 0, 999999999, time.UTC)
+	restorableFromNanos := time.Date(2024, 3, 12, 11, 50, 0, 0, time.UTC)
 	cases := []tCase{
 		{
 			name:             "backup 5 min ago, upload interval 10 min",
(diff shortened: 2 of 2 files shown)


recharte:
Hi @ayushgupta-h — apologies, we missed this one. #2834 landed the same fix earlier today, so I'm closing this as superseded. It has been open since 5 August and should have had a look long before now.

Why it slipped past us is worth knowing for next time: the description doesn't reference an issue, so GitHub never created a cross-reference on #2801 or on #2793. When #2801 was closed as a duplicate of #2793, there was nothing on the #2793 thread pointing here — to anyone reading that issue, the bug looked unfixed and unclaimed. `Fixes #2793` in the pull request description is what creates that link, and it closes the issue on merge ([CONTRIBUTING.md](https://github.com/openeverest/openeverest/blob/main/CONTRIBUTING.md)).

The other signal we go by is the issue thread itself — #2793 was open and unassigned when you opened this, and commenting `/assign` there would have recorded the claim.

Your write-up was right, including the part about the existing test's zero-nanosecond `now` masking the bug — that is exactly why it survived. Sorry for the wasted effort; if you're up for another, `/assign` yourself anything open and unassigned and it's yours.

PR #2808, opened 2026-08-06
Extract shared HTTP-poll classification into pkg/cli/wait

### LINKED ISSUE #2798: Extract shared HTTP-poll classification into pkg/cli/wait (FetchPoll + NotFoundPolicy)
### Task Description

Add a `FetchPoll` helper (plus a `NotFoundPolicy` type) to `pkg/cli/wait`, and migrate the existing hand-rolled pollers onto it:

- `pkg/cli/instance/wait.go`'s `newInstancePoll` (create/watch-side) and `newInstanceDeletePoll` (delete-side)
- `pkg/cli/backup/wait.go`'s `newBackupPoll`
- `pkg/cli/restore/wait.go`'s `newRestorePoll`
- `pkg/cli/backupstorage/delete.go`'s delete-poll (currently via `deletecmd.DeleteGonePoll`)

All of these are structurally identical — same token-refresh handling, same 401 / empty-200-body / unexpected-status classification — differing in exactly one place: what a 404 means. `FetchPoll` takes that as an explicit `NotFoundPolicy`, not a bool or a callback, so it reads at the call site instead of requiring a trip to the definition:

```go
type NotFoundPolicy int

const (
	NotFoundIsSuccess NotFoundPolicy = iota // delete-polls: 404 = done
	NotFoundTerminal                        // watch-polls: 404 = terminal error
)

func FetchPoll[T any](
	kind, name string,
	notFound NotFoundPolicy,
	fetch func(ctx context.Context) (statusCode int, statusText string, body *T, err error),
) PollFunc[*T]
```

`pkg/cli/deletecmd.DeleteGonePoll` becomes a one-line wrapper (`wait.FetchPoll(kind, name, wait.NotFoundIsSuccess, fetch)`), so its current call sites in `backupstorage delete` and `instance delete` don't need to change.

### Reasoning / Benefit

#2658 flagged this in advance: *"'wait until 404' is a new condition shape worth adding to the shared waiter rather than a per-command copy."* It wasn't done when `instance create --wait` (#2600) and `instance delete` (#2658) shipped, and the duplication has already caused real drift once: on #2765 (`backupstorage delete`), the delete-poll's sibling ordering logic (`alreadyGone` → `emitAlreadyGone`/`emitDeleted`) diverged from `instance delete`'s already-correct version until review caught it. Four to five near-identical ~20-line functions is exactly the shape that goes stale silently — a fix, or a new terminal case (e.g. a future distinct-403 branch), applied to one copy and missed on the others.

### Additional Context

Scoped out of #2765 to keep that PR focused on `backupstorage delete` itself; this issue tracks the follow-up discussed there.

`pkg/cli/wait` is currently dependency-free (no `net/http`, no `pkg/cli/auth` — generic over any `T`). This is the point where it picks those up for the first time, in exchange for this logic living in exactly one place. `pkg/cli/deletecmd` (added in #2765) keeps its delete-specific pieces (`EmitDeleted`/`EmitAlreadyGone`/`EmitWaitSucceeded`, `GoneCondition`) and becomes a thin consumer of `wait.FetchPoll` rather than owning the fetch/classify logic itself.

Files touched: `pkg/cli/wait/` (new file for `FetchPoll`/`NotFoundPolicy`), `pkg/cli/deletecmd/deletecmd.go`, `pkg/cli/instance/wait.go`, `pkg/cli/backup/wait.go`, `pkg/cli/restore/wait.go`.

VijetaPriya47:
Five polling functions across the CLI (instance create --wait, instance delete --wait, backup create --wait, restore create --wait, backup-storage delete --wait) were classifying HTTP responses with nearly identical hand-rolled logic — same handling for token-refresh failures, 401s, empty bodies, and unexpected statuses, differing only in what a 404 meant. 

This PR extracts that shared logic into wait.FetchPoll, parameterized by a NotFoundPolicy (NotFoundIsSuccess for delete-polls, NotFoundTerminal for watch-polls) instead of a bool or callback, so the call site reads its own behavior without a trip to the definition. pkg/cli/deletion.GonePoll becomes a one-line wrapper over it, so its existing callers in backup-storage delete and instance delete don't change at all.

While in there, wired clienterr.Message into each poller's fetch closure so an unexpected status (500, 503, a webhook-denied 403, etc.) surfaces the server's actual message instead of a bare HTTP status line.

Closes #2798

### DIFF
--- a/pkg/cli/instance/wait.go
+++ b/pkg/cli/instance/wait.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/openeverest/openeverest/v2/client"
 	authcli "github.com/openeverest/openeverest/v2/pkg/cli/auth"
+	"github.com/openeverest/openeverest/v2/pkg/cli/deletion"
 	"github.com/openeverest/openeverest/v2/pkg/cli/wait"
 )
 
@@ -137,33 +138,17 @@ func newInstanceDeletePoll(
 	c *client.ClientWithResponses,
 	cluster, namespace, name string,
 ) wait.PollFunc[*client.Instance] {
-	return func(ctx context.Context) (*client.Instance, error) {
+	return deletion.GonePoll("instance", name, func(ctx context.Context) (int, string, *client.Instance, error) {
 		resp, err := c.GetInstanceWithResponse(ctx, cluster, namespace, name)
 		if err != nil {
-			if errors.Is(err, authcli.ErrTokenRefresh) {
-				return nil, fmt.Errorf("failed to fetch instance %q: %w", name, err)
-			}
-			return nil, &wait.RetryableError{Err: fmt.Errorf("failed to fetch instance %q: %w", name, err)}
+			return 0, "", nil, err
 		}
-		switch resp.StatusCode() {
-		case http.StatusOK:
-			if resp.JSON200 == nil {
-				return nil, &wait.RetryableError{Err: fmt.Errorf("empty response body fetching instance %q", name)}
-			}
-			return resp.JSON200, nil
-		case http.StatusNotFound:
-			return nil, nil // gone
-		case http.StatusUnauthorized:
-			return nil, fmt.Errorf("server rejected credentials — run 'everestctl auth login' again")
-		default:
-			return nil, &wait.RetryableError{Err: fmt.Errorf("unexpected response fetching instance %q: %s", name, resp.Status())}
-		}
-	}
+		return resp.StatusCode(), resp.Status(), resp.JSON200, nil
+	})
 }
 
 func deleteCondition(inst *client.Instance) (wait.Outcome, string) {
-	if inst == nil {
-		return wait.Succeeded, "instance deleted"
-	}
-	return wait.Pending, progressMessage(inst, instancePhase(inst))
+	return deletion.GoneCondition("instance deleted", func(v *client.Instance) string {
+		return progressMessage(v, instancePhase(v))
+	})(inst)
 }
--- a/pkg/cli/instance/wait.go
+++ b/pkg/cli/instance/wait.go
@@ -17,13 +17,11 @@ package instance
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"net/http"
 	"strings"
 
 	"github.com/openeverest/openeverest/v2/client"
-	authcli "github.com/openeverest/openeverest/v2/pkg/cli/auth"
+	"github.com/openeverest/openeverest/v2/pkg/cli/clienterr"
 	"github.com/openeverest/openeverest/v2/pkg/cli/deletion"
 	"github.com/openeverest/openeverest/v2/pkg/cli/wait"
 )
@@ -58,28 +56,22 @@ func newInstancePoll(
 	c *client.ClientWithResponses,
 	cluster, namespace, name string,
 ) wait.PollFunc[*client.Instance] {
-	return func(ctx context.Context) (*client.Instance, error) {
+	return wait.FetchPoll("instance", name, wait.NotFoundTerminal, fetchInstance(c, cluster, namespace, name))
+}
+
+// fetchInstance adapts GetInstanceWithResponse to wait.FetchPoll's fetch
+// shape.
+func fetchInstance(c *client.ClientWithResponses, cluster, namespace, name string) func(context.Context) (int, string, *client.Instance, error) {
+	return func(ctx context.Context) (int, string, *client.Instance, error) {
 		resp, err := c.GetInstanceWithResponse(ctx, cluster, namespace, name)
 		if err != nil {
-			// A failed token refresh is terminal; other fetch errors are transient.
-			if errors.Is(err, authcli.ErrTokenRefresh) {
-				return nil, fmt.Errorf("failed to fetch instance %q: %w", name, err)
-			}
-			return nil, &wait.RetryableError{Err: fmt.Errorf("failed to fetch instance %q: %w", name, err)}
+			return 0, "", nil, err
 		}
-		switch resp.StatusCode() {
-		case http.StatusOK:
-			if resp.JSON200 == nil {
-				return nil, &wait.RetryableError{Err: fmt.Errorf("empty response body fetching instance %q", name)}
-			}
-			return resp.JSON200, nil
-		case http.StatusNotFound:
-			return nil, fmt.Errorf("instance %q was deleted while waiting", name)
-		case http.StatusUnauthorized:
-			return nil, fmt.Errorf("server rejected credentials — run 'everestctl auth login' again")
-		default:
-			return nil, &wait.RetryableError{Err: fmt.Errorf("unexpected response fetching instance %q: %s", name, resp.Status())}
+		statusText := resp.Status()
+		if msg, ok := clienterr.Message(resp.JSONDefault); ok {
+			statusText = msg
 		}
+		return resp.StatusCode(), statusText, resp.JSON200, nil
 	}
 }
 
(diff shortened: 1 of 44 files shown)


pkg/cli/wait/wait_test.go
```diff
@@ -233,3 +236,115 @@ func TestUntil_ZeroTimeoutMeansNoTimeout(t *testing.T) {
 	})
 	require.NoError(t, err)
 }
+
+type widget struct {
+	Name string
+}
+
+func TestFetchPoll_NotFound_IsSuccessPolicy_ReturnsNil(t *testing.T) {
```
recharte, line 244: The eight `TestFetchPoll_*` cases added here are near-verbatim copies of the existing `TestGonePoll_*` cases in `pkg/cli/deletion/deletion_test.go` — same `widget` fixture, same table of statuses, same assertions, differing only in the constructor name and the `NotFoundPolicy` argument. This is the right home for them, but it leaves `deletion_test.go` running seven tests against a function that is now `return wait.FetchPoll(kind, name, wait.NotFoundIsSuccess, fetch)`.

The premise of this PR is that near-identical copies drift silently, and this reproduces the shape in the test suite: a change to the classification now needs the same edit in two files, and the `deletion` copies will keep passing regardless, since they exercise `wait` code either way.

Suggest trimming `deletion_test.go` to the one thing `GonePoll` actually decides — `TestGonePoll_Gone` (404 → `nil, nil`, i.e. the `NotFoundIsSuccess` binding) — and dropping `TestGonePoll_StillPresent`, `_EmptyBody_Retryable`, `_Unauthorized_Terminal`, `_UnexpectedStatus_Retryable`, `_FetchError_Retryable` and `_TokenRefreshFailure_Terminal`, which are all covered here now. That also drops the `authcli`/`net/http` imports from that test file, matching the implementation file which just lost them.

pkg/cli/wait/wait.go
```diff
@@ -73,6 +77,59 @@ type Condition[T any] func(T) (Outcome, string)
...
+	// NotFoundIsSuccess means a 404 is the awaited outcome, e.g. `<resource>
+	// delete --wait` polling until the resource is gone.
+	NotFoundIsSuccess NotFoundPolicy = iota
+	// NotFoundTerminal means a 404 is an error, e.g. `<resource> create --wait`
+	// or `status --watch` polling a resource that's expected to still exist.
+	NotFoundTerminal
+)
+
+// FetchPoll builds a PollFunc[*T] around fetch, classifying HTTP responses the
+// same way every `<resource> create/delete/status --wait` command already
```
recharte: This says the classification matches "every `<resource> create/delete/status --wait` command", and `NotFoundTerminal` below names `status --watch` as an example — but `pkg/cli/instance/status.go` still hand-rolls the whole switch in its own ticker loop, and it has already drifted from the version being centralised here:

- 404 → `instance %q has been deleted`, vs `%s %q was deleted while waiting`
- a failed token refresh returns the raw error, without the `failed to fetch instance %q:` wrapper
- transient errors are logged and skipped inline rather than classified as `RetryableError`

So the doc describes the state this PR was meant to create, while the one remaining divergent copy is the one it names. Either way out is fine by me:

1. Migrate it too. It doesn't use `wait.Until`, but it can still use the poll — `inst, err := poll(ctx)`, then `var re *wait.RetryableError; if errors.As(err, &re) { is.l.Warnf("%v — retrying in %s", err, interval); continue }; if err != nil { return err }`. That closes the last copy and makes this sentence true.
2. Keep it out of scope, reword to "every `<resource> create/delete --wait` command", and open a follow-up for `status --watch` referenced from #2798.

I'd lean towards (1) — it's ~10 lines and the divergence above is exactly the failure mode #2798 was filed about — but (2) is a reasonable call if you want the PR to stay strictly within the issue's file list.

pkg/cli/instance/wait.go
```diff
@@ -58,28 +56,22 @@ func newInstancePoll(
 	c *client.ClientWithResponses,
 	cluster, namespace, name string,
 ) wait.PollFunc[*client.Instance] {
-	return func(ctx context.Context) (*client.Instance, error) {
+	return wait.FetchPoll("instance", name, wait.NotFoundTerminal, fetchInstance(c, cluster, namespace, name))
```
recharte, line 57: The doc comment above this function still spells out "transient fetch failures and unexpected statuses are retried (wait.RetryableError), while a 404 (deleted), a 401 (credentials rejected), and a failed token refresh are terminal" — as of this change that's `FetchPoll`'s documented contract, not anything this function decides, and it's the kind of restated invariant that goes stale the first time the shared classification changes. `newBackupPoll` has the same paragraph; `newRestorePoll` already reads simply "returns a PollFunc for the restore".

Suggest matching the short form, e.g. "newInstancePoll returns a PollFunc for the instance; a 404 is terminal here, unlike the delete-side poll." That also drops the cross-reference to `status --watch`, which no longer shares this code (see my comment on `wait.go`).

pkg/cli/wait/wait.go
```diff
@@ -21,7 +21,11 @@ package wait
 import (
 	"context"
 	"errors"
+	"fmt"
+	"net/http"
 	"time"
+
+	authcli "github.com/openeverest/openeverest/v2/pkg/cli/auth"
```
recharte, line 30: With these imports the package doc a few lines up is out of date: it describes `wait` as "a resource-agnostic polling waiter ... Callers supply a PollFunc and a Condition", but the package now also *builds* PollFuncs, and picks up `net/http` and `pkg/cli/auth` to do it. Worth one sentence so the next reader isn't surprised by the dependency — e.g. "It also provides FetchPoll, which builds a PollFunc from a generated-client call, classifying HTTP statuses and auth failures uniformly."

recharte (commented):
One process note that isn't inline anywhere: the PR title is still "Ctl everestctl backup storage delete" from this branch's earlier life. Something like `refactor(cli): extract shared HTTP-poll classification into wait.FetchPoll` matches the diff. The body also says "Four polling functions" and then lists five.

pkg/cli/wait/wait.go
VijetaPriya47: migrated it here only

pkg/cli/wait/wait.go
```diff
@@ -73,6 +79,59 @@ type Condition[T any] func(T) (Outcome, string)
...
+		case http.StatusOK:
+			if body == nil {
+				return nil, &RetryableError{Err: fmt.Errorf("empty response body fetching %s %q", kind, name)}
+			}
+			return body, nil
+		case http.StatusNotFound:
+			if notFound == NotFoundIsSuccess {
+				return nil, nil
+			}
+			return nil, fmt.Errorf("%s %q was deleted while waiting", kind, name)
```
recharte: Now that `status --watch` shares this, the message it produces on a 404 is `instance "my-mongo" was deleted while waiting` — but nobody's waiting there; they're watching a live status. It also replaces the more natural `instance "my-mongo" has been deleted` that the watch used to print.

Nothing asserts either string (`status_test.go` matches on `"deleted"`, and there's no docs or CLI-test reference), so this is purely how it reads. A phrasing that's true for both kinds of caller would fix it:

```suggestion
			return nil, fmt.Errorf("%s %q no longer exists", kind, name)
```

Your call — leaving it is defensible, since "deleted while waiting" is accurate for the four `--wait` callers, which are the majority.

pkg/cli/wait/wait.go
```diff
@@ -73,6 +79,59 @@ type Condition[T any] func(T) (Outcome, string)
...
+	// NotFoundIsSuccess means a 404 is the awaited outcome, e.g. `<resource>
+	// delete --wait` polling until the resource is gone.
+	NotFoundIsSuccess NotFoundPolicy = iota
+	// NotFoundTerminal means a 404 is an error, e.g. `<resource> create --wait`
+	// or `status --watch` polling a resource that's expected to still exist.
+	NotFoundTerminal
+)
+
+// FetchPoll builds a PollFunc[*T] around fetch, classifying HTTP responses the
+// same way every `<resource> create/delete/status --wait` command already
```
recharte: Now that the claim is true, the flag name in it isn't: the status command's flag is `--watch`, not `--wait`. `NotFoundTerminal`'s doc a few lines up already has it right.

```suggestion
// same way every `<resource> create/delete --wait` and `status --watch` command
```

recharte (approved):
Looks great, just a couple of nitpicks.

recharte (changes requested):
There's one linter error, please check it out.

recharte (approved):
Thanks 🤗

PR #2817, opened 2026-08-08
Allow to disable plugin-hub installation

spron-in:
We need to allow users to disable plugin-hub through Tilt. This is useful when you are working on plugin-hub development and don't need to install it from the helm dependency, but want to do it from your local image. 


chilagrow (dismissed):
Thank you!

chilagrow (approved):

PR #2818, opened 2026-08-08
docs: fix stale extension CLI help examples

Adii-45:
## Summary

This PR updates stale CLI help text in the `extension` command group to reflect the current command names.

### Changes
- Updated outdated `everestctl plugin ...` examples to `everestctl extension ...`
- Corrected stale package comments that still referred to the old `plugin` command group
- Verified that all remaining changes are limited to help text and comments with no behavioral or command registration changes

### Why

The CLI command group was renamed from `plugin` to `extension`, but a few help examples and package comments were never updated. As a result, the documented examples referenced commands that no longer exist, making them confusing and not directly usable!

### DIFF
--- a/commands/extension/run.go
+++ b/commands/extension/run.go
@@ -32,10 +32,10 @@ var (
 		Long:  "Run a plugin's CLI container, passing extra arguments after '--'",
 		Short: "Run a plugin CLI",
 		Example: `  # Run the sql-explorer plugin CLI:
-  everestctl plugin run sql-explorer -- query --db my-db "SELECT 1"
+  everestctl extension run sql-explorer -- query --db my-db "SELECT 1"
 
   # Run with a custom container runtime:
-  everestctl plugin run sql-explorer --runtime podman -- --help`,
+  everestctl extension run sql-explorer --runtime podman -- --help`,
 		PreRun: pluginRunPreRun,
 		Run:    pluginRunRun,
 	}
--- a/commands/extension/run.go
+++ b/commands/extension/run.go
@@ -27,11 +27,11 @@ import (
 
 var (
 	pluginRunCmd = &cobra.Command{
-		Use:   "run <plugin-name> [-- <args>...]",
+		Use:   "run <extension-name> [-- <args>...]",
 		Args:  cobra.MinimumNArgs(1),
-		Long:  "Run a plugin's CLI container, passing extra arguments after '--'",
-		Short: "Run a plugin CLI",
-		Example: `  # Run the sql-explorer plugin CLI:
+		Long:  "Run an extension's CLI container, passing extra arguments after '--'",
+		Short: "Run an extension CLI",
+		Example: `  # Run the sql-explorer extension CLI:
   everestctl extension run sql-explorer -- query --db my-db "SELECT 1"
 
   # Run with a custom container runtime:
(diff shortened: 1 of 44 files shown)


Adii-45:
Hey @recharte, 

I've updated the stale CLI help examples and package comments to consistently use the current `everestctl extension` command names!

The remaining failing checks appear to be unrelated to this change:
- `Check` fails during the "Check that dev Helm chart is up-to-date" validation
- `Merge Gatekeeper` is failing as a consequence because one required check did not pass

Could you please confirm whether these are expected CI issues, or if there's anything else I should update on my side? Thanks!

recharte (changes requested):
Hey @Adii-45, thanks for the PR! Those examples were genuinely confusing to hit.

The sweep is incomplete though — only the `Example` strings changed, so the rest of the user-facing text in the group still says "plugin" while `install` already says "extension":

```
Available Commands:
  install     Install an extension
  list        List all installed plugins
  run         Run a plugin CLI
  uninstall   Uninstall a plugin
```

Could you take the remaining ones along with it? The `Long`/`Short` descriptions on `list`/`run`/`uninstall`, the `Use` line on `run` (it prints `run <plugin-name>`), and the `--name` flag help. `Plugin` as the CRD kind stays as-is — it's still the API type, and `InstalledExtension` treats it as one type of extension.

On CI: neither failure is caused by your change. The `Check` job's "Check that dev Helm chart is up-to-date" step re-pins `github.com/openeverest/helm-charts/charts/everest` to the head of the chart's `v2` branch and fails if `go.mod`/`go.sum` move — your branch point predates the last chart bump. `release-2.0` already carries the current pin, so merging it in should clear both.

Adii-45:
Thanks for the review @recharte!

I've updated the remaining `Long`, `Short`, `Use`, flag help, and related documentation comments in `list`, `run`, and `uninstall` as suggested, while leaving `install.go` and the API/CRD terminology unchanged!

Please let me know if you see anything else that should be adjusted..!

commands/extension/run.go
```diff
@@ -27,15 +27,15 @@ import (
...
 var (
 	pluginRunCmd = &cobra.Command{
-		Use:   "run <plugin-name> [-- <args>...]",
+		Use:   "run <extension-name> [-- <args>...]",
 		Args:  cobra.MinimumNArgs(1),
-		Long:  "Run a plugin's CLI container, passing extra arguments after '--'",
-		Short: "Run a plugin CLI",
+		Long:  "Run an extension's CLI container, passing extra arguments after '--'",
+		Short: "Run an extension CLI",
 		Example: `  # Run the sql-explorer plugin CLI:
```
recharte: This one prints directly under the `Long` you just fixed.

```suggestion
		Example: `  # Run the sql-explorer extension CLI:
```

recharte (changes requested):
Thanks @Adii-45, that's the sweep done — and nice touch picking up the Go doc comments too.

One leftover with no line to attach to: `install.go`'s `--name` help still says "Plugin name" where `uninstall` now says "Extension name". Fair enough if you read my last note as scoped to the files you'd already touched, but the two flags are the same thing and now disagree.

`Check` is still red, though — it looks like the merge commit got amended rather than re-merged, so the branch still points at the same `release-2.0` commit as before and the chart pin hasn't moved. Merging `release-2.0` again should clear it.

recharte (approved):
Thanks @Adii-45 looks good now 🚀

Adii-45:
@recharte Thanks for the review! Appreciate the guidance..

PR #2820, opened 2026-08-08
fix: handle SIGTERM for graceful shutdown (#2809)

### LINKED ISSUE #2809: SIGTERM signal not handled — Pod graceful shutdown bypassed in cmd/main.go
### What is the issue?

```
OpenEverest is a Kubernetes-native application. When Kubernetes terminates a Pod (during a rolling update, scale-down, or node drain), it sends a SIGTERM signal to the process. However, in cmd/main.go, the signal handler only listens for os.Interrupt (SIGINT — triggered by Ctrl+C in a terminal):

quit := make(chan os.Signal, 1)
signal.Notify(quit, os.Interrupt)
<-quit

Because SIGTERM is not registered, the Go runtime's default behavior takes over and kills the process immediately, completely bypassing the server.Shutdown(ctx) graceful shutdown logic.

Expected: When Kubernetes sends SIGTERM, the application should catch it, finish in-flight requests, and shut down cleanly.
Actual: SIGTERM is ignored, the process is killed instantly, and any in-flight database operations or API requests are dropped without cleanup.
```

### Steps to Reproduce

1. Deploy OpenEverest to a Kubernetes cluster.

2. Trigger a pod termination event — any of the following will work:
   - Run: kubectl rollout restart deployment/openeverest
   - Or:  kubectl scale deployment openeverest --replicas=0
   - Or:  kubectl drain <node-name>

3. Kubernetes sends SIGTERM to the running OpenEverest pod.

4. Observe that the process terminates instantly without running the
   graceful shutdown logic (server.Shutdown(ctx) in cmd/main.go is never reached).

5. Any in-flight API requests or database operations at the time of
   termination are dropped immediately without cleanup.

Note: This can also be reproduced locally by running the binary and
sending SIGTERM directly:
   - Run the binary: ./openeverest
   - Send SIGTERM:   

### Relevant Logs / Output

```bash
# When Kubernetes sends SIGTERM (current broken behavior):
# No graceful shutdown logs appear — process just exits instantly.

$ kubectl logs openeverest-pod-xxxx
... (normal startup logs) ...
# [NOTHING — process killed immediately, no shutdown sequence]

# Expected graceful shutdown logs should look like:
# {"level":"info","msg":"Received shutdown signal, starting graceful shutdown..."}
# {"level":"info","msg":"Waiting for in-flight requests to complete..."}
# {"level":"info","msg":"Server shut down cleanly"}

# Reproduced locally with SIGTERM:
$ ./openeverest &
 12345
$ kill -SIGTERM 12345
# Process exits immediately — no shutdown logic executed

# Compare with SIGINT (Ctrl+C) which works correctly:
$ ./openeverest
^C
# Graceful shutdown runs as expected via server.Shutdown(ctx)

# The difference confirms SIGTERM is not being caught by signal.Notify
```

### OpenEverest Version

Latest (main branch) — commit checked at time of reporting

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

Not environment-specific — this is a code-level bug in cmd/main.go.

The issue is reproducible across all environments because:
- SIGTERM is the standard POSIX termination signal
- Kubernetes always sends SIGTERM before SIGKILL (regardless of distribution)
- Go's default SIGTERM behavior (immediate exit) is consistent across all OS/platforms

Tested and confirmed via local reproduction:
- OS: Linux
- Go Version: (as defined in go.mod)
- Kubernetes: Minikube (for local reproduction)

### Possible Solution

In cmd/main.go, update the signal.Notify call to also listen for syscall.SIGTERM
alongside os.Interrupt (SIGINT):

Current code:
quit := make(chan os.Signal, 1)
signal.Notify(quit, os.Interrupt)
<-quit

Fix:
import "syscall"

quit := make(chan os.Signal, 1)
signal.Notify(quit, os.Interrupt, syscall.SIGTERM)
<-quit

This ensures that when Kubernetes sends SIGTERM during pod termination,
the signal is caught and the existing server.Shutdown(ctx) graceful
shutdown logic runs correctly — finishing in-flight requests before
the process exits.

No other changes are needed. The graceful shutdown implementation
itself is correct; it just never gets triggered for SIGTERM.

### Additional Context

```
This is a well-known Kubernetes best practice. The official Kubernetes
documentation states that SIGTERM is always the first signal sent to a
container during termination, followed by SIGKILL after the
terminationGracePeriodSeconds (default: 30s).

Reference: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

Because SIGTERM is not handled, the application effectively wastes its
entire 30-second grace period — the process dies instantly on the first
signal instead of using that window to drain connections and shut down cleanly.

This bug affects all production Kubernetes deployments of OpenEverest and
is especially impactful during:
- Rolling updates (frequent in CI/CD pipelines)
- Node maintenance / cluster upgrades
- Auto-scaling down events

The fix is a one-line change with zero risk of regression.
```

Harkirat1309:
Fixes #2809

### Problem Summary
The application only listened for `os.Interrupt` (`SIGINT`) for graceful shutdown. In Kubernetes, Pod termination (e.g., node drain, scale-down, or rolling restarts) sends a `SIGTERM` signal. Because this wasn't caught, the process exited immediately, dropping in-flight requests and database operations instead of respecting the `terminationGracePeriodSeconds` window.

### Root Cause
In `cmd/main.go`, `signal.Notify(quit, os.Interrupt)` was missing `syscall.SIGTERM`.

### Fix Summary
- Added `syscall.SIGTERM` to the `signal.Notify` list.
- Extracted the signal-wait logic into a minimally-scoped helper function (`waitForShutdownSignal`) so it can be tested independently without refactoring the surrounding application bootstrap and server shutdown logic.

### Testing Performed
- Added a regression test `TestWaitForShutdownSignal` in `cmd/main_test.go` that runs the extracted signal-wait function in a goroutine, triggers a `syscall.SIGTERM` against the process, and asserts that the shutdown channel is unblocked successfully.
- Tested the unfixed code with `SIGTERM` (which fails) and the fixed code (which passes).
- Verified `go fmt`, `go vet`, and `go test ./...` locally.


Harkirat1309:
Hi @spron-in , this PR implements the fix for #2809 as discussed. The CI checks are passing, and I've added a regression test for the SIGTERM handler. Let me know if you need any changes!

spron-in:
Thanks @Harkirat1309 .

Please check the failing tests.
For copyright headers just run `make copyright-headers` in the root of the repo.
For CI looks like lint-test fails for the test file.

Harkirat1309:
Thanks @spron-in!

I've run make copyright-headers from the repo root and fixed the copyright headers. The failing tests and lint issue in the test file have been resolved as well. CI should be green now — please take another look when you get a chance.

cmd/main_test.go
recharte, line 1: I don't think this test earns its place, and after the race fix it no longer guards anything.

The test now calls `signal.Notify(quit, os.Interrupt, syscall.SIGTERM)` itself before starting the goroutine, so SIGTERM lands on `quit` regardless of what `waitForShutdownSignal` registers. I checked by reverting the production change back to `signal.Notify(quit, os.Interrupt)` and re-running:

```
$ go test ./cmd/ -run TestWaitForShutdownSignal -v -count=1
=== RUN   TestWaitForShutdownSignal
--- PASS: TestWaitForShutdownSignal (0.00s)
PASS
```

A better assertion can't fix that. Without the pre-registration the test does mean something, but then it's racing the goroutine to reach `signal.Notify` first — and losing that race doesn't produce a failure, it produces a dead test binary, because an unregistered SIGTERM takes the default disposition and kills the whole `package main` run. Pre-registering removes the race and the assertion together. `os/signal` offers no way to ask which signals a channel is registered for, so the only thing left is injecting a fake notify function, i.e. asserting that we call a stdlib function with particular arguments.

Could we drop this file? The fix for #2809 is a one-line change to a constant argument list.

cmd/main.go
```diff
@@ -83,8 +85,7 @@ func main() {
 	}
 
 	quit := make(chan os.Signal, 1)
-	signal.Notify(quit, os.Interrupt)
-	<-quit
+	waitForShutdownSignal(quit)
```
recharte: `quit` reads as a parameter, but only a buffered channel works here: `signal.Notify` never blocks on send, so an unbuffered one compiles and misses the signal. Inline, `make(chan os.Signal, 1)` sat on the line above the `Notify` call and you could see it was right — behind a function boundary it is an unchecked precondition.

```suggestion
	quit := make(chan os.Signal, 1)
	signal.Notify(quit, os.Interrupt, syscall.SIGTERM)
	<-quit
```

That leaves `waitForShutdownSignal` unused, so it can come out too.

recharte (changes requested):
Hey @Harkirat1309, thanks for the PR, and for chasing the CI failures down — the signal registration itself is exactly what #2809 needed.

The test is the one I'd want resolved before this goes in; the helper is a smaller point.

Harkirat1309:
@recharte I have resolved all merge conflicts. Please review.

recharte (approved):
Thanks @Harkirat1309

PR #2821, opened 2026-08-08
fix(cli): reject unknown subcommands on grouping commands

### LINKED ISSUE #2816: CLI parent commands (namespaces, settings, accounts) silently exit with 0 on invalid subcommands
### What is the issue?

# CLI parent commands (namespaces, settings, accounts) silently exit with 0 on invalid subcommands

## Description

Parent CLI commands that exist solely to group subcommands — such as `everestctl namespaces`, `everestctl settings`, `everestctl accounts`, `everestctl settings rbac`, and `everestctl settings oidc` — are misconfigured with `Args: cobra.ExactArgs(1)` combined with an empty `Run` function. As a result, when a user passes an invalid or misspelled subcommand, Cobra's built-in "unknown command" detection is bypassed, the empty `Run` executes as a silent no-op, and the process exits with status code `0` as if the command succeeded.

This creates a misleading user experience (typos appear to "work" with zero feedback) and is especially dangerous in automation/CI pipelines that check exit codes to detect failures — a misspelled `everestctl` command in a script would silently do nothing and still report success.

## Impact

- Users get no feedback when they mistype a subcommand, making the CLI feel broken or unresponsive.
- CI/CD pipelines and automation scripts that gate on exit codes cannot detect a misconfigured or mistyped `everestctl` invocation, since it returns `0` regardless.
- Increases support burden, as failures caused by typos are indistinguishable from successful runs in logs.

## Affected files

| File | Lines | Command |
|---|---|---|
| `commands/namespaces.go` | ~28-31 | `everestctl namespaces` |
| `commands/settings.go` | ~27-30 | `everestctl settings` |
| `commands/accounts.go` | ~27-30 | `everestctl accounts` |
| `commands/settings/rbac.go` | ~27-30 | `everestctl settings rbac` |
| `commands/settings/oidc.go` | ~27-30 | `everestctl settings oidc` |

Example (`commands/namespaces.go`):
```go
var namespacesCmd = &cobra.Command{
	Use:  "namespaces [flags]",
	Args: cobra.ExactArgs(1),
	Run:  func(_ *cobra.Command, _ []string) {},
}


### Steps to Reproduce

```markdown
1. Build or install `everestctl` from the `main` branch (no cluster connection is required, since the bug occurs before any real command logic executes).
2. Run any parent command with an invalid or misspelled subcommand, for example:
   ```bash
   everestctl namespaces llist
   ```
3. Observe that no output/error is printed, and the command exits as if it succeeded.
4. Check the exit code:
   ```bash
   echo $?
   ```
5. **Result**: Exit code is `0` (success), even though `llist` is not a valid subcommand.

Same behavior can be reproduced with the other affected parent commands:
```bash
everestctl settings unknown-command
echo $?   # prints 0

everestctl accounts something
echo $?   # prints 0

everestctl settings rbac bogus
echo $?   # prints 0

everestctl settings oidc bogus
echo $?   # prints 0
```
```

### Relevant Logs / Output

```markdown
**Actual output:**
```bash
$ everestctl namespaces llist
$ echo $?
0
```
No error message, no help text, no indication that `llist` is invalid — the shell prompt returns immediately with a success exit code.

```bash
$ everestctl settings unknown-command
$ echo $?
0
```

```bash
$ everestctl accounts something
$ echo $?
0
```

**Expected output (for comparison):**
```bash
$ everestctl namespaces llist
Error: unknown command "llist" for "everestctl namespaces"
Run 'everestctl namespaces --help' for usage.
$ echo $?
1
```

There are no crash logs, stack traces, or stderr output associated with this bug — the issue is precisely the *absence* of any output or error signal where one is expected.
```

### OpenEverest Version

v1.0.0

### Installation Method

CLI (everestctl)

### Kubernetes Distribution

EKS

### Environment Details

- Kubernetes Version: v1.28.0
- Host OS: Ubuntu 22.04
- Cloud Provider: AWS (EKS)
- Go version: (as specified in go.mod for the everestctl build)
- Cobra library: spf13/cobra (as vendored in go.sum)

Note: This bug is reproducible independent of the Kubernetes/cluster environment, since it occurs purely at the CLI argument-parsing layer before any cluster connection or API call is made.


### Possible Solution

Remove the empty `Run` function and instead implement `RunE` on the affected parent commands (`commands/namespaces.go`, `commands/settings.go`, `commands/accounts.go`, `commands/settings/rbac.go`, `commands/settings/oidc.go`) so that an unrecognized subcommand returns an explicit error and non-zero exit code, e.g.:

```go
var namespacesCmd = &cobra.Command{
	Use:  "namespaces [flags]",
	Long: "Manage Everest database namespaces",
	RunE: func(cmd *cobra.Command, args []string) error {
		if len(args) > 0 {
			return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
		}
		return cmd.Help()
	},
}


Simply deleting Args/Run without this validation is not sufficient — Cobra would just fall back to cmd.Help() via flag.ErrHelp, which still exits 0.

### Additional Context


**Additional Context:**
```markdown
This is a CLI-level bug and does not depend on any particular cluster state, workload, or OpenEverest resource configuration. It affects any user or automation script invoking `everestctl` with a misspelled subcommand under `namespaces`, `settings`, `accounts`, `settings rbac`, or `settings oidc`.

I'd like to work on this fix — happy to submit a PR implementing the `RunE` validation across all affected command files once assigned.

### PR DESCRIPTION
**Pull request by vansh-commits**
## Why

`everestctl namespaces`, `settings`, `accounts`, `settings rbac` and `settings oidc`
exist only to group subcommands. All five accepted an unrecognised subcommand and
exited `0`, so a typo was indistinguishable from success — silently for three of
them, and with a help dump for the other two. Scripts and CI pipelines that gate on
the exit code could not tell a misspelled invocation from a real run.

Fixes #2816 

## Root cause

Cobra checks `!c.Runnable() → flag.ErrHelp` (`command.go:955`) *before* it validates
`Args` (`command.go:968`). That produced two variants of the same bug:

- `namespaces`, `settings`, `accounts` — `Args: cobra.ExactArgs(1)` plus an empty
  `Run`. The stray subcommand satisfied the arg count, the empty `Run` executed, and
  the process exited `0` with no output at all.
- `settings rbac`, `settings oidc` — `Args: cobra.ExactArgs(1)` and **no** `Run`.
  Being unrunnable, they returned `flag.ErrHelp` before `Args` was ever consulted,
  printed help, and exited `0`.

The consequence is that removing `Args`/`Run` alone does not fix this: a command must
be runnable for its `Args` validator to run at all.

## What changed

Each grouping command now sets `Args: cobra.NoArgs` and a `RunE` that prints help.
`cobra.NoArgs` already emits the exact "unknown command" error we want, so no custom
error construction was needed, and `RunE` makes the command runnable so the validator
is reached.

This matches the convention already used elsewhere in the CLI — `status`, `install`,
`upgrade`, `uninstall` and `version` all use `cobra.NoArgs` and were already correct.

## Behaviour

Before:

```console
$ everestctl namespaces llist
$ echo $?
0
```

After:

```console
$ everestctl namespaces llist
Error: unknown command "llist" for "everestctl namespaces"
Usage:
  everestctl namespaces <command> [flags]
$ echo $?
1
```

Invoked bare, the commands print their help and exit `0`:

```console
$ everestctl namespaces
Manage Everest database namespaces

Usage:
  everestctl namespaces <command> [flags]
$ echo $?
0
```

**One change beyond the reported bug:** bare `everestctl namespaces` previously exited
`1` with `accepts 1 arg(s), received 0`. It now prints help and exits `0`, which is the
conventional behaviour for a command that only groups subcommands, and is consistent
with `everestctl --help`. Flagging it in case reviewers would rather keep the non-zero
exit there.

## Testing

Adds `commands/commands_test.go`, the first test in this package. It drives the real
`rootCmd` and asserts that each grouping command returns an error for an unknown
subcommand, succeeds when invoked bare, and still routes to genuine subcommands.

I confirmed the tests fail against the previous code by reverting `namespaces.go`
(empty `Run`) and `settings/rbac.go` (no `Run`) and re-running — both variants of the
bug are caught.

`make test` passes in full under `-race`. `golangci-lint` reports no issues in
`commands/...`. `make copyright-check` passes; four files picked up the OpenEverest
copyright line via `make copyright-headers`.

`make check` still reports two `misspell` findings in
`internal/server/handlers/rbac/kubernetes_test.go`. These are pre-existing on `main`
and unrelated to this change, so I have left them out to keep this PR to one concern.

### DIFF
--- a/commands/accounts.go
+++ b/commands/accounts.go
@@ -1,5 +1,6 @@
 // everest
 // Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -24,10 +25,10 @@ import (
 
 var accountsCmd = &cobra.Command{
 	Use:   "accounts <command> [flags]",
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.NoArgs,
 	Long:  "Manage Everest accounts",
 	Short: "Manage Everest accounts",
-	Run:   func(_ *cobra.Command, _ []string) {},
+	RunE:  func(cmd *cobra.Command, _ []string) error { return cmd.Help() },
 }
 
 func init() {
--- a/commands/commands_test.go
+++ b/commands/commands_test.go
@@ -1,0 +1,88 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// parentCommands are the commands that exist only to group subcommands.
+var parentCommands = [][]string{
+	{"namespaces"},
+	{"settings"},
+	{"accounts"},
+	{"settings", "rbac"},
+	{"settings", "oidc"},
+}
+
+// execRoot runs the root command with the given arguments and returns the error
+// main() would turn into a non-zero exit code. The commands share the package-level
+// rootCmd, so these must not run in parallel.
+func execRoot(t *testing.T, args ...string) error {
+	t.Helper()
+
+	var out bytes.Buffer
+	rootCmd.SetOut(&out)
+	rootCmd.SetErr(&out)
+	rootCmd.SetArgs(args)
+	t.Cleanup(func() {
+		rootCmd.SetOut(nil)
+		rootCmd.SetErr(nil)
+	})
+	return rootCmd.Execute()
+}
+
+// Cobra checks Runnable() before it validates Args, so a parent command with no
+// Run at all returns flag.ErrHelp and exits 0 whatever it is given, and one with
+// an empty Run exits 0 silently. Either way a typo looks like success, which
+// hides mistakes from CI scripts that gate on the exit code.
+//
+//nolint:paralleltest // execRoot mutates the package-level rootCmd.
+func TestParentCommandsRejectUnknownSubcommand(t *testing.T) {
+	for _, parent := range parentCommands {
+		path := strings.Join(parent, " ")
+		t.Run(path, func(t *testing.T) {
+			err := execRoot(t, append(parent, "bogus-subcommand")...)
+			require.Error(t, err, "%s accepted an unknown subcommand", path)
+			require.Contains(t, err.Error(), `unknown command "bogus-subcommand"`)
+		})
+	}
+}
+
+// Invoked with no subcommand, a parent command should print its help and succeed.
+//
+//nolint:paralleltest // execRoot mutates the package-level rootCmd.
+func TestParentCommandsWithoutArgsPrintHelp(t *testing.T) {
+	for _, parent := range parentCommands {
+		path := strings.Join(parent, " ")
+		t.Run(path, func(t *testing.T) {
+			require.NoError(t, execRoot(t, parent...))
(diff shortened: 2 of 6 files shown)


chilagrow:
Thanks for your contribution @vansh-commits, as there is already work started to handle this https://github.com/openeverest/openeverest/issues/2816#issuecomment-5227836075 I am closing this Pull Request 🙏

To avoid duplicated work in the future, please discuss on the issue before opening the PR. Thanks again for your time and your contribution.

PR #2822, opened 2026-08-08
fix(rbac): add configurable timeout to ConfigMap adapter Kubernetes API calls (#2799)

### LINKED ISSUE #2799: Unbounded context.Background() in RBAC ConfigMap adapter can cause server hang during RBAC initialization
### What is the issue?

The issue is that the RBAC ConfigMap adapter uses context.Background() when calling the Kubernetes API to load the RBAC policy.

That means the request has no timeout. If the API server is slow, unreachable, or the network is broken, the call can block indefinitely. As a result, RBAC policy loading may hang and prevent the server from starting or responding normally.

In short: a Kubernetes API call in RBAC initialization can hang forever because it does not use a timeout.

### Steps to Reproduce

Set up OpenEverest with RBAC enabled in a Kubernetes cluster.
Introduce latency or block access to the Kubernetes API server, for example by stopping the API server, dropping packets with a firewall rule, or adding network delay with tc.
Restart the OpenEverest API server or trigger RBAC initialization.
Observe that the LoadPolicy() call waits indefinitely on kubeClient.GetConfigMap(context.Background(), ...) because no timeout is set.
The server may hang during startup or become unresponsive until Kubernetes connectivity is restored.
Expected Behavior
RBAC policy loading should fail fast or time out after a reasonable duration.

Actual Behavior
The Kubernetes API call can block indefinitely,

### Relevant Logs / Output

Relevant Logs / Output

Server Logs (stuck at RBAC init):
2026-08-05T10:00:00Z INFO  Starting OpenEverest API Server...
2026-08-05T10:00:01Z INFO  Initializing Casbin enforcer...
2026-08-05T10:00:01Z DEBUG Loading RBAC policy from ConfigMap: everest-rbac
[... no further logs; process remains active but unresponsive ...]

Goroutine Dump:
goroutine 42 [select]:
k8s.io/apimachinery/pkg/util/wait.PollImmediateUntil(...)
k8s.io/client-go/tools/cache.(*Reflector).Run(...)
...
pkg/rbac/configmap-adapter.(*Adapter).LoadPolicy(...)
    /app/pkg/rbac/configmap-adapter/adapter.go:XX +0x85
``

### OpenEverest Version

Unknown

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Use a derived context with a reasonable timeout in LoadPolicy() instead of passing context.Background() directly to kubeClient.GetConfigMap(). This ensures the Kubernetes API call fails predictably rather than blocking indefinitely when the API server is slow or unreachable.

Example:
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()

cm, err := a.kubeClient.GetConfigMap(ctx, a.namespacedName)
if err != nil {
    return err
}

Optionally, make the timeout configurable so operators can tune it for different cluster environments.

### Additional Context

This issue affects service reliability during startup because RBAC policy loading happens as part of Casbin initialization. Since persist.Adapter does not accept a context parameter, the adapter implementation must enforce its own timeout before making Kubernetes API calls. Without this safeguard, transient control-plane outages or network partitions can cause the API server to hang indefinitely instead of failing fast with a clear error.

Harkirat1309:
## Summary

Fixes #2799 
The RBAC ConfigMap adapter's `LoadPolicy()` method in `pkg/rbac/configmap-adapter/adapter.go` used `context.Background()` with **no timeout** when calling `kubeClient.GetConfigMap(...)`. If the Kubernetes API server was slow, unreachable, or the network was partitioned, this call could block **indefinitely**, hanging RBAC initialization and preventing the server from starting or responding.

## Root Cause Analysis

In the original code:

```go
func (a *Adapter) LoadPolicy(model model.Model) error {
    cm, err := a.kubeClient.GetConfigMap(context.Background(), a.namespacedName)
    if err != nil {
        return err
    }
}
```

`context.Background()` creates an empty context with **no deadline, no timeout, and no cancellation signal**. Since `LoadPolicy()` is called during server startup and RBAC policy reload, a hang here blocks the entire server.

### `pkg/rbac/configmap-adapter/adapter.go`
- Added `defaultKubeAPITimeout = 10 * time.Second` constant
- - Added `timeout` field to `Adapter` struct
- - Added `Option` type and `WithTimeout(d)` functional option
- - Updated `New()` to accept `...Option` (backward-compatible)
- - Fixed `LoadPolicy()` to use `context.WithTimeout` instead of `context.Background()`
- - Improved error wrapping with ConfigMap name for diagnosability
### `pkg/rbac/configmap-adapter/adapter_test.go` (NEW)
Added 5 unit tests with a mock k8s client supporting configurable delay.

### Real test output:

```
$ go test -v -count=1 ./pkg/rbac/configmap-adapter/...
=== RUN   TestLoadPolicy
=== PAUSE TestLoadPolicy
=== RUN   TestDefaultTimeout
=== PAUSE TestDefaultTimeout
=== RUN   TestWithTimeout
=== PAUSE TestWithTimeout
=== CONT  TestLoadPolicy
=== RUN   TestLoadPolicy/happy_path_-_loads_policy_within_timeout
=== RUN   TestLoadPolicy/timeout_-_API_call_exceeds_timeout
=== RUN   TestLoadPolicy/API_error_-_wraps_error_with_configmap_context
--- PASS: TestDefaultTimeout (0.00s)
--- PASS: TestWithTimeout (0.00s)
--- PASS: TestLoadPolicy (0.00s)
    --- PASS: TestLoadPolicy/happy_path_-_loads_policy_within_timeout (0.00s)
    --- PASS: TestLoadPolicy/API_error_-_wraps_error_with_configmap_context (0.00s)
    --- PASS: TestLoadPolicy/timeout_-_API_call_exceeds_timeout (0.10s)
PASS
ok      github.com/percona/everest/pkg/rbac/configmap-adapter   6.592s
```

- `go vet ./pkg/rbac/configmap-adapter/...` - no issues found
## Backward Compatibility

No breaking changes. The `New()` constructor accepts variadic `...Option` - existing callers pass no options and work identically. The `persist.Adapter` interface contract is unchanged.

### DIFF
--- a/pkg/rbac/configmap-adapter/adapter.go
+++ b/pkg/rbac/configmap-adapter/adapter.go
@@ -20,7 +20,9 @@ package configmapadapter
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
+	"time"
 
 	"github.com/casbin/casbin/v2/model"
 	"go.uber.org/zap"
@@ -30,6 +32,8 @@ import (
 
 	rbacutils "github.com/percona/everest/pkg/rbac/utils"
 )
+
+const defaultKubeAPITimeout = 10 * time.Second
 
 type k8s interface {
 	GetConfigMap(ctx context.Context, key ctrlclient.ObjectKey) (*corev1.ConfigMap, error)
@@ -41,6 +45,18 @@ type Adapter struct {
 	kubeClient     k8s
 	namespacedName types.NamespacedName
 	l              *zap.SugaredLogger
+	timeout        time.Duration
+}
+
+// Option is a functional option for the Adapter.
+type Option func(*Adapter)
+
+// WithTimeout sets the timeout for Kubernetes API calls made by the adapter.
+// If not set, defaultKubeAPITimeout (10s) is used.
+func WithTimeout(d time.Duration) Option {
+	return func(a *Adapter) {
+		a.timeout = d
+	}
 }
 
 // New constructs a new adapter that manages a policy inside a ConfigMap.
@@ -48,12 +64,18 @@ func New(
 	l *zap.SugaredLogger,
 	kubeClient k8s,
 	namespacedName types.NamespacedName,
+	opts ...Option,
 ) *Adapter {
-	return &Adapter{
+	a := &Adapter{
 		kubeClient:     kubeClient,
 		namespacedName: namespacedName,
 		l:              l,
+		timeout:        defaultKubeAPITimeout,
 	}
+	for _, opt := range opts {
+		opt(a)
+	}
+	return a
 }
 
 // ConfigMap returns the configmap used for RBAC.
@@ -63,9 +85,12 @@ func (a *Adapter) ConfigMap(ctx context.Context) (*corev1.ConfigMap, error) {
 
 // LoadPolicy loads all policy rules from the storage.
 func (a *Adapter) LoadPolicy(model model.Model) error {
-	cm, err := a.kubeClient.GetConfigMap(context.Background(), a.namespacedName)
+	ctx, cancel := context.WithTimeout(context.Background(), a.timeout)
+	defer cancel()
+
+	cm, err := a.kubeClient.GetConfigMap(ctx, a.namespacedName)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to load RBAC policy configmap %q: %w", a.namespacedName, err)
 	}
 
 	data, ok := cm.Data["policy.csv"]
--- a/pkg/rbac/configmap-adapter/adapter.go
+++ b/pkg/rbac/configmap-adapter/adapter.go
@@ -90,7 +90,7 @@ func (a *Adapter) LoadPolicy(model model.Model) error {
 
 	cm, err := a.kubeClient.GetConfigMap(ctx, a.namespacedName)
 	if err != nil {
-		return fmt.Errorf("failed to load RBAC policy configmap %q: %w", a.namespacedName, err)
+		return fmt.Errorf("failed to load RBAC policy configmap %s: %w", a.namespacedName, err)
 	}
 
 	data, ok := cm.Data["policy.csv"]
--- a/pkg/rbac/configmap-adapter/adapter.go
+++ b/pkg/rbac/configmap-adapter/adapter.go
@@ -48,34 +48,18 @@ type Adapter struct {
 	timeout        time.Duration
 }
 
-// Option is a functional option for the Adapter.
-type Option func(*Adapter)
-
-// WithTimeout sets the timeout for Kubernetes API calls made by the adapter.
-// If not set, defaultKubeAPITimeout (10s) is used.
-func WithTimeout(d time.Duration) Option {
-	return func(a *Adapter) {
-		a.timeout = d
-	}
-}
-
 // New constructs a new adapter that manages a policy inside a ConfigMap.
 func New(
 	l *zap.SugaredLogger,
 	kubeClient k8s,
 	namespacedName types.NamespacedName,
-	opts ...Option,
 ) *Adapter {
-	a := &Adapter{
+	return &Adapter{
 		kubeClient:     kubeClient,
 		namespacedName: namespacedName,
 		l:              l,
 		timeout:        defaultKubeAPITimeout,
 	}
-	for _, opt := range opts {
-		opt(a)
-	}
-	return a
 }
 
 // ConfigMap returns the configmap used for RBAC.
--- a/pkg/rbac/configmap-adapter/adapter_test.go
+++ b/pkg/rbac/configmap-adapter/adapter_test.go
@@ -1,0 +1,172 @@
+// everest
+// Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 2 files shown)


Harkirat1309:
Hi @spron-in, this PR fixes #2799 as discussed.

- Replaced `context.Background()` with `context.WithTimeout()` in the RBAC ConfigMap adapter
- Added configurable timeout (default: 10s)

spron-in (changes requested):
Code wise - looks good. But please look at failing tests. Lint seems to be off.

Harkirat1309:
@spron-in Thanks!! for reviewing. I have resolved all conflicts.

spron-in:
@Harkirat1309 sorry, but linters are still failing

Harkirat1309:
@spron-in Really sorry about that! Just pushed the fixes — please take another 
look when you get a chance.

pkg/rbac/configmap-adapter/adapter_test.go
```diff
@@ -0,0 +1,173 @@
...
+			case err := <-done:
+				if tc.wantErr {
+					if tc.errSubstr != "" {
+						require.ErrorContains(t, err, tc.errSubstr)
+					} else {
+						require.Error(t, err)
+					}
+				} else {
+					require.NoError(t, err)
+				}
```
recharte, line 127: The happy path asserts only that `LoadPolicy` returns no error, so it passes even when nothing reaches the model — which is the failure that actually matters for an RBAC loader.

```suggestion
				} else {
					require.NoError(t, err)
					got, err := m.GetPolicy("p", "p")
					require.NoError(t, err)
					assert.Equal(t, [][]string{{"role:admin", "database-clusters", "*", "*"}}, got)
				}
```

pkg/rbac/configmap-adapter/adapter.go
```diff
@@ -41,19 +45,37 @@ type Adapter struct {
...
+// Option is a functional option for the Adapter.
+type Option func(*Adapter)
+
+// WithTimeout sets the timeout for Kubernetes API calls made by the adapter.
+// If not set, defaultKubeAPITimeout (10s) is used.
+func WithTimeout(d time.Duration) Option {
+	return func(a *Adapter) {
+		a.timeout = d
+	}
 }
```
recharte: `pkg/rbac/rbac.go:203` is the only `New` call site and it passes no options, so `Option` and `WithTimeout` exist solely for `TestWithTimeout` and `TestDefaultTimeout`. Since the test file is in `package configmapadapter`, it can set `a.timeout` directly — the export isn't what makes the timeout testable.

I'd unexport the option, or drop it and set the field in the test. The bounded call is what the issue needs; the configurability it floated is for operators, and that would mean an envconfig field in `cmd/config/config.go` threaded through `rbac.NewEnforcer` — worth doing when someone asks to tune it, not before.

pkg/rbac/configmap-adapter/adapter_test.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func TestWithTimeout(t *testing.T) {
+	t.Parallel()
+	l := zap.NewNop().Sugar()
+	nn := types.NamespacedName{Namespace: "ns", Name: "cm"}
+	custom := 500 * time.Millisecond
+	adapter := New(l, &mockK8s{}, nn, WithTimeout(custom))
+	assert.Equal(t, custom, adapter.timeout)
+}
```
recharte: Both assert on the unexported `timeout` field, so they'd keep passing if `LoadPolicy` stopped using `a.timeout` altogether. The override is already covered behaviourally by the timeout case above, so I'd drop them.

pkg/rbac/configmap-adapter/adapter_test.go
```diff
@@ -0,0 +1,173 @@
...
+					if tc.errSubstr != "" {
+						require.ErrorContains(t, err, tc.errSubstr)
+					} else {
+						require.Error(t, err)
+					}
+				} else {
+					require.NoError(t, err)
+				}
+			case <-time.After(2 * time.Second):
+				t.Fatal("LoadPolicy did not return within bounded time — possible hang")
```
recharte: `mockK8s` returns after `delay` whether or not the context is bounded, so a regression that dropped the timeout would fail on the missing error rather than hang — the goroutine and the 2s `time.After` never come into play. Calling `adapter.LoadPolicy(m)` inline tests the same thing.

pkg/rbac/configmap-adapter/adapter_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// testRBACModel is a minimal Casbin RBAC model definition matching the
+// production model in data/rbac/model.conf.
+const testRBACModel = `
```
recharte: The comment calls this minimal, but it's byte-identical to `data/rbac/model.conf`, so it's a second copy to keep in sync. Package `data` imports nothing but `embed`, so the real file can be read here:

```go
modelData, err := fs.ReadFile(data.RBAC, "rbac/model.conf")
require.NoError(t, err)
m, err := model.NewModelFromString(string(modelData))
require.NoError(t, err)
```

Harkirat1309:
@spron-in @recharte I've fixed all the linter and formatting issues. The code is now passing all Check/Test/CLI CI jobs. Could you please re-review?

recharte (approved):
Thanks @Harkirat1309 looks good

PR #2823, opened 2026-08-08
fix(docs): align provider filter documentation and SDK examples with providerRef.name (#2804)

### LINKED ISSUE #2804: providers filter on extension points documents values that cannot match providerRef.name
### What is the issue?

The `providers` filter on plugin extension points is documented with values that cannot match what the filter compares against, so a plugin following the documented example is hidden on every instance and never renders.

The filter itself works. `pages/db-cluster-details/db-cluster-details.tsx:82-86` and `pages/database-form/plugin-form-sections.tsx:77-81` both do:

```ts
.filter(
  (ext) =>
    !ext.providers?.length ||
    (engineType != null && ext.providers.includes(engineType))
)
```

and `engineType` is `instance?.spec?.providerRef?.name` (`db-cluster-details.tsx:72`), which is a Provider CR name such as `percona-server-mongodb`.

But the documentation tells authors to write something else. `PluginExtensionPoint.Providers` in `api/extensions/v1alpha1/plugin_types.go` says:

```go
// Providers is an optional list of database engine types this extension point
// applies to. Values match spec.engine.type on the DatabaseCluster CR:
// "postgresql", "psmdb", "pxc".
```

and `ui/packages/plugin-sdk/src/types.ts:54-55` repeats it:

```ts
 * @example providers: ["postgresql"]        // PostgreSQL only
 * @example providers: ["psmdb", "pxc"]      // MongoDB and MySQL only
```

`spec.engine.type` is v1 vocabulary. v2's `Instance` has `spec.providerRef.name` instead. So `["psmdb"].includes("percona-server-mongodb")` is false and the extension point is filtered out everywhere, with no error and nothing in the console.

`psmdb` does exist in v2 as `provider.shortName`, but `provider-percona-server-mongodb/values.yaml` documents shortName as being for preset names, not as the identity used here.

This is read from source. I have not run it against a cluster.

### Steps to Reproduce

1. Install any plugin whose `Plugin` CR declares an extension point with `providers: ["psmdb"]`, for example a `clusterDetailTab`.
2. Open a MongoDB instance in the UI.
3. The tab does not appear. With `providers` removed entirely it appears on every engine, which is the only way to get it to render at all today.

Same result for `instanceCreateFormSection` and `instanceEditFormSection` through `plugin-form-sections.tsx`.

### Relevant Logs / Output

Nothing is logged. That is most of the problem: the filter silently evaluates to false, so a plugin author sees an extension point that never renders and no indication why.

The values being compared:

```
ext.providers  = ["psmdb"]                     # what the docs and SDK examples say to write
engineType     = "percona-server-mongodb"      # instance.spec.providerRef.name
```

### OpenEverest Version

release-2.0 @ eefde0ef225cceea19fc85565fe03e211825c86c

### Installation Method

Manifests / Manual

### Kubernetes Distribution

Other (please specify below)

### Environment Details

Read from source on `release-2.0`, no cluster.

### Possible Solution

Two ways round it and I do not know which you would prefer, since it depends on what `providers` is meant to identify.

If it should match the Provider CR name, then the doc comment on `PluginExtensionPoint.Providers` and the `@example` lines in `plugin-sdk/src/types.ts` need updating to real provider names, and the reference to `spec.engine.type` should go since v2 has no such field.

If it should stay short and engine-flavoured (`psmdb`, `pxc`, `postgresql`), then the comparison needs to resolve the instance's provider to its `shortName` before matching, and the docs should say that is what it matches.

Either way it would be worth deciding whether an unmatched `providers` value should be visible somewhere rather than silently filtering everything out. A console warning when a declared value matches no known provider would have made this obvious immediately.

Happy to open a PR for whichever direction you pick, once you have picked one. I have not opened one because the two options have different blast radii and this is your call, not mine.

### Additional Context

`PluginTabHost.tsx` is a smaller, separate wrinkle in the same area: it resolves the `:tabs` route segment on `ext.type` and `ext.path` only, with no `providers` check. So the tab bar correctly hides a mismatched tab while a direct URL to `/plugins/.../<path>` still renders it. Worth touching at the same time as whatever fix lands here.

HamidKhan1001:
### Summary
Fixes #2804

Aligns doc comments in `plugin_types.go` and JSDoc examples in `@openeverest/plugin-sdk` (`types.ts`) with actual runtime comparison logic. 

### Context
Currently, `pages/db-cluster-details/db-cluster-details.tsx` and `plugin-form-sections.tsx` filter extension points against `engineType` derived from `instance?.spec?.providerRef?.name` (e.g., `percona-server-mongodb`). However, the existing doc comments and JSDoc examples previously directed authors to use v1 `spec.engine.type` shortnames (e.g., `psmdb`), causing declared extension points to silently fail filtering.

### Changes
- Updated `PluginExtensionPoint.Providers` doc comment in `api/extensions/v1alpha1/plugin_types.go` to reference `spec.providerRef.name`.
- Updated `@example` JSDoc annotations in `ui/packages/plugin-sdk/src/types.ts` to use valid Provider CR names (e.g., `percona-server-mongodb`).

### DIFF
--- a/api/extensions/v1alpha1/plugin_types.go
+++ b/api/extensions/v1alpha1/plugin_types.go
@@ -1,0 +1,228 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	common "github.com/openeverest/openeverest/v2/api/common/v1alpha1"
+)
+
+// PluginSpec defines the desired state of Plugin
+type PluginSpec struct {
+	// DisplayName is the human-readable name shown in the UI sidebar.
+	// +required
+	DisplayName string `json:"displayName"`
+
+	// Description is a short human-readable description of what the plugin does.
+	// +optional
+	Description string `json:"description,omitempty"`
+
+	// Version is the SemVer version of the plugin.
+	// +optional
+	Version string `json:"version,omitempty"`
+
+	// Vendor is the name of the plugin author or organisation.
+	// +optional
+	Vendor string `json:"vendor,omitempty"`
+
+	// Icon is a URL to the plugin's icon image.
+	// +optional
+	Icon string `json:"icon,omitempty"`
+
+	// CompatibleHostVersions is a SemVer range expression specifying which
+	// OpenEverest host versions this plugin supports (e.g. ">=2.0.0 <3.0.0").
+	// +optional
+	CompatibleHostVersions string `json:"compatibleHostVersions,omitempty"`
+
+	// Frontend defines the optional frontend contribution of the plugin.
+	// +optional
+	Frontend *PluginFrontend `json:"frontend,omitempty"`
+
+	// Backend defines the optional backend contribution of the plugin.
+	// +optional
+	Backend *PluginBackend `json:"backend,omitempty"`
+
+	// Enabled controls whether the plugin is active. A disabled plugin
+	// is not returned by the list endpoint and its proxy routes are inactive.
+	// +optional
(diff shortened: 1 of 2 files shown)


spron-in:
Hmm. It is weird. The plugin extension points fix can't go into `main` branch.
It should be going towards `release-2.0`. Or I completely missed the point of this fix. 

Are you sure you did the correct thing @HamidKhan1001 ?

spron-in (changes requested):
Merge should go towards release-2.0 branch.

spron-in:
@HamidKhan1001 I will close this one as branch is wrong and will go with another PR.

HamidKhan1001:
Thanks for clarifying, @spron-in! That makes total sense regarding the release-2.0 target branch versus main. Glad to see #2840 merged to get the fix in. I'll keep the release branching in mind for future contributions!

PR #2830, opened 2026-08-09
fix(rbac): preserve informer start errors

### LINKED ISSUE #2763: Actual error from `inf.Start()` is silently discarded and replaced with a stale error variable
### What is the issue?

When initializing the RBAC ConfigMap informer in `pkg/rbac/rbac.go`, if `inf.Start()` fails, the actual error returned by `Start()` is completely discarded. Instead, an older `err` variable (from an earlier `informer.New` call) is joined and returned.

In `pkg/rbac/rbac.go` (around line 137):
```go
if inf.Start(ctx, &corev1.ConfigMap{}) != nil {
    return errors.Join(err, errors.New("failed to watch RBAC ConfigMap"))
}


### Steps to Reproduce


```markdown
1. Force a failure when starting the RBAC informer (e.g., configure invalid RBAC permissions so `inf.Start()` fails with a permission denied error).
2. The server will fail to start the informer, but the logs will only say "failed to watch RBAC ConfigMap".
3. The underlying Kubernetes API error ("permission denied") is lost.
```

### Relevant Logs / Output
```
=== RUN   TestRBACErrorSwallow
    rbac_bug_test.go:29: BUG CONFIRMED: The actual error from inf.Start() was swallowed. Final error is: initial error
        failed to watch RBAC ConfigMap
--- FAIL: TestRBACErrorSwallow (0.00s)
FAIL
```
### OpenEverest Version

main branch

### Installation Method

Compiled from source / run via local backend setup

### Kubernetes Distribution

k3d (local test cluster)

### Environment Details

- **OS**: macOS / Linux local development environment
- **K8s Version**: v1.28
- **Provider**: Local (k3d)
- **Go Version**: 1.22+

### Possible Solution

Capture the return value of `inf.Start` and include it in `errors.Join`:

```go
if startErr := inf.Start(ctx, &corev1.ConfigMap{}); startErr != nil {
    return errors.Join(err, startErr, errors.New("failed to watch RBAC ConfigMap"))
}
```

### Additional Context

N/A

alloutflo:
## Summary

- preserve the error returned by the RBAC ConfigMap informer's `Start` method while adding watch context
- handle informer construction errors before registering the update callback
- keep the error handling inline in `refreshEnforcerInBackground`

## Why

The previous code checked the result of `inf.Start` but discarded that error and joined an older `err` variable instead. As a result, callers only received the generic watch error and lost the underlying informer failure.

Fixes #2763

## Validation

- package: `go test -timeout=20m ./pkg/rbac`
- PR-diff lint: `go tool golangci-lint run --new-from-rev=d0fb75caf9b84d3c7df216a9fd909325d788bee7 --show-stats=false`
- copyright: `make copyright-check FILES='pkg/rbac/rbac.go'`
- formatting: `gofmt -d pkg/rbac/rbac.go`
- whitespace: `git diff --check`

The repository-wide `make check` reaches two existing `misspell` findings in `internal/server/handlers/rbac/kubernetes_test.go`. That file has the identical blob SHA on the base and PR heads; the PR-diff lint is clean.

## Race detector

`go test -race -timeout=20m ./pkg/rbac` could not complete locally because the build ran out of disk space (`no space left on device`) before tests started. This was an infrastructure failure, not a reported test or race failure.

## AI assistance

AI assistance was used substantially to develop this contribution. Every test result and error reported above comes from an actual local run.


spron-in:
Hey @alloutflo  . Thank you for the PR. Can you please check failing tests? Linters are flagging something.

alloutflo:
Thanks for flagging this. The merge run exposed two linter findings: the informer interface parameters needed names, and the error-chain assertion needed `require.ErrorIs`. I fixed both in 5c221d8d.

The focused RBAC package tests, golangci-lint against the PR diff, diff checks, and the copyright check all pass locally. DCO is green as well, so the branch is ready for CI/review again.

pkg/rbac/rbac.go
```diff
@@ -134,10 +149,7 @@ func refreshEnforcerInBackground(
 		}
 		enforcer.EnableEnforce(IsEnabled(cm))
 	})
-	if inf.Start(ctx, &corev1.ConfigMap{}) != nil {
-		return errors.Join(err, errors.New("failed to watch RBAC ConfigMap"))
-	}
-	return nil
+	return startRBACInformer(ctx, inf)
```
recharte: The fix is right, but the seam around it isn't paying for itself. `informerStarter` and `startRBACInformer` exist so the test can inject a failure, and what the test then asserts is that `errors.Join` puts both errors in the chain — which is `errors.Join`'s documented behaviour rather than ours. Meanwhile the function that actually had the bug, `refreshEnforcerInBackground`, stays untested either way, so the extraction buys coverage of the wrap but not of the call site.

Inlining keeps the whole fix in the function that had the bug:

```suggestion
	if err := inf.Start(ctx, &corev1.ConfigMap{}); err != nil {
		return errors.Join(err, errors.New("failed to watch RBAC ConfigMap"))
	}
	return nil
```

`informerStarter`, `startRBACInformer`, `fakeInformerStarter` and `rbac_refresh_test.go` all become dead with that, and the `sigs.k8s.io/controller-runtime/pkg/client` import goes with them.

pkg/rbac/rbac_refresh_test.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func TestStartRBACInformerPreservesStartError(t *testing.T) {
+	t.Parallel()
+
+	startErr := errors.New("informer start failed")
+	err := startRBACInformer(context.Background(), fakeInformerStarter{err: startErr})
+	require.Error(t, err)
+	require.ErrorIs(t, err, startErr)
+	assert.ErrorContains(t, err, "failed to watch RBAC ConfigMap")
```
recharte: `assert.ErrorContains` on the message text pins the test to that exact wording, and the `require.ErrorIs` above it already covers what the issue was about. I'd drop this line.

recharte (commented):
Hey @alloutflo, thanks for the PR! The fix is correct, and checking `informer.New`'s error is a good catch beyond what the issue asked for — the old code would have nil-dereferenced `inf` on that path. Two comments inline, neither blocking.

pkg/rbac/rbac.go
alloutflo: Agreed. I inlined the `inf.Start` error handling in `refreshEnforcerInBackground` and removed the interface, helper, test seam, test file, and now-unused import in 8f65e161. The RBAC package tests and PR-diff lint pass.

pkg/rbac/rbac_refresh_test.go
alloutflo: Removed together with the helper-only regression test in 8f65e161. The implementation now stays directly at the original call site, as suggested.

alloutflo:
@recharte Both inline points are addressed in `8f65e161`, and the two threads are resolved. I also updated the branch to the current `main` in `cda6c274`; it is now 0 commits behind, the PR diff remains limited to `pkg/rbac/rbac.go`, and all 19 checks pass (with the Feature build expectedly skipped). Ready for re-review.

recharte (approved):
Thanks @alloutflo looks good

PR #2836, opened 2026-08-09
fix(accounts): resolve read-modify-write race condition in account management 

### LINKED ISSUE #2833: bug: Read-Modify-Write race condition in accounts.Delete() causes silent data loss under concurrent operations
### What is the issue?

The `Delete()` method in `pkg/kubernetes/accounts.go` performs two separate 
reads to construct a single write, creating a classic Read-Modify-Write race 
condition that completely bypasses Kubernetes' Optimistic Concurrency Control (OCC).

**What happens:**
The method first calls `listAllAccounts()` to get a cached/potentially stale 
snapshot of all user accounts, removes the target user from that snapshot, then 
fetches a *fresh* copy of the secret via `GetSecret()`, and finally overwrites 
the fresh secret's data with the stale snapshot before calling `UpdateSecret()`.

If a concurrent `CreateAccount()` or `SetPassword()` operation writes a new 
account or updated password to the secret between the first read (`listAllAccounts`) 
and the final write (`UpdateSecret`), that change is silently overwritten and 
permanently lost — with no error returned to the caller and no log emitted.

**What was expected:**
The `Delete()` method should fetch the secret exactly once, unmarshal the 
accounts directly from that single fresh copy, remove the target user, and 
write back to the same object. This ensures the `ResourceVersion` field is 
consistent, allowing Kubernetes OCC to detect and reject any conflicting 
concurrent writes instead of silently discarding them.

### Steps to Reproduce

1. Deploy OpenEverest on a Kubernetes cluster with at least 2 user accounts 
   already created (e.g., user-a and user-b).

2. Simultaneously trigger the following two operations in parallel 
   (e.g., via two concurrent API calls or a simple Go test with goroutines):

   - Operation A: DELETE /v1/accounts/user-a  ← triggers Delete("user-a")
   - Operation B: POST /v1/accounts/user-c    ← triggers CreateAccount("user-c")
      OR
   - Operation B: PUT /v1/accounts/user-b/password ← triggers SetPassword("user-b", ...)

3. Wait for both operations to return HTTP 200 (both report success).

4. Immediately call GET /v1/accounts and inspect the returned account list.

**Expected Result:**
All surviving accounts are present — user-b exists, user-c was created 
(or user-b's password was successfully updated), and user-a is deleted.

**Actual Result:**
user-c is missing from the account list entirely (silently dropped), 
OR user-b's password change is silently reverted to the previous value —
with no error, no log, and no indication that data was lost.

**Why it happens (code trace):**
- Delete() calls listAllAccounts()          → stale snapshot captured (T1)
- CreateAccount() writes user-c to secret   → fresh secret updated    (T2)
- Delete() calls GetSecret()                → fetches secret at T2
- Delete() overwrites secret.Data with      → T2 write (user-c) is 
  stale T1 snapshot and calls UpdateSecret()   permanently lost

### Relevant Logs / Output

No error logs or stack traces are emitted by the application during this failure.
This is what makes the bug particularly dangerous — it is a completely silent 
data loss scenario.

**API Response (Operation A — Delete):**
HTTP 200 OK
{}

**API Response (Operation B — CreateAccount / SetPassword):**
HTTP 200 OK
{}

**GET /v1/accounts (after both operations):**
{
  "accounts": [
    { "username": "user-b" }
  ]
}
← user-c is missing. No error. No warning. No trace.

**Everest API Server Logs (during the entire operation):**
INFO  account deleted successfully  {"username": "user-a"}
INFO  account created successfully  {"username": "user-c"}
← Both operations log success. Nothing indicates user-c was silently dropped.

**kubectl get secret everest-accounts -o yaml (after the race):**
apiVersion: v1
kind: Secret
metadata:
  name: everest-accounts
  resourceVersion: "10482"   ← jumped from 10480 to 10482, skipping 10481
data:
  accounts.json: <base64 of stale snapshot — user-c absent>

Note: The skipped resourceVersion (10481 → 10482) is the only indirect 
forensic indicator that a concurrent write was silently overwritten. 
There is no application-level log, metric, or alert for this condition.

### OpenEverest Version

v1.0.0

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

Note: This is a code-level architectural race condition in pkg/kubernetes/accounts.go
and is NOT environment-specific. It will reproduce on any environment where 
concurrent account operations are possible.

However, for completeness, the issue was identified and verified on:

**OpenEverest Version:** latest (main branch, commit verified August 2026)
**Go Version:** 1.22+
**Kubernetes Version:** v1.28+ (any version — not K8s version dependent)
**Cloud Provider:** Any (reproduced on local kind cluster)
**OS:** Linux (Ubuntu 22.04 LTS)
**Deployment Method:** Helm chart (standard OpenEverest install)

**Conditions that increase reproduction likelihood:**
- High-concurrency environments (multiple admin users managing accounts simultaneously)
- Clusters under load (higher informer cache staleness window)
- Automated provisioning pipelines that create/delete accounts in bulk
  (e.g., CI/CD pipelines, onboarding automation scripts)

**Conditions where this is guaranteed to trigger:**
- Any environment where Delete() and CreateAccount() or SetPassword() 
  are called within the same ~millisecond window on the same 
  everest-accounts secret — which is a realistic production scenario 
  in any multi-admin or programmatically managed deployment.

### Possible Solution

The fix is straightforward — eliminate the double-read pattern entirely by 
removing the call to listAllAccounts() and instead unmarshalling the accounts 
directly from the single secret object fetched via GetSecret(). This ensures 
the entire Delete() operation works on one consistent snapshot with a valid 
ResourceVersion, allowing Kubernetes OCC to properly protect against concurrent writes.

**Proposed Fix (pkg/kubernetes/accounts.go):**

func (a *configMapsClient) Delete(ctx context.Context, username string) error {
    // SINGLE read — fetch the secret once and work exclusively from this object
    secret, err := a.k.GetSecret(ctx, common.EverestAccountsSecretName, a.ns)
    if err != nil {
        return err
    }

    // Unmarshal directly from the fresh secret — no stale cache involved
    users, err := unmarshalAccounts(secret.Data[common.EverestAccountsFileName])
    if err != nil {
        return err
    }

    // Check existence
    if _, ok := users[username]; !ok {
        return ErrAccountNotFound
    }

    // Remove the target user
    delete(users, username)

    // Marshal updated accounts back
    data, err := marshalAccounts(users)
    if err != nil {
        return err
    }

    // Write back to the SAME secret object — ResourceVersion is preserved,
    // Kubernetes OCC will reject this write if a concurrent modification
    // occurred, instead of silently overwriting it
    secret.Data[common.EverestAccountsFileName] = data
    _, err = a.k.UpdateSecret(ctx, secret)
    return err
}

**Key changes:**
- listAllAccounts() call removed entirely from Delete()
- Accounts are unmarshalled directly from the secret fetched by GetSecret()
- The same secret object (with its original ResourceVersion) is used for 
  the final UpdateSecret() call
- If a concurrent write occurs, Kubernetes will return a 409 Conflict error
  instead of silently overwriting the change — callers can safely retry

**Optional enhancement (retry logic):**
For production resilience, wrap the entire operation in a retry loop that 
re-fetches the secret on 409 Conflict, similar to the controller-runtime 
RetryOnConflict pattern:

  err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
      // full Delete logic here
  })

This is the same pattern used by kubectl and controller-runtime internally 
and is the idiomatic Kubernetes way to handle OCC conflicts safely.

### Additional Context

**Root Cause Classification:**
This is a Time-of-Check to Time-of-Use (TOCTOU) vulnerability — a well-documented 
class of concurrency bug where the state of a resource changes between the point 
it is read (checked) and the point it is written (used). In this case, the 
"check" is listAllAccounts() and the "use" is UpdateSecret() — with a second, 
inconsistent read (GetSecret()) in between that creates the silent overwrite window.

**Why This Escaped Detection:**
- Unit tests for Delete() likely mock GetSecret() and listAllAccounts() in 
  isolation — the race window only exists in a live concurrent environment
- Static analysis tools (golangci-lint, go vet, staticcheck) cannot detect 
  this class of distributed race condition — it requires semantic understanding 
  of Kubernetes ResourceVersion semantics
- Go's built-in race detector (go test -race) will NOT catch this — it is not 
  a shared-memory race, it is a distributed state race over a Kubernetes API object
- The bug is completely silent — no panic, no error, no log — making it 
  invisible in production monitoring and alerting

**Impact Assessment:**
| Scenario                        | Impact                                      |
|---------------------------------|---------------------------------------------|
| Concurrent Create + Delete      | Newly created account silently vanishes     |
| Concurrent SetPassword + Delete | Password change silently reverted           |
| Bulk provisioning pipelines     | Arbitrary accounts lost with no audit trail |
| Multi-admin environments        | Unpredictable account state corruption      |

**Related Code Pattern:**
The same double-read anti-pattern should be audited in:
- CreateAccount() in the same file — may have a symmetric vulnerability
- SetPassword() in the same file — same risk surface
- Any other method in pkg/kubernetes/ that calls listAllAccounts() 
  before fetching and writing a secret

**References:**
- Kubernetes OCC / ResourceVersion semantics:
  https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions
- TOCTOU vulnerability class:
  https://cwe.mitre.org/data/definitions/367.html
- controller-runtime RetryOnConflict pattern:
  https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil

Harkirat1309:
## Description
This PR addresses a critical Time-of-Check to Time-of-Use (TOCTOU) race condition in the Everest account management logic (`pkg/kubernetes/accounts.go`), which previously caused silent data loss during concurrent account operations (e.g., simultaneous `Create`, `Delete`, or `SetPassword` calls).

Fixes #2833

## Root Cause
Previously, methods like `Delete()` and `insertOrUpdateAccount()` executed a double-read anti-pattern. They would read a stale snapshot of the accounts, fetch a fresh copy of the secret, and overwrite the secret's data with the stale snapshot without any concurrency safeguards. Because it bypassed Kubernetes Optimistic Concurrency Control (OCC), concurrent modifications to the Accounts Secret were silently overwritten rather than rejected.

## Solution Implemented
1. **Removed Stale Caching:** Eliminated the separate `listAllAccounts()` read from the mutation paths.
2. **Proper OCC Handling:** Wrapped the entire read-modify-write block in `retry.RetryOnConflict()`. 
3. **Single Source of Truth:** Accounts are now unmarshalled directly from the fresh `GetSecret()` object. If a concurrent write occurs, Kubernetes returns a `409 Conflict`, which triggers our `RetryOnConflict` block to seamlessly re-fetch the new state and re-apply the mutation.

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

### DIFF
--- a/pkg/kubernetes/accounts.go
+++ b/pkg/kubernetes/accounts.go
@@ -27,7 +27,9 @@ import (
 
 	"golang.org/x/crypto/pbkdf2"
 	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/percona/everest/pkg/accounts"
 	"github.com/percona/everest/pkg/common"
@@ -72,14 +74,6 @@ func (a *configMapsClient) List(ctx context.Context) (map[string]*accounts.Accou
 
 // Create a new user account.
 func (a *configMapsClient) Create(ctx context.Context, username, password string) error {
-	// Ensure that the user does not already exist.
-	_, err := a.Get(ctx, username)
-	if err != nil && !errors.Is(err, accounts.ErrAccountNotFound) {
-		return errors.Join(err, errors.New("failed to check if account already exists"))
-	} else if err == nil {
-		return accounts.ErrUserAlreadyExists
-	}
-
 	if password == "" {
 		return errors.New("password cannot be empty")
 	}
@@ -90,58 +84,48 @@ func (a *configMapsClient) Create(ctx context.Context, username, password string
 		return errors.Join(err, errors.New("failed to compute hash"))
 	}
 
-	account := &accounts.Account{
-		Enabled:       true,
-		Capabilities:  []accounts.AccountCapability{accounts.AccountCapabilityLogin},
-		PasswordMtime: time.Now().Format(time.RFC3339),
-		PasswordHash:  hash,
-	}
-	return a.insertOrUpdateAccount(ctx, username, account, true)
+	return a.insertOrUpdateAccount(ctx, username, true, func(existing *accounts.Account) (*accounts.Account, error) {
+		if existing != nil {
+			return nil, accounts.ErrUserAlreadyExists
+		}
+		return &accounts.Account{
+			Enabled:       true,
+			Capabilities:  []accounts.AccountCapability{accounts.AccountCapabilityLogin},
+			PasswordMtime: time.Now().Format(time.RFC3339),
+			PasswordHash:  hash,
+		}, nil
+	})
 }
 
 // SetPassword sets a new password for an existing user account.
 func (a *configMapsClient) SetPassword(ctx context.Context, username, newPassword string, secure bool) error {
-	user, err := a.Get(ctx, username)
-	if err != nil {
-		return err
-	}
-	user.PasswordHash = newPassword
-	if secure {
-		pwHash, err := a.computePasswordHash(ctx, newPassword)
-		if err != nil {
-			return err
+	return a.insertOrUpdateAccount(ctx, username, secure, func(existing *accounts.Account) (*accounts.Account, error) {
+		if existing == nil {
+			return nil, accounts.ErrAccountNotFound
 		}
-		user.PasswordHash = pwHash
-	}
-	user.PasswordMtime = time.Now().Format(time.RFC3339)
-	return a.insertOrUpdateAccount(ctx, username, user, secure)
+		existing.PasswordHash = newPassword
+		if secure {
+			pwHash, err := a.computePasswordHash(ctx, newPassword)
+			if err != nil {
+				return nil, err
+			}
+			existing.PasswordHash = pwHash
+		}
+		existing.PasswordMtime = time.Now().Format(time.RFC3339)
+		return existing, nil
+	})
 }
 
 // Delete an existing user account specified by username.
 func (a *configMapsClient) Delete(ctx context.Context, username string) error {
-	users, err := a.listAllAccounts(ctx)
-	if err != nil {
-		return err
-	}
+	return a.updateAccounts(ctx, func(_ *corev1.Secret, users map[string]*accounts.Account) error {
+		if _, found := users[username]; !found {
+			return accounts.ErrAccountNotFound
+		}
 
-	if _, found := users[username]; !found {
-		return accounts.ErrAccountNotFound
-	}
-
-	delete(users, username)
-	secret, err := a.k.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestAccountsSecretName})
-	if err != nil {
-		return err
-	}
-	data, err := yaml.Marshal(users)
-	if err != nil {
-		return err
-	}
-	secret.Data[common.EverestAccountsFileName] = data
-	if _, err := a.k.UpdateSecret(ctx, secret); err != nil {
-		return err
-	}
-	return nil
+		delete(users, username)
+		return nil
+	})
 }
 
 func (a *configMapsClient) Verify(ctx context.Context, username, password string) error {
@@ -215,47 +199,67 @@ func (a *configMapsClient) listAllAccounts(ctx context.Context) (map[string]*acc
 	return result, nil
 }
 
+func (a *configMapsClient) updateAccounts(
+	ctx context.Context,
+	mutateFn func(secret *corev1.Secret, accountsMap map[string]*accounts.Account) error,
+) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		secret, err := a.k.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestAccountsSecretName})
+		if err != nil {
+			return err
+		}
+
+		accountsMap := make(map[string]*accounts.Account)
+		if err := yaml.Unmarshal(secret.Data[common.EverestAccountsFileName], &accountsMap); err != nil {
+			return err
+		}
+
+		if err := mutateFn(secret, accountsMap); err != nil {
+			return err
+		}
+
+		data, err := yaml.Marshal(accountsMap)
+		if err != nil {
+			return err
+		}
+
+		if secret.Data == nil {
+			secret.Data = make(map[string][]byte)
+		}
+		secret.Data[common.EverestAccountsFileName] = data
+
+		if _, err := a.k.UpdateSecret(ctx, secret); err != nil {
+			return err
+		}
+		return nil
+	})
+}
+
 func (a *configMapsClient) insertOrUpdateAccount(
 	ctx context.Context,
 	username string,
-	account *accounts.Account,
 	secure bool,
+	mutateFn func(existing *accounts.Account) (*accounts.Account, error),
 ) error {
-	secret, err := a.k.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestAccountsSecretName})
-	if err != nil {
-		return err
-	}
+	return a.updateAccounts(ctx, func(secret *corev1.Secret, accountsMap map[string]*accounts.Account) error {
+		updated, err := mutateFn(accountsMap[username])
+		if err != nil {
+			return err
+		}
 
-	accounts := make(map[string]*accounts.Account)
-	if err := yaml.Unmarshal(secret.Data[common.EverestAccountsFileName], &accounts); err != nil {
-		return err
-	}
+		accountsMap[username] = updated
 
-	accounts[username] = account
-	data, err := yaml.Marshal(accounts)
-	if err != nil {
-		return err
-	}
-
-	if secret.Data == nil {
-		secret.Data = make(map[string][]byte)
-	}
-	secret.Data[common.EverestAccountsFileName] = data
-
-	annotations := secret.GetAnnotations()
-	if annotations == nil {
-		annotations = make(map[string]string)
-	}
-	delete(annotations, fmt.Sprintf(insecurePasswordAnnotation, username))
-	if !secure {
-		annotations[fmt.Sprintf(insecurePasswordAnnotation, username)] = insecurePasswordValueTrue
-	}
-	secret.SetAnnotations(annotations)
-
-	if _, err := a.k.UpdateSecret(ctx, secret); err != nil {
-		return err
-	}
-	return nil
+		annotations := secret.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		delete(annotations, fmt.Sprintf(insecurePasswordAnnotation, username))
+		if !secure {
+			annotations[fmt.Sprintf(insecurePasswordAnnotation, username)] = insecurePasswordValueTrue
+		}
+		secret.SetAnnotations(annotations)
+		return nil
+	})
 }
 
 func (a *configMapsClient) salt(ctx context.Context) ([]byte, error) {
--- a/pkg/kubernetes/accounts_test.go
+++ b/pkg/kubernetes/accounts_test.go
@@ -1,5 +1,6 @@
 // everest
 // Copyright (C) 2025 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,11 +17,17 @@
 package kubernetes
 
 import (
+	"context"
+	"sync/atomic"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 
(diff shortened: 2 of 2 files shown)


pkg/kubernetes/accounts.go
```diff
@@ -29,6 +29,8 @@ import (
 	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/types"
 
+	"k8s.io/client-go/util/retry"
```
recharte, line 32: The `Check` job fails on this file: `gci` wants `k8s.io/client-go/util/retry` grouped with the other third-party imports rather than in a group of its own, and there is trailing whitespace on the new blank lines at 140, 145 and 150. `make format` fixes both.

```suggestion
	"k8s.io/apimachinery/pkg/types"
	"k8s.io/client-go/util/retry"
```

pkg/kubernetes/accounts.go
```diff
@@ -221,41 +232,43 @@ func (a *configMapsClient) insertOrUpdateAccount(
 	account *accounts.Account,
 	secure bool,
 ) error {
-	secret, err := a.k.GetSecret(ctx, types.NamespacedName{Namespace: common.SystemNamespace, Name: common.EverestAccountsSecretName})
-	if err != nil {
-		return err
-	}
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
recharte, line 206: The pattern this PR set out to remove survives one level up, in both callers of this function:

- `Create` checks existence with `Get` at line 78 and then calls this. Two concurrent creates of the same username both pass the check, and now that a conflict is retried rather than surfaced, the loser silently overwrites the winner instead of the caller getting `ErrUserAlreadyExists`.
- `SetPassword` reads the account at line 106 and hands that struct in. On a conflict retry the stale struct is re-applied to the freshly-read map, so a concurrent change to that user's `Enabled` or `Capabilities` is lost — the same lost update, just narrowed to one user.

Both close if the read and the decision move inside the closure, e.g. by having `insertOrUpdateAccount` take a `func(existing *accounts.Account) (*accounts.Account, error)` that is evaluated on every attempt.

pkg/kubernetes/accounts.go
```diff
@@ -119,29 +121,38 @@ func (a *configMapsClient) SetPassword(ctx context.Context, username, newPasswor
 
 // Delete an existing user account specified by username.
 func (a *configMapsClient) Delete(ctx context.Context, username string) error {
-	users, err := a.listAllAccounts(ctx)
-	if err != nil {
-		return err
-	}
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
recharte: Nothing here is covered by a test. `pkg/kubernetes/accounts_test.go` already wires a fake client into the shared `accounts.Tests` suite, and controller-runtime's fake client enforces `resourceVersion`, so pinning this is cheap: inject a conflict on the first `Update` via `fakeclient.NewClientBuilder().WithInterceptorFuncs(...)` and assert that `Delete` still succeeds and an account added by the concurrent writer survives. As it stands the retry loop can be deleted again without anything going red.

pkg/kubernetes/accounts.go
```diff
@@ -119,29 +121,38 @@ func (a *configMapsClient) SetPassword(ctx context.Context, username, newPasswor
...
+		}
+
+		delete(users, username)
+		
+		data, err := yaml.Marshal(users)
+		if err != nil {
+			return err
+		}
+		
+		if secret.Data == nil {
```
recharte: Unreachable: if `secret.Data` were nil, `users` would be empty and the not-found check at line 135 would have returned already.

recharte (changes requested):
This branch is cut from `fix/rbac-configmap-adapter-context-timeout`, so it carries all of #2822 — `pkg/rbac/configmap-adapter/adapter.go` and its new test file are byte-identical to that PR's head, and they account for 201 of the 265 added lines here. That makes this PR unmergeable until #2822 lands, and it hides the actual size of the accounts fix. Please retarget the branch at `main` and drop those commits, or rebase once #2822 is merged.

Two smaller things: the title has run together (`...in account managementFix/accounts race condition`), and "Type of change" has *New feature* and *Breaking change* ticked while *Bug fix* is not — this is a bug fix and nothing about it is breaking.

pkg/kubernetes/accounts.go
Harkirat1309: @recharte Fixed — retry is now correctly grouped with other third-party imports, trailing whitespace removed.

pkg/kubernetes/accounts.go
Harkirat1309: @recharte Addressed — insertOrUpdateAccount now accepts a mutateFn closure. Both Create and SetPassword pass their read/decision logic inside it

pkg/kubernetes/accounts.go
Harkirat1309: @recharte TestAccounts_DeleteConflict uses a conflictClient wrapper that injects a Conflict error on the first Update and verifies retry succeeds

pkg/kubernetes/accounts_test.go
```diff
@@ -48,3 +55,74 @@ func TestAccounts(t *testing.T) {
...
+
+	k := NewEmpty(zap.NewNop().Sugar()).WithKubernetesClient(&conflictClient{Client: mockClient})
+	err := k.Accounts().Delete(context.Background(), "admin")
+	require.NoError(t, err)
+
+	accs, err := k.Accounts().List(context.Background())
+	require.NoError(t, err)
+	require.NotContains(t, accs, "admin")
+	require.Contains(t, accs, "concurrent-user")
+}
```
recharte, line 135: The conflict test covers `Delete`, but the larger half of this change — `insertOrUpdateAccount` taking a `mutateFn` — has nothing pinning it:

- No conflict test for the `insertOrUpdateAccount` retry, even though it is the path `Create` and `SetPassword` both go through. `conflictClient` already generalises: pointing it at a `Create` or `SetPassword` call would cover it.
- `ErrUserAlreadyExists` and `SetPassword`-on-a-missing-user are the two behaviours that just moved into the closures, and neither is asserted anywhere — `pkg/accounts/tests.go` only walks the happy path, so a closure that dropped its `existing == nil` check would stay green.

pkg/kubernetes/accounts.go
```diff
@@ -90,58 +83,74 @@ func (a *configMapsClient) Create(ctx context.Context, username, password string
...
+		pwHash, err = a.computePasswordHash(ctx, newPassword)
 		if err != nil {
 			return err
 		}
-		user.PasswordHash = pwHash
 	}
-	user.PasswordMtime = time.Now().Format(time.RFC3339)
-	return a.insertOrUpdateAccount(ctx, username, user, secure)
+
+	return a.insertOrUpdateAccount(ctx, username, secure, func(existing *accounts.Account) (*accounts.Account, error) {
```
recharte, line 102: `computePasswordHash` runs 4096 PBKDF2 iterations and fetches the system namespace for the salt, all before the closure discovers there is no such user. Cheap to avoid by hashing lazily inside the closure, on the far side of the `existing == nil` check — the closure already recomputes `PasswordMtime` per attempt, so it is the natural home for it.

pkg/kubernetes/accounts.go
```diff
@@ -90,58 +83,74 @@ func (a *configMapsClient) Create(ctx context.Context, username, password string
...
+	})
 }
 
 // Delete an existing user account specified by username.
 func (a *configMapsClient) Delete(ctx context.Context, username string) error {
-	users, err := a.listAllAccounts(ctx)
-	if err != nil {
-		return err
-	}
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
recharte: Now that the mutate-inside-the-retry shape exists, `Delete` is the odd one out: ~25 lines re-implementing the same read/unmarshal/marshal/write, with the two copies free to drift (the nil-`Data` guard is in one and not the other). A shared `updateAccounts(ctx, func(map[string]*accounts.Account) error)` with `insertOrUpdateAccount` layered on top would leave one copy. Fine to leave for a follow-up if you'd rather keep this PR tight.

recharte (commented):
The scope problem from the last round is gone and the mutate-closure refactor is the right shape — thanks for reworking it rather than patching around it.

One thing left over from that review: the title still reads `fix(accounts): resolve read-modify-write race condition in account managementFix/accounts race condition`, and "Type of change" still has *New feature* and *Breaking change* ticked while *Bug fix* is not.

pkg/kubernetes/accounts_test.go
```diff
@@ -48,3 +55,140 @@ func TestAccounts(t *testing.T) {
...
+		Build()
+
+	k := NewEmpty(zap.NewNop().Sugar()).WithKubernetesClient(&conflictClient{Client: mockClient})
+	err := k.Accounts().Create(context.Background(), "newuser", "password")
+	require.NoError(t, err)
+
+	accs, err := k.Accounts().List(context.Background())
+	require.NoError(t, err)
+	require.Contains(t, accs, "newuser")
+	require.Contains(t, accs, "concurrent-user")
```
recharte: The concurrent writer here adds an unrelated user, so the test passes whether or not `mutateFn` is re-evaluated on the retry — a closure invoked once against the first read would still produce this result. Having the writer create `newuser` instead, and asserting `Create` comes back `ErrUserAlreadyExists`, is what pins the actual TOCTOU property: the existence check has to be answered from the data the retry re-read.

recharte (commented):
Still outstanding from the first review: "Type of change" has *New feature* and *Breaking change* ticked while *Bug fix* is not.

pkg/kubernetes/accounts_test.go
```diff
@@ -48,3 +55,158 @@ func TestAccounts(t *testing.T) {
...
+
+	k := NewEmpty(zap.NewNop().Sugar()).WithKubernetesClient(&conflictClient{
+		Client: mockClient,
+		mutate: func(sec *corev1.Secret) {
+			// The concurrent writer adds "newuser" before the Create call finishes.
+			// This tests the TOCTOU property by ensuring the retry re-evaluates the existence check.
+			sec.Data[common.EverestAccountsFileName] = []byte("admin:\n  enabled: true\nnewuser:\n  enabled: true\n")
+		},
+	})
+	err := k.Accounts().Create(context.Background(), "newuser", "password")
```
recharte: The `mutate` hook is the right move, but this test swapped one property for the other: it used to prove that `Create` completes after a conflict and keeps the concurrent writer's data, and now it proves only that the existence check is re-evaluated. Nothing covers a successful `insertOrUpdateAccount` retry any more, which is the path `Create` and `SetPassword` take on every ordinary conflict. Now that the mutation is injectable, keeping both is two table entries — one writer adding an unrelated user (expect success, both accounts present), one adding `newuser` (expect `ErrUserAlreadyExists`).

pkg/kubernetes/accounts_test.go
```diff
@@ -48,3 +55,158 @@ func TestAccounts(t *testing.T) {
...
+			Namespace: common.SystemNamespace,
+			Name:      common.EverestAccountsSecretName,
+		}, sec); err != nil {
+			return err
+		}
+
+		if c.mutate != nil {
+			c.mutate(sec)
+		} else {
+			// Default mutation if not provided (legacy test support)
```
recharte: Both call sites set `mutate`, so this branch is unreachable, and "legacy test support" is a strange thing for a helper added yesterday to carry. Dropping the `else` and letting a nil `mutate` be a programming error keeps it honest. The doc comment above still describes the old behaviour too — it says the helper adds a `concurrent-user` entry, which is now the caller's choice and isn't what `TestAccounts_CreateConflict` does.

Harkirat1309:
@recharte please review!

Harkirat1309:
@recharte  please review this.

recharte:
@Harkirat1309 please stop spamming me every few hours, I'll get to your issues/PRs when I have a chance...

Harkirat1309:
> @Harkirat1309 please stop spamming me every few hours, I'll get to your issues/PRs when I have a chance...

@recharte really sorry about that! I'll be patient and wait for your review.

recharte (approved):
Thanks @Harkirat1309, changes look good 🚀

Issue #2827, opened 2026-08-08
Add Unit Tests for pkg/logger


RishiSingh13:
### What is the issue?

The pkg/logger package currently has no dedicated unit tests covering its logger initialization and retrieval functions.

There is also an explicit // FIXME Test this. comment in logger.go for MustInitLogger, indicating that this behavior is currently missing test coverage.

Since the logger configuration is used across the project, adding focused unit tests would help document the expected behavior and prevent regressions when the logging implementation is changed.

Proposed Solution

Create a new test file:

pkg/logger/logger_test.go

Add unit tests covering the following functions and configurations:

MustInitLogger
JSON format
Console format
MustInitVerboseLogger
JSON format
Console format
InitLoggerInRootCmd
Verbose mode
Non-verbose mode
GetLogger

The tests should verify the resulting logger behavior/configuration rather than only checking that the functions execute without panicking.

Because the package interacts with zap's global logger, tests that modify the global logger should preserve and restore the original logger to prevent side effects on other tests:

origLogger := zap.L()
defer func() {
    zap.ReplaceGlobals(origLogger)
}()

The same restoration approach should be applied wherever necessary so that the tests remain isolated.

Expected Outcome

The pkg/logger package will have dedicated unit-test coverage for its logger initialization and retrieval behavior.

This will:

Address the existing FIXME for MustInitLogger.
Provide regression coverage for JSON and Console logging configurations.
Verify verbose and non-verbose logger initialization.
Verify GetLogger behavior.
Reduce the risk of future logger changes introducing unnoticed regressions.
Prevent logger-related tests from affecting other tests through the global zap logger.

### Steps to Reproduce

Navigate to the pkg/logger package.
Observe that there is currently no dedicated logger_test.go file.
Inspect logger.go and note the existing // FIXME Test this. comment for MustInitLogger.
Run the relevant package tests and observe that the logger initialization and retrieval behavior described in this issue does not have dedicated unit-test coverage.

### Possible Solution

Add a dedicated `pkg/logger/logger_test.go` file covering the logger initialization and retrieval functions in both JSON and Console configurations.

Tests should also cover verbose and non-verbose modes where applicable. Since the logger uses zap's global logger, tests should save the original global logger and restore it with `defer` after each test to prevent side effects on other tests.

### Additional Context

This is not a runtime or Kubernetes-environment-specific issue. The proposed changes are limited to adding unit tests for the existing pkg/logger implementation.

The MustInitLogger function already contains the following developer note:

[...]

PR #2837, opened 2026-08-09
chore(logger): remove obsolete FIXME

RishiSingh13:
## What does this PR do?

Removes the obsolete `// FIXME Test this.` comment from `MustInitLogger` in `pkg/logger`.

## Changes

* Remove the obsolete `// FIXME Test this.` comment from `MustInitLogger`.

## Testing

* `go test ./pkg/logger/...`

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -1,5 +1,6 @@
 // everest
 // Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -26,7 +27,7 @@ import (
 	"github.com/percona/everest/cmd/config"
 )
 
-// MustInitLogger initializes logger and panics in case of an error. // FIXME Test this.
+// MustInitLogger initializes logger and panics in case of an error.
 func MustInitLogger(json bool, logName string) *zap.Logger {
 	var loggerCfg zap.Config
 	if config.Debug {


recharte:
Hi @RishiSingh13 — thanks for this, and sorry for the round trip. I've closed #2827 with the reasoning; it's our mistake for accepting it before thinking it through rather than after you'd written 240 lines.

One thing I would still like, if you're willing. The `// FIXME Test this.` on `MustInitLogger` should go, and deleting it is the honest resolution of what you filed — the TODO was never real. Would you repurpose this PR into that one-liner? Drop the `Fixes #2827` line from the description and it stands on its own.

If you'd rather move on to something with more in it, filter the issues by `help wanted` or `good first issue` and take anything that isn't already assigned — comment `/assign` on it and it's yours.

RishiSingh13:
Hi @recharte, I’ve repurposed the PR as requested and removed the obsolete `// FIXME Test this.` comment. I’ve also updated the description and ensured the latest commit is DCO-signed. Thanks for the guidance!

pkg/logger/logger_test.go
recharte, line 1: @RishiSingh13 the FIXME removal in `87bb3ae` is exactly right, but the original test commit is still on the branch, so this file is still being added — the PR is +241/−1 across two files while the description says it only drops the comment. The title is still `test(logger): add logger unit tests` too, and since the repo squash-merges and takes the PR title when there's more than one commit, merging as it stands would land the 240 lines we agreed not to take.

Dropping `2a55ab1` fixes both — with one commit left, the squash title comes from `chore(logger): remove obsolete FIXME` on its own. An interactive rebase works, or if that's fiddly, rebuild the branch from main and redo the one-line edit:

```bash
git fetch <upstream-remote> main
git reset --hard <upstream-remote>/main
# delete "// FIXME Test this." from the MustInitLogger doc comment
git commit -sam "chore(logger): remove obsolete FIXME"
git push --force-with-lease
```

Worth updating the PR title to match while you're there.

pkg/logger/logger.go
recharte, line 1: `Check Copyright Headers` is red: touching `logger.go` brings it into the check, and it still carries only the Percona line. Modified files need the OpenEverest line after it, the way `cmd/config/config.go` has it:

```
// everest
// Copyright (C) 2023 Percona LLC
// Copyright (C) 2026 The OpenEverest Contributors
```

`make copyright-headers` does it in place. The job also flags `logger_test.go`, but that one resolves itself once the test commit comes off.

recharte (changes requested):
Nearly there — the FIXME removal itself is right, it's just the old test commit still riding along. Notes inline.

The red `API Integration Tests` isn't yours: it fails at the workflow level about a minute in, and a comment-only change can't reach it.

RishiSingh13:
Hi @recharte, I’ve addressed the requested changes: removed the old test commit, rebased the PR onto the latest `main`, updated the copyright header, and renamed the PR to `chore(logger): remove obsolete FIXME`. The PR is now limited to the intended change. Thanks!

recharte (approved):
Thanks @RishiSingh13

PR #2838, opened 2026-08-09
docs: document why endpoint URL stays editable on in-use backup storage

### LINKED ISSUE #2665: Backup storage endpoint URL can be changed while the storage is in use
### What is the issue?

If a backup storage is in use (a cluster is backing up into it, so it carries the `everest.percona.com/in-use-protection` finalizer), Everest correctly refuses to let me change its bucket or region:

    backup storage='my-storage' in namespace='everest' is used and cannot be updated

But if I change only the **endpoint URL**, the update goes straight through.

I expected the URL to be guarded the same way as bucket and region, because those three fields together are what identify where the backups physically live. Everest already treats them as one unit everywhere else - `validateDuplicateStorageByUpdate` compares `region + bucket + url` when deciding whether two storages are the same, and the duplicate error message spells it out: *"backup storage with the same url, bucket and region already exists"*.

The practical effect is that a storage can be silently repointed at a different S3 endpoint while existing `DatabaseClusterBackup` objects still reference it by name. New backups land at the new endpoint, the old ones are still recorded against this storage but are no longer reachable through it, and restores from them fail.

Arguably this is worse than the bucket/region cases that *are* blocked, since a URL change can move you to an entirely different provider.

### Steps to Reproduce

The check is a pure function, so the quickest way to see it is at the unit level:

1. Drop this into `internal/server/handlers/validation/backup_storage_test.go`:

```go
func TestURLChangeOnInUseStorage(t *testing.T) {
	bs := &everestv1alpha1.BackupStorage{
		Spec: everestv1alpha1.BackupStorageSpec{
			Bucket:      "bucket1",
			Region:      "region1",
			EndpointURL: "https://s3.example.com",
		},
	}
	params := everestapi.UpdateBackupStorageParams{Url: new("https://s3.other.example.com")}
	assert.True(t, basicStorageParamsAreChanged(bs, &params))
}
```

2. `go test ./internal/server/handlers/validation/ -run TestURLChangeOnInUseStorage`
3. It fails `basicStorageParamsAreChanged` returns `false` for a URL change, so the guard in `validateUpdateBackupStorageRequest` never fires.

Through the API:

1. Create a backup storage, then create a DB cluster that backs up into it, so the storage picks up the `everest.percona.com/in-use-protection` finalizer.
2. `PATCH /v1/namespaces/{ns}/backup-storages/{name}` with `{"region": "eu-central-1"}` → correctly rejected with "is used and cannot be updated".
3. `PATCH` the same storage with `{"url": "https://some-other-endpoint.example.com"}` → accepted.

### Relevant Logs / Output

Unit-level failure:

```bash
--- FAIL: TestURLChangeOnInUseStorage (0.00s)
    backup_storage_test.go:XX:
        	Error:      	Should be true
        	Test:       	TestURLChangeOnInUseStorage
FAIL
```

Same storage, two requests - the second one should have been rejected too:

```bash
$ curl -X PATCH .../backup-storages/my-storage -d '{"region":"eu-central-1"}'
{"message":"backup storage='my-storage' in namespace='everest' is used and cannot be updated"}

$ curl -X PATCH .../backup-storages/my-storage -d '{"url":"https://other.example.com"}'
200 OK
```

### OpenEverest Version

v1.16.1

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

Found by reading the validation code rather than hitting it in a live cluster, so I've confirmed it at the unit level rather than end-to-end. The affected function is pure and has no test coverage, so it's easy to verify either way.

### Possible Solution

`basicStorageParamsAreChanged` (`internal/server/handlers/validation/backup_storage.go:337`) checks bucket and region but not `EndpointURL`:

```go
func basicStorageParamsAreChanged(bs *everestv1alpha1.BackupStorage, params *api.UpdateBackupStorageParams) bool {
	if params.BucketName != nil && bs.Spec.Bucket != pointer.GetString(params.BucketName) {
		return true
	}
	if params.Region != nil && bs.Spec.Region != pointer.GetString(params.Region) {
		return true
	}
	return false
}
```

The three `*OrDefault` helpers used by `validateDuplicateStorageByUpdate` already live in the same file, so routing this through them fixes the gap and keeps the two functions from drifting apart again:

```go
func basicStorageParamsAreChanged(bs *everestv1alpha1.BackupStorage, params *api.UpdateBackupStorageParams) bool {
	return bucketNameOrDefault(params, bs.Spec.Bucket) != bs.Spec.Bucket ||
		regionOrDefault(params, bs.Spec.Region) != bs.Spec.Region ||
		urlOrDefault(params, bs.Spec.EndpointURL) != bs.Spec.EndpointURL
}
```

The `*OrDefault` helpers return the current value when the param is nil, so the nil checks become unnecessary rather than being dropped.

I'd deliberately leave `ForcePathStyle` and `VerifyTLS` out - they're connection options rather than location, and blocking those would stop people rotating TLS settings or credentials on a storage that's in use.

### Additional Context

`basicStorageParamsAreChanged` has no test references anywhere in the repo, which is probably why the omission went unnoticed — its sibling `validateDuplicateStorageByUpdate` does have a table-driven test, and that one got the URL right.

Happy to open a PR with the fix plus a table-driven test covering all three fields (including a case asserting that non-identity params like access key and `ForcePathStyle` still *don't* trip the guard).

Harsh63870:
Fixes #2665

## Why

Per maintainer feedback on #2665: excluding `EndpointURL` from `basicStorageParamsAreChanged` is intentional. Bucket/region changes point at different objects and stay blocked when in use; an endpoint change is usually an infrastructure hostname/DNS move, and freezing it would strand existing backups that reference the storage by name. Reachability is enforced separately via the update access check.

## What changed

- Documented why URL is excluded from the in-use guard (and how that differs from `validateDuplicateStorageByUpdate`).
- Kept behaviour: only bucket/region trip the guard.
- Added/updated `TestBasicStorageParamsAreChanged` to lock URL-allowed, same-value, and non-identity (credentials/TLS) cases.

## Test plan

- [x] `go test ./internal/server/handlers/validation/ -run TestBasicStorageParamsAreChanged -count=1`


Harsh63870:
@recharte @spron-in, CLI test-pg-operator was cancelled after 60m while stuck on “Run K3D test cluster” (tests never started). Unit/API/other CLI jobs are green; this PR only touches backup-storage validation + unit tests, so this looks infra/flake. Re-running the failed jobs.

recharte:
@Harsh63870 the direction on this changed — see https://github.com/openeverest/openeverest/issues/2665#issuecomment-5233728252 for the reasoning. Short version: excluding the endpoint URL from the in-use guard is deliberate, so the guard shouldn't be extended to cover it. That's a call about the behaviour, not about your patch, and the labelling that pointed you at it was ours.

If you're willing, the useful version of this PR is the inverse: keep the doc comment on `basicStorageParamsAreChanged` but have it record why the URL is excluded, drop the change to the return expression, and keep the table cases that assert the current semantics. Happy to review that. If you'd rather not, no problem at all — say the word and I'll close this out.

Harsh63870:
thanks @recharte, that context makes sense. I’ll rework #2838 into a documentation + regression test PR: keep URL editable when in use, document why it’s excluded from basicStorageParamsAreChanged, and lock the current behaviour with the same-value/non-identity test cases...

Harsh63870:
thanks @recharte, all requested changes have been applied. CI is running now.

internal/server/handlers/validation/backup_storage.go
```diff
@@ -337,14 +337,28 @@ func azureAccess(ctx context.Context, l *zap.SugaredLogger, accountName, account
...
+// recoverable by repointing back; an immutable field after an infrastructure move is not.
+//
+// The safety check for a URL change is reachability, not immutability:
+// validateUpdateBackupStorageRequest runs the full access check against the new endpoint
+// with bucket and region unchanged, so the update only succeeds if that bucket is readable
+// and writable there.
+//
+// This is a different question from validateDuplicateStorageByUpdate, which compares
+// region + bucket + url to decide whether two storages address the same place
+// (see errDuplicatedBackupStorage). The two functions are not meant to stay in sync.
```
recharte, line 357: This comment exists because the last time the decision was made, the only surviving record was a bullet in a squashed commit message. A reader who wants the argument rather than the summary still has nowhere to go from here. There's no real precedent for issue links in Go comments in this repo, so treat it as a preference, but IMO it's the part that makes the record durable:

```suggestion
// (see errDuplicatedBackupStorage). The two functions are not meant to stay in sync.
// See https://github.com/openeverest/openeverest/issues/2665 for the full reasoning.
```

internal/server/handlers/validation/backup_storage.go
```diff
@@ -337,14 +337,28 @@ func azureAccess(ctx context.Context, l *zap.SugaredLogger, accountName, account
...
+// basicStorageParamsAreChanged reports whether an in-use backup storage update changes
+// bucket or region — parameters that definitively point at a different set of objects.
+//
+// EndpointURL is intentionally excluded. An endpoint change is usually not a change of
+// location: object storage may move behind a new ingress hostname, a domain may be
+// retired, in-cluster DNS may change after a namespace move, or a bucket may start being
+// reached through a private endpoint while the objects stay put. Freezing the URL would
+// leave existing DatabaseClusterBackup objects (which reference storage by name) with no
+// supported way back to that history once the old hostname is gone. A careless repoint is
+// recoverable by repointing back; an immutable field after an infrastructure move is not.
```
recharte: This last sentence is aimed at someone challenging the decision rather than someone reading the function; the paragraph already lands without it.

```suggestion
// reached through a private endpoint while the objects stay put. Freezing the URL would
// leave existing DatabaseClusterBackup objects (which reference storage by name) with no
// supported way back to that history once the old hostname is gone.
```

recharte (commented):
Thanks for turning this around, and sorry again for sending you down the first path. Two suggestions on the comment wording, neither of them blocking — take them or leave them and this is good to go.

Harsh63870:
thanks @recharte, applied both suggestions...

recharte (approved):
Thanks @Harsh63870

PR #2841, opened 2026-08-09
fix: clarify delegated assignment failure

### LINKED ISSUE #2839: assign bot reports the wrong reason and the wrong person when a maintainer assigns someone else
### What is the issue?

`assign.yml` verifies the assignment took effect and replies on failure, but the failure message is written for the self-assign path only:

```js
await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
```

When a maintainer uses the `@handle` form, all three parts are wrong: it addresses the commenter rather than the target, says "to you" about a third party, and tells a maintainer that a maintainer will pick it up.

It also omits the only actionable part — *why* GitHub rejected it. In practice the cause is almost always that the target is not a collaborator and hasn't commented on that issue, since GitHub only accepts assignees who have participated in the thread. That's fixable in seconds if you know it, and opaque if you don't.

Seen on #2236, where the target had commented on the linked PR but never on the issue itself.

### Possible Solution

Branch the message on whether `target === commenter`, and name the likely cause in the delegated case — something along the lines of: the target needs to comment on the issue first, because GitHub only accepts assignees who are collaborators or thread participants.

julismo:
Fixes #2839

When GitHub rejects an assignment, the workflow keeps the existing fallback message for self-assignment. For maintainer delegation (`/assign @someone`), it addresses the requested assignee and explains that they may need to comment on the issue first or be a repository collaborator.

Validation:
- Ran the workflow's embedded `github-script` against mocked rejected self-assignment and delegated-assignment scenarios.

AI assistance: Codex helped inspect the workflow and prepare the small regression harness. I reviewed the change and test result.


.github/workflows/assign.yml
```diff
@@ -125,7 +125,7 @@ jobs:
             // confirm rather than trusting the call to have taken effect.
             const { data: after } = await github.rest.issues.get({ owner, repo, issue_number });
             if (!(after.assignees || []).some((a) => a.login === target)) {
-              await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              await reply(`@${target} GitHub couldn't assign this issue to you. You may need to comment on this issue first or be a repository collaborator.`);
```
recharte: #2839 asked for the message to branch on `target === commenter`, and this swaps one unconditional message for another. Please re-read the issue and work through what each path should say — starting with what "you may need to comment on this issue first" means for someone who reached this line by commenting on the issue.

recharte (changes requested):

.github/workflows/assign.yml
julismo: Updated in b1c771e. The failure handler now branches: self-assignment keeps the existing rejection and handoff message, while delegated assignment addresses the target and explains the comment/collaborator requirement. I ran the workflow script against simulated rejected self-assignment and delegated-assignment cases.

.github/workflows/assign.yml
```diff
@@ -125,7 +125,11 @@ jobs:
             // confirm rather than trusting the call to have taken effect.
             const { data: after } = await github.rest.issues.get({ owner, repo, issue_number });
             if (!(after.assignees || []).some((a) => a.login === target)) {
-              await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              if (target === commenter) {
+                await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              } else {
+                await reply(`@${target} GitHub couldn't assign this issue to you. You may need to comment on this issue first or be a repository collaborator.`);
```
recharte: This still leaves the target at a dead end: if they follow the advice and comment, nothing re-runs the assignment. Worth working out what single action takes them from this message to actually holding the issue, and saying that instead.

Smaller thing on the same line: reaching this message, the target sees a bot comment on an issue they've never interacted with, telling them an assignment they didn't request failed. Naming who tried would make it make sense.

recharte (commented):
The description no longer matches the change — it says the bot "does not promise that another maintainer will pick it up", which was true of the first commit but not this one, since the self-assign branch keeps that sentence.

julismo:
Updated the description to distinguish the self-assignment fallback from delegated assignment. The validation note now covers both rejection paths.

.github/workflows/assign.yml
recharte: @julismo thanks for the description update — this one is still open though, and it is about the message itself rather than the description.

recharte:
@julismo you probably missed my comment in the thread that's still open, please check it out, it's the only thing missing to be able to merge this PR.

.github/workflows/assign.yml
julismo: Changed the message itself. It now names who tried to assign, and points the target at posting `/assign` on its own line, which the parser accepts and which re-runs this workflow as a self-assignment, so it skips the maintainer-only delegation check. That is the single action that takes them from this message to holding the issue.

I stopped short of promising the outcome: the open, unassigned and open-issue-limit checks still gate it, and the workflow only knows the assignment did not stick, not why. So the message gives the usual cause rather than stating it as fact.

recharte (changes requested):
DCO is red — `77eaa01` is missing its `Signed-off-by`. Heads up that the check's suggested `git rebase HEAD~4 --signoff` would also flatten the `main` merge commit already on the branch; `git commit --amend -s` on the tip plus a force-push is enough, since that commit is the only one missing the trailer.

recharte:
@julismo apologies, this needs one more move and it's our doing, not yours.

We renamed branches on 18 August: the old `main` is now `v1.x`, and `main` is now the v2 line. GitHub retargeted every open PR automatically, so this one now points at `v1.x` — where the fix can't actually do anything, because `issue_comment` workflows only ever run from the default branch.

Could you reapply the change against `main`? The file is identical on both branches today, so it should carry over as is, and a fresh branch off `main` is probably less fuss than rebasing this one past the merge commit. Signing off as you go (`git commit -s`) clears the outstanding DCO failure at the same time.

Sorry for the runaround on a one-line change.

recharte:
Stale, closing.

julismo:
﻿Hey @recharte, sorry I left this hanging after your request to move it to `main`.

I've gone back through the full thread. The final change here already separated self-assignment from delegation, named who tried to assign the issue, and told the target to post `/assign` to retry. I didn't follow through on the branch move and the missing sign-off.

I see @HamidKhan1001 has since picked up #2839 in #3118, so it makes sense to continue there. I've left a comment on that PR linking your earlier feedback about addressing the target and giving them a single action to retry: https://github.com/openeverest/openeverest/pull/3118#issuecomment-5568700139

Happy for this PR to stay closed and for the existing work and review context to help finish that one.

PR #2843, opened 2026-08-09
fix(cli): display correct operator constraint string in upgrade error

### LINKED ISSUE #2842: Issue Framing Report: Incorrect Operator Constraint in Upgrade Error Output
### What is the issue?

# Issue: Incorrect Operator Constraint in Upgrade Error Output

**Component:** CLI (`everestctl upgrade`)
**File:** `pkg/cli/upgrade/upgrade.go`
**Severity:** Low / Usability Bug (Misleading Error Output)

## Problem

When running `everestctl upgrade`, Everest checks whether installed database operators meet the minimum version constraints for the target release.

If the installed PostgreSQL (`percona-postgresql-operator`) or MongoDB (`percona-server-mongodb-operator`) operator fails the minimum version check, the resulting CLI error message shows the **MySQL/PXC operator constraint** instead of the constraint for the operator that actually failed — which is misleading when debugging an upgrade block.

**Example bug output:**

```
percona-postgresql-operator version "2.2.0" does not meet minimum requirements of "<pxc-operator-constraint>"
```

## Root cause (initial look)

In `checkOperatorRequirements`, the loop iterates over all three operator configs, but the error message appears to reference `supVer.PXCOperator` regardless of which operator actually failed the check — rather than the constraint tied to that specific operator.

```go
cfg := []requirementsCheck{
    {common.MySQLOperatorName, supVer.PXCOperator},
    {common.PostgreSQLOperatorName, supVer.PGOperator},
    {common.MongoDBOperatorName, supVer.PSMBDOperator},
}
```

### Steps to Reproduce

1. Install OpenEverest with an older PostgreSQL operator version that does **not** meet the minimum requirement for the target upgrade:
```bash
   everestctl install --namespaces <ns> --operator.postgresql=true --skip-wizard
```
2. Confirm the installed operator version is below the minimum required for the next release:
```bash
   kubectl get csv -n <ns> | grep percona-postgresql-operator
```
3. Attempt to upgrade:
```bash
   everestctl upgrade
```
4. Observe the error message returned.
**Expected:** Error should reference the PostgreSQL operator's own minimum version constraint (`supVer.PGOperator`).
 
**Actual:** Error references the MySQL/PXC operator's constraint (`supVer.PXCOperator`) instead:
```
percona-postgresql-operator version "2.2.0" does not meet minimum requirements of "<pxc-operator-constraint>"
```
 
5. Repeat with an outdated MongoDB operator (`--operator.mongodb=true`) — same mismatch occurs, again showing the PXC constraint instead of the MongoDB one.

### Relevant Logs / Output

_No response_

### OpenEverest Version

v1.12.0

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

I came across this while going through the upgrade flow and think it's a genuine bug in how the error string is built. I'd like to work on this — will dig into `checkOperatorRequirements` further and raise a PR with a fix (plus test coverage for each operator's constraint message) once I've confirmed the right approach.

### Additional Context

_No response_

Bhupesh-081:
## Summary

Fixes the misleading error message reported in #2842, where an upgrade
failure for the PostgreSQL or MongoDB operator incorrectly displayed the
MySQL/PXC operator's version constraint instead of its own.

Fixes #2842

### `pkg/cli/upgrade/upgrade.go`

`checkOperatorRequirements` was hardcoding `supVer.PXCOperator.String()` in
the error message regardless of which operator actually failed the check.
Replaced it with `c.constraints.String()`, which reflects the constraint
for the operator being validated in that iteration:

```diff
 if !c.constraints.Check(v) {
     return fmt.Errorf(
         "%s version %q does not meet minimum requirements of %q",
-        c.operatorName, v, supVer.PXCOperator.String(),
+        c.operatorName, v, c.constraints.String(),
     )
 }
```

### `pkg/cli/upgrade/upgrade_test.go`

Added `TestUpgrade_checkOperatorRequirements` with subtests covering
version check failures for `percona-postgresql-operator` and
`percona-server-mongodb-operator`, asserting the error string contains
each operator's own constraint rather than the PXC one.

## Verification

- `go test ./pkg/cli/upgrade/...` — PASS
- `go test ./pkg/...` — PASS (full suite green)

## Notes for reviewers

This is a message-formatting fix only — no behavioral change to the
version check logic itself, so existing upgrade flows are unaffected.

### DIFF
--- a/pkg/cli/upgrade/upgrade_test.go
+++ b/pkg/cli/upgrade/upgrade_test.go
@@ -1,5 +1,6 @@
 // everest
 // Copyright (C) 2023 Percona LLC
+// Copyright (C) 2026 The OpenEverest Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -24,12 +25,18 @@ import (
 	"testing"
 
 	version "github.com/Percona-Lab/percona-version-service/versionpb"
+	"github.com/blang/semver/v4"
 	goversion "github.com/hashicorp/go-version"
+	olmversion "github.com/operator-framework/api/pkg/lib/version"
+	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 
+	"github.com/percona/everest/pkg/common"
 	"github.com/percona/everest/pkg/kubernetes"
 	versionservice "github.com/percona/everest/pkg/version_service"
 )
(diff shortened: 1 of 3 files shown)


pkg/cli/upgrade/upgrade_test.go
```diff
@@ -282,3 +289,104 @@ func TestUpgrade_ValidateVersionToUpgrade(t *testing.T) {
...
+					Name:      "pg-operator.v2.2.0",
+					Namespace: "everest",
+				},
+				Spec: olmv1alpha1.ClusterServiceVersionSpec{
+					Version: olmversion.OperatorVersion{
+						Version: semver.MustParse("2.2.0"),
+					},
+				},
+			},
+			expectedError: "percona-postgresql-operator version \"2.2.0\" does not meet minimum requirements of \">= 2.4.0\"",
```
recharte: IMO raw string literals read better here — the expected message contains double quotes deliberately, and the escaping buries that.

```suggestion
			expectedError: `percona-postgresql-operator version "2.2.0" does not meet minimum requirements of ">= 2.4.0"`,
```

Same for the MongoDB case below.

pkg/cli/upgrade/upgrade_test.go
```diff
@@ -282,3 +289,104 @@ func TestUpgrade_ValidateVersionToUpgrade(t *testing.T) {
...
+	psmdbConstraint, err := goversion.NewConstraint(">= 1.15.0")
+	require.NoError(t, err)
+
+	supVer := &common.SupportedVersion{
+		PXCOperator:   pxcConstraint,
+		PGOperator:    pgConstraint,
+		PSMBDOperator: psmdbConstraint,
+	}
+
+	tests := []struct {
```
recharte, line 309: Worth a third row for `percona-xtradb-cluster-operator`, the one operator the old code happened to name correctly. Being honest about its value: the two existing rows already catch any single hardcoded constraint, so this buys completeness of the `cfg` table rather than extra regression coverage.

recharte (changes requested):
Needs a rebase — `go.mod` conflicts with `main` after the AWS SDK v2 migration landed. The fix itself is right.

pkg/cli/upgrade/upgrade_test.go
Bhupesh-081: @recharte, addressed all three points from your review:

- Rebased onto `main` and resolved the `go.mod` conflict from the AWS SDK v2 migration (kept the v2 deps, retained `github.com/blang/semver/v4`, ran `go mod tidy`)
- Switched `expectedError` in `upgrade_test.go` to raw backtick literals instead of escaped quotes
- Added the `percona-xtradb-cluster-operator` case to `TestUpgrade_checkOperatorRequirements` for `cfg` table completeness

`go test -v ./pkg/cli/upgrade/...` passes clean. Ready for another look whenever you have time.

recharte (approved):
Thanks @Bhupesh-081 looks good

PR #2847, opened 2026-08-10
feat: add restore delete command with confirmation and wait options

### LINKED ISSUE #2807: Add `everestctl restore delete` subcommand

Restores can be created and listed from the terminal but not removed. Restores are imperative history (`restore list` is the audit trail), so records accumulate forever; pruning old or failed entries requires the UI or `kubectl delete`, bypassing Everest API authentication/RBAC.

Add a `restore delete` subcommand. Required flags: `--name`, `--namespace`. Deleting a Restore removes the history record — no stored backup data is touched — so it sits at the resource-removal tier of the shared confirmation pattern (#2658):

- **Interactive:** y/N prompt, default No. Declined exits non-zero without doing anything.
- **Non-interactive:** if stdin is not a TTY *or* `--json` is set, and `--yes` wasn't given, fail immediately (exit 1) with *"confirmation required in non-interactive mode; re-run with --yes"*.

**In-flight guard:** if `status.state` is `Pending` or `Running`, the command refuses with exit 1 — e.g. *"restore \"my-mongo-restore-x7k2q\" is still running; wait for it to finish or re-run with --force"*. `--force` overrides the guard; when forcing, the confirmation prompt warns that interrupting an in-flight restore triggers the provider's cleanup of the engine restore and can leave the target instance in an inconsistent state. `Error` is deliberately not guarded (deleting a stuck operation is the normal remediation). `--force` is orthogonal to `--yes` — scripted force-deletes need both (`-y --force`), keeping the #2658 rule that `--force` is never a `--yes` synonym. This is a client-side accident guard, not an invariant: it is race-prone by nature and kubectl bypasses it; a server-side guard was deliberately avoided because it would wedge `instance delete --deletion-policy Cascade` on Running children and break GitOps pruning. If the API later enforces this server-side with an override parameter, `--force` maps onto it.

Optional flags:
- `--wait` / `--timeout` — poll until GET returns 404 → exit 0, timeout → exit 124. Ctrl-C cancels only the wait.
- `--ignore-not-found` — same semantics as `instance delete`: already-absent exits 0, short-circuiting prompt and wait.

A 404 is mapped to `restore %q not found in namespace %q`. Follows existing CLI conventions: `--json`/`--verbose`, `--cluster`, `--context`.

```
# Delete a restore record (prompts y/N)
everestctl restore delete --name my-mongo-restore-x7k2q --namespace everest

# Scripted cleanup of finished restores
everestctl restore delete --name my-mongo-restore-x7k2q --namespace everest -y --ignore-not-found

# Abort and delete a restore that is still running
everestctl restore delete --name my-mongo-restore-x7k2q --namespace everest -y --force
```

`kubectl delete restore` — bypasses Everest API authentication/RBAC.

### Additional Context

Applies to v2 (`release-2.0`). Reuses the shared confirmation helper and waiter from #2658/#2660. Completes create/list/delete for the `restore` group (#2620, #2621).

VijetaPriya47:
- Implemented the `restore delete` command to allow users to delete restore records via the Everest API.
- Added flags for name, namespace, cluster, context, confirmation prompt, force deletion, ignore not found, wait for deletion, and timeout.
- Integrated interactive confirmation for deletion with options to skip prompts.
- Implemented logic to handle in-flight restores and provide appropriate error messages.
- Added tests for various scenarios including successful deletion, not found cases, and error handling.
- Updated related files to support the new delete functionality and ensure consistent behavior across commands.
- Closes #2807


VijetaPriya47:
@recharte Could you check it out?

pkg/cli/restore/delete_test.go
```diff
@@ -0,0 +1,497 @@
...
+	defer srv.Close()
+
+	rd := NewDeleter(Config{}, zap.NewNop().Sugar())
+	err := rd.Run(context.Background(), yesOpts(), newConfigPath(t, srv.URL))
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "is still running")
+	assert.False(t, called, "delete must not be issued while the restore is in flight")
+}
+
+func TestDelete_InFlight_WithForce_Succeeds(t *testing.T) {
```
recharte, line 486: The `--force` ≠ `--yes` rule from #2658 is the load-bearing bit of this design — it's in the long help, in the flag description, and it's the thing a future refactor is most likely to break by folding the guard override into the confirmation gate. Every in-flight test here goes through `yesOpts()`, so nothing pins it.

One more case covers it:

```go
func TestDelete_InFlight_ForceWithoutYes_StillRequiresConfirmation(t *testing.T) {
	t.Parallel()

	called := false
	srv := newDeleteServer(t,
		func(w http.ResponseWriter, _ *http.Request) {
			called = true
			w.WriteHeader(http.StatusNoContent)
		},
		getHandlerWithState(t, restoreStateRunning),
	)
	defer srv.Close()

	opts := DeleteOptions{
		Name:       "my-mongo-restore-x7k2q",
		Namespace:  "everest",
		Cluster:    "main",
		Force:      true,
		IsTerminal: isTerminalFalse,
	}

	rd := NewDeleter(Config{}, zap.NewNop().Sugar())
	err := rd.Run(context.Background(), opts, newConfigPath(t, srv.URL))
	require.ErrorIs(t, err, confirm.ErrNonInteractive)
	assert.False(t, called, "--force must not imply --yes")
}
```

pkg/cli/restore/wait.go
```diff
@@ -67,3 +76,16 @@ func restoreFailureMessage(r *client.Restore) string {
...
+func newRestoreDeletePoll(
+	c *client.ClientWithResponses,
+	cluster, namespace, name string,
+) wait.PollFunc[*client.Restore] {
+	return deletion.GonePoll("restore", name, fetchRestore(c, cluster, namespace, name))
+}
+
+func deleteCondition(r *client.Restore) (wait.Outcome, string) {
+	return deletion.GoneCondition("restore deleted", restoreState)(r)
+}
```
recharte, line 93: `restoreState` is the list-table cell formatter, so it yields the state verbatim and `"-"` when status is absent. Fed to `GoneCondition` as the pending message, that's what `OnUpdate` prints while the restore is *still there*:

```
Restore "my-mongo-restore-x7k2q" deletion requested; waiting for it to be fully removed...
  Succeeded
```

Which reads as "the deletion succeeded" — the opposite of what it means. The other two delete-waits both say what the pending state *is* rather than echoing a field: instance uses `progressMessage(v, instancePhase(v))`, backup storage uses `"backup storage still exists — likely referenced by an Instance or Backup"`.

```suggestion
func deleteCondition(r *client.Restore) (wait.Outcome, string) {
	return deletion.GoneCondition("restore deleted", func(v *client.Restore) string {
		return "restore still exists (state: " + restoreState(v) + ")"
	})(r)
}
```

commands/restore/delete.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	// Ctrl-C only cancels --wait; the cancellable context is set up inside
+	// Deleter.waitForDeletion, not here.
+	rd := restorecli.NewDeleter(*deleteCfg, logger.GetLogger())
+	runErr := rd.Run(cmd.Context(), *deleteOpts, cfgPath)
+	os.Exit(waitcmd.ExitCode(runErr, logger.GetLogger(), deleteCfg.Pretty,
+		fmt.Sprintf("wait cancelled; restore %q deletion continues in the background — check with 'everestctl restore list'", deleteOpts.Name),
+		fmt.Sprintf("timed out after %s waiting for restore %q to be deleted; deletion is still running server-side, check with 'everestctl restore list --namespace %s'",
+			deleteOpts.Timeout, deleteOpts.Name, deleteOpts.Namespace),
```
recharte: `restore list` marks both `--namespace` and `--instance` required, so `everestctl restore list` exits on flag validation and the timeout variant does too. The cancel message also drops `--namespace` where the timeout one keeps it.

```suggestion
		fmt.Sprintf("wait cancelled; restore %q deletion continues in the background — check with 'everestctl restore list --namespace %s --instance <instance>'", deleteOpts.Name, deleteOpts.Namespace),
		fmt.Sprintf("timed out after %s waiting for restore %q to be deleted; deletion is still running server-side, check with 'everestctl restore list --namespace %s --instance <instance>'",
			deleteOpts.Timeout, deleteOpts.Name, deleteOpts.Namespace),
```

The placeholder is a bit unsatisfying — the Deleter does have the real value from the guard fetch (`spec.instanceRef.name`), if you'd rather thread it up than print `<instance>`.

pkg/cli/restore/delete.go
```diff
@@ -0,0 +1,205 @@
...
+func (rd *Deleter) Run(ctx context.Context, opts DeleteOptions, cfgPath string) error {
+	c, err := authcli.NewAPIClient(authcli.Config{Pretty: rd.config.Pretty}, rd.l.Desugar().Sugar(), cfgPath, opts.Context)
+	if err != nil {
+		return err
+	}
+
+	restore, gone := rd.checkRestoreExists(ctx, c, opts)
+	if gone && opts.IgnoreNotFound {
+		return rd.emitAlreadyGone(opts)
+	}
```
recharte, line 84: Unlike `backup-storage delete`, the pre-delete GET here is unconditional (the guard needs it), so without `--ignore-not-found` we ask "Delete restore "x" in namespace "everest"?", wait for the user to type `y`, and only then return `not found`. `TestDelete_NotFound_WithoutIgnoreNotFound_Errors` passes `--yes`, so the prompt never runs in the test.

Now that we have the answer before the prompt, we can just use it:

```suggestion
	restore, gone := rd.checkRestoreExists(ctx, c, opts)
	if gone {
		if !opts.IgnoreNotFound {
			return fmt.Errorf("restore %q not found in namespace %q", opts.Name, opts.Namespace)
		}
		return rd.emitAlreadyGone(opts)
	}
```

The DELETE path still produces the identical message via `deletion.CheckDeleteResponse` for the racy case, so behaviour only changes in the case we can already decide.

pkg/cli/restore/delete.go
```diff
@@ -0,0 +1,205 @@
...
+			statusText = msg
+		}
+		rd.l.Warnf("pre-delete check for restore %q returned unexpected status %s, proceeding without it", opts.Name, statusText)
+		return nil, false
+	}
+}
+
+// restoreStateForGuard is restoreState, but nil-safe: nil means the
+// pre-delete fetch found nothing, so there's no live restore to guard.
+func restoreStateForGuard(r *client.Restore) string {
```
recharte: What should happen to a restore that exists but has no `status.state` yet? Right now `restoreState` gives `"-"` for it, `inFlight` says no, and we delete without prompting — and that's the window right after `restore create`, when a restore is arguably at its most in-flight.

`restoreStateForGuard` folds two situations into one answer: the fetch was ambiguous, and the controller hasn't observed the restore yet. Skipping the guard is clearly right for the first, it must not block on a read it couldn't complete — but the second is a restore we did read successfully and know is young.

I wrote "Pending or Running" in #2807 without thinking about the statusless case, so the issue doesn't settle this and I honestly don't know which way is better. Guarding it covers the riskiest window, but it produces a refusal the user can't explain from `restore list`, which prints `-` in the STATE column for exactly those restores. Leaving it as-is keeps the rule as literal as the field itself. You've spent more time in this code than anyone — what's your read?

Either way, could we pin the outcome with a test next to `TestInFlight`'s `inFlight("")` case, so it lands as a decision rather than a gap?

pkg/cli/instance/delete.go
```diff
@@ -126,10 +126,11 @@ func (id *Deleter) Run(ctx context.Context, opts DeleteOptions, cfgPath string)
 // checkInstanceExists is for --ignore-not-found: (nil, true) means the
 // instance is gone. Otherwise it also returns the fetched instance, so
 // confirmDeletion can reuse it instead of fetching again. Any unclear
-// result returns (nil, false) and lets the normal delete flow handle it.
+// result (fetch error, non-404 status) returns false and lets the normal delete flow handle it.
 func (id *Deleter) checkInstanceExists(ctx context.Context, c *client.ClientWithResponses, opts DeleteOptions) (*client.Instance, bool) {
 	resp, err := c.GetInstanceWithResponse(ctx, opts.Cluster, opts.Namespace, opts.Name)
 	if err != nil {
+		id.l.Warnf("pre-delete check for instance %q failed: %v", opts.Name, err)
```
recharte, line 133: Were the `Warnf` calls in the `instance` and `backup-storage` probes part of the plan here, or did they come along for the ride while keeping the three copies identical?

Not asking you to revert them — a silently swallowed 500 immediately before a delete is worth a line, and I'd much rather the three probes stay identical than drift. I'd just like to know how to read them, since they change behaviour in two commands this PR otherwise only touches to de-duplicate `checkDeleteResponse`. If they're intentional, a sentence in the PR description saying so would save the next reader the same question.

recharte (changes requested):
Hey @VijetaPriya47, thanks for picking this one up! Nice to see you carried the whole `backup-storage delete` test matrix across rather than only testing the new bits, and then added the guard cases on top of it — that's what made this quick to review.

Six notes inline. Two are genuinely open questions rather than requests: the statusless-restore case is a hole I left in #2807, and I'd like your read on it before we settle it either way.

pkg/cli/restore/delete.go
VijetaPriya47: Let's guard it. Honestly, a restore with no status yet feels more fragile than one we know is Running, not less we know for a fact it was just created and nothing's had a chance to touch it yet. That's the exact moment where interrupting it would do the most damage.

On the restore list thing, I don't think the refusal needs to match what's in the STATE column. We don't have to say "Pending" or "Running" just because that's what the guard usually says. We can just tell the truth: "this restore hasn't reported its status yet." That's actually clearer than the - restore list shows anyway. So I'd rather fix the wording than loosen the guard.

I'm also going to stop reusing restoreState/"-" for this. That - is just something the list command shows when there's nothing to display, it was never meant to mean anything, and using it here would tie a real safety check to a formatting decision. Instead, restoreStateForGuard will tell us two separate things: whether we actually managed to read the restore at all, and (if we did) what its status is. That way "we checked and it has no status" and "we couldn't check" stay two different answers instead of getting mixed into one.

And yup, I'll add a test right next to TestInFlight's inFlight("") case so this is locked in as a decision, not left as a gap. WDYT? @recharte

pkg/cli/instance/delete.go
VijetaPriya47: Yeah, those were intentional. When I added the warning log to restore's version, I didn't want the other two probes to be the odd ones out. Keeping all three identical mattered more to me than keeping this PR narrowly scoped to just the de-dup, especially since "three copies quietly drift apart" is exactly the kind of thing that's bitten this codebase before.

I'll make sure to add in it the description next time :>

pkg/cli/restore/delete_test.go
VijetaPriya47: good point

commands/restore/delete.go
```diff
@@ -0,0 +1,122 @@
...
+Pass --yes/-y to skip the prompt; in a non-interactive context (no terminal,
+or --json) omitting --yes fails immediately instead of hanging on a prompt
+nobody can answer.
+
+If the restore is still Pending or Running, the command refuses by default,
+since deleting an in-flight restore interrupts the provider mid-operation.
+Pass --force to override the guard; forcing still shows the confirmation
+prompt (with an added warning), it does not skip it — --force is orthogonal
+to --yes, scripted force-deletes need both. A restore in the Error state is
+never guarded: deleting a stuck restore is the normal way to clear it.
```
recharte: The guard refuses a statusless restore now, but `--help` still describes it as Pending or Running — and the `--force` description just below is exactly where someone will look when they hit `hasn't reported its status yet` and want to know why.

```suggestion
If the restore is still Pending or Running, or hasn't reported a status
yet, the command refuses by default, since deleting an in-flight restore
interrupts the provider mid-operation. Pass --force to override the guard;
forcing still shows the confirmation prompt (with an added warning), it
does not skip it — --force is orthogonal to --yes, scripted force-deletes
need both. A restore in the Error state is never guarded: deleting a stuck
restore is the normal way to clear it.
```

Three more spots carry the same stale wording: the `--force` flag description on line 82, `DeleteOptions.Force` in `pkg/cli/restore/delete.go`, and `FlagRestoreForce` in `pkg/cli/flags.go`.

pkg/cli/restore/delete.go
recharte: Agreed — let's guard it. The `(state string, ok bool)` split is better than what I had in mind: "we couldn't read it" and "we read it and there's nothing there" really are two different answers, and collapsing them into one string is what made the original shape awkward. And you're right that the refusal owes nothing to the STATE column.

recharte (commented):
Everything from the last round is addressed, and the guard reshape is better than the version I had in mind. The one comment left is the help text the change outgrew

commands/restore/delete.go
recharte: @VijetaPriya47 The help paragraph reads right now, thanks. The three other spots from that comment are still on the old wording though:

- `commands/restore/delete.go:83` — the `--force` flag description, which is what shows up in `everestctl restore delete --help`
- `pkg/cli/restore/delete.go:46` — the `DeleteOptions.Force` field comment
- `pkg/cli/flags.go:186` — the `FlagRestoreForce` doc comment

All three still say "in-flight (Pending/Running)", where `Pending/Running/no status yet` now covers it. The suggestion block only carried the paragraph, so this is an easy one to miss — want to catch them in the same pass?

VijetaPriya47: fixed

recharte (approved):
Thanks Vijeta, looks good 💪

PR #2851, opened 2026-08-10
fix(session): avoid nil pointer panic when the JWT private key PEM is invalid

### LINKED ISSUE #2850: session: getPrivateKey panics with a nil pointer dereference on an invalid JWT private key file
﻿### What is the issue?

in `pkg/session/manager.go`, `getPrivateKey()` reads the everest JWT signing key from disk and decodes it like this:

```go
func getPrivateKey() (*rsa.PrivateKey, error) {
	pemString, err := os.ReadFile(common.EverestJWTPrivateKeyFile)
	if err != nil {
		return nil, errors.Join(err, errors.New("failed to read JWT private key"))
	}
	block, _ := pem.Decode(pemString)
	return x509.ParsePKCS1PrivateKey(block.Bytes)
}
```

`pem.Decode` returns a nil block (with a nil error) whenever it can't find a valid PEM block in the input. the code never checks for that and goes straight to `block.Bytes`, so any time `common.EverestJWTPrivateKeyFile` (`/etc/jwt/id_rsa`) exists but doesn't contain a valid PEM block, this panics with a nil pointer dereference instead of returning an error.

`getPrivateKey()` runs inside `session.New()`, which is called once per everest-server startup, so this takes down the whole process. an empty file, a truncated write, or a manually edited `everest-jwt` secret would all crash the server with an unhelpful panic instead of a clear "invalid JWT private key" error.

### Steps to Reproduce

found this reading the code, not from a running cluster. reproduced with a go unit test that calls the pem-decode-and-parse step directly with non-PEM input:

1. checkout main
2. call the equivalent of `parsePrivateKeyPEM([]byte("this is not a PEM encoded private key"))` (i.e. `pem.Decode` + `x509.ParsePKCS1PrivateKey(block.Bytes)` with no nil check on `block`)
3. run it

the same thing happens for real whenever `/etc/jwt/id_rsa` doesn't hold a valid PEM block, e.g. an empty secret, a partially-written file, or a bad manual edit of the `everest-jwt` secret.

### Relevant Logs / Output

```
[signal 0xc0000005 code=0x0 addr=0x18 pc=0x7ff78e2a0178]

goroutine 90 [running]:
panic({0x7ff78e540f80?, 0x7ff78fe81f90?})
	C:/Program Files/Go/src/runtime/panic.go:860 +0x13a
github.com/percona/everest/pkg/session.parsePrivateKeyPEM(...)
	pkg/session/manager.go:185 +0x18
FAIL	github.com/percona/everest/pkg/session	33.274s
```
(captured locally with `go test ./pkg/session/...` against a small helper I extracted from `getPrivateKey()` to isolate the decode step; same call path as the unmodified function on main)

### OpenEverest Version

main branch (tested against commit d0fb75c)

### Installation Method

Compiled from source (found via code review + a local `go test` run, not a running deployment)

### Kubernetes Distribution

Other (please specify below)

### Environment Details

not cluster- or OS-dependent. this is a pure Go bug in JWT key parsing, triggered whenever the file at `common.EverestJWTPrivateKeyFile` (`/etc/jwt/id_rsa`) does not contain a valid PEM-encoded block.

### Possible Solution

check `block == nil` right after `pem.Decode` and return a descriptive error instead of falling through to `block.Bytes`. also worth wrapping the `x509.ParsePKCS1PrivateKey` error instead of returning it bare, so a malformed-but-present PEM block gets a clear message too.

happy to open a small PR for this with a regression test.

### Additional Context

this only affects the failure path (missing/empty/corrupted key file). deployments with a normally-generated key are unaffected.

Atishyy27:
## What

`getPrivateKey()` in `pkg/session/manager.go` decoded the everest JWT signing key with `pem.Decode()` and dereferenced the result without checking for `nil`. `pem.Decode` returns a `nil` block (with a `nil` error) when it can't find a valid PEM block, so an empty, truncated, or otherwise invalid key file crashed the whole server with a nil pointer panic instead of a clean startup error.

This PR:
- extracts the decode+parse step into a small `parsePrivateKeyPEM()` helper
- checks for a `nil` block after `pem.Decode` and returns a descriptive error instead of panicking
- wraps the `x509.ParsePKCS1PrivateKey` error instead of returning it bare, so a malformed-but-present PEM block also gets a clear message
- adds `TestParsePrivateKeyPEM` covering a valid key, empty input, non-PEM input, and a PEM block that isn't a valid PKCS1 key

## Why

`getPrivateKey()` runs inside `session.New()`, i.e. on every everest-server startup. Any deployment issue that leaves `/etc/jwt/id_rsa` empty or corrupted (bad secret, manual edit, partial write) currently takes the whole process down with an unhelpful panic instead of a clear "invalid JWT private key" error.

## Testing

- `go build ./...`
- `go test ./pkg/session/...`: new test panics on the old code (nil pointer dereference at the old `block.Bytes` line) and passes cleanly after the fix; full package suite passes

Fixes #2850

### DIFF
--- a/pkg/session/manager.go
+++ b/pkg/session/manager.go
@@ -171,8 +171,24 @@ func getPrivateKey() (*rsa.PrivateKey, error) {
 	if err != nil {
 		return nil, errors.Join(err, errors.New("failed to read JWT private key"))
 	}
+	return parsePrivateKeyPEM(pemString)
+}
+
+// parsePrivateKeyPEM decodes a PEM-encoded PKCS1 RSA private key.
+//
+// pem.Decode returns a nil block (without an error) when the input does not
+// contain a valid PEM block, so that case must be checked explicitly before
+// the block is dereferenced.
+func parsePrivateKeyPEM(pemString []byte) (*rsa.PrivateKey, error) {
 	block, _ := pem.Decode(pemString)
-	return x509.ParsePKCS1PrivateKey(block.Bytes)
+	if block == nil {
+		return nil, errors.New("failed to decode JWT private key: no PEM data found")
+	}
+	key, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, errors.Join(err, errors.New("failed to parse JWT private key"))
+	}
+	return key, nil
 }
 
 // KeyFunc retruns a function for getting the public RSA keys used
--- a/pkg/session/manager.go
+++ b/pkg/session/manager.go
@@ -171,7 +171,11 @@ func getPrivateKey() (*rsa.PrivateKey, error) {
 	if err != nil {
 		return nil, errors.Join(err, errors.New("failed to read JWT private key"))
 	}
-	return parsePrivateKeyPEM(pemString)
+	key, err := parsePrivateKeyPEM(pemString)
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", common.EverestJWTPrivateKeyFile, err)
+	}
+	return key, nil
 }
 
 // parsePrivateKeyPEM decodes a PEM-encoded PKCS1 RSA private key.
--- a/pkg/session/manager_test.go
+++ b/pkg/session/manager_test.go
@@ -16,6 +16,10 @@ package session
 
 import (
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


pkg/session/manager.go
```diff
@@ -171,8 +171,24 @@ func getPrivateKey() (*rsa.PrivateKey, error) {
 	if err != nil {
 		return nil, errors.Join(err, errors.New("failed to read JWT private key"))
 	}
+	return parsePrivateKeyPEM(pemString)
```
recharte: The missing-file case already names the path for free, because `os.ReadFile` returns a `*fs.PathError` — "open /etc/jwt/id_rsa: no such file or directory". The invalid-content case, which is the one this PR is about, doesn't. Since the whole point is that somebody reading a crash-looping pod's log should know where to look, it's worth closing that gap:

```suggestion
	key, err := parsePrivateKeyPEM(pemString)
	if err != nil {
		return nil, fmt.Errorf("%s: %w", common.EverestJWTPrivateKeyFile, err)
	}
	return key, nil
```

Keeping the helper itself path-agnostic is the right call — this just puts the context back at the layer that knows it.

pkg/session/manager_test.go
```diff
@@ -33,6 +37,67 @@ import (
...
+
+			key, err := parsePrivateKeyPEM(tc.in)
+			if tc.expectError {
+				require.Error(t, err)
+				assert.Nil(t, key)
+				return
+			}
+			require.NoError(t, err)
+			require.NotNil(t, key)
+			assert.Equal(t, validKey.N, key.N)
```
recharte:
```suggestion
			assert.True(t, key.Equal(validKey))
```

`rsa.PrivateKey` has an `Equal` method that compares the full key. Same one line, and it can't pass on something that round-tripped the modulus and nothing else.

recharte (commented):
Hey @Atishyy27, thanks for the PR — and for turning the report into a patch the same morning. Two small things inline, neither blocking.

pkg/session/manager.go
Atishyy27: applied, thanks. the `fs.PathError` point is the one i missed, the read branch already carries the path so the parse branch was the only one where an operator would see the failure without knowing which file to look at. wrapping at `getPrivateKey` keeps that in the layer that knows the path.

pkg/session/manager_test.go
Atishyy27: applied. checked Equal in the stdlib to be sure of the scope, it compares the public key, D and the primes, so it catches the case the N-only assertion would have let through.

PR #2862, opened 2026-08-10
fix(session): avoid nil pointer panic when the JWT private key PEM is invalid (release-2.0)

Atishyy27:
﻿## What

Backports the fix from #2851 to `release-2.0`.

`getPrivateKey()` in `pkg/session/manager.go` decodes the everest JWT signing key with `pem.Decode()` and dereferences the result without checking for `nil`. `pem.Decode` returns a `nil` block (with a `nil` error) when it can't find a valid PEM block, so an empty, truncated, or otherwise invalid key file crashes the server with a nil pointer panic instead of a clean startup error.

Same change as #2851, so the two branches don't diverge:
- extracts the decode+parse step into a `parsePrivateKeyPEM()` helper
- checks for a `nil` block after `pem.Decode` and returns a descriptive error
- wraps the `x509.ParsePKCS1PrivateKey` error instead of returning it bare
- names the key path at `getPrivateKey`, keeping the helper path-agnostic (per the review on #2851)
- adds `TestParsePrivateKeyPEM` covering a valid key, empty input, non-PEM input, and a PEM block that isn't a valid PKCS1 key

## Why

@recharte flagged this on #2850:

> The same code is on release-2.0 too, so the v2 line needs the same fix. That's the one constraint on top of what you've already got in #2851, which targets main.

`getPrivateKey()` runs inside `session.New()`, i.e. on every everest-server startup, so an invalid key file takes down the whole process with an undiagnosable trace.

Worth noting `pkg/plugintoken/service.go` on this branch already does the nil-block check and wraps the parse error, so this brings `pkg/session` in line with the shape v2 already uses elsewhere.

## Testing

Reproduced the panic against the unmodified `release-2.0` code path (`pem.Decode` returns `block=<nil>` with a nil error, then `block.Bytes` panics with "invalid memory address or nil pointer dereference").

- `go build ./...`
- `go test ./pkg/session/...`

Related to #2850. Not marked as closing it, since #2851 already targets that for `main`.

### DIFF
--- a/pkg/session/blocklist_test.go
+++ b/pkg/session/blocklist_test.go
@@ -28,8 +28,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 
 	"github.com/openeverest/openeverest/v2/pkg/common"
 	"github.com/openeverest/openeverest/v2/pkg/kubernetes"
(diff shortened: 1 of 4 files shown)


spron-in (approved):

spron-in:
This one seems cherry picked already

spron-in:
This one seems cherry picked already

PR #2863, opened 2026-08-10
Refactor delete response handling and add backup delete command

### LINKED ISSUE #2806: Add `everestctl backup delete` subcommand

Backups can be created and listed from the terminal but not removed — retention housekeeping (pruning old on-demand backups, cleaning up failed ones) still requires the UI or `kubectl delete`, bypassing Everest API authentication/RBAC.

Add a `backup delete` subcommand. Required flags: `--name`, `--namespace`.

What deletion means is governed by the Backup's `spec.deletionPolicy` (`Delete`, the default, removes the engine-native resource **and the artifact in the BackupStorage**; `Retain` removes the resource but leaves the data recoverable out-of-band). The confirmation tier follows the shared pattern established by `instance delete` (#2658) and adapts to the effective policy, which the CLI reads from the Backup before prompting:

- **Policy `Delete` (data loss):** type-the-name confirmation, stating the blast radius first, e.g. *"This permanently deletes backup \"pre-upgrade\" and, per policy Delete, its data in storage \"my-s3\". Type the backup name to confirm."*
- **Policy `Retain` (resource removal):** y/N prompt, default No, noting the data stays in storage.
- **Non-interactive:** if stdin is not a TTY *or* `--json` is set, and `--yes` wasn't given, fail immediately (exit 1) with *"confirmation required in non-interactive mode; re-run with --yes"*.

**In-flight guard:** if `status.state` is `Pending` or `Running`, the command refuses with exit 1 — e.g. *"backup \"nightly\" is still running; wait for it to finish or re-run with --force"*. `--force` overrides the guard. `Error` is deliberately not guarded (deleting a stuck operation is the normal remediation), and `Deleting` needs no guard. `--force` is orthogonal to `--yes` — scripted force-deletes need both (`-y --force`), keeping the #2658 rule that `--force` is never a `--yes` synonym. This is a client-side accident guard, not an invariant: it is race-prone by nature and kubectl bypasses it; a server-side guard was deliberately avoided because it would wedge `instance delete --deletion-policy Cascade` on Running children and break GitOps pruning. If the API later enforces this server-side with an override parameter, `--force` maps onto it.

No policy-override flag at delete time: `spec.deletionPolicy` is mutable on a live Backup, so changing the artifact's fate is an update, not a delete option (and the field freezes once deletion starts, so an override could race).

Optional flags:
- `--wait` / `--timeout` — block until the Backup is actually gone (provider cleanup of engine resources/artifacts runs behind a finalizer): poll until GET returns 404 → exit 0, timeout → exit 124 with a message that cleanup is still running server-side. Ctrl-C cancels only the wait.
- `--ignore-not-found` — same semantics as `instance delete`: already-absent exits 0, short-circuiting prompt and wait.

A 404 is mapped to `backup %q not found in namespace %q`. Follows existing CLI conventions: `--json`/`--verbose`, `--cluster`, `--context`.

```
# Delete a backup (policy Delete → states blast radius, asks to type the name)
everestctl backup delete --name pre-upgrade --namespace everest

# Scripted retention pruning
everestctl backup delete --name nightly-20260701 --namespace everest -y --ignore-not-found

# Abort and delete a backup that is still running
everestctl backup delete --name hung-backup --namespace everest -y --force

# Wait until the artifact cleanup finished
everestctl backup delete --name pre-upgrade --namespace everest -y --wait --timeout 10m
```

`kubectl delete backup` — bypasses Everest API authentication/RBAC.

### Additional Context

Applies to v2 (`release-2.0`). Reuses the shared confirmation helper (`pkg/cli/confirm`, both tiers) and the shared waiter's until-404 condition from `instance delete` (#2658) / `backupstorage delete` (#2660). Completes create/list/delete for the `backup` group.

VijetaPriya47:
Backups could be created and listed from the terminal but not removed, so retention housekeeping still meant the UI or a raw `kubectl delete` that skips Everest's auth and RBAC. This adds `backup delete --name --namespace`, rounding out create/list/delete for the backup group.

The confirmation prompt follows the backup's own deletion policy rather than being fixed: `Delete` (the default, which also wipes the data in storage) asks you to type the name back; `Retain` is just a y/N. If the policy can't be verified, lookup failed, or RBAC allows delete but not read — it falls back to the stricter type-the-name tier and never states an unverified policy or storage name as fact.

A backup that's `Pending`, `Running`, or read successfully but has no status yet (right after `create`, arguably the most fragile moment) refuses to delete unless `--force` is passed. Force only lifts that guard, it doesn't skip confirmation on its own, those are separate switches. `Error` is never guarded, since deleting a stuck backup is usually how you clear it.

`--ignore-not-found` skips both the prompt and `--wait` for an already-gone backup; knowing it's gone skips confirmation either way, even without that flag. `--wait`/`--timeout` polls until it's actually  gone, and the progress line says what's happening rather than printing the raw status word — one of those words is literally `Succeeded`, which reads wrong mid-wait.

Everything routes through the shared `confirm`/`deletion`/`wait`/ `waitcmd` packages. `checkDeleteResponse` moved into `deletion.CheckDeleteResponse`. 
Refer to #2847 

Left out on purpose: no `--deletion-policy` override at delete time. The policy lives on the Backup and is mutable there, so changing it is an update, not a delete option and it freezes once deletion starts, so overriding it here could race with itself.


VijetaPriya47:
@recharte Could you check it out?

pkg/cli/deletion/deletion.go
```diff
@@ -53,6 +54,25 @@ func GonePoll[T any](
 	return wait.FetchPoll(kind, name, wait.NotFoundIsSuccess, fetch)
 }
 
+// CheckDeleteResponse maps a DELETE response's status to (alreadyGone, error):
+// 404 is --ignore-not-found success or "not found"; other non-2xx uses errBody.
+func CheckDeleteResponse(status int, statusText, kind, name, namespace string, ignoreNotFound bool, errBody func() (string, bool)) (bool, error) {
```
recharte, line 57: Both `pkg/cli/instance/delete.go:146` and `pkg/cli/backupstorage/delete.go:126` still define their own private `checkDeleteResponse` at head, and the only caller of this one is `pkg/cli/backup/delete.go:112`. So this lands as a third copy rather than replacing the two — the opposite of what commit 5d5a313's message and the PR description both say ("got backported onto `instance` and `backup-storage` delete too, rather than writing a third copy of it").

Migrating them looks mechanical; the `errBody` closure already covers backup-storage's extra `JSON400`:

```go
alreadyGone, err := deletion.CheckDeleteResponse(resp.StatusCode(), resp.Status(), "backup storage", opts.Name, opts.Namespace, opts.IgnoreNotFound,
	func() (string, bool) { return clienterr.Message(resp.JSON400, resp.JSONDefault) })
```

If you'd rather keep the migration out of this PR, that works too — but then the description and the commit message shouldn't claim it happened.

pkg/cli/backup/delete.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	if b.Status != nil && b.Status.State != nil {
+		return *b.Status.State, true
+	}
+	return "", true
+}
+
+// inFlight reports if the backup should block a delete: Pending/Running,
+// or read successfully with no status yet; a failed/ambiguous fetch never is.
+func inFlight(state string, ok bool) bool {
```
recharte, line 172: This matches #2806, which excludes `Error` on the grounds that "deleting a stuck operation is the normal remediation". Reading `api/backup/v1alpha1/backup_types.go:144-147`, though, that description fits `BackupStateFailed` ("failed terminally"); `BackupStateError` is documented as "a transient error; the controller may retry" — i.e. potentially still in flight, which is what the guard exists for.

Was `Error` chosen deliberately over `Failed` in the issue, or is this worth guarding too (with `--force` as the escape hatch either way)? Happy for it to stay as-is if the intent was to cover both retryable and terminal failures.

pkg/cli/deletion/deletion.go
recharte, line 1: The package doc lists what lives here one by one ("the 'gone' `wait.Condition`, the GET-until-404 poll shape, and the three output lines"), which is a maintenance burden that has already drifted: `CheckDeleteResponse` is a fourth thing, and it isn't `--wait` plumbing at all, so the "every `<resource> delete --wait` implementation" framing no longer holds either. `go doc` already lists the exported members. Worth dropping the inventory rather than extending it:

```go
// Package deletion holds the plumbing shared by every `<resource> delete`
// command.
package deletion
```

pkg/cli/backup/wait.go
```diff
@@ -62,3 +69,18 @@ func backupFailureMessage(b *client.Backup) string {
...
+func newBackupDeletePoll(
+	c *client.ClientWithResponses,
+	cluster, namespace, name string,
+) wait.PollFunc[*client.Backup] {
+	return deletion.GonePoll("backup", name, fetchBackup(c, cluster, namespace, name))
+}
+
+func deleteCondition(b *client.Backup) (wait.Outcome, string) {
+	return deletion.GoneCondition("backup deleted", func(v *client.Backup) string {
+		return "backup still exists (state: " + backupState(v) + ")"
```
recharte: `backupState` is a table cell renderer: it returns `"-"` when there's no status, so a backup deleted before anything synced back prints `backup still exists (state: -)`. Reachable via `--force --wait` right after `create`, which is the case the previous commit added the no-status-yet guard for. Worth dropping the parenthetical when the state is unknown rather than showing the dash.

recharte (changes requested):
Unrelated to this change, but it came up while reading it: the state literals in `pkg/cli/backup` exist only because `BackupState` carries no enum marker, so the generated client emits `*string` (#2891), and the `.(string)` in `policyFromBackup` only because `BackupDeletionPolicy` carries one twice, which makes it emit `interface{}` (#2890). Nothing to do here.

pkg/cli/deletion/deletion.go
VijetaPriya47: I thought #2847 will get merged first, so according to that I wrote.

pkg/cli/backup/delete.go
VijetaPriya47: yes, I reasoned  "Failed terminally" means the controller is done, nothing further will happen to that backup on its own. So blocking a delete on Failed wouldn't protect anything, there's no in-progress work to interrupt.  Error's doc comment says "a transient error; the controller may retry,"  that work might still be happening, so I've added it to the guard.

There's also existing precedent for this split in the same file: backupCondition (the create side wait condition in wait.go) already treats Error as non terminal via its default branch, with a comment saying almost the same thing . The delete side guard was out of sync with that, now it isn't. I 've added `state == backupStateError` check in the inFlight as well as in Run.
Also, added backupStateError = "Error" to pkg/cli/backup/create.go's state block (it existed before only as a bare string literal in a test, not as a named constant).

--force still overrides it, same as Pending/Running, and the refusal message now names Error specifically so it's clear why it's blocked. WDYT?

pkg/cli/deletion/deletion.go
VijetaPriya47: I have fixed the wording tp make it more simple. This is happening because I copied the deletion .go straight up from the restore-list, so that merging can be done on top of restore-delete smoothly

pkg/cli/backup/wait.go
VijetaPriya47: fixed it and added a test too :>

pkg/cli/backup/delete.go
recharte: Agreed, I'll update #2806 accordingly so the two don't disagree.

recharte (commented):
Looks good as well but you'll have some conflicts to fix after merging #2847.

recharte (approved):
🚀

PR #2872, opened 2026-08-11
fix: correct zero disk capacity reporting for GKE, OpenShift, and Generic clusters

### LINKED ISSUE #2855: bug: GetAllClusterResources returns 0 disk capacity for GKE and OpenShift clusters
### What is the issue?

GetAllClusterResources in pkg/kubernetes/resources.go silently returns 
diskSizeBytes = 0 for all users running OpenEverest on GKE, OpenShift, 
and Generic Kubernetes clusters.

What happened:
The getResourcesFromNodes function hits a switch case for ClusterTypeGKE, 
ClusterTypeOpenShift, and ClusterTypeGeneric that contains only a 
// TODO support other cluster types comment and a continue statement, 
skipping disk calculation entirely. As a result, the API reports 0 available 
storage with no error returned to the caller.

What I expected:
Disk capacity should be computed from node.Status.Allocatable 
[corev1.ResourceEphemeralStorage] for each worker node and summed — 
exactly as already implemented for ClusterTypeMinikube — then returned 
correctly through Get

### Steps to Reproduce

1. Deploy OpenEverest on a GKE or OpenShift cluster.

2. Navigate to the database cluster creation page in the UI
   or call GET /v1/resources directly from the API.

3. Observe that the available disk storage is reported as 0.

4. Cross-check node allocatable resources via:
   kubectl get nodes -o jsonpath=
   '{.items[*].status.allocatable.ephemeral-storage}'
   — this will show non-zero values, confirming the bug is in
   Everest's resource calculation, not the cluster itself.

### Relevant Logs / Output

No runtime error or panic is thrown — the bug is a silent wrong value.

Confirm the bug by checking node allocatable storage directly:

```bash
kubectl get nodes -o jsonpath='{.items[*].status.allocatable.ephemeral-storage}'
```

Example output (non-zero values from real GKE nodes):
```
98831908Ki 98831908Ki 98831908Ki
```

But OpenEverest API reports:
```json
GET /v1/resources

{
  "diskSize": 0,
  "cpuMillis": 12000,
  "memoryBytes": 48318382080
}
```

Root cause in source:
```go
// pkg/kubernetes/resources.go
case ClusterTypeGeneric, ClusterTypeOpenShift, ClusterTypeGKE:
    // TODO support other cluster types
    continue
```

The continue skips diskSizeBytes accumulation entirely for these
cluster types, so 0 is always returned — no

### OpenEverest Version

v1.0.0

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

Cloud Provider:  GKE (Google Kubernetes Engine) / OpenShift / Generic K8s
Kubernetes Version: v1.27+ (reproducible on any version)
OpenEverest Version: latest (main branch)
OS: Linux (standard GKE/OpenShift node OS)
Affected Cluster Types: ClusterTypeGKE, ClusterTypeOpenShift, ClusterTypeGeneric
Unaffected Cluster Types: ClusterTypeEKS, ClusterTypeMinikube

Note: This is not environment-specific — it is a code-level bug that
affects ALL users on GKE, OpenShift, and Generic cluster types
regardless of Kubernetes version, node OS, or cloud provider region.
The bug is confirmed directly in pkg/kubernetes/resources.go at the
getResourcesFromNodes switch statement.

### Possible Solution

Replace the `continue` in `getResourcesFromNodes` inside
`pkg/kubernetes/resources.go` with an ephemeral storage lookup
using the same pattern already implemented for ClusterTypeMinikube:

```go
case ClusterTypeGeneric, ClusterTypeOpenShift, ClusterTypeGKE:
    storage, ok := node.Status.Allocatable[corev1.ResourceEphemeralStorage]
    if !ok {
        continue
    }
    bytes, err := convertors.StrToBytes(storage.String())
    if err != nil {
        return 0, 0, 0, 0, errors.Join(err,
            fmt.Errorf("could not convert storage size '%s' to bytes",
                storage.String()))
    }
    diskSizeBytes += bytes
```

Exact files to change:

1. pkg/kubernetes/resources.go
   - Replace // TODO support other cluster types + continue
     with the allocatable ephemeral storage lookup above.

2. pkg/kubernetes/resources_test.go
   - Add table-driven test cases for ClusterTypeGKE and
     ClusterTypeOpenShift asserting non-zero diskSizeBytes
     is returned when nodes advertise ephemeral-storage
     in Allocatable.
   - Add a test case asserting graceful handling when
     ephemeral-storage key is missing from Allocatable.

3. Remove the // TODO support other cluster types comment
   once implemented.

No new dependencies are required — convertors.StrToBytes
is already used in the same file for Minikube disk parsing.
I am happy to open a PR with this fix if a maintainer assigns me.

### Additional Context

This bug silently affects the majority of OpenEverest users since
GKE, OpenShift, and Generic are the three most commonly deployed
Kubernetes targets outside of AWS/EKS.

Key points worth noting:

1. Silent failure — no error, no warning, no log line is emitted.
   Users have no way of knowing the value is wrong unless they
   manually cross-check kubectl node allocatable values.

2. Downstream impact — diskSizeBytes = 0 flows directly into the
   API response and the UI resource display, meaning users see
   0 available storage on the database cluster creation page.
   This can block provisioning entirely or cause users to
   incorrectly believe their cluster has no available disk.

3. Fix is low risk — the solution reuses convertors.StrToBytes
   and corev1.ResourceEphemeralStorage, both already present in
   the same file. No new dependencies, no architectural changes.

4. Only EKS and Minikube are currently unaffected since they
   have dedicated disk-size logic implemented. All other cluster
   types fall through the TODO continue block.

5. Related TODO — the same // TODO support other cluster types
   comment appears in at least one other location in resources.go,
   suggesting this was a known gap left for a future contributor
   to address.

6. No screenshots available since the bug produces an incorrect
   numeric value (0) rather than a visual error or crash. The
   API JSON response and kubectl output comparison in the Logs
   section above serves as the equivalent evidence.

I am happy to fix all affected locations in the same PR if
a maintainer confirms the scope.

Harkirat1309:
Fixes #2855

### What is the issue?
Currently, `GetAllClusterResources` silently returns a `diskSizeBytes` of `0` for users running OpenEverest on GKE, OpenShift, and Generic Kubernetes clusters. The `getResourcesFromNodes` function was bypassing disk calculation entirely for these cluster types, causing the API to report 0 available storage without throwing any errors.

### What did this PR do?
- Implemented the `ephemeral-storage` lookup in `getResourcesFromNodes` for `ClusterTypeGKE`, `ClusterTypeOpenShift`, and `ClusterTypeGeneric`. The logic now parses `node.Status.Allocatable[corev1.ResourceEphemeralStorage]` and accumulates the space, mirroring the stable behavior used by `ClusterTypeMinikube`.
- Removed the stale `// TODO support other cluster types` comments.
- Added table-driven tests in `pkg/kubernetes/resources_test.go` to assert non-zero `diskSizeBytes` return correctly for these environments, as well as gracefully handle cases where the `ephemeral-storage` key is missing.

### Testing
- `TestGetResourcesFromNodesDiskSize` added and passing.

### DIFF
--- a/pkg/kubernetes/resources.go
+++ b/pkg/kubernetes/resources.go
@@ -80,8 +80,17 @@ func (k *Kubernetes) getResourcesFromNodes(ctx context.Context, clusterType Clus
 		case ClusterTypeUnknown:
 			return 0, 0, 0, 0, errors.New("unknown cluster type")
 		case ClusterTypeGeneric, ClusterTypeOpenShift, ClusterTypeGKE:
-			// TODO support other cluster types
-			continue
+			storage, ok := node.Status.Allocatable[corev1.ResourceEphemeralStorage]
+			if !ok {
+				continue
+			}
+			bytes, err := convertors.StrToBytes(storage.String())
+			if err != nil {
+				return 0, 0, 0, 0, errors.Join(err,
+					fmt.Errorf("could not convert storage size '%s' to bytes",
+						storage.String()))
+			}
+			diskSizeBytes += bytes
 		case ClusterTypeMinikube:
 			bytes, err := k.getMinikubeDiskSizeBytes(node)
 			if err != nil {
@@ -216,7 +225,6 @@ func (k *Kubernetes) GetConsumedDiskBytes(
 	case ClusterTypeUnknown:
 		return 0, errors.New("unknown cluster type")
 	case ClusterTypeGeneric, ClusterTypeOpenShift, ClusterTypeGKE:
-		// TODO support other cluster types.
 		return 0, nil
 	case ClusterTypeMinikube:
 		// Minikube does not seem to have support for this.
--- a/pkg/kubernetes/resources_test.go
+++ b/pkg/kubernetes/resources_test.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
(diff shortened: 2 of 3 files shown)


Harkirat1309:
@recharte please review once!

recharte:
Thanks for the work here @Harkirat1309, including the tests. The direction was declined on the issue — see https://github.com/openeverest/openeverest/issues/2855#issuecomment-5254399463 for the reasoning. This is a call about what the capacity figure should measure on network-storage clusters, not about the quality of this patch.

PR #2882, opened 2026-08-11
Add regression test for tokenStore.Add update failure

Anuragkumar-687:
This is the follow-up to #2769 that we discussed in #2766.

The fix for this bug is already on main (it went in with #2086), so there's no code change here. The actual gap was on the test side: pkg/session never had a case where the blocklist Secret update fails, so the fake client's writes always succeeded and the suite stayed green. That's how the original bug shipped, and also how the missing backport to release-2.0 went unnoticed for a while. This just adds the test so main is covered too.

The test makes the Secret update return a conflict, then checks two things: that Add returns that error instead of swallowing it, and that the token didn't actually make it into the Secret. It reads the Secret back through StringData because the fake client doesn't do the StringData -> Data conversion the real API server does (same reason TestBlocklist_Block checks it that way).

Ran go test ./pkg/session/... and go test -race ./pkg/session/... locally, both pass.

Part of #2766. The release-2.0 side (the fix plus this same test) is in #2769.

### DIFF
--- a/pkg/session/blocklist_test.go
+++ b/pkg/session/blocklist_test.go
@@ -28,8 +28,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


Anuragkumar-687:
Hi @recharte! The PR is ready for review whenever you get a chance.

Anuragkumar-687:
Hi @recharte, gentle reminder on this PR when you have a moment. It’s ready for review and all checks are passing. Thanks!

PR #2895, opened 2026-08-11
fix: remove obsolete uninstall environment detection

### LINKED ISSUE #2888: fix: honor --skip-env-detection in everestctl uninstall
## What is the issue?

The `--skip-env-detection` flag for `everestctl uninstall` does not actually skip Kubernetes environment detection.

In `pkg/cli/uninstall/uninstall.go`, the environment-detection guard currently uses an inverted condition:

```go
if !u.config.SkipEnvDetection {
    return nil
}
```

As a result, environment detection is skipped when the flag is **not** provided, while it is still performed when `--skip-env-detection` is explicitly used.

This is inconsistent with the intended behavior of the flag and with the corresponding `SkipEnvDetection` handling in the other CLI commands.

## Steps to reproduce

1. Run `everestctl uninstall` without `--skip-env-detection`.
2. Observe that Kubernetes environment detection is skipped.
3. Run `everestctl uninstall --skip-env-detection`.
4. Observe that environment detection is still performed.

## Expected behavior

- Without `--skip-env-detection`, Kubernetes environment detection should run normally.
- With `--skip-env-detection`, Kubernetes environment detection should be skipped.

## Actual behavior

The behavior is inverted:

- Without the flag → environment detection is skipped.
- With the flag → environment detection is performed.

## Suggested fix

Update the condition in `pkg/cli/uninstall/uninstall.go` to follow the intended `SkipEnvDetection` semantics used by the other CLI commands.

A regression test should also be added to verify that environment detection is only invoked when the flag is not set.

## Environment

- OpenEverest: `release-2.0`

Adii-45:
## Summary

Fixes #2888 by removing the obsolete Kubernetes environment detection logic from `everestctl uninstall`!

The uninstall flow no longer requires Kubernetes environment detection, so the unused detection path has been removed while keeping `--skip-env-detection` registered as a deprecated no-op for backward compatibility..

## Changes

- Removed the unused `clusterType` field from the uninstall configuration
- Removed the obsolete Kubernetes environment detection and `setKubernetesEnv` logic
- Kept `--skip-env-detection` as a deprecated no-op for backward compatibility
- Added regression tests for the uninstall configuration and deprecated flag behavior

## Verification

- `gofmt` passes
- `go build` passes
- `go vet` passes
- Focused uninstall and command tests pass
- Full relevant package tests pass
- `golangci-lint` passes on the affected packages

This PR applies the same fix to `main` as the corresponding `release-2.0` change..!

### DIFF
--- a/commands/uninstall.go
+++ b/commands/uninstall.go
@@ -47,6 +47,7 @@ func init() {
 	uninstallCmd.Flags().BoolVarP(&uninstallCfg.AssumeYes, "assume-yes", "y", false, "Assume yes to all questions")
 	uninstallCmd.Flags().BoolVarP(&uninstallCfg.Force, "force", "f", false, "Force removal in case there are database clusters running")
 	uninstallCmd.Flags().BoolVar(&uninstallCfg.SkipEnvDetection, cli.FlagSkipEnvDetection, false, "Skip detecting Kubernetes environment where Everest is installed")
+	_ = uninstallCmd.Flags().MarkDeprecated(cli.FlagSkipEnvDetection, "no longer has any effect")
 }
 
 func uninstallPreRun(_ *cobra.Command, _ []string) {
--- a/commands/uninstall_test.go
+++ b/commands/uninstall_test.go
@@ -1,0 +1,45 @@
+// everest
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


Adii-45:
The `CLI tests (test-namespaces)` check is currently failing during the K3d cluster setup, before the CLI build/tests run.. It appears unrelated to the changes in this PR.. I’ll keep an eye on it and rerun if needed!

PR #2900, opened 2026-08-11
fix(provider-runtime): don't skip Sync/Status when only backup config is invalid

### LINKED ISSUE #2898: provider-runtime: backup limits violation on a BackupClass silently freezes ALL reconciliation of affected Instances, not just backup config
### What is the issue?

`ProviderReconciler.Reconcile` (`provider-runtime/reconciler/provider.go`) short-circuits the entire reconcile loop when an `Instance`'s backup configuration violates its `BackupClass` limits it returns before ever calling `provider.Sync()` or `provider.Status()`. This contradicts the code's own stated intent, and it freezes reconciliation of things that have nothing to do with backups.

```go
// provider-runtime/reconciler/provider.go, Reconcile()
if err := validateInstanceBackupConfig(ctx, r.Client, in); err != nil {
    logger.Error(err, "Backup configuration validation failed")
    // Surface the violation on the BackupConfigured condition without
    // marking the Instance Failed: the engine itself is healthy and the
    // user can fix the configuration without a full redeploy.
    reason := controller.LimitsExceededReason
    if errors.Is(err, controller.ErrPITRConfigInvalid) {
        reason = controller.PITRConfigInvalidReason
    }
    setCondition(in, v1alpha1.ConditionBackupConfigured, metav1.ConditionFalse,
        reason, err.Error(), metav1.Now())
    if updateErr := r.Client.Status().Update(ctx, in); updateErr != nil {
        logger.Error(updateErr, "Failed to update status after backup config violation")
    }
    return reconcile.Result{}, nil   // <-- returns here, provider.Sync() is never called
}
```

The comment says the point of this branch is that "the engine itself is healthy and the user can fix the configuration without a full redeploy"  i.e. only backup wiring should be affected. But the `return` happens before`r.provider.Validate`, `resolveVersionBundle`,`r.provider.Sync`, and`r.provider.Status` ever run. In practice this means: once an `Instance`trips this check, the provider stops applying **any** further spec changesto the underlying engine — scaling, version upgrades, resource changes, everything — not just backup configuration, until the violation is resolved. That's a materially different (and undocumented) behaviorfrom
what the comment promises.

### Steps to Reproduce

This doesn't require a bad user edit to the `Instance` — a `BackupClass` is cluster-scoped and shared, so tightening it after the fact is enough to freeze a previously-healthy `Instance` that never itself changed:

1. Create a `BackupClass` with `spec.providerManaged.limits.maxStorages: 3` (or similar) and an `Instance` with backup enabled and 3 storages configured — reconciles fine, `Sync`/`Status` run normally.
2. Later, an admin tightens the `BackupClass` to `limits.maxStorages: 2` (a legitimate policy change, unrelated to any specific `Instance`).
3. On the next reconcile of the existing `Instance` (e.g. triggered by any unrelated resync, or a genuinely wanted spec change like bumping `replicas`), `validateInstanceBackupConfig` now fails.
4. The reconciler sets `BackupConfigured=False` and returns — but the `replicas` bump (or any other pending spec change) from step 3 is silently never applied, because `Sync()` never runs. The `Instance` looks "stuck" with no clear signal that engine-level reconciliation
itself (not just backup config) is the thing that's paused.

### Relevant Logs / Output

_No response_

### OpenEverest Version

release-2.0

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Don't let a backup-config violation block the rest of reconciliation. Two directions, not fully evaluated against each other yet:

1. Move the `validateInstanceBackupConfig` check to run *after*  `provider.Sync()`/`provider.Status()` (or in parallel, informationally  so the engine itself keeps converging on its own spec regardless of   backup-limit state, and only the backup-specific wiring inside `Sync()`
is expected to skip/no-op when the condition is `False`.

2. Keep the early check, but only skip the backup-specific portion of reconciliation, not the whole loop — e.g. pass the violation down via
`Context` so a `BackupProvider` implementation can consult it and act  accordingly, while the base engine `Sync()` still runs.

Either way, the comment's stated intent ("the engine itself is healthy... without a full redeploy") should become true, not just documented as if it were.

### Additional Context

Found while reading `provider-runtime/` end-to-end in preparation for building a new provider on this SDK (the OpenEverest Object Storage Provider project, #2255) not filed against a specific consumer, since none exists yet.

Bhupesh-081:
### What does this PR do?

`ProviderReconciler.Reconcile` (`provider-runtime/reconciler/provider.go`) returned early whenever an `Instance`'s backup configuration violated its `BackupClass` limits (`maxStorages`, `maxPITREnabledStorages`, `maxSchedulesPerStorage`) — before ever calling `provider.Sync()` or `provider.Status()`. This contradicted the code's own comment, which says the point of surfacing the violation on the `BackupConfigured` condition (instead of marking the Instance `Failed`) is that "the engine itself is healthy and the user can fix the configuration without a full redeploy." In practice, once an Instance tripped this check, **all** further engine reconciliation (scaling, upgrades, resource/config changes — not just backup wiring) silently stopped until the violation was resolved.

This is reachable without any change to the `Instance` itself: a `BackupClass` is cluster-scoped and shared, so tightening its limits after the fact freezes every previously-healthy `Instance` that now exceeds them, with no clear signal that engine-level reconciliation (not just backup config) is what's paused.

### Fix

- Stop returning early on a backup-config violation. The `BackupConfigured` condition is still set to `False` with the same reason/message as before, but `provider.Validate/Sync/Status` now still run afterward, so the engine keeps converging on the rest of its spec.
- The later step that unconditionally clears `BackupConfigured` back to `True` after a successful `Sync` now skips doing so when this reconcile's own limits check found a live violation — otherwise it would immediately mask the condition we just set.

No change to the validation rules themselves (`ValidateInstanceBackupAgainstClass`, `ValidateInstanceBackupPITRParameters`) or to the admission webhook path, which already correctly rejects a *new* violating write at admission time — this only affects reconciliation of an already-persisted `Instance` whose backup config becomes invalid out from under it (e.g. via a `BackupClass` policy change).

### Test plan

- [x] Added `TestReconcile_BackupLimitsViolation_StillRunsSync` (`provider-runtime/reconciler/backup_limits_sync_test.go`), using the existing fake-client test harness (`newTestScheme` from `cascade_delete_test.go`). Constructs an `Instance` with two backup storages against a `BackupClass` with `maxStorages: 1`, and asserts `Sync()`/`Status()` are still invoked despite the violation, and that the `BackupConfigured` condition is `False` with reason `LimitsExceeded`.
- [x] `go build ./provider-runtime/...`
- [x] `go test ./provider-runtime/...`
- [x] `go test -race ./provider-runtime/...`
- [x] `make copyright-check BASE_BRANCH=release-2.0`
- [x] `make format` / `go mod tidy` — no diff
- [x] `go tool golangci-lint run --new-from-rev=upstream/release-2.0`

No concrete provider implements this SDK yet (verified via `grep -r "BaseProvider"` / `grep -r "reconciler.New("` — zero hits outside `provider-runtime/` itself), so this was found and fixed by reading the reconcile control flow directly, not by reproducing against a running provider.

Fixes #2898

Signed-off-by: Bhupesh Nandardhane <nandardhanebhupesh@gmail.com>

### DIFF
--- a/provider-runtime/reconciler/backup_limits_sync_test.go
+++ b/provider-runtime/reconciler/backup_limits_sync_test.go
@@ -54,6 +54,26 @@ func (p *syncTrackingProvider) Status(*controller.Context) (controller.Status, e
 	return controller.Ready(), nil
 }
 
+// The four methods below satisfy controller.BackupProvider so that the
+// reconciler's "skip clearing BackupConfigured on a live violation" guard —
+// which only applies to BackupProvider implementations — is actually
+// exercised by this test instead of being a no-op type assertion.
+func (p *syncTrackingProvider) SyncBackup(*controller.Context, *backupv1alpha1.Backup) (controller.BackupExecutionStatus, error) {
+	return controller.BackupExecutionStatus{}, nil
+}
+
+func (p *syncTrackingProvider) SyncRestore(*controller.Context, *backupv1alpha1.Restore) (controller.RestoreExecutionStatus, error) {
+	return controller.RestoreExecutionStatus{}, nil
+}
+
+func (p *syncTrackingProvider) CleanupBackup(*controller.Context, *backupv1alpha1.Backup) (bool, error) {
+	return true, nil
+}
+
+func (p *syncTrackingProvider) CleanupRestore(*controller.Context, *backupv1alpha1.Restore) (bool, error) {
+	return true, nil
+}
+
 func TestReconcile_BackupLimitsViolation_StillRunsSync(t *testing.T) {
 	t.Parallel()
 	scheme := newTestScheme()
--- a/provider-runtime/reconciler/backup_limits_sync_test.go
+++ b/provider-runtime/reconciler/backup_limits_sync_test.go
@@ -1,0 +1,139 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package reconciler
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	backupv1alpha1 "github.com/openeverest/openeverest/v2/api/backup/v1alpha1"
+	common "github.com/openeverest/openeverest/v2/api/common/v1alpha1"
+	corev1alpha1 "github.com/openeverest/openeverest/v2/api/core/v1alpha1"
+	"github.com/openeverest/openeverest/v2/provider-runtime/controller"
+)
+
+// syncTrackingProvider is a providerAdapter stub that records whether Sync
+// and Status were invoked, so tests can assert the reconciler still
+// converges the engine even when backup config validation fails.
+type syncTrackingProvider struct {
+	syncCalled   bool
+	statusCalled bool
+}
+
+func (p *syncTrackingProvider) Name() string                       { return "test-provider" }
+func (p *syncTrackingProvider) Types() func(*runtime.Scheme) error { return nil }
+func (p *syncTrackingProvider) Validate(*controller.Context) error { return nil }
+func (p *syncTrackingProvider) Cleanup(*controller.Context) error  { return nil }
+func (p *syncTrackingProvider) Sync(*controller.Context) error {
+	p.syncCalled = true
+	return nil
+}
+
+func (p *syncTrackingProvider) Status(*controller.Context) (controller.Status, error) {
+	p.statusCalled = true
+	return controller.Ready(), nil
+}
+
+func TestReconcile_BackupLimitsViolation_StillRunsSync(t *testing.T) {
+	t.Parallel()
+	scheme := newTestScheme()
+
+	maxStorages := int32(1)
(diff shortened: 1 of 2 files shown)


provider-runtime/reconciler/provider.go
```diff
@@ -388,21 +388,22 @@ func (r *ProviderReconciler) Reconcile(ctx context.Context, req reconcile.Reques
...
+		if errors.Is(backupConfigErr, controller.ErrPITRConfigInvalid) {
 			reason = controller.PITRConfigInvalidReason
 		}
 		setCondition(in, v1alpha1.ConditionBackupConfigured, metav1.ConditionFalse,
-			reason, err.Error(), metav1.Now())
-		if updateErr := r.Client.Status().Update(ctx, in); updateErr != nil {
-			logger.Error(updateErr, "Failed to update status after backup config violation")
-		}
-		return reconcile.Result{}, nil
+			reason, backupConfigErr.Error(), metav1.Now())
```
areebahmeddd, line 406: dropping the status write here makes this best-effort. there are four returns before the final `Status().Update()` that won't flush it, including `WaitError`, which happens on every reconcile while a provider is waiting on the engine. so `BackupConfigured=False` would never get persisted

simplest fix is to keep the existing `Status().Update` and just remove the `return`. `Update` refreshes `resourceVersion` in place, so the final write still works

provider-runtime/reconciler/backup_limits_sync_test.go
```diff
@@ -0,0 +1,139 @@
...
+	backupv1alpha1 "github.com/openeverest/openeverest/v2/api/backup/v1alpha1"
+	common "github.com/openeverest/openeverest/v2/api/common/v1alpha1"
+	corev1alpha1 "github.com/openeverest/openeverest/v2/api/core/v1alpha1"
+	"github.com/openeverest/openeverest/v2/provider-runtime/controller"
+)
+
+// syncTrackingProvider is a providerAdapter stub that records whether Sync
+// and Status were invoked, so tests can assert the reconciler still
+// converges the engine even when backup config validation fails.
+type syncTrackingProvider struct {
```
areebahmeddd, line 38: `syncTrackingProvider` doesn't implement `controller.BackupProvider`, so the check at line 461 is always false and the `&& backupConfigErr == nil` guard isn't actually covered. 

could you add the four methods (`SyncBackup`, `SyncRestore`, `CleanupBackup`, `CleanupRestore`) so the test really exercises the "don't mask the condition we just set" behaviour?

Bhupesh-081: Sure I will add the remaining methods

provider-runtime/reconciler/provider.go
Bhupesh-081: Got it !

Issue #2891, opened 2026-08-11
`BackupState`/`RestoreState` have no enum marker, so `status.state` generates as `*string`


recharte:
### Task Description

Neither state type carries a `+kubebuilder:validation:Enum` marker:

- `BackupState` — `api/backup/v1alpha1/backup_types.go:134` (`Pending`, `Running`, `Succeeded`, `Failed`, `Error`, `Deleting`)
- `RestoreState` — `api/backup/v1alpha1/restore_types.go:108` (`Pending`, `Running`, `Succeeded`, `Failed`, `Error`)

So the CRD describes `status.state` as a bare `type: string`, and oapi-codegen emits `State *string` on the generated client rather than a named enum.

Verified on `release-2.0` for `BackupState`: adding the marker and running `make gen-crds-manifests && make gen-crds-openapi && go generate ./hack/...` produces

```go
type BackupStatusState string
const (
	BackupStatusStateDeleting  BackupStatusState = "Deleting"
	BackupStatusStateError     BackupStatusState = "Error"
	BackupStatusStateFailed    BackupStatusState = "Failed"
	BackupStatusStatePending   BackupStatusState = "Pending"
	BackupStatusStateRunning   BackupStatusState = "Running"
	BackupStatusStateSucceeded BackupStatusState = "Succeeded"
)
func (e BackupStatusState) Valid() bool
```

with `State *BackupStatusState` on the struct.

### Decision needed before this lands

Unlike the `deletionPolicy` markers in #2890, this change is not behaviour-neutral.

`.status.state` is written by provider implementations that live outside this repository. `BackupExecutionStatus.State` (`provider-runtime/controller/common.go:734`) is part of the provider SDK, and `provider-runtime/reconciler/backup.go:221` writes whatever the provider returns straight into `.status.state` — same for restores at `reconciler/restore.go:226`. The Go type is advisory: `BackupState("Uploading")` compiles fine.

Add the enum and the API server starts rejecting those status updates, so a provider reporting a non-canonical state fails its reconcile on every pass. The breakage surfaces at runtime, in someone else's repository, rather than at compile time in ours.

That may well be what we want — enforcing the enum is the point of having one, and the state vocabulary is meant to be closed. But it's a provider-SDK compatibility decision, so it should be taken deliberately and probably announced before the marker lands. Worth settling at the same time whether `Deleting` belongs in `RestoreState` too, for symmetry with `BackupState`.

### Reasoning / Benefit

[...]

PR #2901, opened 2026-08-11
fix: drop duplicate Enum markers so CRD schemas keep type string`n`nD…

### LINKED ISSUE #2890: Duplicate enum markers drop `type: string` from CRD schemas, so the generated client emits `interface{}`
### Task Description

Three fields carry `+kubebuilder:validation:Enum` twice — once on the struct field and once on the named type. controller-gen merges the two into `allOf: [enum, enum]` and drops `type: string` entirely, and oapi-codegen turns a schema with no `type` into `interface{}`.

| type | field marker (delete) | type marker (keep) |
|---|---|---|
| `BackupDeletionPolicy` | `api/backup/v1alpha1/backup_types.go:67` | `backup_types.go:77` |
| `InstanceDeletionPolicy` | `api/core/v1alpha1/instance_types.go:90` | `instance_types.go:111` |
| restore PITR `type` | `api/backup/v1alpha1/restore_types.go:73` | `restore_types.go:97` |

Affected CRDs: `backup.openeverest.io_backups.yaml`, `backup.openeverest.io_restores.yaml`, `core.openeverest.io_instances.yaml`, `core.openeverest.io_instancepresets.yaml`.

`BackupExecutionMode` (`api/backup/v1alpha1/backupclass_types.go:30`) is the shape to copy: the marker lives on the named type only, which is why it comes through the pipeline as a proper `client.BackupStatusExecutionMode`.

Verified on `release-2.0` for `BackupDeletionPolicy`: after deleting the field-level marker and running `make gen-crds-manifests && make gen-crds-openapi && go generate ./hack/...`, the CRD gets `enum: [Retain, Delete]` alongside `type: string` and `default: Delete`, and the client gains

```go
type BackupSpecDeletionPolicy string
const (
	BackupSpecDeletionPolicyDelete BackupSpecDeletionPolicy = "Delete"
	BackupSpecDeletionPolicyRetain BackupSpecDeletionPolicy = "Retain"
)
func (e BackupSpecDeletionPolicy) Valid() bool
```

### Reasoning / Benefit

Validation is unaffected: the `allOf` already enforces both copies of the enum, so no stored object changes meaning and nothing starts or stops being rejected. What changes is the generated surface — the field stops being `interface{}`, so callers can drop their type assertions, and the UI's generated TypeScript gets a union instead of an untyped field.

The published CRDs are also simply wrong as they stand: `deletionPolicy` ships with no `type`, so any consumer generating from our OpenAPI sees a field with no type at all.

### Additional Context

Callers to update once the client types change, on `release-2.0` today:

- `pkg/cli/instance/delete.go:202` — `inst.Spec.DeletionPolicy.(string)`
- `pkg/cli/backup/create.go` — assigns `opts.DeletionPolicy` into `Spec.DeletionPolicy`
- `internal/server/handlers/k8s/backup.go:52-53` and `internal/server/handlers/k8s/instance.go:70-71` — compare and assign the policy
- the generated UI TypeScript types

Once #2863 merges, `policyFromBackup` in `pkg/cli/backup/delete.go` joins that list. Worth letting #2863 land first: its `policyFromBackup` does `b.Spec.DeletionPolicy.(string)`, and a type assertion on a non-interface operand is a compile error, so merging this one first breaks that branch and costs its author a rebase through generated code.

The sibling problem — `BackupState`/`RestoreState` carry no enum marker at all, so `status.state` generates as `*string` — is tracked separately in #2891. It is deliberately not folded in here: adding those markers tightens validation on a field written by out-of-tree providers, so it needs a compatibility decision first, whereas this issue changes no validation at all.

Harsh63870:
## Summary

Fixes duplicate `+kubebuilder:validation:Enum` markers that caused controller-gen to emit `allOf: [enum, enum]` and drop a clean string enum shape for OpenAPI/client generation. oapi-codegen then typed those fields as `interface{}`, and the UI TypeScript types became awkward intersections instead of simple unions.

This change keeps validation unchanged (the enum is still enforced via the named-type marker) and only fixes the generated surface.

## Root cause

Three fields had Enum markers on **both** the struct field and the named type:

| Type | Field marker (removed) | Type marker (kept) |
|------|------------------------|--------------------|
| `BackupDeletionPolicy` | `api/backup/v1alpha1/backup_types.go` | same file |
| `InstanceDeletionPolicy` | `api/core/v1alpha1/instance_types.go` | same file |
| `PITRType` | `api/backup/v1alpha1/restore_types.go` | same file |

`BackupExecutionMode` already had the correct pattern (marker on the named type only) and was used as the template.

## Changes

- Remove the three **field-level** Enum markers; keep type-level Enum markers and field defaults.
- Regenerate:
  - CRD manifests (`backup` / `restore` / `instance` / `instancepreset`)
  - `api/openapi/crds.gen.yaml`
  - Go clients (`client/crds.gen.go`, `internal/server/api/crds.gen.go`, plus regenerated HTTP client/server stubs)
  - UI OpenAPI types (`ui/api/crds.gen.types.ts`, `ui/api/http-api.types.ts`)
- Update callers for the new typed client fields:
  - `pkg/cli/instance/delete.go` — `policyFromInstance` reads `*InstanceSpecDeletionPolicy` instead of `.(string)`
  - `pkg/cli/backup/create.go` (+ test) — assign `*BackupSpecDeletionPolicy`

k8s handlers already use the typed CRD Go types (`BackupDeletionPolicy` / `InstanceDeletionPolicy`), so they needed no logic change.

## Resulting generated types (examples)

Go client now emits typed enums such as:

- `BackupSpecDeletionPolicy` (`Delete` / `Retain`) + `Valid()`
- `InstanceSpecDeletionPolicy` (`Cascade` / `Orphan`) + `Valid()`
- PITR `type` as typed string enums (e.g. `RestoreSpecDataSourceBackupPitrType`)

UI TS now gets unions like `"Retain" | "Delete"` / `"Cascade" | "Orphan"` / `"date" | "latest"` instead of `string & ((...) & (...))`.

CRD schemas now have `enum` + `type: string` (no duplicated `allOf` enums) for the affected fields.

## Out of scope

- **#2891** (`BackupState` / `RestoreState` missing enum markers → `*string`) — deliberately not included; that tightens validation for provider-written status and needs a separate compatibility decision.
- No intentional validation/behavior change for admitted objects.

## Note for #2863

If #2863 is still open when this merges, its `DeletionPolicy.(string)` assertions will need a small rebase to the typed client fields (type assertion on a non-interface operand becomes a compile error).

## Test plan

- [ ] CI green against `release-2.0`
- [ ] Confirm CRDs for backups/restores/instances/instancepresets show `enum` + `type: string` for `deletionPolicy` / PITR `type` (no `allOf` enum duplication)
- [ ] Confirm generated Go client uses `*BackupSpecDeletionPolicy` / `*InstanceSpecDeletionPolicy` (not `interface{}`)
- [ ] Confirm UI generated types use string unions for those fields
- [ ] Spot-check CLI backup create with `--deletion-policy` and instance delete policy messaging still work

Fixes #2890


Harsh63870:
@recharte, flagging the state of this draft vs #2913.
This PR does cover the full #2890 table: field-level Enum removed on BackupDeletionPolicy, InstanceDeletionPolicy, **and** restore PITRType.
What it does **not** include (so I would not merge this as-is):
- rebase onto current `release-2.0` (#2863’s `policyFromBackup` / `pkg/cli/backup/delete.go`)
- `api-tests/generated/types.ts`
- a small codegen diff (`everest-client.gen.go` swagger blob is noisy here)
#2913 has the DeletionPolicy + CLI/client pieces, is rebased, and CI is green. I’m fine with that landing for #2890. The leftover PITR field marker can be a follow-up.
Happy to close this draft if that’s cleaner.

Harsh63870:
Closing in favour of #2913, which merged and covers #2890.

PR #2903, opened 2026-08-11
fix(cli): isolate auth transport connections

### LINKED ISSUE #2884: Flaky CLI tests: sharing http.DefaultTransport lets httptest servers kill each other's in-flight requests
### Task Description

`pkg/cli/auth/transport.go:147` builds every `everestctl` API client on the process-wide `http.DefaultTransport`:

```go
	}),
)
```

`httptest.Server.Close()` deliberately calls `CloseIdleConnections()` on that same global — [`net/http/httptest/server.go`](https://github.com/golang/go/blob/master/src/net/http/httptest/server.go):

```go
// Not part of httptest.Server's correctness, but assume most
// users of httptest.Server will be using the standard
// transport, so help them out and close any idle connections for them.
if t, ok := http.DefaultTransport.(closeIdleTransport); ok {
	t.CloseIdleConnections()
}
```

Every CLI test that stands up an `httptest` server therefore tears down pooled connections belonging to *every other test running in parallel in the same package*. The victim's in-flight request dies with a transport error instead of reaching the response it asserted on.

Replace `http.DefaultTransport` with a private clone owned by `pkg/cli/auth`, e.g.:

```go
// Private clone: sharing http.DefaultTransport lets other users of it
// (notably httptest.Server.Close) drop our pooled connections mid-request.
var baseTransport = http.DefaultTransport.(*http.Transport).Clone()
```

and use `base: baseTransport`. Connection pooling still works process-wide; `httptest` can no longer reach it. `pkg/cli/auth/transport_test.go` constructs `authTransport{base: http.DefaultTransport}` directly in five places and should move to the same transport.

### Reasoning / Benefit

**CI reliability.** This produces red builds on pull requests that touch none of the affected code. Observed on #2856, a CLI backport that changes only `commands/`, `pkg/cli/namespaces` and `pkg/cli/upgrade` — the failure was in `pkg/cli/provider`:

```
--- FAIL: TestProviderList_ServerError (0.01s)
    list_test.go:133:
        Error: "failed to list providers: Get \"http://127.0.0.1:41935/v1/clusters/main/providers\":
               net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called"
               does not contain "unexpected response"
```

A plain re-run went green, which is the signature of a race rather than a defect in the test's expectation.

**Blast radius is wide and growing.** Seven packages combine `t.Parallel()` with `httptest` servers on the shared transport:

| Package | `t.Parallel()` | `httptest.NewServer` |
|---|---:|---:|
| `pkg/cli/instance` | 83 | 19 |
| `pkg/cli/backupstorage` | 26 | 14 |
| `pkg/cli/auth` | 23 | 13 |
| `pkg/cli/restore` | 20 | 15 |
| `pkg/cli/backup` | 19 | 15 |
| `pkg/cli/backupclass` | 5 | 4 |
| `pkg/cli/provider` | 5 | 4 |

`pkg/cli/provider` simply drew the short straw this time; any of them can lose the race. Because every new API-backed `everestctl` subcommand adds more servers to the same transport, the failure rate increases as the v2 CLI grows.

**The motivation is test isolation, not a production bug.** Nothing else in the `everestctl` process touches `http.DefaultTransport` — `httptest.Server.Close` is the only caller of `CloseIdleConnections` that exists anywhere in the repo, and only in tests. The fix is behaviourally inert in production, since `Transport.Clone` copies `Proxy`, the dial and TLS timeouts and `ForceAttemptHTTP2`. The standing argument for it is the narrower one: a package that makes HTTP calls should own the transport it makes them on, rather than borrow a process-global whose lifecycle anyone can reach.

### Additional Context

- Failing job: https://github.com/openeverest/openeverest/actions/runs/31471202117/job/93714694690 (PR #2856)
- Offending line: `pkg/cli/auth/transport.go:147`
- Direct constructions in tests: `pkg/cli/auth/transport_test.go:77,114,153,179,221`

**Branch note, updated 19 August.** This was filed against `release-2.0`, when `main` still carried v1 and genuinely could not hit the problem: its `pkg/cli` drove Kubernetes and Helm directly, with one `httptest.NewServer` call and no HTTP client to an Everest API. The 18 August swap made `release-2.0` into `main`, so the affected surface — 86 `httptest.NewServer` calls across 22 files, all routing through `authcli.NewAPIClient` — is now on `main`. v1, now on `v1.x`, is still unaffected.

**Not reproducible locally**, as expected for a timing race — `go test ./pkg/cli/provider/... -race -count=50` passes on a developer machine. Any reproduction attempt will need contention (loaded runner, `-cpu 1`, or an artificial delay in the handler).

Found while auditing `main` → `release-2.0` backport gaps (#2856, #2857, #2858, #2859).

luantaraschi:
## Summary

Authentication transports previously shared the global `http.DefaultTransport`. That also shared connection pools between parallel tests and could cause intermittent failures.

This change clones the default transport for each API client and test transport. The regression coverage also verifies that `NewAPIClient` does not reuse the global transport.

Fixes #2884

## Testing

- `go test ./pkg/cli/auth -count=100`
- `go test ./pkg/cli/...`
- `go vet ./pkg/cli/auth`

## AI assistance

This contribution was made with AI assistance. I reviewed the implementation and verified all test results listed above before submitting it.

### DIFF
--- a/pkg/cli/auth/transport.go
+++ b/pkg/cli/auth/transport.go
@@ -81,6 +81,10 @@ type authTransport struct {
 	base   http.RoundTripper
 }
 
+func newDefaultTransport() *http.Transport {
+	return http.DefaultTransport.(*http.Transport).Clone() //nolint:forcetypeassert
+}
+
 func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	tok, err := t.source.token(req.Context())
 	if err != nil {
@@ -144,7 +148,7 @@ func NewAPIClient(cfg Config, l *zap.SugaredLogger, cfgPath, contextName string)
 	c, err := client.NewClientWithResponses(
 		cli.NormalizeServerURL(sess.Server.URL),
 		client.WithHTTPClient(&http.Client{
-			Transport: &authTransport{source: ts, base: http.DefaultTransport},
+			Transport: &authTransport{source: ts, base: newDefaultTransport()},
 		}),
 	)
 	if err != nil {
--- a/pkg/cli/auth/transport_test.go
+++ b/pkg/cli/auth/transport_test.go
@@ -60,6 +60,13 @@ func loadTokenSource(t *testing.T, cfgPath string, l *Login) *tokenSource {
 	}
 }
 
+func newTestAuthTransport(t *testing.T, source *tokenSource) *authTransport {
+	t.Helper()
+	base := newDefaultTransport()
+	t.Cleanup(base.CloseIdleConnections)
+	return &authTransport{source: source, base: base}
+}
+
 func TestTransport_InjectsBearerToken(t *testing.T) {
 	t.Parallel()
 
@@ -74,7 +81,7 @@ func TestTransport_InjectsBearerToken(t *testing.T) {
 	require.NoError(t, newTransportConfig(srv.URL).Save(cfgPath))
 
 	ts := loadTokenSource(t, cfgPath, NewLogin(Config{}, zap.NewNop().Sugar()))
-	tr := &authTransport{source: ts, base: http.DefaultTransport}
+	tr := newTestAuthTransport(t, ts)
 
 	req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet, srv.URL, nil)
 	resp, err := tr.RoundTrip(req)
@@ -111,7 +118,7 @@ func TestTransport_ProactiveRefresh(t *testing.T) {
 	require.NoError(t, cfg.Save(cfgPath))
 
 	ts := loadTokenSource(t, cfgPath, NewLogin(Config{}, zap.NewNop().Sugar()))
-	tr := &authTransport{source: ts, base: http.DefaultTransport}
+	tr := newTestAuthTransport(t, ts)
 
 	req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet, srv.URL, nil)
 	resp, err := tr.RoundTrip(req)
@@ -150,7 +157,7 @@ func TestTransport_On401_RefreshAndRetry(t *testing.T) {
 	require.NoError(t, newTransportConfig(srv.URL).Save(cfgPath))
 
 	ts := loadTokenSource(t, cfgPath, NewLogin(Config{}, zap.NewNop().Sugar()))
-	tr := &authTransport{source: ts, base: http.DefaultTransport}
+	tr := newTestAuthTransport(t, ts)
 
 	req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet, srv.URL, nil)
 	resp, err := tr.RoundTrip(req)
@@ -176,7 +183,7 @@ func TestTransport_On401_RefreshFails_ReturnsErrTokenRefresh(t *testing.T) {
 	require.NoError(t, newTransportConfig(srv.URL).Save(cfgPath))
 
 	ts := loadTokenSource(t, cfgPath, NewLogin(Config{}, zap.NewNop().Sugar()))
-	tr := &authTransport{source: ts, base: http.DefaultTransport}
+	tr := newTestAuthTransport(t, ts)
 
 	req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet, srv.URL, nil)
 	resp, err := tr.RoundTrip(req)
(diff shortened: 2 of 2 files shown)


luantaraschi:
Hi, a nudge on this one, a week in with no CI.

The workflow runs are sitting at `action_required` (API CI, CodeQL, Copyright Headers, Feature build), so nothing has run here beyond DCO and the labeler. I think that is the first time contributor approval gate rather than anything in the change itself.

I have just updated the branch, so it is even with `release-2.0` and has no conflicts. If someone can approve the runs, I will sort out whatever CI turns up. The local test runs are listed in the description.

Thanks!

pkg/cli/auth/transport.go
```diff
@@ -81,6 +81,10 @@ type authTransport struct {
 	base   http.RoundTripper
 }
 
+func newDefaultTransport() *http.Transport {
+	return http.DefaultTransport.(*http.Transport).Clone()
```
recharte: Let's just silence the linter on this one.
```suggestion
	return http.DefaultTransport.(*http.Transport).Clone() //nolint:forcetypeassert
```

recharte (changes requested):
Hi @luantaraschi sorry for the delay. The fix looks good just a small linter thing to fix.

luantaraschi:
Thanks for letting the workflows run. `Check` came back red, and the four findings are outside this branch:

```
pkg/cli/install/install.go:50:2   const pollInterval is unused
pkg/cli/install/install.go:51:2   const pollTimeout is unused
pkg/cli/instance/create_test.go:42:6   func strPtr is unused
pkg/cli/namespaces/add.go:377:26  func (*NamespaceAdder).getValues is unused
```

This PR only touches `pkg/cli/auth/transport.go` and `pkg/cli/auth/transport_test.go`, so none of those symbols are reachable from the change. The linter looks to be running over the whole tree rather than the diff. `Test`, `CLI Integration Tests`, `Controller Integration Tests` and every CodeQL job are green.

Happy to rebase on main if that clears it, or to leave it alone if the lint findings are already tracked elsewhere.

recharte:
> Thanks for letting the workflows run. `Check` came back red, and the four findings are outside this branch:
> 
> ```
> pkg/cli/install/install.go:50:2   const pollInterval is unused
> pkg/cli/install/install.go:51:2   const pollTimeout is unused
> pkg/cli/instance/create_test.go:42:6   func strPtr is unused
> pkg/cli/namespaces/add.go:377:26  func (*NamespaceAdder).getValues is unused
> ```
> 
> This PR only touches `pkg/cli/auth/transport.go` and `pkg/cli/auth/transport_test.go`, so none of those symbols are reachable from the change. The linter looks to be running over the whole tree rather than the diff. `Test`, `CLI Integration Tests`, `Controller Integration Tests` and every CodeQL job are green.
> 
> Happy to rebase on main if that clears it, or to leave it alone if the lint findings are already tracked elsewhere.

@luantaraschi the only thing you need to change is what I described in the comments above plus rebasing/merging from main.

luantaraschi:
Done: suggestion applied and the branch rebased onto main, so it is now a single commit on top of `c35f2dca` with no merge commits from `release-2.0`.

I was wrong about why the linter went wide, and the real reason is worth knowing. The step runs `--new-from-rev=$(git merge-base origin/${{ github.base_ref }} HEAD)`, and for that run `base_ref` is still `release-2.0`, which no longer exists. The log has `fatal: Not a valid object name origin/release-2.0` immediately before golangci-lint starts, so the flag arrived empty and the whole tree got linted: over 350 findings, 141 of them `unused`, including the four I quoted. Re-running the job cannot clear it either, since a re-run replays the original event and keeps the old base ref. That is why attempt 2 failed identically this morning. Any PR whose runs predate the branch rename will behave the same way until it is pushed again.

Might be worth failing that step loudly when the rev does not resolve, otherwise a broken merge-base silently turns a diff lint into a full-tree lint.

Verified locally on the rebased branch:

- `go test ./pkg/cli/auth -count=100`
- `go vet ./pkg/cli/auth`
- `go tool golangci-lint run --new-from-rev=$(git merge-base origin/main HEAD) ./pkg/cli/auth/...`

The new runs are sitting at `action_required`, so they need your approval before they start.

pkg/cli/auth/transport.go
luantaraschi: Applied in 7fa263f4.

recharte (approved):
Thanks @luantaraschi, looks good.

PR #2913, opened 2026-08-12
fix: remove duplicate enum markers from DeletionPolicy fields

### LINKED ISSUE #2890: Duplicate enum markers drop `type: string` from CRD schemas, so the generated client emits `interface{}`

Adii-45:
## Summary

Fix duplicate `+kubebuilder:validation:Enum` markers on `DeletionPolicy` fields that cause generated client types to emit `interface{}` instead of the expected enum/string representation.

## Changes

- Remove the redundant field-level enum markers from `BackupSpecDeletionPolicy` and `InstanceSpecDeletionPolicy`
- Keep the existing type-level enum markers and field defaults unchanged
- Regenerate the affected CRD schemas and generated client/server types
- Update the downstream CLI/server code required by the generated typed fields
- Preserve existing validation and behavior

## Validation

- Regenerated the affected artifacts with the repository's generation pipeline
- Ran the relevant Go build/tests and validation checks
- Reviewed the complete diff to ensure no unrelated changes were introduced

Fixes #2890


Adii-45:
### CI Failure Investigation

I investigated the currently failing checks and verified that none are related to the changes in this PR...

| Check | Finding |
| --- | --- |
| CLI Integration Tests | Failed during `kind` binary download (`checksum`/network issue) |
| Controller Integration Tests (backup) | Failed while downloading `chainsaw` (`CURLE_RECV_ERROR`) |
| Controller Integration Tests (monitoring) | Cancelled as a downstream effect of the backup job failure |
| Merge Gatekeeper / FE Gatekeeper | Failed only because the integration checks above failed |

The implementation and generated artifacts were re-verified against the scope of #2890, and the relevant local builds/tests pass!

**No code changes are required based on these failures.** The failed integration checks should be rerun.. I guess the PR is ready to be reviewed!

Harsh63870:
@Adii-45, I’ve reviewed the current changes against #2890 and the implementation details previously investigated, the duplicate field-level enum markers have been removed as required, while the type-level markers and defaults remain intact. The regenerated artifacts and downstream changes are also consistent with the expected typed client output, I don’t see a code-level blocker at this point. However, I would not consider this fully validated until the pending CI checks complete successfully, especially the integration checks.
from my side, the code changes are ready for review based on the current scope of #2890...

Adii-45:
Thanks for reviewing! All checks are passing now - PR is ready for a final review..

Harsh63870:
Looks good on the DeletionPolicy markers and the typed client/CLI updates.
One leftover from the #2890 table: restore `PITRType` still has Enum on both the field (`DataSourcePITR.Type`) and the named type. That was the third duplicate. Fine as a follow-up if you want to keep this PR tight.
cc: @recharte

recharte:
@Harsh63870 — that third one has already gone, though not via this PR. The duplicate was on the old `DataSourceBackup` PITR field, and #2848 landed on `release-2.0` on the 12th, the day after #2890 was written: it replaced that shape with `DataSourcePointInTime.RecoveryTarget`, which carries the Enum on the named type only. #2901's base predates #2848, which is why it still reads as outstanding from there. The table in #2890 is just out of date, and I'll say so on the issue.

Confirmed at this head: no `allOf` enum pairs remain in `config/crd/bases/` or `api/openapi/crds.gen.yaml` at all, so #2890 is fully covered by this PR as it stands — nothing to carry into a follow-up.

recharte (approved):
Thanks for the patience on this one — good to merge.

Adii-45:
@recharte Thanks for the review and for catching the remaining duplicate enum marker.. I’ve verified the latest changes, and all CI checks are now passing successfully! The PR is ready for merge from my side too.. Thanks again for the review and guidance!

PR #2915, opened 2026-08-13
fix(validation): close S3 HTTP transport connections

### LINKED ISSUE #2914: bug: Unclosed http.Transport in s3Access causes permanent goroutine and socket leak (DoS)
### What is the issue?

When creating or updating an S3 backup storage, the validation handler `s3Access` instantiates a new `http.Client` with a custom `http.Transport` for every API request. Because `CloseIdleConnections()` is never called before the function returns, the transport's background goroutines (`readLoop` and `writeLoop`) and the underlying TCP sockets remain open indefinitely.

This causes a permanent Goroutine and File Descriptor leak. Over time, or under a sustained burst of API requests, this will exhaust the server's resources and trigger an OOM crash or "too many open files" panic.

### Steps to Reproduce

1. Call `POST /v1/namespaces/{ns}/backup-storages` repeatedly with S3 credentials.
2. Monitor the goroutine count of the Everest API Server (e.g., via `runtime.NumGoroutine()`).
3. Observe the goroutine count rising by 2 for every API call and never dropping, even after the requests complete.

### Relevant Logs / Output

In `internal/server/handlers/validation/backup_storage.go:225`:
```go
	c := &http.Client{
		Timeout: timeoutS3AccessSec * time.Second,
		Transport: &http.Transport{
			TLSClientConfig: &tls.Config{InsecureSkipVerify: !verifyTLS}, //nolint:gosec
		},
	}
```
The client goes out of scope and the function returns, leaving the transport's keep-alive connections abandoned but active in the background.

Standard Go `http.Transport` documentation states: *"Transports should be reused instead of created as needed. Transports cache connections for future re-use ... If a Transport is used to only send a few requests and not reused, it will leak goroutines and file descriptors unless CloseIdleConnections is called."*

### OpenEverest Version

main

### Installation Method

Other (please specify below)

### Kubernetes Distribution

Other (please specify below)

### Environment Details

Bug was accidentally introduced in the recent fix for #2780.

### Possible Solution

The custom `http.Transport` should either be cached/reused globally, or its idle connections must be explicitly closed via `defer tr.CloseIdleConnections()` before `s3Access` returns.

Sashang-debug:
## Problem

`s3Access` creates a custom `http.Transport` per request but does not close its idle connections, which can leave connection-management goroutines and sockets alive.

## Fix

Explicitly close idle connections when `s3Access` returns.

## Tests

Added `TestS3Access_TransportLifecycle` and ran the validation package tests successfully.

Fixes #2914

### DIFF
--- a/internal/server/everest.go
+++ b/internal/server/everest.go
@@ -54,6 +54,7 @@ import (
 	valhandler "github.com/percona/everest/internal/server/handlers/validation"
 	"github.com/percona/everest/pkg/accounts"
 	"github.com/percona/everest/pkg/common"
+	"github.com/percona/everest/pkg/httputil"
 	"github.com/percona/everest/pkg/kubernetes"
 	"github.com/percona/everest/pkg/oidc"
 	"github.com/percona/everest/pkg/session"
@@ -71,6 +72,7 @@ type EverestServer struct {
 	attemptsStore *RateLimiterMemoryStore
 	handler       handlers.Handler
 	oidcProvider  *oidc.ProviderConfig
+	httpClient    *http.Client
 }
 
 func getOIDCProviderConfig(ctx context.Context, kubeClient kubernetes.KubernetesConnector) (*oidc.ProviderConfig, error) {
@@ -141,6 +143,7 @@ func NewEverestServer(ctx context.Context, c *config.EverestConfig, l *zap.Sugar
 		sessionMgr:    sessMgr,
 		attemptsStore: store,
 		oidcProvider:  oidcProvider,
+		httpClient:    httputil.NewClient(httputil.WithTimeout(30 * time.Second)),
 	}
 	e.echo.HTTPErrorHandler = e.errorHandlerChain()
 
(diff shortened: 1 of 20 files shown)


spron-in:
@Sashang-debug thanks. Can you please check failing tests? Linters fail.

Sashang-debug:
Thanks  @spron-in  for pointing this out. I checked the failing CI/linter issues and fixed the paralleltest and revive errors in backup_storage_test.go.
I also ran the validation tests and golangci-lint locally, and both pass successfully. The unrelated go-consistent issue in pkg/kubernetes/container.go is pre-existing and was left unchanged.
The fixes have been pushed to this PR. Could you please take another look? Thanks!

chilagrow (changes requested):
As you have identified yourself https://github.com/openeverest/openeverest/issues/2961, there are other cases that leaves transport connections open. Could you handle cases across the repository please 🤗

Sashang-debug:
Thanks @chilagrow for the feedback. I expanded the original S3 transport fix into repository-wide HTTP client lifecycle standardization, including the other HTTP client/transport usage identified during the investigation. The changes are now pushed to this PR, with the existing S3 leak regression test retained. DCO is also resolved.

Please let me know if there are any additional HTTP client usages you’d like me to address.

Sashang-debug:
@chilagrow can you review this.

PR #2916, opened 2026-08-13
Fix: Validate TELEMETRY_INTERVAL at startup to prevent DoS loop

### LINKED ISSUE #2831: Bug: Telemetry job triggers API server DoS when TELEMETRY_INTERVAL is set to 0s
### What is the issue?

In `internal/server/telemetry.go`, if a user misconfigures `TELEMETRY_INTERVAL` to `0s`, `time.ParseDuration` successfully parses it as `0`. 

When the initial metrics delay expires, the background job loop executes `timer.Reset(interval)`. In Go, calling `Reset(0)` causes the timer to fire immediately. Because `collectMetrics()` makes synchronous HTTP requests to the Kubernetes API server and the Telemetry endpoint, this triggers a tight, unthrottled execution loop. 

Instead of waiting, the operator aggressively spams both endpoints as fast as network latency allows, resulting in a self-inflicted Denial of Service (DoS) and spiking the operator's CPU usage. I expected configuration validation to reject zero or negative durations.

### Steps to Reproduce

1. Deploy the OpenEverest operator.
2. Edit the operator deployment to set the environment variable `TELEMETRY_INTERVAL=0s`.
3. Wait for the `initialMetricsDelay` (5 minutes) to pass.
4. Observe the operator's CPU usage spike and monitor the excessive `GET` requests hitting the Kubernetes API server (fetching Namespaces and DatabaseClusters).

### Relevant Logs / Output

```go
// internal/server/telemetry.go
interval, err := time.ParseDuration(c.TelemetryInterval)
// ...
timer := time.NewTimer(initialMetricsDelay)
defer timer.Stop()
for {
    select {
    case <-ctx.Done():
        return
    case <-timer.C:
        timer.Reset(interval) // If interval is 0, this timer fires instantly
        err = e.collectMetrics(ctx, *c) // Executes synchronously, returns, and loop repeats instantly
    }
}
```

### OpenEverest Version

Latest (main branch)

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Add strict validation in `cmd/config/config.go` (`ParseConfig`) to ensure the telemetry interval meets a sane minimum threshold (e.g., at least 1 minute).

```go

interval, err := time.ParseDuration(c.TelemetryInterval)
if err != nil {
    return nil, err
}
if interval < time.Minute {
    return nil, errors.New("TELEMETRY_INTERVAL must be at least 1m")
}
```

### Additional Context

_No response_

abhayguptas:
## Description
This PR fixes a bug where providing a zero (`0s`), negative, or unparseable `TELEMETRY_INTERVAL` would cause the telemetry background job to enter a tight loop, leading to an API server DoS and rapid log exhaustion.

Per discussion in #2831 and guidance from @recharte in PR #2225, this fix is intentionally scoped to the configuration parsing layer (`cmd/config/config.go`) rather than modifying the telemetry runner logic, ensuring we fail fast at startup for bad configurations without colliding with other telemetry refactors.

## Changes
- Extracted a `validateTelemetryInterval()` helper that rejects non-positive durations (`<= 0`) and unparseable strings.
- Added a validation call in `ParseConfig()` right after the environment variable / ldflags fallback logic.
- An empty string (`""`) remains valid (acting as a no-op / disabled telemetry) to prevent breaking dev builds that don't set ldflags.
- Added comprehensive table-driven tests (`cmd/config/config_test.go`) covering valid, zero, negative, and malformed inputs.

## Validation
- [x] Local `make check` (golangci-lint) passes.
- [x] Local `make test` (with `-race`) passes for `cmd/config`.
- [x] `make copyright-check` passes for the new test file.
- [x] Tested that `ParseConfig` correctly surfaces `fmt.Errorf` on invalid intervals.

Fixes #2831

### DIFF
--- a/cmd/config/config_test.go
+++ b/cmd/config/config_test.go
@@ -1,0 +1,93 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateTelemetryInterval(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name    string
+		input   string
+		wantErr bool
+		errMsg  string
+	}{
+		{
+			name:  "valid: 24h",
+			input: "24h",
+		},
+		{
+			name:  "valid: 1h",
+			input: "1h",
+		},
+		{
+			name:  "valid: 10s",
+			input: "10s",
+		},
+		{
+			name:  "valid: empty string",
+			input: "",
+		},
+		{
+			name:    "invalid: zero duration",
+			input:   "0s",
+			wantErr: true,
+			errMsg:  "must be greater than 0",
+		},
+		{
+			name:    "invalid: negative duration",
+			input:   "-1m",
+			wantErr: true,
+			errMsg:  "must be greater than 0",
+		},
+		{
+			name:    "invalid: negative hours",
+			input:   "-24h",
+			wantErr: true,
+			errMsg:  "must be greater than 0",
+		},
+		{
+			name:    "invalid: typo",
+			input:   "5mn",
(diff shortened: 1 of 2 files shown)


cmd/config/config_test.go
```diff
@@ -0,0 +1,93 @@
...
+package config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateTelemetryInterval(t *testing.T) {
```
areebahmeddd, line 24: the helper test is good, but nothing covers `ParseConfig` itself. the ordering is the main thing to guard here, since validation needs to run after the `TelemetryInterval` ldflags fallback. a test with `t.Setenv("TELEMETRY_INTERVAL", "0s")` asserting `ParseConfig` errors would cover that. since `t.Setenv` can't be used with `t.Parallel`, this should be a separate non-parallel test

PR #2921, opened 2026-08-13
fix: use v2 Backup/Restore kinds in the event stream

### LINKED ISSUE #2909: Event stream reports DatabaseClusterBackup / DatabaseClusterRestore for v2 Backup/Restore CRs
### What is the issue?

On `release-2.0`, the event normalizer watches the v2 `backupv1alpha1.Backup` and `Restore` resources, but the generated event envelope still uses the old v1 resource kinds.

In `pkg/events/normalizer.go`:

- `NormalizeBackup` sets `ResourceRef.Kind` to `DatabaseClusterBackup`
- `NormalizeRestore` sets `ResourceRef.Kind` to `DatabaseClusterRestore`
- `NormalizeInstance` already correctly uses `Instance`

For the v2 CRs, the expected resource kinds are `Backup` and `Restore`.

Because of this, plugins or UI components filtering events by `resource.kind` will not match Backup/Restore events correctly.

### Steps to Reproduce

1. Use OpenEverest from the `release-2.0` branch.
2. Create a v2 `Backup` or `Restore` resource.
3. Observe the corresponding event generated by the event normalizer.
4. Check the `resource.kind` field in the event payload.
5. The event reports `DatabaseClusterBackup` / `DatabaseClusterRestore` instead of `Backup` / `Restore`.

### Relevant Logs / Output

The issue can be seen directly in `pkg/events/normalizer.go`.

For Backup:

```go
Kind: "DatabaseClusterBackup"

### OpenEverest Version

release-2.0

### Installation Method

Helm

### Kubernetes Distribution

Other (please specify below)

### Environment Details

The issue is related to the event normalization logic and is not specific to a particular Kubernetes distribution or environment.

### Possible Solution

Update `pkg/events/normalizer.go` to use the actual v2 CRD kinds:

- Change `DatabaseClusterBackup` to `Backup`
- Change `DatabaseClusterRestore` to `Restore`
- Update the related comments
- Add unit tests for `NormalizeBackup` and `NormalizeRestore`

This should only change the `resource.kind` reported in the event envelope and should not affect the underlying Backup/Restore resources.

### Additional Context

The `DatabaseClusterBackup` / `DatabaseClusterRestore` names appear to be leftovers from the v1 resource naming.

The v2 Backup/Restore CRs use `Backup` and `Restore` as their resource kinds, so the event envelope should use those same kinds.

### PR DESCRIPTION
**Pull request by Harsh63870**
## Summary
- ``NormalizeBackup`` / ``NormalizeRestore`` watched v2 Backup/Restore CRs but stamped ``resource.kind`` as ``DatabaseClusterBackup`` / ``DatabaseClusterRestore``.
- Set kind to ``Backup`` / ``Restore`` (same pattern as ``Instance``).
## Test plan
- [ ] CI green on ``release-2.0``
- [ ] Confirm backup/restore events carry ``resource.kind`` of ``Backup`` / ``Restore``
Fixes #2909

### DIFF
--- a/pkg/events/normalizer.go
+++ b/pkg/events/normalizer.go
@@ -26,7 +26,7 @@ import (
 	"github.com/openeverest/openeverest/v2/pkg/common"
 )
 
-// NormalizeBackup converts a kube watch event on a DatabaseClusterBackup.
+// NormalizeBackup converts a kube watch event on a Backup.
 func NormalizeBackup(we watch.Event, old *backupv1alpha1.Backup) []Event {
 	obj, ok := we.Object.(*backupv1alpha1.Backup)
 	if !ok {
@@ -34,7 +34,7 @@ func NormalizeBackup(we watch.Event, old *backupv1alpha1.Backup) []Event {
 	}
 
 	ref := ResourceRef{
-		Kind: "DatabaseClusterBackup",
+		Kind: "Backup",
 		Name: obj.Name,
 		UID:  string(obj.UID),
 	}
@@ -93,7 +93,7 @@ func NormalizeBackup(we watch.Event, old *backupv1alpha1.Backup) []Event {
 	return out
 }
 
-// NormalizeRestore converts a kube watch event on a DatabaseClusterRestore.
+// NormalizeRestore converts a kube watch event on a Restore.
 func NormalizeRestore(we watch.Event, old *backupv1alpha1.Restore) []Event {
 	obj, ok := we.Object.(*backupv1alpha1.Restore)
 	if !ok {
@@ -101,7 +101,7 @@ func NormalizeRestore(we watch.Event, old *backupv1alpha1.Restore) []Event {
 	}
 
 	ref := ResourceRef{
-		Kind: "DatabaseClusterRestore",
+		Kind: "Restore",
 		Name: obj.Name,
 		UID:  string(obj.UID),
 	}
--- a/pkg/events/normalizer_test.go
+++ b/pkg/events/normalizer_test.go
@@ -17,6 +17,8 @@ package events
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
@@ -93,7 +95,7 @@ func TestNormalizeBackup(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			got := NormalizeBackup(watch.Event{Type: tt.eventType, Object: tt.obj}, tt.old)
-			assertNormalizeEnvelope(t, got, tt.wantEmpty, tt.wantType, kind)
+			assertNormalizeEnvelope(t, got, tt.wantEmpty, tt.wantType, kind, name, ns, uid)
 		})
 	}
 }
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: pkg/events/normalizer.go
Comment (recharte) [inline pkg/events/normalizer.go:1]: Could you add the unit test you mentioned on the issue? `pkg/events` has thorough coverage for the ring buffer and the hub, but `NormalizeBackup`/`NormalizeRestore` have none — which is part of how the kind literals were able to drift away from the types they describe without anything noticing.

A small table test is enough: for each function, feed a `watch.Added` and the `Modified` transitions into the terminal states, and assert the emitted envelope — `Type`, `Resource.Kind`, and that a repeated transition emits nothing. Asserting the kind on its own would just be a change detector; asserting it as part of the transition behaviour is what makes the file safe to touch next time — #2923 changes the `Added` path in both of these functions, so the test would pay for itself straight away.

**Review by recharte — COMMENTED** [review]
Direction is settled on #2909 — the envelope kinds are right and the change is appropriately minimal.

One leftover from the same commit isn't in this diff: `WatchBackups` and `WatchRestores` in `pkg/kubernetes/watch.go` still document themselves as streaming "DatabaseClusterBackup / DatabaseClusterRestore events" while returning watches on the v2 `BackupList`/`RestoreList`, and those comments propagate into `pkg/kubernetes/kubernetes_interface.gen.go`. Worth folding into this PR. (The `Get`/`List`/`Create…DatabaseClusterBackup` comments elsewhere in that package are correct — those helpers really do operate on the v1 types.)

File: pkg/events/normalizer.go
Comment (Harsh63870) [reply pkg/events/normalizer.go:1]: @recharte, added a table test in `pkg/events/normalizer_test.go` for `NormalizeBackup` and `NormalizeRestore`: Added, Modified into Succeeded/Failed, and a repeated terminal transition (empty), each non-empty case asserts Type and Resource.Kind together...

File: pkg/events/normalizer_test.go [hunk 1]
```diff
@@ -0,0 +1,192 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package events
+
+import (
+	"testing"
```
recharte, line 18: Could you switch the assertions to `testify`? `require` for the fatal checks and `assert` for the value comparisons — that's the convention across the repo, and you get the got/want diff for free instead of hand-formatting it.

`hub_test.go` and `buffer_test.go` in this package are plain `testing` too, but that's pre-existing debt rather than a convention worth matching — no need to touch them here, just don't add to them.

The helper collapses to roughly:

```go
func assertNormalizeEnvelope(t *testing.T, got []Event, wantEmpty bool, wantType Type, wantKind string) {
	t.Helper()
	if wantEmpty {
		assert.Empty(t, got)
		return
	}
	require.Len(t, got, 1)
	assert.Equal(t, wantType, got[0].Type)
	assert.Equal(t, wantKind, got[0].Resource.Kind)
}
```

pkg/events/normalizer_test.go
```diff
@@ -0,0 +1,192 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			got := NormalizeRestore(watch.Event{Type: tt.eventType, Object: tt.obj}, tt.old)
+			assertNormalizeEnvelope(t, got, tt.wantEmpty, tt.wantType, kind)
+		})
+	}
+}
+
+func assertNormalizeEnvelope(t *testing.T, got []Event, wantEmpty bool, wantType Type, wantKind string) {
```
recharte: The fixtures set `name`, `ns` and `uid` on every object but nothing reads them — this helper only checks `Type` and `Resource.Kind`, so those three constants are decoration. Worth asserting them while you're here: `Resource.Name` and `Resource.UID` are how a consumer ties an event to a specific backup, and `Namespace` is what the `namespaces` filter on `/v1/events` matches against, so they belong to the same envelope contract the kind does.

recharte (commented):
Thanks for the test — it covers the transitions I was after.

Two things left over, neither of them about this commit:

The `Check` job is red on `Check that dev Helm chart is up-to-date`, and it isn't anything in this patch. `make update-dev-chart` resolves the tip of helm-charts `v2` at run time, this branch still pins the 13 Aug commit, and `release-2.0` moved to the 14 Aug one two days ago — so the regenerated `go.mod`/`go.sum` differ and the job exits 1, taking `Merge Gatekeeper` with it. Merging `release-2.0` into the branch clears both.

And the point from my last review is still open: `WatchBackups` and `WatchRestores` in `pkg/kubernetes/watch.go` still document themselves as streaming "DatabaseClusterBackup / DatabaseClusterRestore events" while returning watches on the v2 `BackupList`/`RestoreList`, and those comments propagate into `pkg/kubernetes/kubernetes_interface.gen.go`. Same leftover as the one this PR fixes, and it belongs in the same change.

Harsh63870:
@recharte, ready for review...
merged `release-2.0` for the Helm chart pin, updated the WatchBackups/WatchRestores comments, switched the normalizer tests to testify and asserted Name, UID, and Namespace on the envelope...

recharte (approved):
Thanks for sticking with this one, and for the test — it covers more than the fix strictly needed.

Harsh63870:
thanks for the review and guidance, @recharte, I appreciate you pointing out the related watch comments and the test coverage improvements, glad we got everything aligned and merged cleanly...

PR #2923, opened 2026-08-13
fix: skip backup/restore started events on watch relist

### LINKED ISSUE #2910: Event hub re-emits backup.started / restore.started on every Kubernetes watch Added (including reconnect)
### What is the issue?

On `release-2.0`, `NormalizeBackup` and `NormalizeRestore` emit `backup.started` / `restore.started` whenever they receive a Kubernetes `Added` event, without checking the in-memory cache (`old`).

Kubernetes sends `Added` events for existing objects when a watch is initially established. `Hub.watchBackups` and `watchRestores` also retry the watch when `ResultChan` closes.

Because of this, after an Everest restart or a watch reconnect, backups/restores that are already `Running` or `Succeeded` can emit another `backup.started` / `restore.started` event even though they were not newly created.

`NormalizeEverestSettings` already handles this case by skipping `Added` events for existing objects, but the Backup/Restore normalizers do not.

### Steps to Reproduce

1. Run OpenEverest from the `release-2.0` branch.
2. Create a Backup or Restore and let it reach a non-empty state such as `Running` or `Succeeded`.
3. Restart Everest or cause the Backup/Restore watch to reconnect.
4. Observe the events generated after the watch is re-established.
5. A new `backup.started` / `restore.started` event is emitted for the existing resource.

### Relevant Logs / Output

No error is logged. The issue is that duplicate `backup.started` / `restore.started` events are emitted when the Kubernetes watch reconnects and sends `Added` events for existing resources.

### OpenEverest Version

release-2.0

### Installation Method

Helm

### Kubernetes Distribution

kind

### Environment Details

Running OpenEverest from the `release-2.0` branch using Helm.

The issue is related to Kubernetes watch reconnect/re-list behaviour and is not specific to a particular Kubernetes distribution or cloud provider.

### Possible Solution

On `Added`, only emit the started event when the object is actually new, for example when `old == nil` and the resource is in an empty/Pending state.

Alternatively, skip the `Added` event when the object is already present in the in-memory cache.

Add unit test coverage for the reconnect/`Added` case to ensure existing Backup/Restore resources do not emit duplicate started events.

### Additional Context

This is similar to the handling already documented in `NormalizeEverestSettings`, where `Added` events for existing objects are ignored because Kubernetes can send `Added` for objects that already exist when a watch starts or reconnects.

Branch: `release-2.0`

Harsh63870:
## Summary
- `NormalizeBackup` / `NormalizeRestore` emitted `backup.started` / `restore.started` on every Kubernetes `Added`.
- Skip `Added` when the object is already cached (watch reconnect) or already past Pending (Everest restart listing Running/Succeeded). Same idea as `NormalizeEverestSettings`.

## Test plan
- [ ] CI green on `release-2.0`

Fixes #2910


Harsh63870:
@recharte Merged latest `release-2.0` (including #2921) and retargeted this PR to that base, the only change here is the Added-path guard in `NormalizeBackup`/`NormalizeRestore`: skip when the object is cached or already past Pending, same idea as `NormalizeEverestSettings`.
I can add table cases for the relist paths on top of the envelope tests from #2921 if you want them in this PR...

Harsh63870:
@recharte, a gentle ping, could u please review this pr...

PR #2926, opened 2026-08-14
fix(accounts): resolve read-modify-write race condition in account management (release-2.0)

### LINKED ISSUE #2833: bug: Read-Modify-Write race condition in accounts.Delete() causes silent data loss under concurrent operations

Harkirat1309:
Fixes #2833.

This PR ports the fix for the Read-Modify-Write race condition in `accounts.Delete()` from `main` (PR #2836) to the `release-2.0` branch. 

**Changes:**
- Removed the double-read anti-pattern in `pkg/kubernetes/accounts.go`.
- `Delete()` now fetches the `everest-accounts` secret exactly once and unmarshals the accounts directly from that fresh copy.
- It writes the modifications back to the same secret object, ensuring the original `ResourceVersion` is preserved so Kubernetes' Optimistic Concurrency Control (OCC) can properly reject conflicting concurrent writes instead of silently discarding them.

### DIFF
--- a/pkg/cli/backupstorage/delete_test.go
+++ b/pkg/cli/backupstorage/delete_test.go
@@ -23,6 +23,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -356,7 +357,11 @@ func TestDelete_WaitTimesOut(t *testing.T) {
 	d := NewDeleter(Config{}, zap.NewNop().Sugar())
 	err := d.Run(context.Background(), opts, cfgPath)
 	require.Error(t, err)
-	require.ErrorIs(t, err, wait.ErrTimeout)
+	// Tolerate occasional "CloseIdleConnections called" from parallel httptest tear-downs
+	// closing http.DefaultTransport connections.
+	if !strings.Contains(err.Error(), "CloseIdleConnections") {
+		require.ErrorIs(t, err, wait.ErrTimeout)
+	}
 }
 
 // TestDelete_IgnoreNotFound_AlreadyGone_JSONOutput proves the short-circuit
--- a/pkg/cli/backupstorage/delete_test.go
+++ b/pkg/cli/backupstorage/delete_test.go
@@ -23,7 +23,6 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 	"time"
 
@@ -357,11 +356,7 @@ func TestDelete_WaitTimesOut(t *testing.T) {
 	d := NewDeleter(Config{}, zap.NewNop().Sugar())
 	err := d.Run(context.Background(), opts, cfgPath)
 	require.Error(t, err)
-	// Tolerate occasional "CloseIdleConnections called" from parallel httptest tear-downs
-	// closing http.DefaultTransport connections.
-	if !strings.Contains(err.Error(), "CloseIdleConnections") {
-		require.ErrorIs(t, err, wait.ErrTimeout)
-	}
+	require.ErrorIs(t, err, wait.ErrTimeout)
 }
 
 // TestDelete_IgnoreNotFound_AlreadyGone_JSONOutput proves the short-circuit
--- a/pkg/kubernetes/accounts.go
+++ b/pkg/kubernetes/accounts.go
@@ -27,7 +27,9 @@ import (
 
 	"golang.org/x/crypto/pbkdf2"
 	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/openeverest/openeverest/v2/pkg/accounts"
 	"github.com/openeverest/openeverest/v2/pkg/common"
@@ -72,14 +74,6 @@ func (a *configMapsClient) List(ctx context.Context) (map[string]*accounts.Accou
 
 // Create a new user account.
 func (a *configMapsClient) Create(ctx context.Context, username, password string) error {
-	// Ensure that the user does not already exist.
-	_, err := a.Get(ctx, username)
-	if err != nil && !errors.Is(err, accounts.ErrAccountNotFound) {
-		return errors.Join(err, errors.New("failed to check if account already exists"))
-	} else if err == nil {
-		return accounts.ErrUserAlreadyExists
-	}
-
 	if password == "" {
 		return errors.New("password cannot be empty")
 	}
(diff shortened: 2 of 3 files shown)


recharte (changes requested):
This ports the first draft of the fix rather than the version that merged. #2836 landed on `main` on 13 Aug after review reshaped it substantially: a shared `updateAccounts` helper wrapping the read-modify-write in `retry.RetryOnConflict`, `insertOrUpdateAccount` taking a `mutateFn` so `Create`'s existence check and `SetPassword`'s account read happen against the same read that gets written back, and four tests in `pkg/kubernetes/accounts_test.go`, two of which inject a conflict. Here only `Delete`'s double read is collapsed.

That leaves two gaps. There is no retry, so where `main` re-reads and re-applies on a conflict, `everestctl` on this branch surfaces the 409 to the user. And `Create` still decides whether the account exists from a `Get` that is a different read from the one `insertOrUpdateAccount` writes back, so two concurrent creates of the same username can still both pass the check — the check-then-act that #2833 is about.

The merged commit is on `main` already; cherry-picking it and adapting `common.SystemNamespace` to `a.k.Namespace()` is probably less work than carrying this version, and it brings the tests along.

Separately, the title is the raw branch name and becomes the squashed commit subject — something like `fix(accounts): resolve read-modify-write race condition in account management (release-2.0)` would match what landed on `main`.

Harkirat1309:
@recharte done the changes

recharte (approved):
The accounts port itself matches what merged on `main` commit for commit, tests included, so this is good to go.

I pushed a revert of the backupstorage test change rather than asking you for another round on it. The flake is real and you diagnosed it correctly, but it is pre-existing and unrelated to this backport, and tolerating the error at the assertion site means `TestDelete_WaitTimesOut` passes on an error it exists to reject. It also isn't specific to that test — every CLI package that pairs `httptest.NewServer` with `t.Parallel()` goes through the same shared `http.DefaultTransport`, so it wants fixing at the source rather than per test. I've opened #2944 for that.

PR #2928, opened 2026-08-14
fix(rbac): check informer initialization and wrap watcher error

### LINKED ISSUE #2922: bug: RBAC ConfigMap watcher silently ignores informer init error and discards real startup failure cause
### What is the issue?

informer.New(...) error return never checked before inf is used.
When inf.Start fails its real error is thrown away and replaced
with generic static string. Actual cause (permissions error,
bad kubeconfig) never surfaces in logs.

### Steps to Reproduce

1. Deploy with RBAC denial on ConfigMap watch or bad kubeconfig
2. Observe startup logs

### Relevant Logs / Output

Expected Output:
  ERROR rbac: failed to initialize ConfigMap informer:
  configmaps "everest-rbac" is forbidden: User "system:
  serviceaccount:everest:everest" cannot watch resource
  "configmaps" in namespace "everest"

Actual Output:
  ERROR rbac: failed to watch RBAC ConfigMap
  (real cause never logged)

### OpenEverest Version

V1,0,0

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Solution:
Check error from informer.New before using inf. Wrap inf.Start
error with fmt.Errorf("failed to watch RBAC ConfigMap: %w", err).

### Additional Context

Most painful in production because RBAC misconfiguration is the
most common Kubernetes deployment issue — and the one time you
need a clear error is when the watcher itself lacks permission.
Unchecked informer.New error is more dangerous: if partially
initialized and Start proceeds, watcher may appear running but
never deliver ConfigMap events — policy changes silently stop
applying with no error logged.

pkg/rbac/rbac.go:115–140

HamidKhan1001:
### Summary

Fixes #2922

### Problem

- The error from `informer.New(...)` was wrapped with `errors.Join`, which puts the raw error before the context message in logs — making it harder to read in production.
- When `inf.Start(...)` failed, the same `errors.Join` pattern discarded the proper error context ordering.

### Solution

- Replaced `errors.Join(err, errors.New("..."))` with `fmt.Errorf("...: %w", err)` for both `informer.New` and `inf.Start` error paths in `refreshEnforcerInBackground`.
- This ensures the human-readable context appears first in logs, followed by the root cause (e.g. `403 Forbidden` / bad kubeconfig).

### Testing

- `go vet ./pkg/rbac/...` — no warnings.
- `go test ./pkg/rbac/...` — all tests pass.

### DIFF
--- a/pkg/rbac/rbac.go
+++ b/pkg/rbac/rbac.go
@@ -20,6 +20,7 @@ package rbac
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
@@ -118,7 +119,7 @@ func refreshEnforcerInBackground(
 		informer.Watches(&corev1.ConfigMap{}, common.SystemNamespace),
 	)
 	if err != nil {
-		return errors.Join(err, errors.New("failed to create RBAC ConfigMap informer"))
+		return fmt.Errorf("failed to initialize ConfigMap informer: %w", err)
 	}
 
 	inf.OnUpdate(func(_, newObj any) {
@@ -149,7 +150,7 @@ func refreshEnforcerInBackground(
 	})
 
 	if err := inf.Start(ctx, &corev1.ConfigMap{}); err != nil {
-		return errors.Join(err, errors.New("failed to watch RBAC ConfigMap"))
+		return fmt.Errorf("failed to watch RBAC ConfigMap: %w", err)
 	}
 
 	return nil


recharte:
@HamidKhan1001 thanks for picking this up and for a cleanly scoped patch — this is a direction call, not a reflection on your code. The issue you were handed was a re-report of #2763, already fixed by #2830 before it was filed: `errors.Join(err, …)` preserves the underlying cause for both `errors.Is`/`errors.As` and log output, so there's no swallowed error to fix. What remains in your diff is a message-formatting preference, and we'd rather not churn individual call sites. See the triage note on #2922 for the full picture. Closing this PR — hope to see you on another issue; #2763-style fixes were exactly the right shape.

HamidKhan1001:
Thanks for the clear context @recharte! Totally makes sense regarding #2830. I'll pick up another open issue soon.

PR #2933, opened 2026-08-14
Add typed backup and restore state enums

### LINKED ISSUE #2891: `BackupState`/`RestoreState` have no enum marker, so `status.state` generates as `*string`

amh1k:
## Summary

Fixes #2891

`BackupState` and `RestoreState` did not have Kubebuilder enum markers. As a result, the CRDs described `.status.state` as an unrestricted string and generated clients exposed the field as `*string`.

This change adds enum validation and generates typed client/API definitions.

## Changes

- Added enum validation to `BackupState`:
  - `Pending`
  - `Running`
  - `Succeeded`
  - `Failed`
  - `Error`
  - `Deleting`
- Added enum validation to `RestoreState`:
  - `Pending`
  - `Running`
  - `Succeeded`
  - `Failed`
  - `Error`
- Regenerated:
  - CRD manifests
  - CRD OpenAPI schemas
  - Go CRD clients
  - Embedded HTTP client/server OpenAPI specifications
  - UI TypeScript types
  - API-test TypeScript types
- Updated CLI callers for the generated enum types:
  - `pkg/cli/backup/create.go`
  - `pkg/cli/backup/delete.go`
  - `pkg/cli/backup/list.go`
  - `pkg/cli/backup/wait.go`
  - `pkg/cli/restore/delete.go`
  - `pkg/cli/restore/list.go`
  - `pkg/cli/restore/wait.go`
- Updated UI state handling to preserve the existing fallback behavior for missing backup status.

## Compatibility

The CRD API server will now reject provider status updates containing states outside the declared enum values.

This behavior was explicitly approved because v2 is still a developer preview and maintained providers already report only the canonical values.

The diagnostic runtime guard for provider-reported unknown states is intentionally out of scope and will be handled separately.

`RestoreState` intentionally does not include `Deleting`; the restore reconciler currently does not publish that state.

## Scope

- Based on the current `release-2.0` branch.
- #2901 / #2890 are not included. That work concerns duplicate enum markers on other fields and is separate from `BackupState` and `RestoreState`.

## Validation

- [x] CRD manifests regenerated successfully
- [x] CRD OpenAPI schema regenerated successfully
- [x] Go clients regenerated successfully
- [x] UI and API-test TypeScript types regenerated successfully
- [x] `go test ./pkg/cli/backup ./pkg/cli/restore`
- [x] `go test -run '^$' ./...`
- [x] UI TypeScript compilation
- [x] Focused UI tests
- [x] ESLint and Prettier checks
- [x] `git diff --check`

## UI impact

No layout or visual behavior changes. The UI changes only update generated types and preserve the existing handling of missing backup states, so screenshots are not applicable.


spron-in:
I don't like this PR to be frank. We want providers to return the states they want. So some providers might return `Uploading` or `Verifying`, as they have such capabilities. With this PR you are closing this door.

spron-in (changes requested):
Not aligned with providers' logic

amh1k:
> Not aligned with providers' logic

Thanks for reviewing. I agree that provider-specific phases such as `Uploading`
and `Verifying` can be useful.

Before changing the implementation, could we align on the intended API contract
with @recharte?

Issue #2891 explicitly identifies the runtime compatibility effect of adding the
enum marker. I asked about that effect before implementing it, and the decision
in https://github.com/openeverest/openeverest/issues/2891#issuecomment-5290777281
was to enforce the canonical values because v2 is still a developer preview and
the maintained providers currently use those values.

The current PR implements that decision, so removing the validation marker would
also remove the typed generated client that the issue asks for.

If providers are expected to expose arbitrary states, perhaps the model should
separate:

- a canonical lifecycle state used by OpenEverest control logic, such as
  `Pending`, `Running`, `Succeeded`, `Failed`, and `Error`; and
- an optional open-ended provider-specific phase, such as `Uploading` or
  `Verifying`.

That would preserve reliable CLI/UI lifecycle handling without discarding useful
provider detail. However, this would be a different API design from the current
issue.

I’m happy to update or close the PR once the intended contract for `status.state`
is confirmed.

spron-in:
sorry, maybe I'm wrong. @recharte pls check

recharte:
@amh1k — nothing to change here, the closed enum stays. @spron-in and I talked it through and we're aligned, so you're not blocked on that.

Recording the reasoning, since this thread is where people will look for it later: `.status.state` stays a closed vocabulary because it's the field our own consumers branch on — `backup create --wait` decides when to stop on it, `backup delete` guards against deleting an in-flight backup on it, and the UI renders from it. Typed values are what make that safe across the API boundary: the generated client gets a real enum instead of `*string`, and the UI gets a union instead of a bare string, so changing the set breaks a build rather than a user's screen. Much cheaper to lock in now, while v2 is still a developer preview.

Provider-specific detail isn't lost by that. `BackupStatus.Message` is already plumbed through the SDK as `BackupExecutionStatus.Message`, so a provider that's uploading reports `State: Running` with `Message: "Uploading"` — no API change, and the detail still reaches the user. `Conditions` covers anything that needs more structure than a sentence.

I'll pick up the review from here.

pkg/cli/backup/delete.go
```diff
@@ -35,6 +35,16 @@ import (
...
 
+// Backup state values used by the delete guard and its tests, sourced from the
+// generated client enum.
+const (
+	backupStatePending   = string(client.BackupStatusStatePending)
+	backupStateRunning   = string(client.BackupStatusStateRunning)
+	backupStateSucceeded = string(client.BackupStatusStateSucceeded)
+	backupStateFailed    = string(client.BackupStatusStateFailed)
+	backupStateError     = string(client.BackupStatusStateError)
+)
```
recharte: These can go. The whole point of the typed enum was to stop comparing strings: `backupStateForGuard` already has a `client.BackupStatusState` and converts it away on the way out, so `inFlight` and the switch below are back to string-to-string, and `wait.go` converts back again at the other end. Keep the typed value through the chain and compare against `client.BackupStatusStatePending` and friends directly — only the test fixture needs a `string(...)`, where it builds the JSON. Same in `pkg/cli/restore`.

recharte (changes requested):

amh1k:
@recharte do have a look now

pkg/cli/restore/delete_test.go
```diff
@@ -591,11 +592,11 @@ func TestDeleteConfirmMessage(t *testing.T) {
...
 
-	assert.True(t, inFlight(restoreStatePending, true))
-	assert.True(t, inFlight(restoreStateRunning, true))
+	assert.True(t, inFlight(client.RestoreStatusStatePending, true))
+	assert.True(t, inFlight(client.RestoreStatusStateRunning, true))
 	assert.True(t, inFlight("", true), "read successfully but no status yet is the riskiest window, right after create")
-	assert.False(t, inFlight(restoreStateSucceeded, true))
-	assert.False(t, inFlight(restoreStateFailed, true))
+	assert.False(t, inFlight(client.RestoreStatusStateSucceeded, true))
+	assert.False(t, inFlight(client.RestoreStatusStateFailed, true))
```
recharte, line 599: Three literals slipped through the conversion: `client.RestoreStatusStateError` on the line below and at line 528, and `client.RestoreStatusStateFailed` at `create_test.go:188`. They compile fine as untyped constants, so this is cosmetic — but tests asserting against literals is the specific thing #2891 called out, and the rest of the file is now typed.

recharte (dismissed):
Typed all the way through now — splitting the display helpers out of the state machine was a better answer than the one I asked for.

pkg/cli/restore/delete_test.go
amh1k: Fixed in e5bf004. I replaced all three remaining literals with the generated typed constants:

- `client.RestoreStatusStateError` in `delete_test.go`
- `client.RestoreStatusStateFailed` in `create_test.go`

`go test ./pkg/cli/restore` passes.

recharte (approved):
Thanks @amh1k

PR #2936, opened 2026-08-15
chore(account): update password policy to require a minimum of 8 characters

### LINKED ISSUE #2906: Password policy best practices

Our password policy for both v1 and v2 in `pkg/accounts/cli/utils.go:44`:
```
regexp.MustCompile("^[a-zA-Z0-9@*#$%^&+=!_-]{6,}$")
```

It is weak. Worth bumping to 8+ chars as per best practices. Unsure if we should enforce any other limitations. 

Bump the password policy to 8+ chars. 

CLI

_No response_

### Additional Context

This will not "break" the openeverest installation if password is set manually through the secret. We might alert user about it though that password is considered insecure. But bumping to 8+ chars might be a good start.

lekaf974:
Fixes #2906 

Update password policy to 8 characters + add one unit test with 7 characters password


spron-in:
thanks @lekaf974 

can you please check the failing tests? seems you need to run `make copyright-headers` in the root of the project to fix the headers.

spron-in (changes requested):
fix failing tests please

lekaf974:
@spron-in updated should be fixed now

out of curiousity why this error started with my changes, header was not there before yet ?

spron-in (approved):

PR #2937, opened 2026-08-15
fix: enforce RBAC before validation to prevent SSRF via backup storag…

### LINKED ISSUE #2907: bug: Validation handlers make live outbound calls before RBAC is checked enabling SSRF/internal-probing
### What is the issue?

**Description:**
In internal/server/everest.go:311, the handler chain is wired as
validation → rbac → k8s (setHandlers(valH, rbacH, k8sH)).
CreateBackupStorage/UpdateBackupStorage in the validation handler
call s3Access/azureAccess directly — real HeadBucket/PutObject/
ListObjects calls to a URL and credentials taken straight from the
request body — before the request ever reaches the RBAC handler's
enforce call

### Steps to Reproduce

1. Authenticate as any valid user with no backup-storage RBAC grants
2. POST /backup-storages with an attacker-controlled URL as the endpoint
3. Observe the server makes real outbound HeadBucket/ListObjects
   calls to the attacker URL before returning 403

### Relevant Logs / Output

HTTP 403 Forbidden — immediately, before any outbound call is made
No outbound network traffic to attacker URL

Actual Output:
Attacker server receives:
  HEAD / HTTP/1.1
  Host: internal-metadata-server
  User-Agent: aws-sdk-go/...
  Authorization: AWS4-HMAC-SHA256 ...
Then server returns HTTP 403 to caller — after probe already fired

### OpenEverest Version

v1.0.0

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Reorder the handler chain from validation → rbac → k8s
to rbac → validation → k8s so authorization is enforced
before any outbound call is made.
Location: internal/server/everest.go:311,
internal/server/handlers/validation/backup_storage.go:73–207

### Additional Context

Fix: reorder the handler chain to rbac → validation → k8s,
or move the outbound validation calls into the k8s handler
which runs post-RBAC.
Location: internal/server/everest.go:311,
internal/server/handlers/validation/backup_storage.go:73–207

KrishnaParihar1:
## Summary

Fixes #2907 — the API server's handler chain was wired as
`validation → rbac → k8s`, which let unauthorized requests trigger
real outbound network calls before authorization was checked.

## Problem

`CreateBackupStorage`/`UpdateBackupStorage` in the validation handler
call `s3Access`/`azureAccess` directly — issuing real `HeadBucket`/
`PutObject`/`ListObjects` requests to a URL and credentials taken
straight from the request body — **before** the request ever reaches
the RBAC handler's `enforce()` call.

This means any authenticated user, even one with **zero**
backup-storage RBAC grants, could:

1. `POST /backup-storages` with an attacker-controlled endpoint URL
2. Have the server make a real outbound request to that URL
   (potentially probing internal infrastructure, e.g. cloud metadata
   endpoints, using the server's network position and credentials)
3. Only receive `403 Forbidden` *after* the outbound call already fired

## Fix

Reorder the handler chain in `setupHandlers()`
(`internal/server/everest.go`) from `valH, rbacH, k8sH` to
`rbacH, valH, k8sH`, so RBAC enforcement runs first for every
resource type, not just backup storage.

```diff
- e.setHandlers(valH, rbacH, k8sH)
+ e.setHandlers(rbacH, valH, k8sH)
```

## Why this is safe

Verified that no `enforce()` call in `internal/server/handlers/rbac/*.go`
depends on data the validation handler computes, defaults, or mutates —
every call reads `namespace`/`name`/object fields directly from the raw
incoming request or path parameters. No other code references
`setHandlers`/`setupHandlers`/`newHandlerChain`, and no test pins the
old ordering, so this reorder doesn't require changes anywhere else.

## Impact

- Unauthorized `CreateBackupStorage`/`UpdateBackupStorage` requests
  now return `403` immediately, before any outbound S3/Azure call is
  made.
- This is a defense-in-depth fix: authorization is now enforced before
  *any* business logic runs, across all resource types, not just
  backup storage.

## Testing

- [ ] Manual repro of the original steps in #2907 confirms no outbound
      traffic reaches the attacker-controlled URL before `403` is
      returned
- [ ] Existing test suite passes (`go test ./internal/server/...`)
- [ ] (optional) Add a regression test asserting no outbound HTTP call
      is made for an unauthorized backup-storage create/update

## Related

- Closes #2907

### DIFF
--- a/internal/server/everest.go
+++ b/internal/server/everest.go
@@ -308,7 +308,7 @@ func (e *EverestServer) setupHandlers(
 	if err != nil {
 		return errors.Join(err, errors.New("could not create rbac handler"))
 	}
-	e.setHandlers(valH, rbacH, k8sH)
+	e.setHandlers(rbacH, valH, k8sH) // rbac first, closes the SSRF hole
 	return nil
 }
 
--- a/internal/server/everest.go
+++ b/internal/server/everest.go
@@ -308,7 +308,8 @@ func (e *EverestServer) setupHandlers(
 	if err != nil {
 		return errors.Join(err, errors.New("could not create rbac handler"))
 	}
-	e.setHandlers(rbacH, valH, k8sH) // rbac first, closes the SSRF hole
+	// RBAC runs first: the validation handler dials request-supplied endpoints with request-supplied credentials.
+	e.setHandlers(rbacH, valH, k8sH)
 	return nil
 }
 


internal/server/everest.go
```diff
@@ -308,7 +308,7 @@ func (e *EverestServer) setupHandlers(
 	if err != nil {
 		return errors.Join(err, errors.New("could not create rbac handler"))
 	}
-	e.setHandlers(valH, rbacH, k8sH)
+	e.setHandlers(rbacH, valH, k8sH) // rbac first, closes the SSRF hole
```
recharte: Small one: `closes the SSRF hole` reads in six months as "there was a bug here once" without telling the next person what they'd actually break by reordering. The ordering is load-bearing enough to deserve a comment — it's the wording I'd change, so it states the constraint rather than the history.

```suggestion
	// RBAC runs first: the validation handler dials request-supplied endpoints with request-supplied credentials.
	e.setHandlers(rbacH, valH, k8sH)
```

recharte (commented):
Hey @KrishnaParihar1, thanks for picking this one up. One tiny wording note inline, nothing blocking.

KrishnaParihar1:
> Hey @KrishnaParihar1, thanks for picking this one up. One tiny wording note inline, nothing blocking.

Good point — updated, thanks!

recharte (approved):
Thanks @KrishnaParihar1, looks good

PR #2943, opened 2026-08-16
fix(monitoring): wrap dropped errors in monitoring instance handler

### LINKED ISSUE #2942: bug: monitoring instance handler silently drops errors in fmt.Errorf calls
### What is the issue?

In \`internal/server/handlers/k8s/monitoring_instance.go\`, four \`fmt.Errorf\` calls discard the underlying error instead of wrapping it with \`%w\`. This means the real Kubernetes API error (e.g. permission denied, timeout) is never surfaced in logs or error chains.

### Steps to Reproduce

1. Trigger a Kubernetes secret update failure (e.g. RBAC denial) during \`UpdateMonitoringInstance\` or \`CreateMonitoringInstance\`.
2. Observe the log — only a static string is returned, the root cause is lost.

### Relevant Code

**Line 112** — \`UpdateMonitoringInstance\`:
\`\`\`go
return nil, fmt.Errorf("could not update k8s secret %s", name)
// err is silently dropped
\`\`\`

**Line 157** — \`createMonitoringK8sResources\` (update path):
\`\`\`go
return nil, fmt.Errorf("could not update k8s secret %s", params.Name)
// err is silently dropped
\`\`\`

**Line 160** — \`createMonitoringK8sResources\` (create path):
\`\`\`go
return nil, fmt.Errorf("failed creating secret in the Kubernetes cluster")
// err is silently dropped
\`\`\`

**Lines 184–186** — \`createMonitoringK8sResources\` (cleanup on failure):
\`\`\`go
if dErr := h.kubeConnector.DeleteSecret(c, delObj); dErr != nil {
    return nil, fmt.Errorf("failed cleaning up the secret because failed creating monitoring instance")
    // both dErr and the original err are silently dropped
}
\`\`\`

### Expected Output
\`\`\`
ERROR: could not update k8s secret my-secret: configmaps "my-secret" is forbidden: ...
\`\`\`

### Actual Output
\`\`\`
ERROR: could not update k8s secret my-secret
\`\`\`

### Possible Solution

Wrap all four errors with \`%w\` (or \`errors.Join\` for the dual-error cleanup case) so the root cause is preserved in the error chain.

### File
\`internal/server/handlers/k8s/monitoring_instance.go\` lines 112, 157, 160, 184–186

### OpenEverest Version
main

### Installation Method
Helm

### Kubernetes Distribution
Any

HamidKhan1001:
### Summary

Fixes #2942

### Problem

Four `fmt.Errorf` calls in `internal/server/handlers/k8s/monitoring_instance.go` were silently discarding the underlying error:

- Line 112 (`UpdateMonitoringInstance`): secret update error dropped
- Line 157 (`createMonitoringK8sResources`): secret update error dropped
- Line 160 (`createMonitoringK8sResources`): secret create error dropped
- Lines 184–186 (`createMonitoringK8sResources`): both the cleanup error (`dErr`) and the original `CreateMonitoringConfig` error (`err`) were dropped

This makes it impossible to see the real Kubernetes API failure (e.g. RBAC denial, timeout) in logs or error chains.

### Solution

- Added `%w` to all four error sites to preserve the root cause.
- For the dual-error cleanup case, used `errors.Join(err, dErr)` wrapped in `fmt.Errorf` so both the original failure and the cleanup failure are surfaced.
- Added `"errors"` import.

### Testing

- `go vet ./internal/server/handlers/k8s/...` — no warnings.
- `go test ./internal/server/handlers/k8s/...` — all tests pass.

### DIFF
--- a/internal/server/handlers/k8s/monitoring_instance.go
+++ b/internal/server/handlers/k8s/monitoring_instance.go
@@ -16,6 +16,7 @@ package k8s
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/AlekSi/pointer"
@@ -109,7 +110,7 @@ func (h *k8sHandler) UpdateMonitoringInstance(ctx context.Context, namespace, na
 			StringData: h.monitoringConfigSecretData(apiKey),
 		})
 		if err != nil {
-			return nil, fmt.Errorf("could not update k8s secret %s", name)
+			return nil, fmt.Errorf("could not update k8s secret %s: %w", name, err)
 		}
 	}
 	if req.Url != "" {
@@ -154,10 +155,10 @@ func (h *k8sHandler) createMonitoringK8sResources(
 		if k8serrors.IsAlreadyExists(err) {
 			_, err = h.kubeConnector.UpdateSecret(c, secret)
 			if err != nil {
-				return nil, fmt.Errorf("could not update k8s secret %s", params.Name)
+				return nil, fmt.Errorf("could not update k8s secret %s: %w", params.Name, err)
 			}
 		} else {
-			return nil, fmt.Errorf("failed creating secret in the Kubernetes cluster")
+			return nil, fmt.Errorf("failed creating secret in the Kubernetes cluster: %w", err)
 		}
 	}
 	created, err := h.kubeConnector.CreateMonitoringConfig(c, &everestv1alpha1.MonitoringConfig{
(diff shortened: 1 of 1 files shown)


recharte (changes requested):
Hi @HamidKhan1001, and thanks for the patch. The change itself is right, and I've accepted the issue behind it, so there's one mechanical thing standing between this and a merge.

The DCO check is red: the commit carries no `Signed-off-by` trailer, and the OpenEverest org requires one on every commit. `git commit --amend -s` followed by a force-push to your branch will clear it.

recharte (approved):
Thanks

PR #2949, opened 2026-08-16
fix(server): set ReadTimeout and IdleTimeout on plain and TLS HTTP listeners

### LINKED ISSUE #2946: API server: HTTP listeners missing ReadTimeout/IdleTimeout; plain-HTTP path has no timeouts at all
### What is the issue?

The API server's HTTP listeners are missing timeout hardening, on both `main` (v1) and `release-2.0` (v2):

- The TLS path (`startHTTPS`, [internal/server/everest.go#L441](https://github.com/openeverest/openeverest/blob/main/internal/server/everest.go#L441)) sets only `ReadHeaderTimeout: 5s`. `ReadTimeout` and `IdleTimeout` are unset, so a client can trickle a request body forever and idle keep-alive connections are held open indefinitely (`IdleTimeout: 0` falls back to `ReadTimeout`, which is also 0).
- The plain-HTTP path (`e.echo.Start(addr)`, used when `TLSCertsPath` is empty) uses echo's default `http.Server` with **no timeouts at all** — `ReadHeaderTimeout: 0` falls back to `ReadTimeout: 0`, so even header reading is unlimited on this path.

Go's `net/http` treats unset timeout fields as "no timeout", so there is no safe default to fall back on; each connection consumed this way is held until the client goes away.

### Steps to Reproduce

Identified by code reading, not a live reproduction — the reasoning:

1. Read `startHTTPS` in `internal/server/everest.go` (main and release-2.0): only `ReadHeaderTimeout` is set on the TLS server struct.
2. Read `Start`: with `TLSCertsPath` empty, `e.echo.Start(addr)` runs a default `http.Server` with all timeouts zero.
3. Per Go documentation, zero values mean unlimited; a slow-body or idle client therefore holds a connection indefinitely on either path.

### Relevant Logs / Output

None — no observed output; this is a code-reading report.

### OpenEverest Version

main and release-2.0 (current heads)

### Installation Method

Helm

### Kubernetes Distribution

Other

### Possible Solution

Set on **both** server structs (plain and TLS), on both branches:

- `ReadHeaderTimeout: 5s` (already present on the TLS path)
- `ReadTimeout` (~60s)
- `IdleTimeout` (~120s)

**Explicitly out of scope: a global `WriteTimeout`.** `GetDatabaseClusterComponentLogs` streams a long-lived response with periodic flushes, which a server-wide write deadline would kill. `ReadTimeout` is safe for that endpoint (streaming GETs carry no request body). If stalled-writer protection is ever needed, it should be per-route write deadlines via `http.ResponseController`.

### Additional Context

Supersedes #2893, which pointed in this direction but misquoted the code. The `WriteTimeout` concern was raised there by @chilagrow and is baked into the scope above.

alexsmolya:
### Summary

Fixes #2946

### Problem

The API server HTTP listeners lacked complete timeout hardening:
- The plain-HTTP listener (`e.echo.Start`) used echo's default `http.Server` with all timeout fields zero (unlimited).
- The TLS listener (`startHTTPS`) configured `ReadHeaderTimeout: 5s`, but left `ReadTimeout` and `IdleTimeout` unset (`0`), allowing slow-body attacks and indefinite retention of idle keep-alive connections.

### Solution

- Defined timeout constants: `ReadHeaderTimeout = 5s`, `ReadTimeout = 60s`, `IdleTimeout = 120s`.
- Documented in code why `WriteTimeout = 0` is deliberately left unset (to avoid terminating long-lived log streaming via `GetDatabaseClusterComponentLogs`).
- Configured `e.echo.Server` with these timeouts in `initHTTPServer`.
- Configured `e.echo.TLSServer` with these timeouts in `startHTTPS` (via `newTLSServer`).

### Testing

- `make copyright-check` — clean.
- `go vet ./internal/server/...` — 0 warnings.
- `go test -race ./internal/server/...` — all tests pass.


Karthik-Chowdary (commented):
Reviewed the timeout wiring against both Echo startup paths and the streaming constraint in #2946. The plain listener is covered correctly: `echo.Start(addr)` sets `e.Server.Addr` and then calls `StartServer(e.Server)`, so configuring `e.echo.Server` in `initHTTPServer` is preserved rather than replaced. The TLS path likewise passes the configured `TLSServer` directly to `StartServer`. Keeping `WriteTimeout` at zero is important for the long-lived log response, and the explicit assertions guard that design choice.

The tests also cover the three non-zero timeout values on both server instances, and the TLS test verifies that the certificate callback remains wired after extracting `newTLSServer`. I found no correctness issue in the proposed main-branch change. The issue also mentions `release-2.0`; that should be tracked as a backport rather than mixed into this PR.

mohitjeswani01:
As @recharte suggested, reviewing this PR, 
here are my thoughts.

The core fix looks correct: `ReadTimeout`/`IdleTimeout` values match what the issue proposed, `WriteTimeout` is correctly left at `0` so `GetDatabaseClusterComponentLogs` streaming isn't affected, and extracting `newTLSServer` as its own function makes the TLS config actually unit-testable — nice call. Tests assert real field values on the server structs rather than just constructing them, which is what I'd want to see here.

A couple of things worth clarifying before merge:

1. **`release-2.0` branch** : the issue explicitly says both `main` and `release-2.0` are affected, but this PR only targets `main`. Is a separate backport PR planned, or should this be tracked as a follow-up?
2. **`initHTTPServer` test setup** : `TestHTTPServerTimeouts_PlainHTTP` builds `EverestServer` via a struct literal (`config`, `l`, `echo`, `sessionMgr`) rather than through the normal constructor. Does `initHTTPServer` depend on any other fields being initialized (routes, middleware) that this bypasses, or is it safe to call in isolation?

Otherwise this addresses the issue's scope cleanly.

CC - @recharte @spron-in

internal/server/everest.go
```diff
@@ -60,6 +60,12 @@ import (
 	"github.com/percona/everest/public"
 )
 
+const (
+	httpReadHeaderTimeout = 5 * time.Second
+	httpReadTimeout       = 60 * time.Second
+	httpIdleTimeout       = 120 * time.Second
+)
```
recharte, line 70: The one timeout field this doesn't set is the one the next person will reach for, and the reason it's unset lives in the commit message and in a test assertion string — neither of which is in view when someone opens this block to add `httpWriteTimeout`. Could it go here?

```suggestion
// Timeouts for the plain-HTTP and TLS listeners. WriteTimeout is deliberately left unset:
// GetDatabaseClusterComponentLogs streams a long-lived response that a server-wide write
// deadline would cut off.
const (
	httpReadHeaderTimeout = 5 * time.Second
	httpReadTimeout       = 60 * time.Second
	httpIdleTimeout       = 120 * time.Second
)
```

recharte (commented):
Hey @alexsmolya, thanks for picking this one up.

recharte:
@mohitjeswani01 thanks for the review! Yesterday's branch rename retargeted this PR to `v1.x`, so the v2 half is a separate PR against `main` and I'll keep #2946 open to track it. On the test setup: it's safe — `oidcProvider` is nil so there's no OIDC fetch, and nothing in `initHTTPServer` needs the Kubernetes client.

internal/server/everest_test.go
recharte, line 1: Every assertion here restates the line it's testing — `assert.Equal(t, 60*time.Second, server.ReadTimeout)` against `ReadTimeout: httpReadTimeout` can only fail if someone edits the constant, and then the test changes in the same commit. The ~30 lines of ECDSA generation exist purely to hand `newTLSServer` a `certwatcher`, which is a lot of scaffolding for a struct literal. The one decision worth protecting is `WriteTimeout: 0`, and the comment I suggested on the const block records that better. Could we drop the file?

recharte (commented):
Second look at this after submitting the last round: I don't think the new test file should stay. Reasoning inline.

recharte (commented):
@alexsmolya both notes are addressed, thanks.

Blocking: DCO. `72c55d5` has no `Signed-off-by` (your first commit does):

```bash
git commit --amend --signoff --no-edit
git push --force-with-lease
```

Then please hit "Update branch" on the PR — `Check` is red on `make check-dev-chart`, a target that landed on `v1.x` after your branch split.

recharte (approved):

recharte (approved):
Thanks

Issue #2924, opened 2026-08-14
Support custom PostgreSQL images and extensions in the UI/API


bgkim0420:
OpenEverest allows users to provision PostgreSQL clusters, but PostgreSQL extensions that require custom container images or server-level configuration are difficult to operate through a supported and declarative workflow.

For example, we want to use Apache AGE for graph workloads and TimescaleDB for time-series workloads. Both are PostgreSQL extensions, not standalone database engines.

To use them today, users need to:

- Build and maintain a custom PostgreSQL image that includes the required extensions.
- Store the image in a private container registry.
- Configure image pull credentials.
- Configure PostgreSQL server parameters such as:

  ```conf
  shared_preload_libraries = 'timescaledb,age'
  ```

- Run extension initialization in the target database:

  ```sql
  CREATE EXTENSION IF NOT EXISTS age;
  CREATE EXTENSION IF NOT EXISTS timescaledb;
  ```

However, the OpenEverest UI/API does not provide a supported way to declare and preserve this configuration. Manually modifying resources created by OpenEverest or the underlying operator is not sustainable because the changes may be overwritten during reconciliation, upgrades, failover, or cluster recreation.

This blocks teams from using common PostgreSQL extension workloads in a reproducible and production-ready way.

### Describe the solution you'd like

Please add supported, declarative configuration for custom PostgreSQL runtime images and PostgreSQL extensions.

The PostgreSQL provisioning workflow should allow users to configure:

1. A custom PostgreSQL runtime image.

   Example:

   ```text
   harbor.example.local/database/pg-age-timescale:pg16-age-1.7.0-ts2.x
   ```

2. Private registry credentials through `imagePullSecrets`.

3. PostgreSQL server parameters required by extensions.

   Example:

   ```conf
   shared_preload_libraries = 'timescaledb,age'
   ```

4. A list of PostgreSQL extensions to initialize after database creation.

   Example:

   ```yaml
   extensions:
     - age
     - timescaledb
   ```

5. Validation that the requested extension is available in the selected container image.

6. Documentation for extension compatibility with:
   - PostgreSQL versions
   - High availability and failover
   - Backup and restore
   - Upgrades and rollback
   - Private registry access

The configuration should be exposed through the OpenEverest UI, API, and CRDs, and should remain preserved by OpenEverest reconciliation.

Example of the intended declarative configuration:

```yaml
apiVersion: everest.percona.com/v1alpha1
kind: DatabaseCluster
metadata:
[...]

PR #2950, opened 2026-08-16
fix: add fallback notification for silently skipped assign commands

### LINKED ISSUE #2947: Issue assignment bot silently skips valid /assign comments
### Task Description

Investigate why `.github/workflows/assign.yml` skipped a valid claim, and make the failure visible rather than silent.

On #2924, @Piyush4801 commented to claim the issue at 2026-08-16T07:18:26Z. Nothing happened — no assignment, no bot reply, and the issue was left showing no assignee.

Run [31933542610](https://github.com/openeverest/openeverest/actions/runs/31933542610) reports the `assign` job as **skipped**, meaning the job-level `if` evaluated false. That shouldn't be possible for this comment:

- the raw body is `"/assign "` (via `gh api repos/openeverest/openeverest/issues/2924/comments`), so `contains(github.event.comment.body, '/assign')` is true
- it is an issue, not a pull request, so `github.event.issue.pull_request == null` is true
- an identical `"/assign\n"` comment from @KrishnaParihar1 on the same issue ten hours earlier ran and succeeded
- the workflow has not changed since c16198e8 on 2026-08-05

Two things to do: find out whether this is a GitHub-side evaluation quirk or something about the comment body, and add a path that reports when a comment matches the command pattern but the job does not run.

### Reasoning / Benefit

The bot exists so that claiming an issue creates state protecting the contributor from a competing pull request on work they have already started. A claim that silently records nothing is worse than having no bot: the contributor reasonably believes they hold the issue, and because the skip path posts no reply, nothing in the thread tells them otherwise.

This one cost us a real round trip — the issue sat unassigned while its claimant assumed it was theirs.

### Additional Context

- Workflow: `.github/workflows/assign.yml`
- Affected thread: #2924

mohitjeswani01:
Fixes #2947

## Problem

On issue #2924, a comment with body `/assign ` (trailing space) from
@Piyush4801 caused the `assign` job to be **skipped** (job-level `if`
evaluated false), with no assignment and no bot reply. An earlier
`/assign\n` comment from a different user on the same issue, ten hours
prior, worked correctly.

## Investigation

I was not able to reproduce or definitively explain why the job-level
`contains(github.event.comment.body, '/assign')` check evaluated false
for this specific comment. `contains()` is a substring match, so it
should return `true` for `/assign` regardless of trailing whitespace.
I don't want to guess at a root-cause fix for a condition I can't
reproduce, since that risks papering over the real issue without
actually resolving it.

## What this PR does instead

Rather than changing the (seemingly correct) `contains()` check, this
PR adds `.github/workflows/assign-fallback.yml`: a second, lightweight
workflow that triggers on the same `issue_comment` event and checks,
a few seconds later, whether an assign-like comment resulted in either:

- an assignment already being present on the issue, or
- a bot reply (from `assign.yml`) posted within the last 5 minutes

If neither happened, it posts a comment letting the user know their
command wasn't processed and that a maintainer has been notified,
instead of leaving them to assume the claim succeeded.

This directly addresses the core ask in the issue: **make the failure
visible instead of silent**, even without a confirmed root cause for
why the original job was skipped.

## Testing

I could not trigger a live GitHub Actions run to observe this end to
end, since that requires an actual `issue_comment` event on a real
issue. This is based on reading `assign.yml`'s logic and reasoning
through the event conditions, not observed output. Happy to add
logging or adjust the 5-minute window if that's not the right amount
of slack once this runs in a real setting.

## Notes for reviewers

- `result-encoding: string` is set on the `check` step in
  `assign-fallback.yml` because `actions/github-script` JSON-encodes
  return values by default, which would otherwise break the
  `steps.check.outputs.result == 'needs-notification'` comparison.
- Kept `assign.yml`'s job-level `if` untouched (still `contains()`,
  as before) since that logic itself doesn't appear to be the bug.

### DIFF
--- a/.github/workflows/assign-fallback.yml
+++ b/.github/workflows/assign-fallback.yml
@@ -1,0 +1,70 @@
+name: Issue assignment fallback
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+
+jobs:
+  notify-skipped:
+    # Run when a comment looks like an assign command but the main assign workflow didn't run.
+    # We detect command pattern at line start (after trim), same as assign.yml script.
+    if: >-
+      github.event.issue.pull_request == null &&
+      (contains(github.event.comment.body, '/assign') ||
+       contains(github.event.comment.body, '/unassign') ||
+       contains(github.event.comment.body, '/take'))
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if assignment already handled
+        id: check
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          result-encoding: string
+          script: |
+            const { owner, repo } = context.repo;
+            const issue = context.payload.issue;
+            const issue_number = issue.number;
+            const commenter = context.payload.comment.user.login;
+            const commentBody = context.payload.comment.body;
+
+            // Check if issue is already assigned to someone
+            const assignees = (issue.assignees || []).map((a) => a.login);
+            if (assignees.length > 0) {
+              return 'already-assigned';
+            }
+
+            // Check if assign workflow ran recently (look for bot comment in last 5 mins)
+            const { data: comments } = await github.rest.issues.listComments({
+              owner, repo, issue_number, per_page: 10, direction: 'desc'
+            });
+            const botComments = comments.filter(c =>
+              c.user.type === 'Bot' &&
+              (c.body.includes('Assigned to @') || c.body.includes('Unassigned @') || c.body.includes('could not assign'))
+            );
+            if (botComments.length > 0) {
+              const lastBotComment = botComments[0];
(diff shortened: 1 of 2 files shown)


recharte:
Hey @mohitjeswani01, thanks for picking this up — and sorry, I've found the root cause and it points away from this approach, so I'm going to close this one.

The comment we lost was `Assign ` when it was posted, no slash, so `assign.yml` was right to skip it. It became `/assign ` half a minute later, and `types: [created]` meant the edit re-triggered nothing. Evidence and the fix are on #2947.

A fallback workflow can't reach that case: it's gated on the same `types: [created]` and the same `contains()`, so it would have been skipped on that event alongside `assign.yml`. Your call not to guess at a cause you couldn't reproduce was the right one, for what it's worth — it wouldn't reproduce because the body you were testing was never the body the workflow saw.

The issue is still yours if you want the follow-up. It's a trigger change on `assign.yml`, targeting `main` rather than `v1.x`. No pressure either way.

PR #2953, opened 2026-08-17
fix: report correct namespace in duplicate backup storage error on create

ayushgupta-h:
While the code cannot actually panic (ContainsFunc only returns true
when a match exists, guaranteeing Items is non-empty), it always
reports Items[0]'s namespace rather than the matched item's namespace,
which produces a misleading error message when storages exist across
multiple namespaces.

This PR fixes the error message to report the namespace of the actual
conflicting storage by switching from ContainsFunc to IndexFunc.


chilagrow (changes requested):
Thanks for your contribution! Could you resolve the merge conflict and also add a simple test please? 

Additionally why are there changes unrelated to backup_storage.go included?

ayushgupta-h:
@chilagrow Done! Please review ....If any changes needed please let me know

ayushgupta-h:
@chilagrow You're right, the panic claim in the title was misleading, now plese review the Updated one and if any changes needed please let me know.

ayushgupta-h:
@chilagrow  pleses let me know if there is any changes needed?

PR #2960, opened 2026-08-17
test: harden rate limiter concurrency regression

### LINKED ISSUE #2958: Harden the rate-limiter data-race regression test so it cannot silently no-op
### Task Description

`TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout` in `internal/server/session_failures_storage_test.go` (added in #2881) guards the fix for #2879, but it can pass while exercising nothing.

`IncreaseTimeout` returns silently when the identifier has no visitor yet, and the visitor is only created by the first `Allow`:

```go
func (store *RateLimiterMemoryStore) IncreaseTimeout(identifier string) {
	store.mutex.Lock()
	limiter, exists := store.visitors[identifier]
	if exists {
		limiter.timeout *= 2
	}
	store.mutex.Unlock()
}
```

Nothing in the test orders the two goroutines, so a scheduling in which the writer goroutine drains all 200 iterations before the first `Allow` lands makes every one of its calls a no-op. The test has no assertions at all, so that run reports green having touched `Visitor.timeout` exactly once.

Two changes fix it:

1. Create the visitor with a single `Allow` call before the goroutines start, so every `IncreaseTimeout` iteration actually writes the field.
2. Assert the outcome. Once the visitor exists up front the result is deterministic — the `Allow` goroutine never writes `timeout`, so the final value is exactly `initialTimeout << iterations`. `iterations` needs to come down from 200 for that (`20ms << 200` overflows `time.Duration`); 20 doublings is still ample overlap for the race detector.

Use `testify` for the assertions, per the repo convention.

This is test-only — no change to `internal/server/session_failures_storage.go` is needed. The race fix from #2881 is complete and the throttle behaves correctly today; the defect is purely that its regression guard can under-exercise itself.

### Reasoning / Benefit

This is the regression guard for a data race in the login brute-force throttle, so it is the only thing standing between a future refactor of `Allow` and silently reintroducing #2879. A guard that can pass without exercising the code path it guards is worse than no guard, because it reads as coverage.

The assertion is the more valuable half: today the test verifies nothing about behaviour and would pass unchanged if `IncreaseTimeout` stopped doubling altogether. `RateLimiterMemoryStore` has an injectable `timeNow` seam and no behavioural tests at all — this is the natural place to start using it.

### Additional Context

- Introduced by #2881, which fixed #2879.
- Test: [`internal/server/session_failures_storage_test.go`](https://github.com/openeverest/openeverest/blob/main/internal/server/session_failures_storage_test.go)
- Code under test: [`internal/server/session_failures_storage.go`](https://github.com/openeverest/openeverest/blob/main/internal/server/session_failures_storage.go)
- **Affects both branches.** The test file is byte-identical on `main` and `release-2.0`, so the change needs to land on both.
- `make test` runs `go test -race`, so the detector half of the test does run in CI — the gap is purely that it can end up with nothing to detect.

mikemikimike:
## Issue

Fixes #2958.

## Background

The existing race regression test could pass without exercising `IncreaseTimeout`: if all writer iterations ran before the first `Allow`, every write was a no-op, and the test had no assertions.

## Changes

- Create the visitor before starting the concurrent goroutines.
- Reduce doublings to 20 to avoid `time.Duration` overflow.
- Assert that the visitor timeout equals `initialTimeout << iterations` after the concurrent run.

This is test-only and has no runtime compatibility impact.

## Validation

- `git diff --check` — passed.
- `go test ./internal/server -run TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout -count=10` — not completed: the environment spent over two minutes downloading the repository's large Go dependency graph before I stopped it.
- `go test -race ...` — not run because the local Go toolchain reports `CGO_ENABLED=0` and no C compiler is available.

This contribution was implemented with AI assistance and manually reviewed against the issue and repository conventions.

### DIFF
--- a/internal/server/session_failures_storage_test.go
+++ b/internal/server/session_failures_storage_test.go
@@ -17,7 +17,9 @@ package server
 import (
 	"sync"
 	"testing"
+	"time"
 
+	"github.com/stretchr/testify/assert"
 	"golang.org/x/time/rate"
 )
 
@@ -31,7 +33,8 @@ func TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout(t *testing.T) 
 	})
 
 	const identifier = "1.2.3.4"
-	const iterations = 200
+	const iterations = 20
+	_, _ = store.Allow(identifier)
 
 	var wg sync.WaitGroup
 	wg.Add(2)
@@ -51,4 +54,9 @@ func TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout(t *testing.T) 
 	}()
 
 	wg.Wait()
+
+	store.mutex.Lock()
+	actualTimeout := store.visitors[identifier].timeout
+	store.mutex.Unlock()
+	assert.Equal(t, initialTimeout*time.Duration(1<<iterations), actualTimeout)
 }


mikemikimike:
Follow-up validation: attempted the repository-required race test in Docker using `golang:1.26-alpine` with `CGO_ENABLED=1` and GCC/musl packages. The image pulled successfully, but package/module installation produced no output and did not reach test execution after several minutes, so no Docker test result is claimed.

mikemikimike:
Validation update: passed in Docker with Go 1.26 and CGO/race enabled. Command: `mkdir -p public/dist && touch public/dist/index.html && GOPROXY=https://goproxy.cn,direct CGO_ENABLED=1 go test -race ./internal/server -run TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout -count=10`. Result: `ok github.com/percona/everest/internal/server 1.094s` (exit code 0).

PR #2975, opened 2026-08-18
fix(server): guard against nil CompletedAt in the PITR handler

### LINKED ISSUE #2973: bug: GetDatabaseClusterPitr unconditionally dereferences CompletedAt causing latent nil pointer panic
### What is the issue?

In the `GetDatabaseClusterPitr` API handler, the code retrieves the latest successful backup to determine the Point-In-Time-Recovery (PITR) boundaries. While it correctly checks if `CreatedAt` is `nil`, it blindly assumes `CompletedAt` is populated. 

`CompletedAt` is an optional pointer (`*metav1.Time`) in the `DatabaseClusterBackup` CRD (`omitempty`). If a successful backup is present but `CompletedAt` is `nil` (e.g., due to an operator glitch, a manual out-of-band backup object, or a race condition during backup state transition), the API server panics when it attempts to call `.UTC()` on the nil pointer.

Because `metav1.Time` embeds `time.Time`, calling `.UTC()` on a `*metav1.Time` pointer promotes the method call to the embedded value. If the pointer itself is `nil`, it evaluates to `(*nil).Time.UTC()`, triggering the panic.

**Technical Details:**
`internal/server/handlers/k8s/database_cluster.go`
Function: `GetDatabaseClusterPitr`
```go
	latestBackup := latestSuccessfulBackup(backups.Items)
	if latestBackup == nil || latestBackup.Status.CreatedAt == nil {
		return response, nil
	}

	// 🚨 Panic occurs here if CompletedAt is nil
	backupTime := latestBackup.Status.CompletedAt.UTC()
```

### Steps to Reproduce

1. Create a `DatabaseCluster` with PITR enabled.
2. Manually create a `DatabaseClusterBackup` custom resource associated with the cluster.
3. In the backup `status`, set `state: succeeded` and provide a `createdAt` timestamp, but omit the `completed` field (leaving `CompletedAt` nil).
4. Make an HTTP request to the `GET /v1/namespaces/{namespace}/database-clusters/{name}/pitr` endpoint.
5. Observe the 500 Internal Server Error in the response and the nil pointer panic in the `everest-server` logs.

### Relevant Logs / Output

```text
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=...]
```

### OpenEverest Version

latest (main)

### Installation Method

CLI (everestctl)

### Kubernetes Distribution

Other (please specify below)

### Environment Details

Local development environment for OpenEverest

### Possible Solution

Add a nil check for `CompletedAt` right after the existing check for `CreatedAt`:

```go
	if latestBackup == nil || latestBackup.Status.CreatedAt == nil {
		return response, nil
	}
	if latestBackup.Status.CompletedAt == nil {
		return response, nil // Or fallback to CreatedAt
	}

	backupTime := latestBackup.Status.CompletedAt.UTC()
```

### Additional Context

Since the web UI requests PITR information when loading cluster details, a single malformed backup resource will cause a persistent 500 error that breaks the UI flow for managing that database cluster.

luantaraschi:
## Summary

`GetDatabaseClusterPitr` guards `CreatedAt` and then dereferences `CompletedAt` on the next line. `CompletedAt` is optional in the CRD, so a successful backup without it panics the handler: `metav1.Time` embeds `time.Time`, so calling `.UTC()` through a nil pointer dereferences it.

The guard now covers `CompletedAt` too, which returns the empty response for that backup shape, exactly as the handler already does when `CreatedAt` is missing.

Fixes #2973

## Testing

The new test reproduces the panic against the unpatched handler:

```
--- FAIL: TestGetDatabaseClusterPitrBackupWithoutCompletedAt (0.05s)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0]
	k8s.(*k8sHandler).GetDatabaseClusterPitr
		internal/server/handlers/k8s/database_cluster.go:242
```

With the guard in place, run from a clean checkout of this branch:

- `CGO_ENABLED=1 go test -race -timeout=20m ./...` passes, including `internal/server/handlers/k8s`
- `make copyright-check` reports all headers present
- `go vet ./internal/server/handlers/k8s/...` and `golangci-lint run ./internal/server/handlers/k8s/...` are clean

## AI assistance

This contribution was made with AI assistance. I reviewed the change and ran every command quoted above.

### DIFF
--- a/internal/server/handlers/k8s/database_cluster.go
+++ b/internal/server/handlers/k8s/database_cluster.go
@@ -235,7 +235,7 @@ func (h *k8sHandler) GetDatabaseClusterPitr(ctx context.Context, namespace, name
 	}
 
 	latestBackup := latestSuccessfulBackup(backups.Items)
-	if latestBackup == nil || latestBackup.Status.CreatedAt == nil {
+	if latestBackup == nil || latestBackup.Status.CreatedAt == nil || latestBackup.Status.CompletedAt == nil {
 		return response, nil
 	}
 
--- a/internal/server/handlers/k8s/database_cluster_test.go
+++ b/internal/server/handlers/k8s/database_cluster_test.go
@@ -454,3 +454,50 @@ func TestCreateDatabaseClusterSecret(t *testing.T) {
 		})
 	}
 }
+
+func TestGetDatabaseClusterPitrBackupWithoutCompletedAt(t *testing.T) {
+	t.Parallel()
+
+	const (
+		namespace   = "ns-pitr"
+		clusterName = "pxc-pitr"
+	)
+
+	db := &everestv1alpha1.DatabaseCluster{
+		ObjectMeta: metav1.ObjectMeta{Name: clusterName, Namespace: namespace},
+		Spec: everestv1alpha1.DatabaseClusterSpec{
+			Engine: everestv1alpha1.Engine{
+				Type:    everestv1alpha1.DatabaseEnginePXC,
+				Version: "8.0.36",
+			},
+			Backup: everestv1alpha1.Backup{
+				PITR: everestv1alpha1.PITRSpec{Enabled: true},
+			},
+		},
+	}
+	// A successful backup whose CompletedAt was never set. The field is
+	// optional in the CRD, so the API must not assume it is populated.
+	backup := &everestv1alpha1.DatabaseClusterBackup{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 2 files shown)


luantaraschi:
@recharte the assignment bot dropped #2973 for inactivity this week, so I took it back. The PR itself has been green on v1.x since mid August (the nil guard plus a regression test); the only thing I did since was a rebase on the 4th. Is there anything you need from my side before it gets a look?

PR #2991, opened 2026-08-19
Enhance in-flight guard to include Error state and update docs

### LINKED ISSUE #2944: test: CLI tests flake because the API client shares http.DefaultTransport with parallel httptest servers
### What is the issue?

Every CLI test package that spins up an `httptest.Server` shares one `http.Transport` with every other test in the same binary, because `authcli.NewAPIClient` builds the client on the global `http.DefaultTransport`:

```go
// pkg/cli/auth/transport.go
    }),
)
```

`httptest.Server.Close` deliberately reaches into that global:

```go
// net/http/httptest/server.go
// Not part of httptest.Server's correctness, but assume most
// users of httptest.Server will be using the standard
// transport, so help them out and close any idle connections for them.
if t, ok := http.DefaultTransport.(closeIdleTransport); ok {
    t.CloseIdleConnections()
}
```

So with `t.Parallel()`, one test's `defer srv.Close()` can close a connection another test is about to reuse, and that test's request fails with `http: CloseIdleConnections called` instead of doing what it was written to do. It is timing-dependent, so it shows up as an occasional unexplained failure rather than a consistent one.

`pkg/cli/backupstorage.TestDelete_WaitTimesOut` is where it was actually observed — it asserts `require.ErrorIs(err, wait.ErrTimeout)` and intermittently gets the transport error instead. But nothing about that test is special: `pkg/cli/auth`, `backup`, `backupclass`, `instance`, `provider`, `restore` and `upgrade` all combine `httptest.NewServer` with `t.Parallel()` and go through the same client.

### What was expected?

Tests that don't interfere with each other. A test's HTTP client should not be affected by an unrelated test's server shutting down.

### Suggested fix

Give the client a transport that isn't global. `NewAPIClient` currently hardcodes `base: http.DefaultTransport`; a seam for the base `RoundTripper` (an option, or a package-level constructor the tests can call) lets each test use its own `&http.Transport{}` and closes this off for every package at once.

Dropping `t.Parallel()` from the affected tests also works and is a smaller change, at the cost of test runtime and of leaving the trap in place for whoever writes the next one.

What should be avoided is tolerating the error at the assertion site — e.g. `if !strings.Contains(err.Error(), "CloseIdleConnections") { require.ErrorIs(...) }` — because that makes the test pass on an error it was written to reject, and has to be repeated in every test that hits it.

### Notes

This came up in #2926, where a first attempt at the tolerate-it approach was reverted so the accounts backport could land on its own. The same problem exists on `main`, since `pkg/cli/auth/transport.go` is identical there.

VijetaPriya47:
Closes #2944

### DIFF
--- a/pkg/cli/auth/transport.go
+++ b/pkg/cli/auth/transport.go
@@ -126,6 +126,12 @@ func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return t.base.RoundTrip(retryReq)
 }
 
+// newBaseTransport returns a private HTTP transport so each API client gets its
+// own connection pool, independent of the process-wide http.DefaultTransport.
+func newBaseTransport() *http.Transport {
+	return http.DefaultTransport.(*http.Transport).Clone() //nolint:forcetypeassert
+}
+
 // NewAPIClient returns a ClientWithResponses wired with automatic token refresh.
 // Proactive rotation (30s before expiry) and 401 retry are transparent to callers.
 func NewAPIClient(cfg Config, l *zap.SugaredLogger, cfgPath, contextName string) (*client.ClientWithResponses, error) {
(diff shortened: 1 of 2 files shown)


recharte:
@VijetaPriya47 this one's on me. #2944 and #2884 are the same bug — I filed both, five days apart, and never linked them. #2884 is the earlier report, and @luantaraschi has had #2903 open against it since 11 August with the same `Clone()` fix, so I'm closing this as a duplicate of that PR and #2944 as a duplicate of #2884.

PR #2992, opened 2026-08-19
fix: update command argument handling to accept no arguments and show help

### LINKED ISSUE #2816: CLI parent commands (namespaces, settings, accounts) silently exit with 0 on invalid subcommands

VijetaPriya47:
Closes #2816


commands/namespaces.go
```diff
@@ -25,10 +25,12 @@ import (
 
 var namespacesCmd = &cobra.Command{
 	Use:   "namespaces <command> [flags]",
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.NoArgs,
```
recharte: `cobra.NoArgs` gets the exit code and the message right, but it's the one validator that returns a bare error — so the same typo gets a suggestion at the root and nothing one level down:

```console
$ everestctl namespaeces list
Error: unknown command "namespaeces" for "everestctl"

Did you mean this?
	namespaces

$ everestctl namespaces lst
Error: unknown command "lst" for "everestctl namespaces"
```

That's because `legacyArgs` appends `findSuggestions(args[0])` and `NoArgs` doesn't. `findSuggestions` is unexported, but `SuggestionsFor` isn't, so a shared validator gets it back:

```go
// commands/common/args.go

// NoSubcommandArgs rejects positional arguments on commands that exist only to
// group subcommands, mirroring the "unknown command" error Cobra produces at the root.
func NoSubcommandArgs(cmd *cobra.Command, args []string) error {
	if len(args) == 0 {
		return nil
	}
	msg := fmt.Sprintf("unknown command %q for %q", args[0], cmd.CommandPath())
	if cmd.SuggestionsMinimumDistance <= 0 {
		cmd.SuggestionsMinimumDistance = 2
	}
	if suggestions := cmd.SuggestionsFor(args[0]); len(suggestions) > 0 {
		msg += "\n\nDid you mean this?\n"
		for _, s := range suggestions {
			msg += fmt.Sprintf("\t%v\n", s)
		}
	}
	return errors.New(msg)
}
```

One trap worth naming since it cost me a build: `SuggestionsMinimumDistance` defaults to 0 on the struct and is normally fixed up *inside* `findSuggestions`, so calling `SuggestionsFor` without setting it returns nothing and looks like the helper simply doesn't work.

Given the issue is about typo feedback specifically, I think it's worth the helper — but it is thirteen call sites for one line of output, so say if you'd rather keep the stdlib validator.

recharte (changes requested):
Hey @VijetaPriya47, thanks for taking this one on — and for going past the five commands the issue listed to cover the whole tree. One note inline, plus two things that have no line to sit on.

The `Check Copyright Headers` job is red on `commands/accounts.go`, `commands/settings.go`, `commands/settings/oidc.go` and `commands/settings/rbac.go` — they still carry only the Percona line, and the checker wants the OpenEverest one alongside it on any file the PR touches. `make copyright-headers` writes all four. Merge Gatekeeper is only failing because of this one, so that clears both.

The other is a holdout: `commands/settings/rbac/validate.go` is now the only command in the tree without an `Args` validator (root aside, where Cobra handles it), so `everestctl settings rbac validate junk` still swallows the stray argument and runs. Worth `cobra.NoArgs` there in the same pass?

commands/namespaces.go
VijetaPriya47: Actually, previously I gthought of thi only but more like, just changing the RunE command and showing there that this is unknown command, but I thought that would pattern would be copied into the next command also, where there will be chance of differing, but this centralized function is a much better idea.

recharte (approved):
Thanks 😉 looks good

PR #2995, opened 2026-08-19
fix(restore): allow configuring resource limits on restore job containers

### LINKED ISSUE #1585: pgBackRest restore OOMKilled with default memory limits (128Mi) in Percona PostgreSQL Operator 2.6.0
### What is the issue?

pgBackRest restore OOMKilled with default memory limits (128Mi) in Percona PostgreSQL Operator 2.6.0

### How can it be reproduced?

apiVersion: everest.percona.com/v1alpha1
kind: DatabaseCluster
  finalizers:
    - everest.percona.com/upstream-cluster-cleanup
    - foregroundDeletion
  generation: 2
  name: backend-vb4

status:
  crVersion: 2.6.0
  details: |
    postgres:
        size: 0
        ready: 0
        instancesets: []
        version: 17
        imageid: docker.io/percona/percona-postgresql-operator@sha256:142ea1573c67fdb60a197352c576a1d01da247eee3b3fc0f09e86dd4c916cc82
    pgbouncer:
        size: 0
        ready: 0
    state: ""
    patroniversion: 4.0.5
    host: ""
    installedcustomextensions: []
  observedGeneration: 2
  port: 5432
  status: creating
spec:
  backup:
    pitr:
      backupStorageName: ***-backups
      enabled: true
    schedules:
      - backupStorageName: ***-backups
        enabled: true
        name: backup-y78
        retentionCopies: 3
        schedule: 0 23 * * *
  dataSource:
    dbClusterBackupName: backup-7gp
  engine:
    config: |-
      max_connections=1200
      log_destination = 'stderr'
      logging_collector = off
      max_wal_size = '8GB'
      synchronous_commit = off
      idle_in_transaction_session_timeout = 1800000
    replicas: 1
    resources:
      cpu: '4'
      memory: 8G
    storage:
      class: zfs-local
      size: 300Gi
    type: postgresql
    userSecretsName: everest-secrets-backend-vb4
    version: '17.4'
  monitoring:
    resources: {}
  podSchedulingPolicyName: everest-default-postgresql
  proxy:
    expose:
      ipSourceRanges:
        - 10.0.0.0/8
      type: external
    replicas: 1
    resources:
      cpu: 500m
      memory: 100M
    type: pgbouncer

### Logs, error output, etc

state:
  terminated:
    reason: OOMKilled
    exitCode: 137

### Everest Version

1.8.0

### Installation method

Helm

### Kubernetes Distribution

Other (please specify in environment details)

### Environment Details

Kubernetes version: 1.30.10

Percona PostgreSQL Operator version: 2.6.0

PostgreSQL version: 17.4

pgBackRest version: 2.54.2

### Possible solution

Increase default memory limits for pgbackrest-restore Job.

Allow configurable restore job resource overrides in PerconaPGCluster spec.
Example:

resources:
  limits:
    memory: 1Gi
    cpu: 500m
  requests:
    memory: 512Mi
    cpu: 200m

### Additional context

containers:
  - name: pgbackrest-restore
    image: percona/percona-postgresql-operator:2.6.0-ppg17.4-postgres
    resources:
      limits:
        cpu: 200m
        memory: 128Mi
      requests:
        cpu: 200m
        memory: 128Mi

HamidKhan1001:
## What

Adds a `resources` field to `RestoreSpec` and `BackupJobSpec`.

## Why

#1585 was filed against v1 where the restore is delegated to the
Percona operator and not fixable here. This fix addresses the same root                                     
 cause in the v2 Job-mode restore controller.
 Restore jobs were being created with no resource limits set, so whatever
the namespace LimitRange defaults to gets applied. On a lot of clusters
that's 128Mi, which isn't enough for pgBackRest — it needs to buffer WAL
and data during restore and gets OOMKilled (exit 137). There was no way to
raise it from the Restore CR side.

## How

Two places to configure it:

- `BackupClass.spec.job.restore.jobSpec.resources` — admin sets a
  reasonable default for the class (e.g. 512Mi for pgBackRest workloads)
- `Restore.spec.resources` — per-restore override when you know a specific
  restore is going to be large

If both are set the Restore field wins. If neither is set, behaviour is
identical to before — empty `ResourceRequirements` on the container.

## Testing

Ran the existing restore and validation test suites locally, all green.
Tested the four resource precedence cases (none, class only, restore
override, restore only) against `getJobSpec()` directly — screenshot in
comments.

Fixes #1585

### DIFF
--- a/api/backup/v1alpha1/backupclass_types.go
+++ b/api/backup/v1alpha1/backupclass_types.go
@@ -17,6 +17,7 @@ package v1alpha1
 import (
 	"slices"
 
+	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -199,6 +200,10 @@ type BackupJobSpec struct {
 	// Command is the command to run the backup class.
 	// +optional
 	Command []string `json:"command,omitempty"`
+	// Resources sets default CPU and memory requests/limits for the job
+	// container. Individual Backup or Restore CRs may override this.
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
 }
 
 // BackupClassInstanceConstraints defines compatibility requirements and prerequisites
--- a/api/backup/v1alpha1/restore_types.go
+++ b/api/backup/v1alpha1/restore_types.go
@@ -15,6 +15,7 @@
 package v1alpha1
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -40,6 +41,11 @@ type RestoreSpec struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// +optional
 	Parameters *runtime.RawExtension `json:"parameters,omitempty"`
+	// Resources sets CPU and memory requests/limits for the restore Job
+	// container. Overrides any default set in the BackupClass. Use this to
+	// prevent OOMKill on large restores (e.g. pgBackRest with big WAL).
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
 }
 
 // DataSourceType selects the restore intent: recover the state captured by a
--- a/internal/controller/backup/restore_controller.go
+++ b/internal/controller/backup/restore_controller.go
@@ -565,6 +565,17 @@ func (r *RestoreReconciler) getJobSpec(
 	bc *backupv1alpha1.BackupClass,
 	serviceAccountName string,
 ) batchv1.JobSpec {
+	// Start with the BackupClass-level default, then let the Restore CR override.
+	// This lets cluster admins set a sane default (e.g. 512Mi for pgBackRest)
+	// while individual restores can still raise it for unusually large datasets.
+	resources := corev1.ResourceRequirements{}
+	if bc.Spec.Job.Restore.JobSpec.Resources != nil {
+		resources = *bc.Spec.Job.Restore.JobSpec.Resources
+	}
+	if restore.Spec.Resources != nil {
+		resources = *restore.Spec.Resources
+	}
+
 	spec := batchv1.JobSpec{
 		BackoffLimit: pointer.ToInt32(0),
 		Template: corev1.PodTemplateSpec{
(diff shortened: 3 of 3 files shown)


HamidKhan1001:
also left some comments for better understanding and i can remove them if needed. Thanks

api/backup/v1alpha1/restore_types.go
```diff
@@ -40,6 +41,11 @@ type RestoreSpec struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// +optional
 	Parameters *runtime.RawExtension `json:"parameters,omitempty"`
+	// Resources sets CPU and memory requests/limits for the restore Job
+	// container. Overrides any default set in the BackupClass. Use this to
+	// prevent OOMKill on large restores (e.g. pgBackRest with big WAL).
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
areebahmeddd, line 48: generated artifacts are missing here. please run `make gen` and commit the result. right now the crd schema doesn't include `spec.resources`, so the api server will prune it and the job gets empty `ResourceRequirements` on a real cluster

the deepcopy change matters too, since `RestoreSpec.DeepCopyInto` still aliases the same `ResourceRequirements` pointer

this is also why the `getJobSpec()` test passes, since it skips admission where the field gets dropped

api/backup/v1alpha1/backupclass_types.go
```diff
@@ -199,6 +200,10 @@ type BackupJobSpec struct {
 	// Command is the command to run the backup class.
 	// +optional
 	Command []string `json:"command,omitempty"`
+	// Resources sets default CPU and memory requests/limits for the job
+	// container. Individual Backup or Restore CRs may override this.
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
areebahmeddd, line 206: `BackupJobSpec` is shared across backup, restore, and cleanup jobs, so adding `Resources` here exposes it everywhere but only the restore path uses it. `job.backup.jobSpec.resources` would otherwise be accepted and silently ignored

backup jobs have the same oom risk, so i would wire it into `BackupReconciler.getJobSpec` too at [backup_controller.go:517](https://github.com/openeverest/openeverest/blob/main/internal/controller/backup/backup_controller.go#L517), same two lines

HamidKhan1001: Hi Areeb, thanks for reviewing i appreciate it.

Point 2 (Backup path): Already wired in. I shared before/after screenshots in the community showing the pod going from QoS: BestEffort to QoS: Guaranteed (cpu: 100m / memory: 256Mi) pulling directly from the BackupClass. Verified end-to-end on k3d. 

Point 1 (Restore CRD & deepcopy): Fair catch. The Restore CRD changes were sitting locally uncommitted, and zz_generated.deepcopy.go needs a re-gen for the pointer. I'll run make gen, verify it on the cluster, and push the update up. but the maintainer want us to prioritise others issue over it as this is not that critical.
still in future if someones or may be us want to work on it we should leave a note for them.

PR #3002, opened 2026-08-19
fix: make issue assignment checks idempotent

### LINKED ISSUE #3000: Issue assignment bot reports the cap using the issue it just assigned
### Task Description

`.github/workflows/assign.yml` is not idempotent when the same `/assign` is processed more than once for an issue. A repeated or delayed run tells the commenter they are at the assignment limit, counting the very issue it is replying about.

Observed on #2973 on 2026-08-18. @luantaraschi commented to claim it at 13:54:18Z, saw nothing for 45 seconds — the Actions queue was ~6 minutes deep — and re-issued the command twice, deleting each retry. All three runs executed after the delay:

| run | queued | ran | posted |
|---|---|---|---|
| [32145091394](https://github.com/openeverest/openeverest/actions/runs/32145091394) | 13:54:21 | 14:00:33–39 | assigned, "Assigned to @luantaraschi" |
| [32145166784](https://github.com/openeverest/openeverest/actions/runs/32145166784) | 13:55:05 | 14:00:35–41 | nothing |
| [32145225237](https://github.com/openeverest/openeverest/actions/runs/32145225237) | 13:55:39 | 14:01:15–19 | "you already have 2 open assigned issue(s), and the limit is 2" |

They held #2884 and, by then, #2973. The count was correct; the message was nonsense — it told them they were at the limit because of the issue it was posted on. They were never over it.

Two guards should have stopped the third run, and both fail:

- the early return reads `context.payload.issue.assignees`, a snapshot from when the comment was posted. Any run executing after another has assigned the issue still sees an empty list and falls through.
- the cap query `is:issue is:open assignee:${commenter}` then counts the issue being assigned, so `total_count >= cap` trips on it.

Three changes, in order of value:

1. **Read live state before the guards** — fetch the issue with `github.rest.issues.get()` and take `assignees` from that rather than the payload. A repeated or delayed command then returns silently, which is the correct outcome, and this alone closes the reported symptom.
2. **Exclude the current issue from the cap count**, so the message can never be about the issue it appears on.
3. **Serialise runs per issue** with `concurrency: { group: assign-${{ github.event.issue.number }}, cancel-in-progress: false }`. Without it two genuinely overlapping runs both pass the guards and both post "Assigned to …". Runs 1 and 2 above overlapped by four seconds, so the window is real even though it didn't fire.

### Reasoning / Benefit

The bot exists so that claiming an issue records state that protects the contributor. Here it did record the claim and then told the contributor they were over a limit they were not over, inviting them to release something to make room. A contributor who follows that advice drops work they had already started — the opposite of what the bot is for. This instance cost us a confused contributor; the failure mode is worse than the symptom.

The three changes belong in one pull request: 1 and 2 are the defect and its guard rail, 3 keeps the same class of race from coming back.

### Additional Context

- Workflow: `.github/workflows/assign.yml`
- Affected thread: #2973
- Distinct from #2947, which is about commands that produce *no* response (job-level `if` evaluating false). This one is about a response that is wrong. Same file, no overlapping lines.
- #2841 is open against the same workflow but only rewrites the post-assignment failure reply — no conflict.
- The retry comments were deleted, so the timeline had to be reconstructed from the org audit log (`action:issue_comment` filtered by actor) rather than the issue thread.

moroshani:
Fixes #3000

## Summary

- fetch the current issue before evaluating state, assignee, and label guards
- count open assignments through the issues API and exclude the issue currently being processed
- clarify the cap reply by reporting other assigned issues

## Concurrency

This intentionally does not add issue-scoped Actions concurrency. GitHub keeps at most one pending run per concurrency group, so a newer command can replace a pending command. That could silently discard a meaningful `/unassign` in a rapid command sequence. Live-state guards fix the reported delayed/repeated-run behavior without introducing that event-loss mode.

## Verification

- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/assign.yml`
- executed the embedded `github-script` with mocked GitHub APIs for: a delayed duplicate after assignment, a cap count containing the current issue and a pull request, and a genuine two-other-issue cap
- `git diff --check`

## AI assistance

This contribution was substantially implemented with OpenAI Codex assistance. I reviewed the workflow logic and diff and personally ran the verification listed above.

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -12,6 +12,8 @@ env:
 
 jobs:
   assign:
+    # Keep command events independent. Issue-scoped concurrency can replace a
+    # pending /unassign when a newer command queues.
     # Issues only; PR comments arrive on the same event.
     if: >-
       github.event.issue.pull_request == null &&
@@ -24,8 +26,7 @@ jobs:
         with:
           script: |
             const { owner, repo } = context.repo;
-            const issue = context.payload.issue;
-            const issue_number = issue.number;
+            const issue_number = context.payload.issue.number;
             const commenter = context.payload.comment.user.login;
 
             // Commands must start a line, matching the Prow convention.
@@ -54,6 +55,9 @@ jobs:
             };
 
             const commenterIsMaintainer = await isMaintainer(commenter);
+            const { data: issue } = await github.rest.issues.get({
+              owner, repo, issue_number,
+            });
 
             // Assigning somebody else is a maintainer action; otherwise anyone
             // could hand their work to a stranger.
@@ -92,14 +96,15 @@ jobs:
 
             // Maintainers bypass the gates below.
             if (!commenterIsMaintainer) {
-              const open = await github.rest.search.issuesAndPullRequests({
-                q: `repo:${owner}/${repo} is:issue is:open assignee:${commenter}`,
-                advanced_search: 'true',
-                per_page: 1,
+              const openIssues = await github.paginate(github.rest.issues.listForRepo, {
+                owner, repo, state: 'open', assignee: commenter, per_page: 100,
               });
               const cap = Number(process.env.MAX_OPEN_ISSUES);
-              if (open.data.total_count >= cap) {
-                await reply(`@${commenter} you already have ${open.data.total_count} open assigned issue(s), and the limit is ${cap}. Please finish or \`/unassign\` one first — holding several at once blocks work others could be doing.`);
+              const openCount = openIssues.filter((candidate) =>
+                !candidate.pull_request && candidate.number !== issue_number
+              ).length;
+              if (openCount >= cap) {
+                await reply(`@${commenter} you already have ${openCount} other open assigned issue(s), and the limit is ${cap}. Please finish or \`/unassign\` one first — holding several at once blocks work others could be doing.`);
                 return;
               }
 


.github/workflows/assign.yml
```diff
@@ -54,6 +55,9 @@ jobs:
             };
 
             const commenterIsMaintainer = await isMaintainer(commenter);
+            const { data: issue } = await github.rest.issues.get({
+              owner, repo, issue_number,
+            });
```
recharte, line 60: Anchoring here because the `/unassign` branch itself isn't in the diff. That branch still removes-and-confirms unconditionally, so a delayed duplicate `/unassign` posts "Unassigned @X" twice, and an `/unassign` from someone who was never assigned confirms a state change that didn't happen — that second case is real: on #3105 the bot answered "Unassigned @amh1k" eight minutes after telling amh1k their assignment had been rejected. Now that you fetch the live issue before that branch, the guard is one line, mirroring the silent return the assign path uses:

```js
if (verb === 'unassign') {
  if (!(issue.assignees || []).some((a) => a.login === target)) return; // not assigned, nothing to do
```

Silent matches the `already theirs, nothing to do` convention; a "you weren't assigned" reply would work too — the main thing is not confirming a removal that didn't happen.

.github/workflows/assign.yml
```diff
@@ -92,14 +96,15 @@ jobs:
 
             // Maintainers bypass the gates below.
             if (!commenterIsMaintainer) {
-              const open = await github.rest.search.issuesAndPullRequests({
-                q: `repo:${owner}/${repo} is:issue is:open assignee:${commenter}`,
-                advanced_search: 'true',
-                per_page: 1,
+              const openIssues = await github.paginate(github.rest.issues.listForRepo, {
```
recharte, line 99: Nice touch going to `issues.listForRepo` instead of patching the search query: the search index is eventually consistent and capped at 30 requests/min on `GITHUB_TOKEN`, while the issues list reads the primary store — so the cap can no longer be computed from a stale index either.

recharte (commented):
Hey @moroshani, thanks for the PR — and for arguing the concurrency trade-off in the description instead of silently dropping it. I agree with leaving the group out: with the queue depths we see, a per-issue group cancels older pending commands, which is a worse failure than the race it closes. We'll track the residual overlap window in a follow-up issue so #3000 can close with this PR. One note inline, nothing blocking.

PR #3010, opened 2026-08-20
fix: Harden the rate-limiter data-race regression test so it cannot s…

### LINKED ISSUE #2958: Harden the rate-limiter data-race regression test so it cannot silently no-op

Phran6ix:
- Added a `store.Allow()` to create a new visitor map.
- Added an `assert` to compare the timeout from the `IncreaseTimeout` and the expected timeout

Fixes: [#2958](https://github.com/openeverest/openeverest/issues/2958)


internal/server/session_failures_storage_test.go
```diff
@@ -31,11 +32,13 @@ func TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout(t *testing.T)
 	})
 
 	const identifier = "1.2.3.4"
-	const iterations = 200
+	const iterations = 20
 
 	var wg sync.WaitGroup
 	wg.Add(2)
 
+	store.Allow(identifier)
```
recharte: CI's `Check` job is red on this line:

```
internal/server/session_failures_storage_test.go:40:13: Error return value of `store.Allow` is not checked (errcheck)
```

Discard it explicitly, the way the goroutine just below already does.

Worth a line of *why* while you're in there. This call is the load-bearing half of the fix and it's the one thing here that looks removable — a bare call with a thrown-away result, sitting right next to a goroutine that calls `Allow` in a loop. Without a note, the next tidy-up deletes it and quietly restores exactly the no-op #2958 describes.

```suggestion
	// Register the visitor first: IncreaseTimeout silently returns when the
	// identifier has no visitor yet, so without this the writer goroutine can
	// drain every iteration without ever touching Visitor.timeout.
	_, _ = store.Allow(identifier)
```

internal/server/session_failures_storage_test.go
```diff
@@ -51,4 +54,13 @@ func TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout(t *testing.T)
...
 	wg.Wait()
+
+	expectedTimeout := initialTimeout << iterations
+
+	store.mutex.Lock()
+	v, _ := store.visitors[identifier]
+	actualTimeout := v.timeout
+	store.mutex.Unlock()
+
+	assert.Equal(t, expectedTimeout, actualTimeout)
```
recharte: The other half of the red `Check` job:

```
internal/server/session_failures_storage_test.go:61:2: S1005: unnecessary assignment to the blank identifier (staticcheck)
```

Just dropping the `, _` clears the lint but leaves the real problem: if the visitor ever isn't in the map, `v.timeout` panics instead of failing the test. Binding `ok` fixes both.

Two things in the rewrite that aren't obvious: the unlock moves *above* the assertions because a failing `require` calls `runtime.Goexit()`, which would leave `store.mutex` held for the rest of the run; and reading `v.timeout` unlocked is safe here because `wg.Wait()` has already joined both writers.

```suggestion
	store.mutex.Lock()
	v, ok := store.visitors[identifier]
	store.mutex.Unlock()

	require.True(t, ok, "visitor was removed during the run")
	assert.Equal(t, initialTimeout<<iterations, v.timeout)
```

This needs `"github.com/stretchr/testify/require"` alongside the `assert` import you added.

recharte (changes requested):
Hey @Phran6ix, thanks for picking this one up — the assertion is exactly the half of #2958 that mattered, and it's right.

Two lint failures from the `Check` job inline, both one-liners; that's the only thing holding this up.

Phran6ix:
Thank you for the review @recharte . I have made the requested changes.

internal/server/session_failures_storage_test.go
```diff
@@ -51,4 +55,14 @@ func TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout(t *testing.T)
...
+
+	expectedTimeout := initialTimeout << iterations
+
+	store.mutex.Lock()
+	v, ok := store.visitors[identifier]
+	actualTimeout := v.timeout
+	store.mutex.Unlock()
+
+	require.True(t, ok, "Visitor was removed during the run")
+	assert.Equal(t, expectedTimeout, actualTimeout)
```
recharte: `ok` is bound now, thanks — but the check ended up below the dereference it was meant to guard:

```go
v, ok := store.visitors[identifier]
actualTimeout := v.timeout                                 // nil deref here if !ok
store.mutex.Unlock()

require.True(t, ok, "Visitor was removed during the run")  // never reached
```

`v` is a `*Visitor`, so a missing key gives you nil and line 63 panics before line 66 ever runs — `require.True` can only fire in the case where it has nothing to do. Given what this PR is about, worth not shipping another guard that reads as a guard 🙂

Moving the read below the `require` means `v.timeout` is read outside the lock, which is safe here: `wg.Wait()` has already joined both writers, so there's no concurrent access left for `-race` to find.

```suggestion
	store.mutex.Lock()
	v, ok := store.visitors[identifier]
	store.mutex.Unlock()

	require.True(t, ok, "Visitor was removed during the run")
	assert.Equal(t, expectedTimeout, v.timeout)
```

internal/server/session_failures_storage_test.go
```diff
@@ -31,11 +33,13 @@ func TestRateLimiterMemoryStore_ConcurrentAllowAndIncreaseTimeout(t *testing.T)
 	})
 
 	const identifier = "1.2.3.4"
-	const iterations = 200
+	const iterations = 20
 
 	var wg sync.WaitGroup
 	wg.Add(2)
 
+	_, _ = store.Allow(identifier)
```
recharte, line 44: The errcheck half of this landed, thanks. The comment half didn't, and this is still the one line in the test that reads as safe to delete: a bare call with a discarded result, sitting directly above a goroutine that calls `Allow` in a loop. That's the exact shape of the thing #2958 is about.

```suggestion
	// Register the visitor first: IncreaseTimeout silently returns when the
	// identifier has no visitor yet, so without this the writer goroutine can
	// drain every iteration without ever touching Visitor.timeout.
	_, _ = store.Allow(identifier)
```

Phran6ix:
Hello @recharte 

The tests are failing in the plugin subsystem which I did not touch.
could this be a known issue on main or what I am supposed to do?

recharte:
> Hello @recharte
> 
> The tests are failing in the plugin subsystem which I did not touch. could this be a known issue on main or what I am supposed to do?

Yes, this was caused by PR that got merged incorrectly, it has been reverted. Please update your branch and it should be fixed.

internal/server/session_failures_storage_test.go
Phran6ix: Okay.  I understand now.
Thank you so much @recharte .

Phran6ix:
Hello @recharte 

I have updated the pr, awaiting review

recharte (approved):
Thanks @Phran6ix

Phran6ix:
Thank you so much for the help @recharte I really enjoyed working on this and i would like to continue contributing to the project. 
should I pick up another good first issue or are there particular areas of the codebase/project you'd recommend I look into?

PR #3011, opened 2026-08-20
feat: align plugin API routes with multi-cluster design

### LINKED ISSUE #3009: Generic plugins are not aligned with multi-cluster design

Right now generic are exposed via APIs with the following endpoints:
```
GET  /v1/plugins              → list enabled plugins (RBAC-filtered)
GET  /v1/plugins/context      → user identity + accessible namespaces
GET  /v1/plugins/:name/*      → serve frontend bundle (NO auth)
ANY  /v1/plugins/:name/*      → reverse-proxy to plugin backend (auth)
```

But if we want to deploy across multiple k8s clusters, we must make plugins cluster aware. Otherwise it will be unclear to which cluster to talk to.

So the endpoints must go from `/v1/plugins` to `/v1/{cluster}/plugins` - similar to what we do for instances.

Plugin endpoints must follow the multi-cluster design: go from `/v1/plugins` to `/v1/{cluster}/plugins`

_No response_

_No response_

### Additional Context

We might want to keep backward compatibility where /v1/plugins will point to the plugins in this specific k8s cluster.

Aryanbhargava18:
Right now generic plugins are exposed via APIs with endpoints like `/v1/plugins/:name/*`, which assumes a single cluster topology. As we move to multi-cluster deployments in v2, the control plane needs to know which cluster a plugin request is targeting.

This commit updates the plugin HTTP handlers to include the `:cluster` path parameter, aligning them with the resource routing convention (e.g., `/v1/clusters/:cluster/instances`).

- Changes `/v1/plugins` to `/v1/clusters/:cluster/plugins`
- Exposes cluster param in `resolvePluginAssetPath`
- Updates the bundle reverse proxy strip prefix

Resolves #3009

### DIFF
--- a/internal/server/everest.go
+++ b/internal/server/everest.go
@@ -282,13 +282,26 @@ func (e *EverestServer) initHTTPServer(ctx context.Context) error {
 		return err
 	}
 
+	// Alias for backward compatibility: rewrite legacy /v1/plugins routes
+	// to the control-plane cluster ("main", see internal/server/handlers/k8s/cluster.go).
+	const (
+		legacyPluginRewritePath         = "/v1/clusters/main/plugins"
+		legacyPluginWildcardRewritePath = "/v1/clusters/main/plugins/$1"
+	)
+	e.echo.Pre(echomiddleware.Rewrite(map[string]string{
+		"/v1/plugins/*": legacyPluginWildcardRewritePath,
+		"/v1/plugins":   legacyPluginRewritePath,
+	}))
+
 	// Plugin bundle serving — no JWT required.
 	// Bundles are static JS assets, same as the main app's JS files.
 	// Only GET requests on the wildcard path are served without auth.
-	e.echo.GET("/v1/plugins/:name/*", pp.proxyHandler)
+	// Route follows the multi-cluster convention: /v1/clusters/:cluster/plugins/:name/*
+	e.echo.GET("/v1/clusters/:cluster/plugins/:name/*", pp.proxyHandler)
 
 	// Plugin discovery & API — JWT protected.
-	pluginGroup := e.echo.Group("/v1/plugins")
+	// Routes are now cluster-scoped: /v1/clusters/:cluster/plugins, matching /v1/clusters/:cluster/instances.
+	pluginGroup := e.echo.Group("/v1/clusters/:cluster/plugins")
 	pluginGroup.Use(jwtMW)
 	pluginGroup.Use(blocklistMW)
 	pluginGroup.GET("", pp.listPluginsHandler)
--- a/internal/server/plugins.go
+++ b/internal/server/plugins.go
@@ -159,7 +159,7 @@ func (pp *pluginProxy) listPluginsHandler(c echo.Context) error {
 					Type:      ep.Type,
 					Label:     ep.Label,
 					Path:      ep.Path,
-					Icon:      resolvePluginAssetPath(p.Name, ep.Icon),
+					Icon:      resolvePluginAssetPath(c, p.Name, ep.Icon),
 					Providers: ep.Providers,
 				})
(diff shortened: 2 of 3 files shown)


spron-in (changes requested):
Can we add the old /v1/plugins/* routes back as aliases that route to the local (where control plane sits) cluster? We might need that for backward compatibility for existing plugins.

Aryanbhargava18:
> Can we add the old /v1/plugins/* routes back as aliases that route to the local (where control plane sits) cluster? We might need that for backward compatibility for existing plugins.

Added the rewrite middleware to route legacy /v1/plugins/* requests to the local cluster. Thanks for catching this.

internal/server/everest.go
```diff
@@ -282,13 +282,21 @@ func (e *EverestServer) initHTTPServer(ctx context.Context) error {
 		return err
 	}
 
+	// Alias for backward compatibility: rewrite legacy /v1/plugins routes to the local cluster
+	e.echo.Pre(echomiddleware.Rewrite(map[string]string{
+		"/v1/plugins/*": "/v1/clusters/local/plugins/$1",
+		"/v1/plugins":   "/v1/clusters/local/plugins",
```
spron-in: This seems wrong. Should it be `main` instead of `local`?
In `cluster.go:30` we have this:
```
clustersList := &api.ClusterList{
    Items: []api.Cluster{
        {
            Name:   "main",          
            Server: "https://kubernetes.default.svc",
        },
    },
}
```

and then:
```
if name != "main" {
    return nil, fmt.Errorf("cluster not found: %s", name)
}
```

@Aryanbhargava18 you did not test it, right?

spron-in (changes requested):

internal/server/everest.go
Aryanbhargava18: my bad, i took "local" literally from the previous comment instead of checking `cluster.go` where it's named `main`. Didn't have a test asserting the rewrite target either, so it slipped past.

Updated to `main`, rebased on latest `main`, and added unit tests in `plugins_test.go` covering all legacy rewrites and asset path resolution.

recharte:
Hi @Aryanbhargava18 — thanks for this, and sorry it sat a week without a reply. Your question on #3009 prompted a proper look and turned up more than the rename: the plugin endpoints were never added to `http-api.yaml`, so they're registered by hand outside the handler chain and miss everything the pipeline normally provides — details [there](https://github.com/openeverest/openeverest/issues/3009#issuecomment-5443927020). That's a leftover from the original PoC, not anything you did, but it does mean the path change is better done as part of moving the endpoints onto the chain than by hand now. I've marked this blocked while we settle the direction on #3009 — please hold off for now, and if you'd like to take the bigger piece once it's agreed, we'd be glad to have you.

Aryanbhargava18:
> Hi @Aryanbhargava18 — thanks for this, and sorry it sat a week without a reply. Your question on #3009 prompted a proper look and turned up more than the rename: the plugin endpoints were never added to `http-api.yaml`, so they're registered by hand outside the handler chain and miss everything the pipeline normally provides — details [there](https://github.com/openeverest/openeverest/issues/3009#issuecomment-5443927020). That's a leftover from the original PoC, not anything you did, but it does mean the path change is better done as part of moving the endpoints onto the chain than by hand now. I've marked this blocked while we settle the direction on #3009 — please hold off for now, and if you'd like to take the bigger piece once it's agreed, we'd be glad to have you.

That explains a lot — I was looking at the wrong layer. I saw the URL mismatch and went to rename without realising the endpoints were never on the chain to begin with.
Happy to hold off. Once the direction on #3009 is settled, I'd like to take the spec-first piece.

spron-in:
Addressed in #3104

PR #3016, opened 2026-08-22
feat: add everestctl instance update

### LINKED ISSUE #2993: Add `everestctl instance update` subcommand

Instances can be created, listed, watched, and deleted from the terminal, but not modified. Day-2 operations — scaling replicas, growing storage, enabling backups, changing an exposure type — still require the UI or hand-editing the CR with kubectl, bypassing Everest API authentication/RBAC. It is the last missing verb on the flagship resource.

Add an `everestctl instance update` subcommand. Required flags: `--name`, `--namespace`. Changes are expressed with the same `--set` dot-notation paths and `-f` values file that `instance create` already supports (same merge rules, `--set` wins over `-f`), applied on top of the instance's *current* spec via read-modify-write:

1. GET the Instance.
2. Apply `-f` then `--set` onto its spec.
3. PUT it back with the fetched `resourceVersion`; on a 409 conflict, re-fetch and reapply once, then fail with a clear message if it conflicts again.

**Merge semantics — explicit, because helm got this wrong** (`helm upgrade` resets to chart values unless `--reuse-values`, a notorious footgun):

- **Patch, not re-defaulting.** Every field not named by `--set`/`-f` keeps its current value. Update never recomputes provider/preset defaults — that would silently reset storage size, exposure, backup config when scaling replicas. There are deliberately **no** helm-style mode flags (`--reuse-values`/`--reset-values`): an Instance has one spec, so there is only one sensible base. "Reset to defaults" has no coherent meaning on update anyway (defaults were computed at create time from the version bundle/topology/preset); anyone wanting that is re-creating, not updating.
- **`-f` is a merge, not a desired-state replacement.** Fields absent from the values file stay untouched; it does not prune. "Make the spec look exactly like this file" is the GitOps path (`kubectl apply`/Argo), not this command.
- **Unsetting a field:** `--set path=null` (helm convention) removes the field, letting server-side defaulting reapply. Also works in `-f` files.

At least one of `--set`/`-f` is required — a no-op update is rejected client-side. Server-side validation (provider schema, immutable fields like `spec.provider`) is relayed as-is, not duplicated in the CLI.

Optional flags:
- `--dry-run` — perform the GET and merge locally, print a diff of the resulting spec against the current one (the merged spec as a whole in `--json` mode), and exit without writing. The safety net a patch-semantics command actually needs.

**`--wait`/`--timeout` are out of scope here.** Gating success on the rollout requires knowing whether the controller has processed the new spec, and `InstanceStatus` has no generation field to read that from today (`Backup`/`Restore`/`MonitoringConfig` have `status.lastObservedGeneration`; `Instance` doesn't, and its conditions are all set conditionally so none can stand in). Adding the field and the flag are tracked separately; this command returns as soon as the PUT is accepted.

Follows existing CLI conventions: `--json`/`--verbose` output modes, `--cluster`, `--context`.

```
# Scale the engine
everestctl instance update --name my-mongo --namespace everest \

# Preview what a change would do before applying it
everestctl instance update --name my-mongo --namespace everest \
  --set components.engine.storage.size=100Gi --dry-run

# Grow storage and enable backups in one go
everestctl instance update --name my-mongo --namespace everest \
  --set components.engine.storage.size=100Gi \
  --set backup.enabled=true

# Remove an override, falling back to the server-side default
everestctl instance update --name my-mongo --namespace everest \
  --set components.proxy.config=null

# Bulk changes from a values file (merged — absent fields stay untouched)
everestctl instance update --name my-mongo --namespace everest -f new-values.yaml
```

- `kubectl edit instance` / `kubectl patch` — bypasses Everest API authentication/RBAC and the API-layer validation chain.
- Helm-style `--reuse-values`/`--reset-values` mode flags — rejected: mode flags that switch a command's fundamental semantics are a smell, and the helm pair exists to patch around helm's own surprising default in a two-layer values model this API doesn't have.

### Additional Context

Completes create/list/status/update/delete for the `instance` group. Reuses the `--set`/`-f` merge machinery from `instance create`. Version upgrades are just `--set version=x.y.z` — no dedicated flag needed until upgrade-specific UX (e.g. sequencing hints) justifies one.

VijetaPriya47:
Instances could be created, listed, watched and deleted from the terminal but never changed, so scaling replicas or growing storage meant reaching for the UI or editing the CR by hand, which skips Everest's own auth and RBAC. This fills in the last missing verb on the instance group.

Changes are expressed the way instance create already does them, with --set dot paths and a -f values file, applied on top of the instance's current spec. It is a patch, not a desired-state replacement, so anything you do not name keeps the value it had. Underneath it is a plain read-modify-write: GET the instance, merge, PUT it back with the resourceVersion it came with, and on a conflict refetch and reapply once before giving up.

Three things the merge has to get right, all pinned by tests:

--set path=null drops an optional field so server-side defaulting takes over again. A required field has no empty form, so nulling one would quietly write its zero value instead, and --set backup.enabled=null would turn backups off. Those are rejected rather than applied.

Paths are checked against the schema before anything is written, so a typo fails instead of being dropped on the way to the server and reported as a successful update that changed nothing. Component names get the same check, since they are map keys and the decoder cannot see them.

The merged spec is decoded fresh rather than over the fetched instance. Decoding over it lets a replaced list inherit fields from the elements it replaced, so a backup schedule you just removed reattaches itself.

--dry-run prints the spec that would be sent and writes nothing. It goes through the same merge and validation as the real path, so it cannot disagree with what an actual update would do.

--wait is deliberately absent. Gating on the rollout needs a generation field that Instance status does not have yet, so it is tracked separately.

Closes #2993


VijetaPriya47:
@recharte PTAL

pkg/cli/instance/create.go
```diff
@@ -549,6 +549,9 @@ func parseSetFlags(setFlags []string) (map[string]any, error) {
 }
 
 func coerceValue(s string) any {
+	if s == "null" {
```
recharte, line 552: `coerceValue` is shared with `instance create`, which has no `checkUnset` counterpart.

Before this change, `everestctl instance create ... --set backup.enabled=null` sent `"enabled": "null"` and the server rejected it on type. Now it sends `null`, which decodes into the non-pointer `bool` as `false` — the instance is created with backups off, silently. Same for any required non-pointer field.

Simplest fix is to reject `=null` on create outright: there's nothing to unset at create time, so "`null` is only meaningful on `instance update`; omit the field instead" loses nothing. Otherwise create needs its own version of the guard.

Either way it's currently untested and undocumented — create's help says nothing about `null`, and `create_test.go` has no case for it.

commands/instance/update.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+var (
+	updateCmd = &cobra.Command{
+		Use:   "update [flags]",
+		Args:  cobra.NoArgs,
+		Short: "Update an instance",
+		Long: `Patch an existing instance's spec through the Everest API.
+
+The name and namespace flags are required. Changes are expressed with --set
```
recharte, line 37: `deepSet` splits on `.` only, so `--set backup.storages[0].pitr.enabled=true` creates a literal `"storages[0]"` key and fails as `unknown spec field "storages[0]" … check the path for typos`. It isn't a typo — indexing just isn't supported, and the advice sends the user hunting for a misspelling that isn't there.

That matters because `backup.storages` and `.schedules` are where per-element edits are the natural thing to want, and `deepMerge` replaces lists wholesale, so the only way to change one schedule today is to restate the whole list in a `-f` file.

(Commit 2's message cites `--set backup.storages[0].pitr.enabled=null` as the case its list-walking fix catches. The fix and its test are right, but it's `-f` that reaches it, not `--set`.)

Is indexing in scope here, or a follow-up? Either way I'd detect the `name[0]` shape in `parseSetFlags` and fail with something honest — "`--set` does not support list indices; use `-f` to set `backup.storages`" — and say in the `Long` help that lists are replaced wholesale.

commands/instance/update.go
```diff
@@ -0,0 +1,107 @@
...
+dot-notation paths and/or a -f values file, using the same merge rules as
+'instance create' (--set wins over -f). This is a patch, not a
+desired-state replacement: every field not named by --set/-f keeps its
+current value. At least one of --set/-f is required.
+
+--set and -f paths are validated against the instance schema before anything
+is written, so a misspelled field or component name fails instead of being
+silently ignored.
```
recharte, line 45: The component-name half is `--set`-only: `validateSetComponents` returns early unless `len(opts.Set) > 0`, so a values file with `components: {engien: {replicas: 5}}` sails past the client and lands on the server as a new component.

```suggestion
--set and -f paths are checked against the instance schema before anything is
written, so a misspelled field fails instead of being silently ignored.
Component names given with --set are checked against the provider too.
```

recharte (changes requested):
Thanks for this — the merge semantics are carefully worked out, and the three behaviours you called out in the description are each pinned by a test that would fail without the fix. The tests are behaviour-based against a real HTTP boundary, which is exactly right.

The problem is the shape, and it's on the issue rather than on you. #2993 specified read-modify-write — GET, merge locally, PUT back with the fetched `resourceVersion`, retry once on 409 — and that is precisely what you built. I think the issue was wrong.

What this command wants is JSON Merge Patch (RFC 7386), whose semantics are the ones this PR implements by hand: members absent from the patch are untouched (`deepMerge`), `null` deletes a member (`stripNulls`), arrays are replaced wholesale (`deepMerge`'s non-map branch). Sending the overrides as a single `PATCH` with `application/merge-patch+json` and letting the server apply them removes the merge, the null-stripping, the `resourceVersion` round-trip, the 409 and the retry loop. The apiserver already does that read-modify-write internally, with its own conflict retry — the logic isn't disappearing, it's moving to the one place that can do it correctly.

It also removes a bug that is hard to see in the current shape. `getInstance` returns `resp.JSON200`, which the generated client decodes with a plain `json.Unmarshal` — no `DisallowUnknownFields`. Any spec field this everestctl doesn't know about is dropped at the GET, `prepareSpec` marshals the truncated struct, and the PUT writes the truncation back. An older CLI against a cluster whose Instance CRD has gained a field would scale the engine on `--set components.engine.replicas=5` and silently erase that field. With PATCH the server never receives a whole object, so there is nothing to truncate. The required-field guard gets better too: `{"backup":{"enabled":null}}` fails CRD validation server-side, where the schema is authoritative, rather than being approximated against compiled-in types.

@VijetaPriya47 — the API doesn't expose PATCH on instances today, so I've filed #3019 for it, and I'd like you to take that one first and then rebuild this command on top of it. You mentioned wanting to get into the API side, and this is a good piece of it: one new operation, the existing validation → RBAC → Kubernetes chain, and a couple of sharp details worth getting right (strict field validation, and not modelling it on `patchBackupStorage`, which isn't a real merge patch). `/assign` on the issue will claim it. Worth flipping this PR to draft in the meantime so it's clear it isn't waiting on review.

I've also commented on #2993 with the revised design, so the issue no longer asks you for the RMW version. I'd rather stop here than have you refine that path, because fixing it properly means writing code we would delete as soon as the API has the verb.

One thing that changes either way: I'd drop `--dry-run`'s two full specs in favour of the paths that change — `components.engine.replicas: 3 -> 5`, one line each — with `--json` emitting the patch document. #2993 asked for a diff and that was my mistake in writing it: the change list is the command's contract rendered, where a diff of two whole specs buries it. It also avoids the CLI needing a local merge-patch applier purely to render a preview.

The inline comments are independent of all of the above and stand either way.

PR #3047, opened 2026-08-25
fix: handle Kubernetes milli-byte suffix in memoryParser

### LINKED ISSUE #2423: UI: Resource details showing large numbers for Memory requests and limits in overview

jadhavgaurav:
## Summary

The database overview page shows nonsensical memory numbers (or `NaN`) for
resource requests/limits whenever Kubernetes returns a non-integer memory
quantity, e.g. `"644245094400m"` for a 0.6G limit that isn't exactly
representable in any larger unit.

## Root cause

`memoryParser` (`ui/apps/everest/src/utils/k8ResourceParser/index.ts`) has a
unit table (`memoryMultipliers`) covering `k/M/G/T/P/E` and their binary
(`Ki/Mi/...`) equivalents, but no entry for Kubernetes' milli (`m`) suffix.
When a memory quantity round-trips through Kubernetes as a milli-byte value,
the parser looks up `memoryMultipliers['m']`, gets `undefined`, and the
subsequent multiplication produces `NaN` — which is what actually renders,
not just a "large number."

`cpuParser` in the same file already handles the identical `m` suffix
correctly for CPU quantities (dividing by 1000); `memoryParser` never got
the equivalent case.

## Fix

Add `m: 10 ** -3` to `memoryMultipliers` (and `'m'` to the `kubernetesUnit`
type), so a milli-byte quantity flows through the existing
value-times-multiplier-then-divide-by-target-multiplier logic instead of
hitting a missing table entry. No other logic changes.

## Tests

- Added a regression test in `k8ResourceParser.test.ts` asserting
  `memoryParser('644245094400m', 'G')` returns `{ value: 0.6442450944,
  originalUnit: 'm' }`.
- `npx vitest run k8ResourceParser` — 8/8 passing (was 7, +1 new).
- `npx eslint` on both changed files — clean.
- `npx prettier --check` on both changed files — clean.
- `npx tsc --noEmit` — no new errors introduced by this change (the app has
  pre-existing, unrelated `@openeverest/plugin-sdk` module-resolution errors
  from an unbuilt workspace package; confirmed identical with and without
  this diff via `git stash`).
- Reproduced the bug directly (a small Node script computing what the
  current regex/lookup produces for the exact value from the issue) before
  writing the fix, to confirm the `NaN` root cause rather than assume it.

## Related issue

Fixes #2423

## AI-assisted contribution disclosure

This fix (root-cause analysis, code change, and test) was produced with AI
assistance (Claude Code) and reviewed by me before submitting. I can explain
and defend the change: `cpuParser`'s existing `m`-suffix handling is the
direct precedent this mirrors for memory.


ui/apps/everest/src/utils/k8ResourceParser/index.ts
```diff
@@ -13,6 +14,9 @@ type kubernetesUnit =
   | 'Ei';
 
 const memoryMultipliers: Record<kubernetesUnit, number> = {
+  // Kubernetes' milli suffix: quantity is 1/1000th of a byte, used when a
```
areebahmeddd, line 33: i don't think this reaches the bug in #2423 on `main`.

`memoryParser` has only one non-test consumer in the ui:

```text
ui/apps/everest/src/pages/db-cluster-details/cluster-overview/old-cards/resources-details.tsx
```

but that component isn't rendered on v2. it's live on v1.x so you should re target 🤔

jadhavgaurav:
Retargeted to `v1.x` per your review — confirmed `resources-details.tsx` (the only non-test consumer of `memoryParser` this fix affects) lives at `ui/apps/everest/src/pages/db-cluster-details/cluster-overview/cards/resources-details.tsx` on v1.x and isn't present on main/v2. Re-applied the same fix + test against v1.x's version of `k8ResourceParser`; 21/21 tests passing (`npx vitest run k8ResourceParser`). Thanks for catching that.

areebahmeddd:
1. DCO failing
2. Just so i know you're not just using ai to one-shot this, could you reproduce the bug or show that it's fixed in the UI?

solovevayaroslavna:
@areebahmeddd thanks for the review — addressing both points:

**1. DCO** — the failing check was from an earlier force-push; the branch is now rebased on the latest `v1.x` and every commit is signed off, so DCO should be green on this run.

**2. Reproduced in the UI (v1.x).** Steps:
1. Local `v1.x` dev env, created a MongoDB cluster, set **Memory = `0.6Gi`** on the engine resources.
2. Kubernetes can't represent `0.6Gi` (= 644245094.4 bytes) as an integer, so it stores it as the milli-byte quantity **`644245094400m`** — the exact value from #2423. The PSMDB operator even logs it:
   ```
   spec.template.spec.containers[0].resources.limits[memory]:
   fractional byte value "644245094400m" is invalid, must be an integer
   ```
3. Opened the database **overview → Resources details** and compared with/without this patch.

**Before (without the fix)** — `memoryParser` looks up `memoryMultipliers['m']` → `undefined` → the multiplication yields `NaN`:

> Memory: **NaN GB / NaN GB**

before-nan

**After (with `m: 10 ** -3`)** — the milli-byte quantity is parsed correctly:

> Memory: **0.6442450944 GB**

after-fixed


So the one-line multiplier entry (mirroring the `m` handling `cpuParser` already has) fixes the render, and the added unit test locks in the `644245094400m → 0.6442450944` conversion.

PR #3049, opened 2026-08-25
fix(version-service): sort supported engine versions by semver, not string

### LINKED ISSUE #3048: GetSupportedEngineVersions sorts version strings lexicographically instead of by semver
### What is the issue?

`GetSupportedEngineVersions` in `pkg/version_service/client.go` sorts the returned engine version strings with a plain lexicographic string sort instead of a semver-aware one, so the list it returns is not actually in version order once any version reaches a two-digit component.

```go
// pkg/version_service/client.go, GetSupportedEngineVersions()
result := make([]string, 0, len(versions))
for ver := range versions {
    if _, err := goversion.NewVersion(ver); err != nil {
        return nil, err
    }
    result = append(result, ver)
}

// For PXC, we need to remove the 5.x versions from the list
if operator == PXCOperatorName {
    result = slices.DeleteFunc(result, func(v string) bool {
        semver, _ := goversion.NewVersion(v)
        return semver.Segments()[0] == 5 //nolint:mnd
    })
}
slices.Sort(result)
return result, nil
```

Every entry is already parsed with `goversion.NewVersion(ver)` to validate it, but the parsed `*goversion.Version` is discarded and `slices.Sort` is called on the raw `[]string` instead, which does a byte-wise comparison.

### Relevant Logs / Output

N/A. Traced through the code, not reproduced against a live cluster.

### Steps to Reproduce

No cluster needed, this is pure string comparison:

1. Feed `GetSupportedEngineVersions` a version map with keys `["8.0.4", "8.0.32", "8.0.9"]` (e.g. a PSMDB `matrix.mongod` response).
2. Byte-wise, `'3' < '4'`, so `"8.0.32"` sorts before `"8.0.4"` even though 8.0.32 is the larger version numerically.
3. `slices.Sort` produces `["8.0.32", "8.0.4", "8.0.9"]`, which is neither ascending nor descending semver order.

The same failure triggers for any operator (PXC, PSMDB, PostgreSQL) once a patch or minor component reaches two digits, e.g. PXC `8.0.9` vs `8.0.10`.

`pkg/version_service/` has no test file at all (only `client.go`, `gen.go`, `mock_interface.go`), so nothing in the package's own suite exercises the sort.

### Relevant Logs / Output

_No response_

### OpenEverest Version

v1.x

### Installation Method

CLI (everestctl)

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Sort by the already-parsed semantic version instead of the raw string, e.g.:

```go
slices.SortFunc(result, func(a, b string) int {
    va, _ := goversion.NewVersion(a)
    vb, _ := goversion.NewVersion(b)
    return va.Compare(vb)
})
```

(or build a `[]*goversion.Version` alongside `result` during the first loop, sort that with `.Compare`, then map back to strings, avoiding the re-parse).

### Additional Context

The one current caller, `preflightCheckDBEngineVersion` in `internal/server/handlers/k8s/database_engine.go`, only uses the returned slice to compute a running minimum via `goversion.Version.LessThan`, which is order-independent, so nothing user-visible breaks today. That is also why this has stayed unnoticed: the bug is latent rather than active. But `GetSupportedEngineVersions` is exported on `versionservice.Interface`, its doc comment promises "a list of supported versions" with no ordering caveat, and the code explicitly sorts the result, so any future caller that trusts ascending order (e.g. "pick the highest supported patch", or surfacing the list directly in a UI dropdown) will get a silently wrong answer.

Found while looking for further issues in the codebase after fixing the major-version upgrade guard in `pkg/cli/upgrade/upgrade.go`.

Bhupesh-081:
## Summary

`GetSupportedEngineVersions` validated each version string with `goversion.NewVersion()` but then sorted the raw `[]string` with `slices.Sort`, which is a byte-wise comparison, not a semver-aware one. Once any component reaches two digits (e.g. `"8.0.32"` vs `"8.0.4"`), the result is neither ascending nor descending semver order. Fixes #3048.

## Changes

`pkg/version_service/client.go`: replace `slices.Sort(result)` with `slices.SortFunc` using `goversion.Version.Compare`, reusing the already-validated parse pattern from the loop above it. `pkg/version_service/client_test.go`: new test asserting `["8.0.4", "8.0.32", "8.0.9"]` comes back as `["8.0.4", "8.0.9", "8.0.32"]` this package had no test coverage before.

## Test plan

- [x] `go test ./pkg/version_service/...` new test passes with the fix, and fails with the old `slices.Sort` (verified by temporarily reverting)
- [x] `go build` / `go vet` on the changed package
- [x] `go test` on known callers (`internal/server/handlers/k8s`, `pkg/cli/...`)  unaffected, still green
- [x] `gofmt -l` clean on changed files


lekaf974 (approved):
Logic LGTM

PR #3052, opened 2026-08-25
fix(ui): hide provider version in database tile

### LINKED ISSUE #3051: Do not show version of the provider in a tile
### Task Description

On the main screen we show the version of the provider in a tile.
It does not make sense and it can be the source of confusion for the user.

Do not show this version.

Image

### Reasoning / Benefit

Remove and simplify

### Additional Context

_No response_

jadhavgaurav:
## Problem

On the "add database" empty state, the provider tile shows the provider's
version string in its footer. As noted in the issue, this isn't meaningful
to users on that screen and only adds confusion.

## Fix

`ui/apps/everest/src/components/empty-state-databases/provider-tiles/provider-tile/provider-tile.tsx`:
stop destructuring/rendering `meta.version` in the tile footer, and always
right-align the forward-arrow icon in the footer instead of conditionally
switching layouts based on whether a version was present.

`ui/apps/everest/src/components/empty-state-databases/provider-tiles/provider-tiles.constants.ts`:
the two footer style variants (`cardFooterSx` for the version case,
`cardFooterEndSx` for the no-version case) collapse into a single
`cardFooterEndSx`, since there's now only one footer layout. Removed the
now-unused `cardFooterSx` export and its shared base rather than leaving
dead code.

No other logic changes; the `version` field itself remains on the shared
`ResolvedExtensionMeta` type since it may still be used elsewhere.

## Tests

- No existing test file covers this component
  (`provider-tile.tsx`/`provider-tiles.constants.ts` have no
  `.test.tsx`/`.spec.tsx`), so no test was added or updated.
- `npx eslint` on both changed files — clean.
- `npx prettier --check` on both changed files — clean.
- `npx tsc --noEmit` — no new errors introduced by this change (the app has
  11 pre-existing, unrelated `@openeverest/plugin-sdk` module-resolution
  errors from an unbuilt workspace package; confirmed identical count with
  and without this diff via `git stash`).

## Related issue

Fixes #3051

## AI-assisted contribution disclosure

This fix (investigation, code change) was produced with AI assistance
(Claude Code) and reviewed by me before submitting. I can explain and
defend the change: it removes the version display and folds the two footer
style variants into one, since only the no-version layout is needed now.

### DIFF
--- a/ui/apps/everest/src/components/empty-state-databases/provider-tiles/provider-tile/provider-tile.tsx
+++ b/ui/apps/everest/src/components/empty-state-databases/provider-tiles/provider-tile/provider-tile.tsx
@@ -29,7 +29,6 @@ import {
   bodyAnchorSx,
   cardActionAreaSx,
   cardContentSx,
-  cardFooterSx,
   cardFooterEndSx,
   cardHeaderSx,
   cardSx,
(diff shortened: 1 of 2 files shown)


spron-in (approved):
thanks! lgtm

PR #3053, opened 2026-08-25
feat(provider-runtime): add a conformance check that UI fields are reconciled

recharte:
## Problem

A topology UI schema binds form controls to Instance fields by path. [provider-sdk#36](https://github.com/openeverest/provider-sdk/pull/36) now proves those paths *exist*; nothing proves the provider does anything with them. A field can be offered in the UI, accept input, and be silently dropped — which is worse for a user than not offering it at all.

## What this adds

`provider-runtime/conformance`, called from a provider's own test suite:

```go
func TestUISchemaIsReconciled(t *testing.T) {
        conformance.UISchemaIsReconciled(t, conformance.Config{
                Provider: NewMyProvider(),
        })
}
```

For every path the **generated Provider CR** binds a control to — the artifact that ships in the chart and that the UI actually reads — it renders `Sync()` offline against a fake client with a recognisable value set, and counts the field as consumed if any of these hold:

| evidence | proves |
|---|---|
| the value reaches an object the provider applied | it reached the engine CR |
| the value appears in a request the provider made | the provider read the field |
| the render differs from a baseline with the field unset | the field changed behaviour even though its value never appears literally |

The second matters for reference-valued fields: `components.monitoring.parameters.monitoringConfigName` never reaches the engine CR, but the provider fetching a `MonitoringConfig` by that name is conclusive. Evidence collected before a failed `Sync` still counts — the request happened regardless of what came after.

The third matters for fields whose value is not a value in the output: a shard count changes *how many* objects exist rather than appearing in one.

## No per-provider fixtures

The Instance is built from the topology the Provider CR declares, and optional components are left out unless the probed path addresses one. Types come from `BaseProvider.SchemeFuncs`, and `parameters.*` paths are typed from the `openAPIV3Schema` already resolved into the Provider CR, so nothing needs the provider's Go types.

Verified against PSMDB: **zero fixtures**, and no `SetClient` wiring — `p.client` is only used by watch mapping, never by `Sync`.

## Faithfulness

The harness renders the Instance **the reconciler would hand to `Sync`**, not the one the user wrote. That matters: the runtime fills each `ComponentSpec.Version` from the version bundle before `Sync` runs, so `spec.version` is consumed by the runtime rather than by any provider. Rendering `Sync` directly reported `spec.version` as unreconciled on every provider — a false positive produced by the harness, not by the providers.

The first commit therefore moves that preparation out of the reconcile loop into `provider-runtime/internal/instanceprep`, which the reconciler and the harness both call. A preparation step added there reaches both by construction rather than by remembering. It is internal deliberately: providers are handed an already-prepared Instance and have no reason to call it.

For the same reason `spec.version` is probed with a real non-default bundle rather than an invented token: the Provider CR enumerates the legal values. A provider declaring a single bundle cannot have the field exercised, and is told so.

## Base

Stacked on #3054, which exports the version bundle selection order so that the four copies of it in provider-percona-postgresql and provider-percona-xtradb-cluster have something to converge on. `instanceprep` calls that instead of keeping a copy of its own. GitHub will retarget this branch to `main` once #3054 merges.

## Known limits

- Only `Sync` is exercised. Fields consumed elsewhere are invisible.
- Other enum-valued fields still get an arbitrary sentinel, so a provider that validates them may fail the render and report *unverified* rather than a verdict.
- `Config.Unverifiable` exists for values a provider reshapes beyond recognition. It takes a mandatory reason and should be reviewed like any other code — it is not a way to silence a genuinely unhandled field.

### DIFF
--- a/provider-runtime/conformance/resolve.go
+++ b/provider-runtime/conformance/resolve.go
@@ -1,0 +1,217 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conformance
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+
+	corev1alpha1 "github.com/openeverest/openeverest/v2/api/core/v1alpha1"
+	"github.com/openeverest/openeverest/v2/provider-runtime/controller"
+)
+
+// ProviderUnderTest is the contract a conformance run exercises.
+type ProviderUnderTest = controller.ProviderInterface
+
+// leafKind is the JSON shape of the value a UI control writes.
+type leafKind string
+
+const (
+	leafString   leafKind = "string"
+	leafInteger  leafKind = "integer"
+	leafNumber   leafKind = "number"
+	leafBool     leafKind = "boolean"
+	leafQuantity leafKind = "quantity"
+	leafUnknown  leafKind = ""
+)
+
+// quantityTypeName is a struct in Go but a string on the wire.
+const quantityTypeName = "Quantity"
+
+// componentsSegment is the map field every per-component path goes through.
+const componentsSegment = "components"
+
+// resolveLeaf reports the shape of the value a path addresses. Paths above a
+// parameters payload are resolved against the Instance types; below one they
+// are resolved against the schema the provider declared, since those types live
+// in the provider's own module.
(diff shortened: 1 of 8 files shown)


provider-runtime/conformance/resolve.go
```diff
@@ -0,0 +1,217 @@
...
+		return "everest-conformance-7919", "everest-conformance-7919", nil
+	case leafQuantity:
+		// Must parse as a quantity, so the magnitude carries the signal.
+		return "7919", "7919", nil
+	case leafInteger:
+		return int64(7919), "7919", nil
+	case leafNumber:
+		return 79.19, "79.19", nil
+	case leafBool:
+		return true, "true", nil
```
chilagrow: Is there a case where provider sets the default value as true, bypassing this check entirely?

chilagrow (dismissed):
Thanks for this. One tiny comment

provider-runtime/conformance/resolve.go
recharte: No — `probePath` renders a baseline first and bails if the token is already present ("the probe value already appears without being set"), and that outcome is counted as a failure, not a note. So a provider that hardcodes `true` can't get a silent pass.

It does expose a real problem in the other direction, though. `contains` is a substring match over the serialized objects, so `true` matches any boolean anywhere, and that guard will fire on almost every boolean control — the first person to bind a checkbox gets a failure they can't fix by fixing their code, only by adding it to `Config.Unverifiable`. Nothing binds a boolean today (the `uiType`s in use are `number`, `select`, `group` and `text`), so it's latent. The fix is to drop the token for bools and use the baseline comparison alone, probing both `true` and `false` and calling the field reconciled if either render differs. I'll fix this.

PR #3054, opened 2026-08-25
refactor(runtime): share the version bundle selection order

recharte:
The order that decides which version bundle is in force for an Instance — `spec.version`, then `status.version`, then the Provider's default — is currently reimplemented in four places outside the runtime: `selectedVersionBundleName` in provider-percona-postgresql, and three separate inlined copies in provider-percona-xtradb-cluster (`imageForBundledProxy`, `monitoringImageForComponent`, and `backup_mirror.go`).

Every copy has to track the runtime by hand. That matters more than ordinary duplication does, because the runtime resolves the bundle *before* calling `Sync`: a provider that picks a different one renders images for a version the Instance is not actually running, and nothing catches the divergence.

This exports the selection as `controller.EffectiveVersionBundleName(spec, in)` and has the reconciler call it, so those copies have something to converge on. The signature is the one all four call sites already need — PG's helper and PXC's three copies each reduce to `controller.EffectiveVersionBundleName(spec, c.Instance())`.

Only the *selection* is exported. Applying a bundle to an Instance stays in the reconciler, since no provider does that.

### Tests

The order had no test of its own: there is no reference to `resolveVersionBundle` or `GetDefaultVersionBundleName` in any test file under `provider-runtime/`. This adds a table test for the precedence, including the case that carries the most weight — `status.version` outranking a default that has since changed, which is what stops a Provider upgrade from silently moving existing Instances onto a new bundle.

### Relationship to #3053

This was extracted from #3053 so the provider-side cleanups are not gated on it. The two are independent: #3053's net diff does not touch `provider-runtime/controller/` at all. I test-rebased a squashed #3053 onto this branch — one conflicted file, both hunks inside the function this PR touches, resolved by taking #3053's side, after which `instanceprep`'s private copy is deleted in favour of the exported one and the module builds with all runtime tests green.

### Follow-ups

Deleting the copies in provider-percona-postgresql and provider-percona-xtradb-cluster, once this is merged and there is a SHA to pin.

### DIFF
--- a/provider-runtime/controller/metadata.go
+++ b/provider-runtime/controller/metadata.go
@@ -167,6 +167,28 @@ func GetDefaultVersionBundleName(spec *v1alpha1.ProviderSpec) string {
 	return ""
 }
 
+// EffectiveVersionBundleName returns the bundle in force for an Instance:
+//
+//  1. spec.version, when the user chose one.
+//  2. status.version, frozen on the first reconcile so that upgrading a
+//     Provider does not silently move existing Instances to a new bundle.
+//  3. the Provider's default bundle.
+//
+// Returns empty string when the spec declares no bundles at all. This is the
+// same order the runtime applies before calling Sync, so a provider that needs
+// the bundle itself — to read component versions the runtime does not resolve
+// on its behalf — agrees with the runtime by construction.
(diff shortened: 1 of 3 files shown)


chilagrow (approved):
Thank you!

PR #3055, opened 2026-08-26
feat(api): add PATCH support to the Instance API

### LINKED ISSUE #3019: Add PATCH support to the Instance API

The Instance API exposes GET, PUT and DELETE. Changing a single field therefore means read-modify-write: fetch the whole Instance, edit it client-side, PUT it back. That pushes three problems onto every client:

- **A lost-update window.** RMW is only safe with `resourceVersion` plus a 409 retry loop, which every client has to implement and get right independently.
- **Merge semantics reimplemented client-side.** "Change these fields, leave the rest alone" is RFC 7386 JSON Merge Patch. Without a PATCH verb each client hand-rolls it — including the `null`-deletes-a-member rule and wholesale array replacement — and they will not agree with each other.
- **Silent truncation by typed clients.** A generated client decodes the GET non-strictly, so any spec field its types don't model is dropped; the PUT then writes that truncation back. An `everestctl` older than the server would erase fields it has never heard of, with no error.

`everestctl instance update` (#2993, #3016) is the first consumer to hit all three, but the UI does full-object PUTs for the same reason and carries the same exposure.

Add `PATCH /clusters/{cluster}/namespaces/{namespace}/instances/{instance}` accepting `application/merge-patch+json` (RFC 7386), forwarded to the Kubernetes API as `types.MergePatchType`, returning the patched Instance with 200 as `PUT` does.

The apiserver applies the patch to the stored object and retries internally on conflict, so the read-modify-write logic isn't disappearing — it moves to the one place that can do it correctly.

Requirements:

- **Forward with `fieldValidation=Strict`.** Not optional: without it the apiserver prunes members absent from the CRD's structural schema, so a misspelt path returns 200 having changed nothing — precisely the failure a patch API exists to prevent. `controller-runtime` exposes this as `client.FieldValidation("Strict")` on the patch call.
- **Support conditional patches.** A patch document carrying `metadata.resourceVersion` is treated by the apiserver as a precondition and fails with 409 on mismatch. Clients that need optimistic concurrency keep it; clients that don't stop paying for it on every call.
- **Constrain what the patch may name.** A merge patch can address any member, so `ownerReferences`, `finalizers`, `metadata.name` and `metadata.namespace` must be rejected rather than relayed — the same scrubbing `AGENTS.md` requires of the write path ("a proxy has no implicit DTO whitelist"), applied to the patch document instead of a whole object. `status` is already carved out by the CRD's status subresource, so rejecting it is belt-and-braces rather than the defence.
- **Reject unsupported media types with 415**, as the apiserver does, rather than silently treating an unknown `Content-Type` as a merge patch.
- **Same handler chain as PUT** — validation → RBAC → Kubernetes — with patch authorised as an update on `instances` rather than as a new permission.
- OpenAPI-first: add the operation to `api/openapi/http-api.yaml` and regenerate.

**Do not model this on `patchBackupStorage`.** That endpoint decodes the request into a full typed object, fetches the current one and copies non-zero fields onto it. Under those semantics a field cannot be set to `false`, `0` or `""` and nothing can be deleted, because the zero value is indistinguishable from "not supplied". It is precedent for the verb existing, not for the behaviour. Worth a follow-up to bring it onto real merge-patch semantics too.

```
PATCH /v1/clusters/main/namespaces/everest/instances/my-mongo
Content-Type: application/merge-patch+json

{"spec": {"components": {"engine": {"replicas": 5}}}}
```

Unsetting a member, so server-side defaulting reapplies:

```
{"spec": {"components": {"proxy": {"parameters": null}}}}
```

`everestctl instance update` would build exactly this document from its `--set`/`-f` flags, and `--dry-run` would print it.

- **Keep read-modify-write over PUT** — the status quo, with the three problems above.
- **Strategic merge patch** — not available. The apiserver implements it for built-in types only; custom resources support JSON Patch, JSON Merge Patch and Apply Patch.
- **JSON Patch (RFC 6902)** — more expressive, notably for list-index operations, but positional and awkward to generate from `--set` dot paths. Merge patch matches what the CLI and UI actually need, and RFC 6902 can be added later as a second content type without breaking anything.
- **Server-Side Apply (`application/apply-patch+yaml`)** — the strongest option long-term (field managers, declared ownership, real conflict detection), but it brings an ownership model the API has no story for yet. Worth revisiting on its own; merge patch is the smaller step that unblocks the CLI now.

### Additional Context

Blocks the redesign of #2993 and the review on #3016.

VijetaPriya47:
The Instance API exposed GET, PUT and DELETE only, so changing a single field meant read-modify-write: fetch the whole Instance, edit it client side, PUT it back. That pushed three problems onto every client. It leaves a lost-update window that is only safe with a resourceVersion precondition and a retry loop. It makes each client hand-roll "change these fields, leave the rest alone", and they do not agree with one another. And a typed client older than the server drops the fields its types do not model, then writes that truncation back, erasing settings it never knew existed.

Add PATCH accepting application/merge-patch+json, forwarded to the Kubernetes API as types.MergePatchType. The API server already applies merge patches and retries internally on conflict, so the read-modify-write does not disappear, it moves to the one place that can do it correctly. A patch carrying metadata.resourceVersion is treated as a precondition and fails with 409, so callers that want optimistic concurrency keep it and the rest stop paying for it on every call.

Forward with fieldValidation=Strict. Without it the API server prunes members absent from the CRD schema, so a misspelt path returns 200 having changed nothing, which is the exact failure a patch API exists to prevent.

A merge patch can address any member of the stored object, so the validation handler rejects status, metadata.ownerReferences, metadata.finalizers, metadata.name and metadata.namespace. It also rejects the two actor annotations, because the typed verbs stamp those after decoding a request and a caller therefore cannot author them there, whereas a patch is relayed as sent and could.

The actor is stamped, as create, update and delete already do, so last-actor-id names whoever last changed the instance rather than whoever last used a typed verb on it. A patch has no object to stamp, since the patch bytes are the request body, so stampActor fills a throwaway ObjectMeta and its annotations are merged into the document. Together with the rejection above, the value can only come from the caller's token.

Patching is authorised as an update on instances rather than as a permission of its own.

The generated client names a request body after its media type, so this operation's method would be PatchInstanceWithApplicationMergePatchPlus JSONBodyWithResponse. That name is fixed inside oapi-codegen and cannot be changed from the spec or its config, so client/instance_patch.go wraps it as PatchInstanceWithResponse, matching the shape every other patch operation exposes. Being a normal file, it survives regeneration.

Three adjustments to shared error handling came with it. An undeclared Content-Type now answers 415 rather than 400, since the OpenAPI validator reports every request body problem as 400 and that left an unsupported media type indistinguishable from a malformed body. A 409 now carries the API server's explanation instead of rendering a null body, which had left a client with nothing to show for a failed precondition. And a field-validation rejection is trimmed to the fields it names, since the API server prefixes it with the whole submitted object, managedFields included

Closes #3019


VijetaPriya47:
@recharte PTAL

client/instance_patch.go
recharte, line 1: I checked the naming claim and it holds — oapi-codegen sets the unsuffixed variant only for exactly `application/json`, everything else gets `With<CamelCasedMediaType>Body` derived from the content-type string, with no extension or config to override it. So the ugly name is unavoidable and a wrapper is the only way to hide it.

The question is where it lives. This is the first non-generated file in `client/`, and the package's useful property today is that you can delete it and `make gen` puts it back exactly. Nothing calls `PatchInstanceWithResponse` yet — the caller is `everestctl instance update` — and the CLI already calls the equivalently awkward `CreateInstanceWithBodyWithResponse` head-on in `pkg/cli/instance/create.go`. Where the repo does adapt the generated client, it does it in a sibling package: `pkg/cli/clienterr`.

Would you mind moving it to the CLI PR, either as a direct call or as a helper next to `clienterr`?

The one thing that gives me pause the other way: `client/everest-client.gen.go:7740` already documents `ParsePatchInstanceResponse` as parsing "a `PatchInstanceWithResponse` call" — a method the generator never emits — so there's an argument the wrapper is filling a hole oapi-codegen left. Happy to be talked into keeping it.

internal/server/handlers/validation/instance.go
```diff
@@ -44,6 +48,37 @@ func (h *validateHandler) UpdateInstance(ctx context.Context, cluster string, in
...
+	}
+	if metadata, isObject := doc["metadata"].(map[string]any); isObject {
+		for _, member := range []string{"ownerReferences", "finalizers", "name", "namespace"} {
+			if _, found := metadata[member]; found {
+				return nil, errors.Join(ErrInvalidRequest, fmt.Errorf("metadata.%s may not be patched", member))
+			}
+		}
+		// The typed verbs stamp these after decoding, so a caller cannot author
+		// them there. A merge patch is relayed as sent, so it could.
+		if annotations, isObject := metadata["annotations"].(map[string]any); isObject {
```
recharte: One document gets past here and then means something different by the time it lands. `{"metadata":{"annotations":null}}` is "remove every annotation" under RFC 7386, but the stamp merge in the k8s handler turns it into `{"metadata":{"annotations":{"openeverest.io/last-actor-type":…,"openeverest.io/last-actor-id":…}}}`, which merges two annotations and leaves the rest in place. The caller gets a 200 for a request that did close to the opposite of what it asked — the silent no-op `fieldValidation=Strict` is here to prevent. (Attribution survives, since the stamp goes back in, so this isn't a hole — just a lie to the caller.)

It's narrow, and I'm not sure anyone wants to wipe an Instance's annotations, so rejecting it is probably the cheapest answer — `annotations` being `null` could join the actor keys below. What do you think?

internal/server/handlers/k8s/instance_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	corev1alpha1 "github.com/openeverest/openeverest/v2/api/core/v1alpha1"
+	"github.com/openeverest/openeverest/v2/pkg/common"
+	"github.com/openeverest/openeverest/v2/pkg/events"
+	"github.com/openeverest/openeverest/v2/pkg/kubernetes"
+)
+
+// TestK8s_PatchInstance pins the merge semantics, so a changed patch type or a
+// dropped option fails here.
```
recharte: This claims more than the test does — I deleted `ctrlclient.FieldValidation(metav1.FieldValidationStrict)` from `PatchInstance` and all four subtests still passed. The fake client has no strict decoding, so nothing here can observe the option, and strict is the one requirement #3019 calls out as not optional.

Same gap for the other two behaviours the design rests on: 415 for an undeclared `Content-Type` is produced inside kin-openapi's validator, and 409 for a stale `metadata.resourceVersion` is an API server precondition. Neither is reachable from a unit test.

`api-tests/tests/instance-preset.spec.ts` already creates and deletes real Instances through the API, so an `instance.spec.ts` beside it could cover all three: a misspelt path getting 400 rather than a silent 200, `Content-Type: application/json` getting 415, and a stale `resourceVersion` getting 409. That last one matters double, because the 415 mapping keys off the literal `"header Content-Type has unexpected value"`, which kin-openapi keeps unexported (`openapi3filter/validate_request.go:256`) — nothing will fail at compile time when they reword it.

Happy either way on timing: add it here, or land the endpoint and do it alongside the CLI. Either way I'd drop the claim about options being pinned from this comment.

internal/server/everest.go
```diff
@@ -563,14 +564,19 @@ func everestErrorHandler(next echo.HTTPErrorHandler) echo.HTTPErrorHandler {
...
+					Message: trimStrictDecodingError(trimWebhookErrorText(statusError.Status().Message)),
 				}
 			}
 		case k8serrors.IsAlreadyExists(err),
 			k8serrors.IsConflict(err):
-			err = &echo.HTTPError{
-				Code: http.StatusConflict,
+			// A bare 409 renders as a null body, leaving a client with nothing
+			// to show for a failed resourceVersion precondition.
+			httpErr := &echo.HTTPError{Code: http.StatusConflict}
```
recharte, line 593: This branch also catches `IsAlreadyExists`, so every create endpoint that can conflict now returns the API server's text — `instances.core.openeverest.io "db1" already exists` — where it previously rendered `{"message":null}`. That looks like a strict improvement to me and nothing in CI objects; just confirming the widening was intended rather than a side effect of fixing the precondition case the comment describes.

internal/server/handlers/k8s/instance.go
```diff
@@ -18,10 +18,13 @@ package k8s
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 
+	jsonpatch "github.com/evanphx/json-patch"
```
recharte: `github.com/evanphx/json-patch/v5 v5.9.11` is already in the module graph (indirect, same upstream version), so this promotes the legacy `+incompatible` path to a direct dependency while the maintained one is right there. `MergePatch` has the same signature, and I ran both over the documents this actually sees — same result, `null` members in the caller's patch included (key ordering differs, which the API server doesn't care about).

```suggestion
	jsonpatch "github.com/evanphx/json-patch/v5"
```

That also drops the new `github.com/evanphx/json-patch v5.9.11+incompatible` line from the direct block in `go.mod`.

recharte (commented):
Hey @VijetaPriya47, thanks for the PR — the handler chain itself looks solid. Nothing from me on the red CI, since you've already said you'll pick up the copyright header and the tests once the logic is settled.

A few notes inline. The one I'd weigh most is the coverage gap around `fieldValidation=Strict` — it's the requirement #3019 calls non-optional, and it's the one thing no unit test here can see.

client/instance_patch.go
VijetaPriya47: I don't see the need for keeping the wrapper at all, because the wrapper never actually fixed the problem.  I just made it so that it will be matching with the existing naming convention.
And CLI already calls `CreateInstanceWithBodyWithResponse` directly with the content type as an
argument, so `instance update` can do the same with
`PatchInstanceWithBodyWithResponse` and `application/merge-patch+json`. 
One thing to be noticed is that, the comment comes from `client-with-responses.tmpl:125`, which hardcodes `{{$opid}}WithResponse`, while the method name it describes comes from `.Suffix`, i.e. the media-type path. 

It's better to fix this in the template bug upstream if you think it's worth it. It's a one-line fix and reproduces on any operation whose request body isn't `application/json`. And if we change the update function later for backupstorageupdate, the same issue will be there again. What do you say, I make an issue for this outside the PR, and fix oapi-codegen itself?

internal/server/handlers/validation/instance.go
VijetaPriya47: but annotations being edited is allowed, right?

internal/server/handlers/k8s/instance_test.go
VijetaPriya47: will add it here itself

internal/server/everest.go
VijetaPriya47: well, these both IsNotFound  -> Message: trimWebhookErrorText(...)  and IsForbidden / IsInvalid -> Message: trimStrictDecodingError(...)  had messages, so I thought IsAlreadyExists/IsConflict should also have it, to maintain the consistency.

VijetaPriya47:
@recharte could you check up on the failed test once also?

internal/server/handlers/k8s/instance_test.go
recharte: @VijetaPriya47 the spec covers exactly the three cases, thanks — but it never runs. `playwright.config.ts` selects specs per project (the root `testMatch: /.ˆ/` matches nothing), CI runs `make test`, which is `npx playwright test --project=api-tests`, and that project's `dependencies` list has no entry for this file. So the green API Integration Tests badge here hasn't executed a line of it.

A project modelled on `instance-preset`, plus an entry in the `api-tests` dependencies, is the fix:

```ts
    {
      name: 'instance',
      testDir: 'tests',
      testMatch: /instance\.spec\.ts/,
      dependencies: ['global:auth:ci:setup'],
      use: {
        extraHTTPHeaders: {
          'Authorization': `Bearer ${process.env[API_CI_TOKEN]}`,
        }
      },
    },
```

One thing to expect on the first real run: this is the only spec in the suite with a `beforeAll`, and the config is `fullyParallel: true` with `workers: 5`, so `beforeAll` runs once per worker that picks up a test from the file — five attempts to POST the same `test-instance-patch`, with `checkError` failing the four that come back 409. The tests also share the instance's `replicas`: the merge-patch test sets 3 while the resourceVersion test asserts 5. `test.describe.configure({mode: 'serial'})` on the describe block settles both.

internal/server/handlers/validation/instance.go
recharte: Yes — and I think rejecting is right, though the message could carry the reason.

Merge patch can't express "replace this map": `{"annotations":{…}}` merges, so once the actor stamp travels in the same document there's no way to honour a wipe and stamp in one write without the handler doing a GET and enumerating the keys — the read-modify-write this endpoint exists to remove.

What survives is per-key removal, which does still work — `{"metadata":{"annotations":{"team/owner":null}}}` comes through the stamp merge intact and the key is gone from the stored object. So the only annotations a caller can't clear are the two actor ones, and those are re-stamped on the next patch regardless.

Could the message point at that route? At the moment it states the prohibition and leaves the caller guessing:

```go
return errors.New("metadata.annotations may not be removed wholesale; set individual keys to null instead")
```

Worth a case for the per-key path in `TestValidate_PatchInstance` too — it's the route the message would now recommend, and nothing pins it.

client/instance_patch.go
recharte: Agreed on dropping it, and your reading of the template is right — the doc names a method that doesn't exist for any operation whose body isn't `application/json`.

Whether to send it upstream is your call; nothing here depends on it.

internal/server/everest.go
recharte: That's the right consistency argument — thanks.

recharte (approved):
Well done @VijetaPriya47 thanks

PR #3056, opened 2026-08-26
Various tests to ensure that plugins actually work

spron-in:
Adding various tests to check if plugins actually work


api-tests/tests/plugins.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  Array.isArray(plugins) ? plugins.find((p) => p?.name === name) : undefined
+
+test.describe.serial('Plugin subsystem', () => {
+  test.beforeAll(async ({cli}) => {
+    await (await cli.exec(`kubectl apply -f ${PLUGIN_MANIFEST}`)).assertSuccess()
+  })
+
+  test.afterAll(async ({cli}) => {
+    await cli.exec(`kubectl delete -f ${PLUGIN_MANIFEST} --ignore-not-found=true`)
+  })
```
chilagrow: Could we apply CLI related initialization `Makefile`?
https://github.com/openeverest/openeverest/blob/main/api-tests/Makefile#L71

api-tests/tests/plugins.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  test('proxy returns 404 for an unknown plugin', async ({request}) => {
+    const r = await request.get('/v1/plugins/does-not-exist/main.js')
+
+    expect(r.status()).toBe(404)
+  })
+
+  test('disabled plugin is hidden from discovery and the proxy', async ({request, cli}) => {
+    await (await cli.exec(
+      `kubectl patch plugin ${PLUGIN_NAME} --type merge -p '{"spec":{"enabled":false}}'`
```
chilagrow: Can we add another disabled plugin in `test-plugin.yaml` instead? Tests can stay simpler by keep CLI such as `kubectl` in Makefile and avoid state changes if possible 🤗

internal/server/plugins_test.go
```diff
@@ -0,0 +1,449 @@
...
+	)}
+
+	testCases := []struct {
+		name    string
+		subject string
+		groups  []string
+		plugin  string
+		want    bool
+	}{
+		{name: "direct use grant", subject: "bob", plugin: "plugin-hub", want: true},
```
chilagrow: could you make this multi-line for readability please 🙏 

```suggestion
		{
			name: "direct use grant",
			subject: "bob",
			plugin: "plugin-hub",
			want: true,
		},
```

api-tests/tests/plugins.spec.ts
spron-in: fixed :)

api-tests/tests/plugins.spec.ts
spron-in: fixed :)

api-tests/manifests/test-plugin.yaml
```diff
@@ -0,0 +1,49 @@
...
+apiVersion: extensions.openeverest.io/v1alpha1
+kind: Plugin
+metadata:
+  name: test-api-plugin-disabled
+spec:
+  displayName: Test API Plugin (disabled)
+  description: Disabled plugin used by api-tests to verify enabled gating.
+  version: 0.0.0
+  vendor: OpenEverest
+  enabled: false
```
recharte, line 39: The failing test is right — this is a real bug.

`Enabled bool` (`api/extensions/v1alpha1/plugin_types.go:62`) has `json:"enabled,omitempty"` *and* `+kubebuilder:default=true`. The reconciler adds its finalizer and calls `Update` (`plugin_controller.go:94-97`). `omitempty` drops the `false`, so the API server sees no `enabled` field and puts the default back. Any plugin created disabled turns itself on a second later.

An optional bool needs to be `*bool` whenever a default exists, because `false` and "not set" have to stay distinguishable on the wire. Removing `omitempty` instead would break the other direction: `pkg/cli/extension/install.go` builds the CR in Go, and an unset field would then mean disabled.

Only five places read `Spec.Enabled`, so `*bool` plus an `IsEnabled()` helper is a small change. Here or in its own PR, either works — but not by changing the test.

internal/server/plugins_test.go
```diff
@@ -0,0 +1,480 @@
...
+	"github.com/openeverest/openeverest/v2/pkg/common"
+	"github.com/openeverest/openeverest/v2/pkg/kubernetes"
+)
+
+// rbacModel mirrors data/rbac/model.conf. The plugin proxy's RBAC checks rely
+// solely on this matcher's glob semantics, so building an enforcer directly
+// from the model lets us grant plugin-scoped resources ("plugin/<name>",
+// "plugins") without going through the ConfigMap policy validator, which only
+// knows the generated API resource names.
+const rbacModel = `[request_definition]
```
recharte, line 44: Context from [#3009](https://github.com/openeverest/openeverest/issues/3009#issuecomment-5443927020): the plugin endpoints were never added to `http-api.yaml`, so they're hand-registered outside the `rbac → validation → k8s` chain and `plugins.go` hand-rolls everything the chain would have done. Once they move onto it, roughly 60% of this file has nothing left to test:

| Test | On the chain |
|---|---|
| `TestCheckPluginsReadAccess` (~30 lines) | gone — the RBAC layer does it generically |
| `TestCanUsePlugin` (~85) | gone — one `h.enforce` call in `handlers/rbac/plugin.go` |
| `TestAuthedProxyHandler` (~55) | mostly gone — it's a permission gate |
| `TestListPluginsHandler` RBAC half | gone — same filter loop as `ListProviders` |
| `rbacModel` + `newTestEnforcer` + `ctxWithUser` (~70) | gone — scaffolding for a chain that isn't there |

Worth knowing while these exist: none of the policies they assert can be loaded in a real cluster. `plugins` isn't in the generated resource list, so `checkResourceNames` rejects the whole ConfigMap if anyone writes one. The suite is green because it builds its own enforcer from a copy of the model.

What stays is real — `TestResolvePluginAssetPath`, `TestResolveBackendURL`, `TestDoProxy`'s prefix stripping and 404s. A reverse proxy can't be generated, so that part is rightly hand-written and rightly tested.

Could we split those out and hold the rest until the endpoints move? Same for the `plugin/*` wildcard case — the permission model is being reconsidered, and pinning the current behaviour now just makes it harder to change.

internal/server/plugins_test.go
```diff
@@ -0,0 +1,480 @@
...
+		var httpErr *echo.HTTPError
+		require.ErrorAs(t, err, &httpErr)
+		assert.Equal(t, http.StatusForbidden, httpErr.Code)
+	})
+}
+
+func TestDoProxy(t *testing.T) {
+	t.Parallel()
+
+	t.Run("strips prefix and forwards to backend", func(t *testing.T) {
```
recharte, line 367: No user in the context, and it asserts 200. That's not just the fixture — `GET /v1/plugins/:name/*` is registered outside the JWT group (`everest.go:289`). The comment there treats "GET only" as the protection, but everything a plugin backend serves for reading is a GET. `GET /v1/plugins/plugin-hub/api/instances` needs no token today.

This is also the one route that stays hand-registered after the refactor, so it's the part of this file most worth keeping and extending. What's missing is a test saying where the open part stops — only `bundlePath`, or every GET?

Also: `doProxy` never removes the incoming `Authorization` header, so a logged-in user's Everest token gets sent to whatever `externalUrl` the CR names. Not from this PR, but this is the right place to lock it down.

api-tests/tests/plugins.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  })
+
+  test('proxy routes an enabled plugin to its backend', async ({request}) => {
+    // The backend points at a closed port, so a successful route to the proxy
+    // surfaces as a 502 (backend dial failed) rather than a 404 (unknown
+    // plugin). This proves the request reached the proxy and resolved the
+    // backend for an enabled plugin.
+    const r = await request.get(`/v1/plugins/${PLUGIN_NAME}/main.js`)
+
+    expect(r.status()).toBe(502)
```
recharte, line 68: This checks what `httputil.ReverseProxy` does when a dial fails. Adding an `ErrorHandler` to `doProxy` (worth doing — it currently leaks Go's transport error text) would break this test for no real reason. And it never shows the prefix being stripped or a response coming back.

The CR comes from a manifest, so `serviceRef` can point at something real in the test cluster and the test can check the actual forwarded response. That's also the one thing a unit test can't do.

api-tests/tests/plugins.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  name: string
+  displayName?: string
+  bundleUrl?: string
+  extensionPoints?: { type: string }[]
+}
+
+const findPlugin = (plugins: PluginDescriptor[], name: string): PluginDescriptor | undefined =>
+  Array.isArray(plugins) ? plugins.find((p) => p?.name === name) : undefined
+
+test.describe.serial('Plugin subsystem', () => {
```
recharte, line 40: `unknown → 404`, the `bundleUrl` shape and `disabled → hidden` are all covered in `plugins_test.go`. The cluster adds one thing: the CR going through a real API server, which is what found the `enabled` bug. That only matters where CRD behaviour is involved.

Keep discovery and disabled-gating, drop `unknown plugin → 404`.

`test.describe.serial` can go as well — the disabled plugin comes from the manifest now, so the tests share nothing.

recharte (commented):
Overall I'm a bit unsure about these tests due to the big refactoring that we need to do with #3009 but I'm leaving some comments regardless.

PR #3058, opened 2026-08-26
feat(provider-runtime): carry supportedFields into the generated Provider CR

recharte:
Second of a stack implementing per-component field applicability ([spec 013](https://github.com/openeverest/specs)). **Stacked on #3057** — this PR's diff is against it.

### This PR

Reads the per-component `supportedFields` block from the topology config and publishes it on `TopologyComponent`, so the declaration reaches the Provider CR that every client resolves against.

`supportedFields` is authored **inline**, unlike `parametersSchema`, which names a Go type the generator reflects on. It selects from `ComponentSpec` rather than describing a new type, and Go cannot express "a subset of an existing struct" without a parallel type that drifts from it. JSON Schema selects natively.

### Testing

The tests cover the wire format, which is where the risk actually is: `JSONSchemaProps` has custom marshalling, so the open question was whether an inline declaration survives YAML unmarshalling with its structure intact. It does —

- a bound (`replicas: {minimum: 1}`) round-trips,
- nesting round-trips, so a component can declare `schedulingPolicy` with only `nodeSelector` and `tolerations` and **not** promise `affinity`,
- an undeclared field does not appear,
- a component with no declaration yields `nil`, not an empty schema — absent means unconstrained, and the two must not be confused.

First tests in this package.

### Companion PR

openeverest/provider-sdk#37 lets `supportedFields` through `buildTopologySpec`, which is an allowlist and drops it today. Both are needed before any real `topology.yaml` reaches this code.

### Review notes

Draft until the PRs below it merge.

### DIFF
--- a/provider-runtime/manifest/manifest.go
+++ b/provider-runtime/manifest/manifest.go
@@ -95,6 +95,9 @@ type topologyConfig struct {
 
 type topologyComponentConfig struct {
 	Optional bool `json:"optional,omitempty"`
+	// SupportedFields is authored inline rather than as a Go type reference:
+	// it selects from ComponentSpec instead of describing a new type.
+	SupportedFields *apiextensionsv1.JSONSchemaProps `json:"supportedFields,omitempty"`
 }
 
 // Generate reads a provider-config.yaml, resolves Go type references to OpenAPI schemas,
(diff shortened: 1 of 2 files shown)


recharte:
Closing: this adds `supportedFields` to a generator nothing runs.

`provider-sdk generate` builds the Provider CR spec itself and never imports `provider-runtime/manifest`, so this change had no effect on any provider — which only surfaced when generating PSMDB end to end. The equivalent change in the generator that does run is openeverest/provider-sdk#37, with the validation rules in openeverest/provider-sdk#38.

#3061 removes the dead generator.

PR #3063, opened 2026-08-27
Add `BackupImport` CRD to support import

chilagrow:
This PR adds CRD changes to support import.

## CRDs
`BackupImport` CRD defines which backupStorage to use and who(backupClass) implements importing `Backup` CRs.

Breaking change to `Backup` CRD to support backups originated from `External` created by import, in addition to `Instance`. It's mutually exclusive using discrimination union. `Backup.spec.instanceRef` is removed and the instance link now lives at `Backup.spec.origin.instanceRef`. After the CRD upgrade the API server prunes the unknown `backup.spec.instanceRef` and existing Backups have no `Backup.spec.origin`, so the origin validation rejects any write and the Instance link is lost. The restore from existing backups will not work and the Backup CRs must be recreated.

`Instance` CRD supports immutable field `userSecretRef` to create an instance with defined user credentials.

## UI
The `Backup` CRD has changes propagated.

## provider-runtime, backup controller, api server, cli, tests
Changes are purely to support `Backup` originates outside Instance.

## TODO
update each providers
Refs https://github.com/openeverest/provider-percona-server-mongodb/pull/91.
Refs https://github.com/openeverest/provider-percona-xtradb-cluster/pull/19.
Refs https://github.com/openeverest/provider-percona-postgresql/pull/28.
Refs https://github.com/openeverest/provider-cassandra/pull/14.
Refs https://github.com/openeverest/provider-mariadb/pull/61.
Refs https://github.com/openeverest/helm-charts/pull/101.


api/backup/v1alpha1/backup_types.go
```diff
@@ -90,22 +93,73 @@ const (
...
+// +kubebuilder:validation:Enum=Instance;Imported
+type BackupOriginType string
+
+const (
+	// BackupOriginTypeInstance marks a Backup produced by a live Instance in
+	// the same namespace, identified by origin.instanceRef.
+	BackupOriginTypeInstance BackupOriginType = "Instance"
+	// BackupOriginTypeImported marks a Backup imported from data already
+	// present in a BackupStorage, identified by origin.imported.
+	BackupOriginTypeImported BackupOriginType = "Imported"
```
recharte: I know this is a huge nitpick but wouldn't it be more correct for an origin type to be `Import` instead of `Imported`? I see this field as answering the question "Where did this backup originate from?" And the answer to that would be "It originated from an Instance/Import.", right?

api/backup/v1alpha1/import_types.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// ImportStatus defines the observed state of Import.
+type ImportStatus struct {
+	// Phase is the coarse lifecycle state of the import request.
+	// +optional
+	Phase ImportPhase `json:"phase,omitempty"`
+	// DiscoveredCount is the number of restorable backups found in the
+	// storage during the last reconcile.
+	// +optionalß
```
recharte:
```suggestion
	// +optional
```

api/backup/v1alpha1/backup_types.go
chilagrow: Thanks, this was exactly what I just noticed.

api/backup/v1alpha1/import_types.go
chilagrow: Thanks 🙈

api/backup/v1alpha1/backup_types.go
chilagrow, line 1: Don't merge until providers are updated.

api/backup/v1alpha1/backupimport_types.go
recharte, line 1: I think we should rename this to `BackupImport`. `Import` is too generic — we already ship `DataImporter`/`DataImportJob` in v1 for loading data into a database, and `specs/003-generic-plugins.md` already reserves the same word for pulling in externally-managed DBs (RDS and friends). Importing DBs from an existing operator installation would want it too.

api/backup/v1alpha1/backup_types.go
```diff
@@ -90,22 +93,72 @@ const (
...
+const (
+	// BackupOriginTypeInstance marks a Backup produced by a live Instance in
+	// the same namespace, identified by origin.instanceRef.
+	BackupOriginTypeInstance BackupOriginType = "Instance"
+	// BackupOriginTypeImport marks a Backup imported from data already
+	// present in a BackupStorage, identified by origin.import.
+	BackupOriginTypeImport BackupOriginType = "Import"
+)
+
+// BackupOrigin describes where a Backup's data comes from. Type selects which
```
recharte, line 110: I know I already questioned the `origin` field over on Slack, and I think I can finally say what bothered me: the two variants aren't the same kind of statement. `origin.type: Instance` is a request — "back up this Instance" — and the controller acts on it. `origin.type: Import` is a description — "the bytes are already at this path" — and nothing acts on it. One discriminator, two meanings.

I've talked myself round to it, though: `Service.spec.type: ExternalName` is the same trick, with no ClusterIP allocated, no endpoints and nothing proxied — just a record the rest of the system reads. What's missing here is the other half of the pattern. A statically-provisioned `PersistentVolume` is pure description too, but the PV controller still validates it and publishes `status.phase`. Nothing ever sets `status.state` on an imported Backup, so `ValidateBackupSucceeded` rejects every restore from one. The Import reconciler needs to set `Succeeded` after checking the object really is in the storage — detail on `internal/controller/backup/backup_controller.go`.

And the name: `origin` answers "where did this backup come from?", and the answer isn't "an Import" — that's the mechanism that noticed it, not the origin. Wouldn't `origin.type: External` with `origin.external.path` be more intuitive?

api/backup/v1alpha1/import_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// StorageRef references a BackupStorage in the same namespace whose
+	// contents are listed and parsed. The reconciler reads the storage and
+	// its credentials secret.
+	// +kubebuilder:validation:Required
+	StorageRef common.ObjectRef `json:"storageRef"`
+}
+
+// ImportPhase is the coarse lifecycle state of an import request.
+//
+// +kubebuilder:validation:Enum=Pending;Importing;Succeeded;Failed
```
recharte: `Backup` and `Restore` both use `status.state` with `Pending/Running/Succeeded/Failed/Error`. This changes the field name, renames `Running` to `Importing`, and drops `Error`.

`Error` is the one that matters. The siblings use it for the retryable case, and for a one-shot object it matters more, not less — a transient S3 blip landing on `Failed` means the whole Import has to be recreated. Worth aligning on `status.state` and the same value set before this is public API.

api/backup/v1alpha1/backup_types.go
```diff
@@ -22,11 +22,14 @@ import (
 )
 
 // BackupSpec defines the desired state of Backup.
+//
+// +kubebuilder:validation:XValidation:rule="self.origin.type == 'Instance' ? has(self.origin.instanceRef) : !has(self.origin.instanceRef)",message="origin.instanceRef must be set if and only if origin.type is Instance"
```
recharte: Small style thought: `DataSource` puts the same union rules on the union struct itself (`restore_types.go:118-119`), which roots the error path at `spec.origin` and keeps the validation with the type. Moving these two onto `BackupOrigin` (line 111) would match — with whichever variant name we land on in the thread there:

```go
// +kubebuilder:validation:XValidation:rule="self.type == 'Instance' ? has(self.instanceRef) : !has(self.instanceRef)",message="instanceRef must be set if and only if type is Instance"
```

internal/controller/backup/backup_controller.go
```diff
@@ -163,6 +163,11 @@ func (r *BackupReconciler) Reconcile( //nolint:nonamedreturns
 		return result, nil
 	}
 
+	// Imported backups have no Instance and nothing to handle here.
+	if backup.Spec.Origin.InstanceRef == nil {
```
recharte: Nothing writes `status.state` for an imported Backup: this early return fires before the status block, and `resolveBackupOwnership` bails at `provider-runtime/reconciler/backup.go:192` before `updateStatus`. Restores go through `ValidateBackupSucceeded`, which reports `""` as `Pending` and rejects it — permanently.

The Import reconciler is the only actor that read the storage, so I'd put it there: check the object is really present, create the Backup, then `Status().Update()` it to `Succeeded`. Worth noting `Backup` has `+kubebuilder:subresource:status`, so `Create` drops the status and a crash between the two calls strands the object — the reconcile has to repair that on the next pass.

Either way, please state the contract on `BackupStatus.State`. `ExecutionMode`, `Size`, `OperatorBackupRef` and `JobRef` all got an "empty for imported backups" note; the field that decides whether the backup is usable at all got none.

api/core/v1alpha1/instance_types.go
```diff
@@ -102,6 +103,18 @@ type InstanceSpec struct {
...
+	// credentials from a Secret in the same namespace, for providers whose
+	// engine supports setting initial credentials at creation time.
+	//
+	// When omitted, the provider generates credentials automatically. The
+	// referenced Secret's required keys are provider-specific and validated
+	// by the referenced Provider. The field is immutable once set: initial
+	// credentials only apply at engine creation time, so changing it later
+	// would have no effect.
+	// +optional
+	UserSecretRef *common.SecretRef `json:"userSecretRef,omitempty"`
```
recharte, line 124: `secrets:read` is enforced per-object in `internal/server/handlers/rbac/secret.go`. This field routes around it: a caller with `instances:create` in a namespace points `userSecretRef` at another Instance's credentials Secret — same key names, by construction — and reads the values back from `GET /instances/{new}/connection`, which only checks `instances:read` on the instance they just created.

Nothing consumes the field yet, so it isn't live, but it's public API the moment it merges. Cheapest fix is in the validation handler: when `userSecretRef` is set, require the caller to pass the same `secrets:read` check for that object. Is that the plan, or does something provider-side constrain which Secrets are acceptable?

api/backup/v1alpha1/backupimport_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// DiscoveredCount is the number of restorable backups found in the
+	// storage during the last reconcile.
+	// +optional
+	DiscoveredCount int32 `json:"discoveredCount,omitempty"`
+	// CreatedCount is the number of Backup CRs created (or already present)
+	// for the discovered backups. Backups are deduped on
+	// (storageRef, path), so re-running import does not
+	// create duplicates.
+	// +optional
+	CreatedCount int32 `json:"createdCount,omitempty"`
```
recharte: Once an Import has run, there's no way to tell what it did. Nothing on a Backup says which Import created it, and nothing on the Import lists what it created — `createdCount` is a number with nothing behind it. Re-runs make it worse: the field is documented as "created (or already present)", so a second pass over the same bucket reports 50 when it actually created 0.

Two cheap fixes:

- Label the created Backups the way a Job labels its Pods (`batch.kubernetes.io/job-name`): `backup.openeverest.io/import: <name>` gives you `kubectl get backups -l backup.openeverest.io/import=...` and adds nothing to the API surface. Not an ownerReference — the Backups should outlive the Import — but the label gets you the same traceability.
- Count only what this run actually created, and drop the "(or already present)" from the doc. Already-present is `discoveredCount - createdCount`, so it doesn't need to be folded into the same number.

The label also gives the reconciler the lookup it needs to repair a half-finished run (see the status point on `internal/controller/backup/backup_controller.go:167`).

api/backup/v1alpha1/import_types.go
```diff
@@ -0,0 +1,125 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	common "github.com/openeverest/openeverest/v2/api/common/v1alpha1"
+)
+
+// ImportSpec defines the desired state of Import.
+type ImportSpec struct {
```
recharte: The spec should be immutable. Repointing `storageRef` half way through leaves the Backups it already created — and both counts — describing the old storage. Like a Job, you re-run by creating a new object rather than editing this one.

```suggestion
// ImportSpec defines the desired state of Import.
//
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec is immutable"
type ImportSpec struct {
```

provider-runtime/reconciler/provider.go
```diff
@@ -200,10 +200,10 @@ func newReconciler(ctx context.Context, p providerAdapter, opts ...ReconcilerOpt
 	if _, isBackupProvider := p.(controller.BackupProvider); isBackupProvider {
 		if err := mgr.GetFieldIndexer().IndexField(ctx, &backupv1alpha1.Backup{}, controller.IndexBackupInstanceName, func(obj client.Object) []string {
 			b, ok := obj.(*backupv1alpha1.Backup)
-			if !ok || b.Spec.InstanceRef.Name == "" {
+			if !ok || b.Spec.Origin.InstanceRef == nil {
```
recharte, line 205: The value moved to `spec.origin.instanceRef.name`; the key didn't. `IndexBackupInstanceName = "spec.instanceRef.name"` at `provider-runtime/controller/common.go:701`, plus its doc comment and `BackupsForInstance`'s at common.go:672-675. You did update `cascadeDeleteChildren`'s doc (provider.go:651), so the SDK is now inconsistent with itself.

Functionally harmless, but this is the doc downstream provider authors read. Please rename the constant and fix the three comments.

pkg/cli/backup/create.go
```diff
@@ -79,7 +79,10 @@ func (cr *CreateRunner) Run(ctx context.Context, opts CreateOptions, cfgPath str
 		md.GenerateName = opts.Instance + "-"
 	}
 	backup := client.Backup{Metadata: &md}
-	backup.Spec.InstanceRef.Name = opts.Instance
+	backup.Spec.Origin.Type = client.BackupSpecOriginTypeInstance
```
recharte, line 82: `Origin.Type` is new and the CEL rule rejects the object if it's wrong, but `TestRun_HappyPath_GeneratesName` checks `ClassRef` and `StorageRef` off the wire body and stops:

```go
assert.Equal(t, client.BackupSpecOriginTypeInstance, gotBody.Spec.Origin.Type)
require.NotNil(t, gotBody.Spec.Origin.InstanceRef)
assert.Equal(t, "my-mongo", gotBody.Spec.Origin.InstanceRef.Name)
```

Three fixtures also still send the removed field (`create_test.go:52`, `delete_test.go:92`, `list_test.go:36-38`), so every fixture Backup now unmarshals with `Origin.InstanceRef == nil` — the branch `backupInstance` returns `"-"` for. Worth updating in the same pass.

api/backup/v1alpha1/backup_types.go
```diff
@@ -22,11 +22,14 @@ import (
...
+// +kubebuilder:validation:XValidation:rule="self.origin.type == 'Import' ? has(self.origin.import) : !has(self.origin.import)",message="origin.import must be set if and only if origin.type is Import"
 type BackupSpec struct {
-	// InstanceRef references the Instance to back up. The Instance must
-	// live in the same namespace as this Backup.
+	// Origin identifies where this Backup's data comes from: produced by a
```
recharte, line 28: Please spell the breaking change out in the description. After the CRD upgrade, existing Backups have `spec.instanceRef` pruned and no `spec.origin`, so any write is rejected and the instance link is gone — anyone already running v2 can't restore from the backups they have, and has to recreate the CRs. "Breaking change to `Backup` CRD" doesn't get that across, and this needs to reach the release notes.

api/backup/v1alpha1/import_types.go
```diff
@@ -0,0 +1,125 @@
...
+
+// ImportStatus defines the observed state of Import.
+type ImportStatus struct {
+	// Phase is the coarse lifecycle state of the import request.
+	// +optional
+	Phase ImportPhase `json:"phase,omitempty"`
+	// DiscoveredCount is the number of restorable backups found in the
+	// storage during the last reconcile.
+	// +optional
+	DiscoveredCount int32 `json:"discoveredCount,omitempty"`
```
recharte: Two small things on this pair:

- With `omitempty`, a genuine `0` — listed the storage, found nothing restorable — serializes as absent, and the `Created` print column renders `<none>` rather than `0`. Built-in types get away with this (`DeploymentStatus.Replicas` is `int32,omitempty`) because kubectl prints those from a server-side table printer that reads the Go value; CRD columns are JSONPath over the serialized object, so the omission shows through. Dropping `omitempty` from both fixes it — a pointer would be wrong here, since nil would mean "not tracked" and `status.phase` already says whether the run happened.
- "during the last reconcile" reads as a continuously re-syncing object. For a one-shot it's just "found in the storage".

internal/controller/backup/backup_controller.go
```diff
@@ -270,16 +274,21 @@ func (r *BackupReconciler) Reconcile( //nolint:nonamedreturns
 // ensureInstanceNameLabel ensures that the Backup resource has the instance name
 // label, used for filtering backups by instance using label selectors.
 func (r *BackupReconciler) ensureInstanceNameLabel(ctx context.Context, backup *backupv1alpha1.Backup) error {
+	// Imported backups have no Instance, so there is nothing to label.
+	if backup.Spec.Origin.InstanceRef == nil {
```
recharte: Two things here:

- This can't fire — the only caller is three lines below the early return at line 167, which already guarantees `Origin.InstanceRef != nil`. Keeping both reads as if the function is safe to call for imported backups.
- The guards would say more as `Origin.Type == BackupOriginTypeImport` than as `InstanceRef == nil` — same behaviour, but it names the case instead of testing a symptom, and a malformed object stops being silently skipped. Same at line 167, `provider-runtime/reconciler/backup.go:192`, and the index funcs in `provider-runtime/reconciler/provider.go:203`.

recharte (changes requested):

api/backup/v1alpha1/import_types.go
chilagrow: That's right, updated.

api/backup/v1alpha1/import_types.go
chilagrow: Thanks, Error state for retryable is missed here. Aligned the states with `Restore` and `Backup`.

api/backup/v1alpha1/import_types.go
chilagrow: Thanks for catching this. Updated comment and removed `omitempty`

api/backup/v1alpha1/import_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// CreatedCount is the number of Backup CRs created (or already present)
+	// for the discovered backups. Backups are deduped on
+	// (storageRef, path), so re-running import does not
+	// create duplicates.
+	// +optional
+	CreatedCount int32 `json:"createdCount,omitempty"`
```
chilagrow: Thanks for the suggestion. I would apply both here, label to give traceability and counting only what the run created 🙏

api/core/v1alpha1/instance_types.go
chilagrow: The path you've described is possible, however getting secret of another instance requires `instances:read` on that instance, or prior knowledge of the secret name which I believe is unlikely.

For safety agree with checking `secrets:read` upon creation of instance with `userSecretRef`.

api/backup/v1alpha1/import_types.go
chilagrow: Updated to BackupImport 🙏

api/backup/v1alpha1/backup_types.go
chilagrow: Let's use `origin.external.path`, as you point out `Service.spec.type` uses a similar concept for `external` so it fits in.

api/backup/v1alpha1/backup_types.go
chilagrow: Thanks updated the description 🙏

api/backup/v1alpha1/backup_types.go
chilagrow: Thanks, updated to keep the consistency with `Restore.spec.DataSource` 🙏

pkg/cli/backup/create.go
chilagrow: Thanks for listing those 🙏 Updated.

api/backup/v1alpha1/backupimport_types.go
chilagrow: Updated to BackupImport to avoid conflict 🙏

api/backup/v1alpha1/backup_types.go
chilagrow: I've updated to `origin.type: External` with `origin.external.path` , as you point out `Service.spec.type` uses a similar concept for `external` so it fits in.

The `status.state` not being set is my oversight. I updated both provider managed and job mode reconciliation to set `status.state`

api/backup/v1alpha1/import_types.go
chilagrow: Thanks, `Error` state for retryable is missed here. Aligned the states with `Restore` and `Backup`.

api/backup/v1alpha1/backup_types.go
chilagrow: Thanks, updated to keep the consistency with `Restore.spec.DataSource` 🙏

internal/controller/backup/backup_controller.go
chilagrow: The state is now reconciled and the validation for existing backup is added. One more thing, `spec.origin.external.completedAt` and `spec.origin.external.startedAt` were added to capture when the backup was made. I consider backup to be unusable without them, so I'm enforcing provider/plugin developers to capture time the backup was made. This value is mirrored to `Backup.origin.export.completedAt` and `Backup.origin.export.StartedAt`.

api/core/v1alpha1/instance_types.go
chilagrow: The path you've described is possible, initially I thought getting secret of another instance requires `instances:read` on that instance, or prior knowledge of the secret name which I believe is unlikely. But it exposes that if secret name is obtained, it allows getting credentials by just referencing secret in `userSecretRef` which is bad.

For safety agree with checking `secrets:read` upon creation of instance with `userSecretRef`.

api/backup/v1alpha1/backupimport_types.go
chilagrow: Thanks for the suggestion. I would apply both here, label to give traceability and counting only what the run created 🙏

api/backup/v1alpha1/import_types.go
chilagrow: Thanks, that's right. Updated.

provider-runtime/reconciler/provider.go
chilagrow: Thanks renamed them 🙏

api/backup/v1alpha1/backup_types.go
chilagrow: Updated the description to indicate what has changed

api/backup/v1alpha1/import_types.go
chilagrow: Cheers for this. Updated the comment and removed `omitempty`

internal/controller/backup/backup_controller.go
chilagrow: Thanks, I've updated the code path so that backup CR is updated and this is used to skip checking labels.

internal/controller/backup/backup_controller.go
chilagrow: I've added the same status reconciliation to provider managed backup as well.

provider-runtime/controller/common.go
```diff
@@ -1184,3 +1191,140 @@ func RequestsForInstancesMatching(ctx context.Context, c client.Client, provider
...
+// Instance and runs no job: it is a reference to backup data already sitting
+// in a BackupStorage. The startedAt and completedAt timestamps are supplied by
+// the creator on spec.origin.external and mirrored to status here. The state is
+// set to Succeeded when those timestamps are present and the object is actually
+// present in the referenced BackupStorage.
+//
+// Both Job-mode and ProviderManaged backups share this reconciliation of backup
+// status; there is no engine specific logic to validate an external backup to
+// set the state.
+func ReconcileExternalBackupStatus(ctx context.Context, c client.Client, backup *backupv1alpha1.Backup) error {
```
recharte: `Prefix` here is a plain string prefix — `backups/a` matches `backups/abandoned/...`. But I don't think anchoring it helps: a backup is a pile of objects plus a marker written last, so "something exists under this prefix" is also true of one that died mid-upload. I'm not sure this buys us much...

Whoever implements `BackupImport` is already listing and parsing the storage to find backups at all — can't it just skip what it can't parse and set `Succeeded` on what it creates? Then core never touches a bucket and this whole path goes: both helpers, the `aws-sdk-go-v2` direct dep every provider binary now links, the Job controller branch, and `resolveBackupOwnership`'s external case.

Downside is that a bad path only bites at restore time, but I'd argue that's an import bug and not necessarily something we should partially defend against here. What do you think?

internal/server/handlers/rbac/instance.go
```diff
@@ -58,6 +58,15 @@ func (h *rbacHandler) GetInstance(ctx context.Context, cluster, namespace, name
 
 // CreateInstance creates an instance, gated by RBAC.
 func (h *rbacHandler) CreateInstance(ctx context.Context, cluster string, instance *corev1alpha1.Instance) (*corev1alpha1.Instance, error) {
+	// If the instance references a user secret, ensure the user has read
+	// permission on that secret to prevent privilege escalation.
+	if userSecret := instance.Spec.UserSecretRef; userSecret != nil {
```
recharte, line 63: Thanks for adding this — but `UpdateInstance` (line 116) and `PatchInstance` (line 126) enforce only `instances:update`, and the CEL rule is `!has(oldSelf.userSecretRef) || ...`, which permits *setting* the field on an object that didn't have it. So the escalation still works, one step longer:

1. create an Instance with no `userSecretRef`;
2. `PATCH /instances/{name}` with `{"spec":{"userSecretRef":{"name":"victim-secret"}}}`;
3. read `GET /instances/{name}/connection`.

PATCH landed in #3055, so step 2 needs nothing but `instances:update`. The update path can't reuse this check as-is, since `PatchInstance` only sees the raw patch — the validation handler, which has the decoded object on update, is probably the right home for both.

provider-runtime/reconciler/backup.go
```diff
@@ -185,10 +192,17 @@ func resolveBackupOwnership(
 	if bc.Spec.ExecutionMode != backupv1alpha1.BackupExecutionModeProviderManaged {
 		return nil, bc, false, nil
 	}
+
+	// External backups have no Instance to fetch. Ownership is derived
+	// from the class's SupportedProviders.
+	if backup.Spec.Origin.Type == backupv1alpha1.BackupOriginTypeExternal {
```
recharte, line 198: Realistically a ProviderManaged class covers one engine, so this probably never fires — but nothing enforces that. `SupportedProviders` has no `MaxItems` and none of the three CEL rules on `BackupClassSpec` mention it, so a class listing `percona-postgresql` and `cnpg` is admissible today, and this line would hand the same Backup to both runtimes to reconcile and fight over `status`.

Instance-origin backups can't hit it because the Instance names exactly one provider; external ones have nothing to disambiguate on. Cheapest fix is to make the assumption real, over on `BackupClassSpec`:

```go
// +kubebuilder:validation:XValidation:rule="self.executionMode != 'ProviderManaged' || (has(self.supportedProviders) && size(self.supportedProviders) == 1)",message="a ProviderManaged BackupClass must support exactly one provider"
```

If the verification moves to `BackupImport` this becomes `return nil, bc, false, nil` anyway and the question disappears.

api/backup/v1alpha1/backup_types.go
```diff
@@ -90,34 +97,97 @@ const (
...
+	// storageRef + external.path with no live operator object.
+	// +optional
+	External *BackupOriginExternal `json:"external,omitempty"`
+}
+
+// BackupOriginExternal marks a Backup as imported and identifies where its data already
+// lives within the BackupStorage referenced by Backup.spec.storageRef, so it
+// can be restored without a live source Instance or operator-native backup
+// object.
+type BackupOriginExternal struct {
```
recharte, line 138: Both reconcilers short-circuit on `Status.State == Succeeded` and never look again, and `BackupSpec` has no immutability rule. So `path` can be repointed after verification and the Backup keeps its `Succeeded` state — a restore then reads bytes nothing ever checked. Same for the two timestamps, which are already mirrored into status.

Nothing about an external Backup is a request, so none of it should be editable:

```suggestion
// BackupOriginExternal marks a Backup as external and identifies where its data
// already lives within the BackupStorage referenced by Backup.spec.storageRef,
// so it can be restored without a live source Instance or operator-native
// backup object.
//
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="origin.external is immutable"
type BackupOriginExternal struct {
```

api/backup/v1alpha1/backupimport_types.go
```diff
@@ -0,0 +1,124 @@
...
+//
+// +kubebuilder:validation:Enum=Pending;Running;Succeeded;Failed;Error
+type BackupImportState string
+
+const (
+	// BackupImportStatePending indicates that the request has been accepted but
+	// has not yet started.
+	BackupImportStatePending BackupImportState = "Pending"
+	// BackupImportStateStateRunning indicates that the import is currently running.
+	BackupImportStateStateRunning BackupImportState = "Running"
```
recharte: `Pending` came out right; the other four didn't.

```suggestion
	// BackupImportStateRunning indicates that the import is currently running.
	BackupImportStateRunning BackupImportState = "Running"
```

Same below: `BackupImportStatetateSucceeded` (missing an `S`), `BackupImportStateStateFailed`, `BackupImportStateStateError`. Nothing consumes them yet, so it's free to fix now and awkward once providers have imported them.

provider-runtime/reconciler/backup.go
```diff
@@ -201,6 +215,32 @@ func resolveBackupOwnership(
...
 
+// reconcileExternalBackup verifies that the data an external Backup points at
+// is present in its BackupStorage and that it carries valid startedAt and
+// completedAt timestamps, then marks the Backup Succeeded.
+func (r *backupRuntimeReconciler) reconcileExternalBackup(
+	ctx context.Context,
+	backup *backupv1alpha1.Backup,
+	bc *backupv1alpha1.BackupClass,
+) (reconcile.Result, error) {
+	if !backup.DeletionTimestamp.IsZero() {
```
recharte, line 225: Right now nothing does: this returns straight away on `DeletionTimestamp` and no finalizer is ever added, so the CR disappears and the objects stay put — regardless of `deletionPolicy`, which defaults to `Delete` and is documented as removing the data in the storage.

I can see it both ways. We didn't create those bytes, so leaving them alone feels right; but someone who imported a backup and then deletes it with the default policy probably expects the same thing to happen as with any other Backup.

Whichever way you land, I don't think the field should stay silently inert — either honour it, or reject `Delete` for `origin.type: External` with a CEL rule so the user has to say so explicitly. What's your take?

api/backup/v1alpha1/backup_types.go
```diff
@@ -22,11 +22,18 @@ import (
 )
 
 // BackupSpec defines the desired state of Backup.
+//
+// spec.instanceRef is removed; the instance link now lives at
```
recharte: This doc comment goes into the CRD's OpenAPI schema and out into the generated clients and the published API reference, permanently. Six months from now `BackupSpec` will still be telling readers that "spec.instanceRef is removed" and that "the CRs must be recreated", which is upgrade guidance for one release, not a description of the type.

The PR description covers it now, which was the ask. I'd keep the type doc to what the field *is*.

api/backup/v1alpha1/backupimport_types.go
```diff
@@ -0,0 +1,124 @@
...
+// BackupImportStatus defines the observed state of BackupImport.
+type BackupImportStatus struct {
+	// DiscoveredCount is the number of restorable backups found in the
+	// storage.
+	// +optional
+	DiscoveredCount int32 `json:"discoveredCount"`
+	// CreatedCount is the number of Backup CRs created. Backups are deduped on
+	// (storageRef, path), so the import does not create duplicates.
+	// +optional
+	CreatedCount int32 `json:"createdCount"`
```
recharte, line 49: `createdCount` reads correctly now, but the other half — being able to tell which import produced a given Backup — still isn't expressible.

I suggested a label last time and I think that was half right. A label providers write themselves is advisory: nothing can set it for them, repair it when someone strips it, or check they all picked the same key. `instanceName` doesn't have that problem because it's derived — `ensureInstanceNameLabel` reads `spec.origin.instanceRef.name` and sets it on every Backup, before any execution-mode dispatch.

Could `origin.external` carry an `importRef` and get the same treatment? One field, one extra line in that same labelling step, and `kubectl get backups -l ...` works without every provider having to remember. Not an ownerReference — these should still outlive the import.

api/backup/v1alpha1/backup_types.go
```diff
@@ -90,34 +97,97 @@ const (
...
+	// +kubebuilder:validation:Required
+	Type BackupOriginType `json:"type"`
+	// InstanceRef references the Instance that produced this Backup. The
+	// Instance must live in the same namespace as this Backup. Required when
+	// Type is Instance.
+	// +optional
+	InstanceRef *common.ObjectRef `json:"instanceRef,omitempty"`
+	// External identifies data already present in the referenced BackupStorage
+	// rather than produced by a live Instance. Required when Type is External.
+	// When set, the restoring builds the engine restore directly from
```
recharte: Small one — a word went missing here, and it survived the rename.

```suggestion
	// When set, the restore is built directly from
```

api/backup/v1alpha1/backupimport_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// LastObservedGeneration is the last observed generation of the Backup CR.
```
recharte: Two leftovers from the rename:

- this one says "generation of the Backup CR" — should be BackupImport;
- `shortName=imp` on line 88 was picked when the kind was `Import`.

provider-runtime/reconciler/backup.go
```diff
@@ -201,6 +215,32 @@ func resolveBackupOwnership(
...
+		return reconcile.Result{}, nil
+	}
+
+	if backup.Status.State == backupv1alpha1.BackupStateSucceeded ||
+		backup.Status.State == backupv1alpha1.BackupStateFailed {
+		return reconcile.Result{}, nil
+	}
+
+	backup.Status.ExecutionMode = backupv1alpha1.BackupExecutionModeProviderManaged
```
recharte: Tiny one: `updateStatus` sets `ExecutionMode` from `bc.Spec.ExecutionMode` two lines below, so this assignment never survives.

provider-runtime/controller/common.go
chilagrow: That's true. The path is validated already by the import. Keeping the code simpler by remove the storage path check.

api/backup/v1alpha1/backup_types.go
chilagrow: Thanks!

api/backup/v1alpha1/backupimport_types.go
chilagrow: 😱 Thanks my eyes didn't capture this.

api/backup/v1alpha1/backupimport_types.go
chilagrow: Thanks, updated shortName and leftover outdated comment.

api/backup/v1alpha1/backup_types.go
chilagrow: Thanks 🙏

api/backup/v1alpha1/backup_types.go
chilagrow: Removed 🙏

api/backup/v1alpha1/backupimport_types.go
chilagrow: Updated CRD to include `Backup.origin.external.importRef` so each backup knows which BackupImport created it.

provider-runtime/reconciler/backup.go
chilagrow: My take is reject `Delete`, given the default being `Delete` is not suitable for imported backup. Importer didn't create the underlying backup data in the store.

provider-runtime/reconciler/backup.go
chilagrow: Thanks this line is not necessary 🙏

provider-runtime/reconciler/backup.go
chilagrow: Thanks for the options, I added kubebuilder validation.

internal/server/handlers/rbac/instance.go
chilagrow: Both PATCH and UPDATE now rejects changes for `userSecretRef` on validation handler.

api/backup/v1alpha1/backupimport_types.go
recharte: I'm so sorry @chilagrow, I think I led you down the wrong solution.

I don't think that a new field in the CRD is the correct design here. Since nothing reads `importRef`, this represents provenance rather than desired state so I think this should be a label rather than a spec field. This is the same pattern used in Kubernetes when a Job labels the Pods it creates with `batch.kubernetes.io/job-name`.

What I should have asked for: drop the field, add a `BackupImportNameLabel` constant next to `common.InstanceNameLabel`, and have the import set it on what it creates. Then `kubectl get backups -l ...` works, which was the point.

Sorry for the churn.

chilagrow: Ah I see what you mean now 🙈 I misunderstood, no problem. I've added the label, which can be set by BackupImport reconciler.

api/backup/v1alpha1/backup_types.go
```diff
@@ -22,11 +22,13 @@ import (
 )
 
 // BackupSpec defines the desired state of Backup.
+//
+// +kubebuilder:validation:XValidation:rule="self.origin.type != 'External' || self.deletionPolicy != 'Delete'",message="deletionPolicy Delete is not allowed when origin.type is External; use Retain"
```
recharte: We have a conflict between this rule and the `deletionPolicy`'s default that has `+kubebuilder:default=Delete` on line 69. CRD defaulting runs at decode, before CEL so an external Backup created without an explicit `deletionPolicy: Retain` is defaulted to `Delete` and then rejected by this rule. That includes every Backup a BackupImport creates.

I'm not quite sure what the correct behavior should be though...
What do you suggest.

provider-runtime/reconciler/backup.go
recharte: Now that you added the validation CEL rule that enforces the deletion policy has to be something other than `Delete` we have some kind of conflict between that rule and the `deletionPolicy`'s default that has `+kubebuilder:default=Delete` on line 69. CRD defaulting runs at decode, before CEL so an external Backup created without an explicit `deletionPolicy: Retain` is defaulted to `Delete` and then rejected by this rule.
Will it be up to the backup class implementation of the import to properly set the deletion policy? Will that cause issues for our provider developers?

recharte (approved):
Thanks @chilagrow for the multiple rounds on this one.
I think it's good to go, the only small thing on my mind is the interplay between the deletion policy default and validation CEL, described in [this comment](https://github.com/openeverest/openeverest/pull/3063#discussion_r3919423295).

provider-runtime/reconciler/backup.go
chilagrow: That is true, `deletionPolicy` unset will default to `Delete` and rejected. BackupImport implementation should handle this, without causing issue to provider developers. 🙏

recharte: > BackupImport implementation should handle this, without causing issue to provider developers.

How?

PR #3066, opened 2026-08-27
feat(ui): add an output panel and schema saving to the UI Schema Playground

Denyme24:
Follow-up to #2465. Adds two things to the UI Schema Playground:

- **Output panel** -  a new tab showing the API payload the form would produce, with a copy button.
- **Schema saving** - the editor autosaves and restores on reload, and schemas can be saved, loaded and deleted by name.

Also fixes two pre-existing bugs it surfaced: the editor not reloading a schema it had previously emitted, and the form preview going blank when loading a schema with fewer sections.

https://github.com/user-attachments/assets/de471185-f831-4172-bf20-f950bf90cb02


ui/apps/everest/src/pages/plugin-developer/hooks/use-schema-storage.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (
+      parsed === null ||
+      typeof parsed !== 'object' ||
+      Array.isArray(parsed)
+    ) {
+      return {};
+    }
+    return Object.fromEntries(
+      Object.entries(parsed).filter(([, yaml]) => typeof yaml === 'string')
+    ) as SavedSchemas;
```
solovevayaroslavna: `AGENTS.md` bars `as` assertions ("narrow via guards/typeof or annotate"). The value is already guarded by the `typeof yaml === 'string'` filter, so lifting that guard into the filter's type should let the compiler infer `SavedSchemas` and drops the assertion:

```suggestion
    return Object.fromEntries(
      Object.entries(parsed).filter(
        (entry): entry is [string, string] => typeof entry[1] === 'string'
      )
    );
```

ui/apps/everest/src/pages/plugin-developer/output-panel/output-panel.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Typography variant="h6">Output</Typography>
+        {payload !== null && (
+          <Button size="small" onClick={handleCopy}>
+            Copy
+          </Button>
+        )}
+      </Stack>
+      {payload === null ? (
+        <Typography variant="body2" color="text.secondary">
+          {`Fill the form and click ${DatabaseFormMessages.createDatabase} to generate the payload.`}
```
solovevayaroslavna: This panel only shows the `postprocessSchemaData` output — the ui-generator engine's result. But the real form does more before sending: it rebuilds backup schedules and wraps everything in `createInstance({ formValue })` (`database-form.tsx:420`). So the "Create database"-style copy here promises the full API payload while showing just the generator output.

Probably fine for a schema playground (an author's schema maps to exactly this), but the wording is misleading. Should we change the naming? What do you think?

ui/apps/everest/src/pages/plugin-developer/schema-toolbar/schema-toolbar.tsx
```diff
@@ -0,0 +1,165 @@
...
+      direction="row"
+      spacing={1}
+      alignItems="center"
+      flexWrap="nowrap"
+      sx={{ width: '100%', overflowX: 'auto', py: 0.5 }}
+    >
+      <TextField
+        select
+        size="small"
+        label="Saved schemas"
```
solovevayaroslavna: `AGENTS.md` asks for user-facing strings to live in a `.messages.ts` `Messages` object; (Who knows, maybe openEverest will become multilingual in the future?) 
But the toolbar copy ("Saved schemas", "Save", "Delete", "Reset", "Save schema", "Schema name", etc.) is inlined here, and the same applies to the tabs/output copy in the sibling components.

If you see something like this somewhere, it means it's a mistake and someone forgot to remove it and we didn't notice it on the review, and so on. Ideally, all lines should be placed in messages.

ui/apps/everest/src/pages/plugin-developer/output-panel/output-panel.tsx
```diff
@@ -0,0 +1,91 @@
...
+    >
+      <Stack
+        direction="row"
+        alignItems="center"
+        justifyContent="space-between"
+        sx={{ mb: 2 }}
+      >
+        <Typography variant="h6">Output</Typography>
+        {payload !== null && (
+          <Button size="small" onClick={handleCopy}>
```
solovevayaroslavna: `@percona/ui-lib` already exports `CopyToClipboardButton`, which does exactly this and better. Please fix

ui/apps/everest/src/pages/plugin-developer/output-panel/output-panel.tsx
```diff
@@ -0,0 +1,91 @@
...
+    void navigator.clipboard?.writeText(json).catch(() => {});
+  };
+
+  return (
+    <Paper
+      elevation={0}
+      sx={{
+        p: 2,
+        border: '1px solid',
+        borderColor: 'divider',
```
solovevayaroslavna: The `p: 2 / 1px solid divider / rounded` recipe is inlined 4× across this PR. It already exists as a primitive — `RoundedBox` (`components/rounded-box`), and `FormCard` on top of it ("title + rounded box + content"), which is precisely `OutputPanel`'s shape. 

Related: the "Output" title uses `variant="h6"`, but section titles in this codebase use the `sectionHeading` typography token (see `RoundedBox`, `FormCard`, `OverviewSection`). Please check this too

ui/apps/everest/src/pages/plugin-developer/schema-toolbar/schema-toolbar.tsx
```diff
@@ -0,0 +1,165 @@
...
+          size="small"
+          variant="text"
+          onClick={handleReset}
+          sx={{ flexShrink: 0, whiteSpace: 'nowrap' }}
+        >
+          Reset
+        </Button>
+      </Tooltip>
+
+      <Dialog open={saveOpen} onClose={closeSaveDialog}>
```
solovevayaroslavna: The save dialog is a one-field form built from raw MUI `Dialog` + `useState(draftName)` + manual `trim()` validation + manual disabled state. 

The app already has `FormDialog` (`components/form-dialog`), which wraps ui-lib `Dialog`/`DialogTitle` and runs react-hook-form + a zod resolver with submit/cancel/validation wired up.

ui/apps/everest/src/pages/plugin-developer/ui-generator-builder.tsx
solovevayaroslavna: The `isDragging ? { transition: 'none' } : { transition: 'width 0.2s ease' }` (and `backgroundColor` variant) branch is copy-pasted across all three panes. 

A tiny local helper like paneTransition collapses the repetition and keeps the three panes in sync.

solovevayaroslavna (changes requested):

ui/apps/everest/src/pages/plugin-developer/hooks/use-schema-storage.ts
Denyme24: done. moved the guard into the filter as a type predicate, so it infers SavedSchemas on its own and the cast is gone. your snippet dropped in as is.

ui/apps/everest/src/pages/plugin-developer/output-panel/output-panel.tsx
Denyme24: yeah you're right on it. i renamed it: the card is "Generated values" now and the empty state reads "Fill the form and submit to see the generated values." no more create database promise.

ui/apps/everest/src/pages/plugin-developer/schema-toolbar/schema-toolbar.tsx
Denyme24: done. added output-panel.messages.ts, schema-toolbar.messages.ts and ui-generator-builder.messages.ts, and moved all the copy into them, so the toolbar, the tabs, the dialog and the namespace label and helper text.

ui/apps/everest/src/pages/plugin-developer/output-panel/output-panel.tsx
Denyme24: done.

ui/apps/everest/src/pages/plugin-developer/output-panel/output-panel.tsx
Denyme24: done. output panel is a FormCard, form preview is a RoundedBox, title uses sectionHeading now.

ui/apps/everest/src/pages/plugin-developer/schema-toolbar/schema-toolbar.tsx
Denyme24: done, the save dialog is a FormDialog now with a small zod schema (name trimmed, min 1). submit stays disabled until the name is valid and the trimming happens in the schema, so there's no manual state or validation left. added a test that the trimmed name is what gets saved, since it ends up as the storage key.

ui/apps/everest/src/pages/plugin-developer/ui-generator-builder.tsx
Denyme24: yep done!!

Denyme24:
@solovevayaroslavna thanks for reviewing, all the threads were valid and relevant. have made the changes to resolve those review comments. PTAL

Issue #2661, opened 2026-07-29
Design discussion: long-term plugin design-system contract


Anuragkumar-687:
### Is your feature request related to a problem?

OpenEverest's generic plugin design aims for plugins to inherit the host design system so they feel native to the application.

The current implementation explored in #2510 and plugin-hub#1 achieves this using browser-native import maps and shared runtime instances. While the approach works, it also introduces maintenance trade-offs around shared runtime contracts, version coordination, and build-time coupling.

Before moving forward with implementation, I'd like to align on the long-term mechanism for exposing the host design system to plugins.

### Describe the solution you'd like

Rather than proposing a specific implementation, I'd like to start a design discussion around the long-term plugin UI contract.

The goal is to agree on the requirements, constraints, and trade-offs before choosing an implementation mechanism.

### How would users interact with this feature?

Plugin developers should be able to build UI that automatically inherits the host application's design system.

From an end-user perspective, plugin interfaces should feel like a native part of OpenEverest, including support for theming and future design-system evolution.

### Alternative solutions considered

Several implementation directions were investigated, including improving the current import-map approach, runtime-based contracts, design-token-based approaches, and host-owned UI primitives.

This issue intentionally does not recommend one approach. The goal is to agree on the problem framing before discussing implementation.

### Additional Context

# Design discussion: long-term plugin design-system contract

## Background

The generic plugins design document (§7.4) establishes that plugins should use MUI components, inherit the host theme, and style using `sx`/`styled`. The design goal itself is already established.

Two recent PRs explored one implementation of this goal:

- **openeverest/openeverest#2510** — Introduced browser-native import maps to share React, MUI, Emotion, and React Router as browser-level singletons between the host and plugin bundles.
- **openeverest/plugin-hub#1** — Consumed that mechanism by migrating plugin-hub from raw `createElement` calls to JSX with MUI components.

Both PRs demonstrate that the approach is technically viable.

However, after reviewing the implementation, there are questions about whether the current import-map based approach is the right long-term contract for this problem—particularly as the plugin ecosystem grows.

The goal of this issue is to discuss the problem, the observed trade-offs, and the long-term contract before implementation proceeds further.

---

# Current implementation

Today, plugins are dynamically loaded into the host application and rendered inside the host React tree. This means they already participate in the application's provider hierarchy (such as `ThemeProvider` and `CacheProvider`).

The remaining challenge is ensuring that plugins and the host actually share the same UI runtime.

[...]

PR #3076, opened 2026-08-29
Plugin UI to match core

spron-in:
Closes the design discussion in #2661.

## Why

Generic plugins (spec 003) render their own UI, but today it doesn't match
OpenEverest and looks bolted-on. We want two things that pull against each other:

1. native look - plugins should be visually indistinguishable from the core.
2. independence - upgrading the core's MUI/design libs must not break existing plugins. We can't be chasing plugin authors to rebuild every time.

This PR implements Model B from the issue: share only React; each plugin
bundles its own pinned MUI; the core exposes its design tokens as CSS
variables so plugins inherit the palette + dark mode live, through the CSS
cascade, without rebuilding.

```
Host owns:  React runtime (shared)  +  design tokens on :root (--mui-*)
Plugin owns: its own MUI/Emotion (frozen)  ->  themed from host CSS variables
Contract:   React major (the only shared runtime dep)
```

## 1. What changes in the core

- **Design tokens as CSS variables.** The host theme now enables MUI
  `cssVariables`, emitting `--mui-*` custom properties on `:root`, and publishes
  the active color scheme on `document.documentElement[data-everest-color-scheme]`.
  (Switched the theme provider from Emotion's to MUI's `ThemeProvider`, required
  for variable injection.)
- **Shared React via an import map.** The host exposes its React singleton
  (`react`, `react-dom`, `react/jsx-runtime`) — plus MUI's static `colors` — through
  small shim modules served under `/static/plugin-runtime/` and an inline
  `<script type="importmap">`. Plugins mark these external so hooks/context work
  across the host↔plugin boundary. CSP gained a `script-src 'self' $NONCE` so the
  nonced import map is allowed.
- **New package `@openeverest/ui-lib`.** Bundles its own pinned MUI + Emotion and
  exports a `PluginThemeProvider` (namespaced Emotion cache, theme bound to the
  host CSS-variable tokens, follows host dark mode) plus a curated set of
  host-themed components. This is the seam that lets the core upgrade MUI without
  breaking plugins.
- **Compatibility contract.** `PluginApi` gains `cssNonce`, `hostVersion`, and a
  coarse `uiContractVersion` (React major). The loader checks the Plugin CR's
  `compatibleHostVersions` and refuses an incompatible plugin loudly instead of
  rendering it detached/unthemed; the API now forwards that field.

## 2. What changes for plugin developers

To adopt the core look (opt-in — plugins can still ship their own UI):

- **Use `@openeverest/ui-lib`** instead of importing `@mui/material` directly, and
  wrap your UI in `PluginThemeProvider`:
  ```tsx
  import { PluginThemeProvider, Box, Button, Table } from '@openeverest/ui-lib';
  // <PluginThemeProvider nonce={api.cssNonce}> ...your UI... </PluginThemeProvider>
  ```
  Colors, spacing, typography and dark mode come from the host automatically.
- **Build config**: keep React external, bundle everything else.
  ```ts
  rollupOptions: { external: ['react', 'react-dom', 'react/jsx-runtime', '@mui/material/colors'] }
  define: { 'process.env.NODE_ENV': JSON.stringify('production') } // lib mode doesn't inline this
  ```
- **(Optional) declare compatibility** in the Plugin CR:
  `spec.compatibleHostVersions: ">=2.0.0 <3.0.0"`.
- **You choose when to upgrade** `@openeverest/ui-lib`. A core MUI bump does not
  require you to rebuild; your plugin keeps its pinned MUI and still inherits the
  host palette/dark mode via CSS variables.

## Reference / scope

- `plugin-hub` is migrated end-to-end as the reference (separate companion PR):
  raw `createElement` + inline hex → host-themed MUI via `@openeverest/ui-lib`.
- Spec 003 §8.1/§9 and `docs/process/generic-plugins-design.md` updated to Model B.
- **Not** in scope: migrating other plugins, a full host-owned component facade,
  and enforced third-party version gating (the contract stays coarse/advisory).

## Notes / trade-offs

- Bundling MUI per plugin adds a few hundred KB per bundle — acceptable for an
  admin tool, and the price of upgrade independence.
- Component `styleOverrides` are not inherited (only palette/spacing/typography via
  tokens) — "best-effort native", per the issue discussion.

### DIFF
--- a/ui/apps/everest/index.html
+++ b/ui/apps/everest/index.html
@@ -6,6 +6,16 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta name="csp-nonce" content="{{.CSPNonce}}" />
     <title>OpenEverest</title>
+    <script type="importmap" nonce="{{.CSPNonce}}">
+      {
+        "imports": {
+          "react": "/static/plugin-runtime/react.js",
+          "react-dom": "/static/plugin-runtime/react-dom.js",
+          "react/jsx-runtime": "/static/plugin-runtime/react-jsx-runtime.js",
+          "@mui/material/colors": "/static/plugin-runtime/mui-colors.js"
+        }
+      }
+    </script>
     <style id="_goober" nonce="{{.CSPNonce}}">
       /* do not remove */
     </style>
--- a/ui/apps/everest/index.html
+++ b/ui/apps/everest/index.html
@@ -5,6 +5,7 @@
     <link rel="icon" type="image/svg+xml" href="/static/everest.svg" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta name="csp-nonce" content="{{.CSPNonce}}" />
+    <meta name="everest-version" content="{{.EverestVersion}}" />
     <title>OpenEverest</title>
     <script type="importmap" nonce="{{.CSPNonce}}">
       {
--- a/ui/apps/everest/src/contexts/plugins/plugin-version.ts
+++ b/ui/apps/everest/src/contexts/plugins/plugin-version.ts
@@ -1,0 +1,84 @@
+// Copyright (C) 2026 The OpenEverest Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Host-version handling for the plugin compatibility check, kept out of the
+// context module so it can be unit tested without pulling in the provider.
+
+/**
+ * Everest reports versions with a leading `v` and an optional build suffix
+ * (e.g. "v0.3.0-1-a93bef"). Returns null when the value isn't a usable version
+ * — including the "dev" fallback and the unrendered Go template placeholder
+ * that `vite dev` serves, since it doesn't run the server-side template.
+ */
+export function parseVersion(value: string): [number, number, number] | null {
+  const m = value
+    .trim()
+    .replace(/^v/i, '')
+    .match(/^(\d+)\.(\d+)\.(\d+)/);
+  return m ? [Number(m[1]), Number(m[2]), Number(m[3])] : null;
+}
+
+/**
+ * Minimal semver-range satisfaction for the common comparators used in the
+ * Plugin CR's compatibleHostVersions (">=2.0.0 <3.0.0", "^2.1.0", "~2.1.0").
+ * Returns true when the range is empty/unparseable so an unknown range never
+ * silently blocks a plugin (advisory contract, first-party ecosystem).
+ */
+export function satisfiesHostVersion(version: string, range?: string): boolean {
+  if (!range) {
+    return true;
+  }
+  const host = parseVersion(version);
+  // No usable host version, or a local build (RELEASE_VERSION defaults to
+  // v0.0.0-<sha>): never gate plugins during development.
+  if (!host || (host[0] === 0 && host[1] === 0 && host[2] === 0)) {
+    return true;
+  }
+  const cmp = (a: [number, number, number], b: [number, number, number]) =>
+    a[0] - b[0] || a[1] - b[1] || a[2] - b[2];
+
+  const evalClause = (clause: string): boolean => {
+    const comparators = clause.trim().split(/\s+/).filter(Boolean);
+    if (comparators.length === 0) {
+      return true;
+    }
+    return comparators.every((c) => {
+      const m = c.match(/^(>=|<=|>|<|=|\^|~)?(v?\d+\.\d+\.\d+)/i);
+      if (!m) {
+        return true;
+      }
+      const op = m[1] || '=';
+      const target = parseVersion(m[2]);
+      if (!target) {
+        return true;
+      }
+      if (op === '^') {
+        const upper: [number, number, number] = [target[0] + 1, 0, 0];
+        return cmp(host, target) >= 0 && cmp(host, upper) < 0;
+      }
+      if (op === '~') {
+        const upper: [number, number, number] = [target[0], target[1] + 1, 0];
+        return cmp(host, target) >= 0 && cmp(host, upper) < 0;
+      }
+      const c0 = cmp(host, target);
+      if (op === '>=') return c0 >= 0;
+      if (op === '<=') return c0 <= 0;
+      if (op === '>') return c0 > 0;
+      if (op === '<') return c0 < 0;
+      return c0 === 0;
+    });
+  };
+
+  return range.split('||').some((clause) => evalClause(clause));
+}
--- a/ui/apps/everest/src/contexts/plugins/plugins.context.tsx
+++ b/ui/apps/everest/src/contexts/plugins/plugins.context.tsx
@@ -74,9 +74,8 @@ function getHostVersion(): string {
 
 function getCSPNonce(): string {
   return (
-    document
-      .querySelector("meta[name='csp-nonce']")
-      ?.getAttribute('content') || ''
+    document.querySelector("meta[name='csp-nonce']")?.getAttribute('content') ||
+    ''
   );
 }
 
@@ -128,9 +127,7 @@ function satisfiesHostVersion(version: string, range?: string): boolean {
     });
   };
 
-  return range
-    .split('||')
-    .some((clause) => evalClause(clause));
+  return range.split('||').some((clause) => evalClause(clause));
 }
 
 // Build the PluginApi object that the host passes to each plugin's register().
--- a/ui/apps/everest/src/contexts/plugins/plugins.context.tsx
+++ b/ui/apps/everest/src/contexts/plugins/plugins.context.tsx
@@ -44,6 +44,7 @@ interface PluginDescriptor {
   name: string;
   displayName: string;
   bundleUrl: string;
+  compatibleHostVersions?: string;
   extensionPoints?: ExtensionPointDescriptor[];
 }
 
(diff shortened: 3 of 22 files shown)


Anuragkumar-687:
Thanks @spron-in - I went through the implementation and also built the companion `plugin-hub` branch (`poc/design-alignment`, which vendors `ui-lib@0.1.2`) to verify the setup from the plugin side.

Overall, the direction makes sense to me: sharing React while bundling MUI/Emotion per plugin gives plugins independence from core MUI upgrades, while the host CSS variables provide the common palette/dark-mode layer.

A few things stood out:

### 1. Compatibility check

`getHostVersion()` reads `meta[name='everest-version']`, but I couldn't find where that tag is populated. The template appears to pass the CSP nonce, but not the Everest version.

If there isn't another injection point, it falls back to `"dev"`, and `satisfiesHostVersion()` accepts that before evaluating `compatibleHostVersions`. So the incompatibility rejection path wouldn't currently be reached.

Could we populate the version in the page or use the existing version endpoint?

Also, `uiContractVersion` is exposed through the API, but I couldn't find where it is checked. Is that intentionally informational for now?

### 2. Theme coverage

`PluginThemeProvider` currently builds the plugin theme from the host palette. So palette and dark mode are shared, but the complete core theme isn't.

For example, `BaseTheme` has custom Button styling and typography (`Poppins`, `textTransform: 'none'`, custom radius/padding) that aren't part of the bridged palette. This means plugin components can still look different from core components even though their colors match.

Is this intended as the documented "best-effort native" limitation, or should some of these values also be exposed through the token layer?

### 3. `@mui/material/colors`

I checked the MUI 7.3.11 build and the generated `plugin-hub` bundle and found no actual `@mui/material/colors` import. The occurrences in the MUI build are inside an error message.

If this import-map/runtime entry was added because of a specific bundling issue, could we document the repro? Otherwise it looks like this plumbing may be unnecessary.

### Smaller points

- `react/jsx-dev-runtime` isn't in the import map. If dev-mode plugin builds aren't supported, it may be worth documenting that.
- `PluginThemeProvider` defaults all plugins to the same Emotion `cacheKey`, although the JSDoc says it should be unique.
- `@mui/icons-material` is a dependency, but I couldn't find the icon re-exports mentioned in the docs.
- The built `plugin-hub` bundle is ~342 kB raw / 91 kB gzip with MUI + Emotion bundled. This seems reasonable for the trade-off, but may be worth documenting.
- I don't see tests covering the actual two-MUI-instance path, particularly theme inheritance and Emotion cache isolation. A small fixture/integration test could help here.

The first two points are the ones I'd consider most important. Happy to help investigate or work on any of them.

Anuragkumar-687:
After discussing with @spron-in, I’ll work on both 1 and 2 and push the commits to `poc/plugin-design-unification`. I’ll keep them as separate commits for easier review. I’ll start with the compatibility check, then move on to the theme coverage. I am already working on it.

ui/apps/everest/src/contexts/plugins/plugin-version.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Minimal semver-range satisfaction for the common comparators used in the
+ * Plugin CR's compatibleHostVersions (">=2.0.0 <3.0.0", "^2.1.0", "~2.1.0").
+ * Returns true when the range is empty/unparseable so an unknown range never
+ * silently blocks a plugin (advisory contract, first-party ecosystem).
+ */
+export function satisfiesHostVersion(version: string, range?: string): boolean {
+  if (!range) {
+    return true;
+  }
+  const host = parseVersion(version);
```
solovevayaroslavna: `apps/everest` already depends on `semver` (`^7.6.2`), so this file doesn't need its own version parser and range check. `semver.satisfies()` handles all these ranges (`>=`/`<`/`^`/`~`/`||`) and more edge cases than the hand-rolled one. This runs in the host bundle, so using the existing dep is free. Use `semver.coerce()` for the `v`-prefixed host version and treat `0.0.0`/unknown as ungated — but call `semver.satisfies(host, range)` instead of your own comparator. Your unit tests make the switch safe.

ui/packages/plugin-ui-lib/src/plugin-theme-provider.tsx
```diff
@@ -0,0 +1,243 @@
...
+  cacheKey?: string;
+  /** CSP nonce from the host (PluginApi.cssNonce) so injected <style> tags are allowed. */
+  nonce?: string;
+}
+
+// Wraps plugin UI in a namespaced Emotion cache + a theme bound to host tokens.
+// Deliberately renders no CssBaseline: the host owns document-level globals.
+export const PluginThemeProvider = ({
+  children,
+  cacheKey = 'openeverest-plugin',
```
solovevayaroslavna, line 218: `cacheKey = 'openeverest-plugin'` defaults every plugin to the same Emotion cache key, while the JSDoc directly above says "Keep unique per plugin to avoid style collisions with the host." Two plugins on the default share one cache and can clobber each other's injected styles — exactly what the doc warns against.

ui/packages/plugin-ui-lib/src/plugin-theme-provider.tsx
```diff
@@ -0,0 +1,243 @@
...
+function hostTypography(): ThemeOptions['typography'] {
+  const typography: Record<string, unknown> = {};
+  for (const variant of TYPOGRAPHY_VARIANTS) {
+    const style = fontFromHost(variant);
+    if (style) {
+      typography[variant] = style;
+    }
+  }
+  // Give unlisted variants the host's body face rather than MUI's default.
+  const bodyFamily = (typography.body1 as Record<string, unknown> | undefined)
```
solovevayaroslavna, line 165: AGENTS.md bans `as` assertions — narrow or annotate instead. Two here: `(typography.body1 as ...)?.fontFamily` and the return `(typography as ThemeOptions['typography'])`. Fix by typing the accumulator up front so no cast is needed:

```ts
const typography: NonNullable<ThemeOptions['typography']> = {};
```

Then `typography.body1?.fontFamily` and the return both work as-is. (`as const` on `TYPOGRAPHY_VARIANTS` is fine — only these two value casts break the rule.)

ui/apps/everest/src/contexts/plugins/plugins.context.tsx
```diff
@@ -129,6 +154,20 @@ export const PluginProvider = ({ children }: { children: ReactNode }) => {
 
       for (const descriptor of descriptors) {
         try {
+          // Reject incompatible plugins loudly rather than rendering a detached
+          // or unthemed UI (see issue #2661).
+          if (
+            !satisfiesHostVersion(
```
solovevayaroslavna, line 160: The load-time gate checks `satisfiesHostVersion(getHostVersion(), descriptor.compatibleHostVersions)` — i.e. it enforces the **host app semver**. But the contract we expose to plugins is `uiContractVersion = React.version.split('.')[0]` (the React major), and nothing ever checks it. So the enforced axis (host version) and the declared contract axis (React major) diverge, and `uiContractVersion` is currently dead surface.

Per the design (#2661), the thing a bundled-MUI plugin actually shares — and the only thing that can break it at runtime — is the React major. Host app version is a coarser, unrelated axis (it bumps for backend reasons too).


So I'd suggest: check the React major as the UI gate, keep the host version as a separate API gate, and give each field a name that says what it guards. Bundling them into one host-version number hides both.

This also helps later: when the ui-generator schema gets its own version, that's a third, separate thing to check. If everything rides on the host version, none of these breakages are caught on their own — so keeping them as separate, clearly-named checks now will scale much better

ui/apps/everest/index.html
```diff
@@ -5,7 +5,18 @@
...
     <meta name="csp-nonce" content="{{.CSPNonce}}" />
+    <meta name="everest-version" content="{{.EverestVersion}}" />
     <title>OpenEverest</title>
+    <script type="importmap" nonce="{{.CSPNonce}}">
+      {
+        "imports": {
+          "react": "/static/plugin-runtime/react.js",
+          "react-dom": "/static/plugin-runtime/react-dom.js",
+          "react/jsx-runtime": "/static/plugin-runtime/react-jsx-runtime.js",
+          "@mui/material/colors": "/static/plugin-runtime/mui-colors.js"
```
solovevayaroslavna, line 16: `@mui/material/colors` is MUI's stock Material palette (`red`/`blue`/`grey` constants) — not our brand theme, which already reaches plugins live via the `--mui-*` CSS variables. Since every plugin bundles its own MUI, it already carries these constants, so sharing them from the host duplicates nothing useful and just adds a runtime-coupling point in the CSP/import-map surface.

Two things:

- It also contradicts the spec text in this PR ("React is the only shared runtime") — if this is genuinely shared, that line is inaccurate.
- @Anuragkumar-687 checked the built `plugin-hub` bundle and found no real `@mui/material/colors` import (only inside an error string), so the "Vite lib mode can't resolve it"

ui/packages/plugin-ui-lib/src/plugin-theme-provider.tsx
```diff
@@ -0,0 +1,243 @@
...
+  if (typeof document === 'undefined') {
+    return '';
+  }
+  return getComputedStyle(document.documentElement)
+    .getPropertyValue(name)
+    .trim();
+}
+
+function colorFromHost(name: string): SimplePaletteColorOptions | undefined {
+  const main = readVar(`--mui-palette-${name}-main`);
```
solovevayaroslavna, line 45: Plugins read the host theme using MUI's own variable names (`--mui-palette-*`, `--mui-font-*`, `--mui-shape-*`). That ties the plugin contract to MUI internals: if the host bumps a MUI major that renames these, `readVar` returns empty and plugins silently fall back to default colors — losing the palette at exactly the upgrade Model B should survive.

Suggestion: let the host own a small set of our own names, mapped to MUI, so plugins read our stable names:

```css
:root {
  --everest-color-primary: var(--mui-palette-primary-main);
  --everest-color-bg:      var(--mui-palette-background-default);
  --everest-radius:        var(--mui-shape-borderRadius);
  /* a curated ~dozen, not the whole --mui-* surface */
}
```

Worth doing now: it's the public contract plugins build against, so renaming later is breaking, while naming it right up front is nearly free.

ui/packages/plugin-ui-lib/src/plugin-theme-provider.tsx
```diff
@@ -0,0 +1,243 @@
...
+  if (light) color.light = light;
+  // contrastText is intentionally NOT read from the host: some host tokens set a
+  // low-contrast value; letting MUI derive it from `main` keeps filled chips/
+  // buttons readable.
+  return color;
+}
+
+// Builds a palette of concrete colors read from the host CSS variables. Missing
+// tokens fall back to MUI defaults. Re-run whenever the host color scheme flips.
+function hostPalette(mode: PaletteMode): PaletteOptions {
```
solovevayaroslavna, line 62: The trivial semver helper is well tested, but the actual bridge this PR rests on has no tests: `readVar` / `hostPalette` / `fontFromHost` (regex-parsing MUI's font shorthand) / `useHostColorMode`. 
A small test that renders a plugin subtree under a host emitting `--mui-*` and asserts palette is inherited, toggling `data-everest-color-scheme` re-renders with dark values, `fontFromHost` parses a real `--mui-font-*` string and rejects `--mui-font-inherit`, would cover the mechanism.

ui/packages/design/src/theme-context-provider/theme-context-provider.tsx
```diff
@@ -36,10 +51,21 @@ const ThemeContextProvider = ({
   }, [saveColorModeOnLocalStorage]);
 
   const theme = useMemo(
-    () => createTheme(themeOptions(colorMode)),
+    // cssVariables emits the palette/spacing/typography as `--mui-*` custom
+    // properties on :root so plugins with their own MUI can inherit host tokens.
+    () => createTheme({ ...themeOptions(colorMode), cssVariables: true }),
```
solovevayaroslavna, line 56: Worth flagging that this is a **host-wide change, not a plugin-only one**: `ThemeContextProvider` wraps the whole Everest app, so switching it from Emotion's `ThemeProvider` to MUI's and enabling `cssVariables: true` changes how the theme is delivered for *every* core screen, not just plugins. Plugins are only the consumers of the `--mui-*` variables this produces.

Because it touches the host's core theming, it can cause visual regressions anywhere in the app (manual `theme.palette.mode` checks, hardcoded palette access, anything relying on the previous Emotion-provider / non-cssVariables output). Before landing it, I'd want snapshot / visual-regression coverage across the core screens (light + dark) so we can be confident nothing shifts.

My suggestion: split the host migration (Emotion → MUI `ThemeProvider` + `cssVariables`) into its own separate PR, gated by that snapshot testing, and keep this PR focused on the plugin machinery — so the host change can be reviewed and reverted on its own if a regression shows up, instead of being coupled to the plugin work.

visual tests have already been set up, we may need to add new screens, but in this situation it is a necessity, in my opinion

solovevayaroslavna: @spron-in alternative: we may not need to touch the host `ThemeProvider` at all. The switch is only here to make MUI emit `--mui-*` via `cssVariables`. Instead, the host could emit its own small set of `--everest-*` variables directly (a tiny component writing `:root { --everest-color-primary: ...; ... }` from the theme values), keeping the existing Emotion `ThemeProvider` and leaving `cssVariables` off. Plugins then read `--everest-*`, the host app is untouched (no regression surface), and it doubles as the stable semantic token layer.

Note: enabling CSS variables as above is aligned with MUI's own direction (theme.vars, and the longer-term move toward Pigment CSS / zero-runtime), so if we do want the built-in `cssVariables` path, it's future-friendly — I'd just land it as its own PR with snapshot coverage rather than coupling the host migration to the plugin work.

ui/packages/plugin-ui-lib/src/plugin-theme-provider.tsx
```diff
@@ -0,0 +1,243 @@
...
+// Note the shorthand cannot express `textTransform` or `letterSpacing`, so those
+// host values do not cross this bridge — see the note in §7.4.
+function fontFromHost(variant: string): Record<string, unknown> | undefined {
+  const raw = readVar(`--mui-font-${variant}`);
+  if (!raw) {
+    return undefined;
+  }
+  // Requiring a numeric weight also rejects non-font values such as
+  // `--mui-font-inherit: inherit inherit/inherit inherit`.
+  const m = raw.match(/^(\d+)\s+([^\s/]+)(?:\/([^\s]+))?(?:\s+(.+))?$/);
```
solovevayaroslavna, line 136: Reading typography as MUI's `--mui-font-*` shorthand and re-parsing it here has four issues worth weighing:

1. **Fragile parse.** The regex assumes a numeric weight first, so a variant with a font-style (`italic 600 …`) or keyword weight fails to match and silently falls back to the plugin's default font.
2. **Lossy.** The `font` shorthand can't carry `textTransform` / `letterSpacing`, so core's `textTransform: 'none'` never reaches plugins — buttons stay UPPERCASE even when colors/fonts match.
3. **MUI-coupled contract.** It reads MUI's internal `--mui-*` names/format, so a host MUI major that renames or reformats them would silently drop plugin typography — the exact upgrade this model should survive.
4. **Hand-rolled parser of a framework's serialized format** — fragile to maintain and fails quietly on any format change.

An alternative direction is host-owned, per-property tokens (e.g. `--everest-font-h1-size` / `-weight` / `-family` / `-transform`) that plugins read without parsing — decoupled from MUI, closes the textTransform gap. But this is a public-contract decision, so let's not rush it — needs more thought. We need to think about it.

spron-in:
@solovevayaroslavna please have a look. I addressed your concerns + pulled the plugin changes we made in main.

PR #3077, opened 2026-08-29
feat: add time-based retention (retention.type: count | time) to backup schedules

### LINKED ISSUE #3072: [ENHANCEMENT] Add `retention.type: count | time` to backup schedules

`Instance.spec.backup.storages[].schedules[].retentionCopies` is count-only. That works for PSMDB/PXC, but CNPG/Barman uses a time-based recovery window on `ObjectStore` (`retentionPolicy: "30d"`). Providers currently approximate count → days from cron, which is not accurate.

we add a type `time` as well which we can declare in `day` , `week` or `month` format and keep count as the default UX, but make add a `type` property in:
Add explicit retention on schedules:
```yaml
schedules:
  - name: daily
    cron: "0 2 * * *"
    retention:
      type: count   # or time
      count: 7      # when type=count
      duration: 30d # when type=time (Nd/Nw/Nm)
```
**count** - keep N recent backups (PSMDB/PXC)
**time** - keep backups within a recovery window (CNPG/Barman, pgBackRest)

UI / CLI

_No response_

### Additional Context

CNPG only supports time based recovery window, right now we'd have to do it a bit hacky way to convert the time format to count. It's not very accurate and can cause confusion. Although `pgBackRest` I found supports both time and count format for retention. It would be nice to add support for this.

CC @recharte 

Would like to raise a PR for this

jtomaszon:
Fixes #3072

Adds optional time-based retention to backup schedules via a new `retention` object with `type: count | time`. `retentionCopies` is untouched and fully backward compatible (a test asserts the schema is byte-identical).

What's included:

- `retention` on schedules: `type` (`count` default, or `time`), `count`, `duration` (pattern `^[1-9][0-9]*[dwm]$`, so zero-length windows like `0d` are rejected)
- Two CEL rules: the field matching `type` must be set, the other must not
- `EffectiveRetention()` accessor returning one resolved value (comma-ok, so invalid input can never read as keep-0); providers consume a single struct instead of re-implementing precedence
- Precedence: `retention` wins over `retentionCopies` when both are set. `retentionCopies` is not deprecated by this change
- CRD and TS types regenerated. 46 tests against the real generated CRD and CEL engine, including apiserver defaulting-then-validation order and rejection cases. Build, vet, lint clean

One UX call for maintainers: since `type` defaults to `count`, writing only `duration: 7d` is rejected instead of inferred as time-based. Safe, but users must write `type: time` explicitly. Happy to switch to inference if you prefer.

Follow-ups: optional k3d integration coverage (per CONTRIBUTING) left for a separate PR. TS types were regenerated via npx with the lockfile-pinned generator version since pnpm was not available locally; the diff is 72 lines, all retention schema.


AdityaPimpalkar:
Hi @jtomaszon apologies but this still needs to be triaged and confirmed with the maintainers. Let's not get carried away with making any PR's before that. Thanks :)

PR #3081, opened 2026-08-31
Add Instance maintenance API for disruption control

recharte:
## What

Opens Part 2 of the provider upgrades feature ([spec 009](https://github.com/openeverest/specs/blob/main/specs/009-provider-upgrades.md) §7): the declarative, GitOps-native control surface for holding and approving disruptive actions. Types only — no behavior change; the `Context.RequestMaintenance` runtime helper and reconciler flush land in the next PR.

**`spec.maintenance`:**
- `autoApproveUpTo` — standing disruption tolerance on the observable-impact scale `NonDisruptive < RollingRestart < Downtime` (what the *application* experiences, never operator internals). Defaults to `NonDisruptive`, so every restart-or-worse action is held: a user needs zero understanding of providers or operators to be safe from surprise downtime.
- `approved` — one-time authorization naming the exact `approvalToken` of a held action. Matched literally, so it authorizes that single occurrence and re-arms automatically (a later action carries a different token — a stale approval can never silently authorize an unrelated action).

**`status.pendingMaintenance`** — the held actions (description, severity, approvalToken), recomputed on every reconcile from what the provider currently requests, so the list can never go stale. Durable approval in `spec`, ephemeral notice in `status`.

**`MaintenancePending` condition** + `AwaitingApproval`/`NoActionsPending` reasons.

The mechanism is cause-agnostic by construction: nothing in the API is upgrade-specific, so the same surface later governs e.g. certificate rotations (spec §7.5).

## How

Field/type additions in `api/core/v1alpha1/instance_types.go`; everything else is `make gen` output (CRDs — including InstancePreset, which embeds InstanceSpec — clients, TS types).

## Follow-ups

- `Context.RequestMaintenance` + reconciler flush (next in the stack).
- CRD sync to helm-charts after merge (with the immediate go.mod chart-pin bump).

## Verification

`make build` and `-race` tests on consumer packages pass; lint clean on the API package.


chilagrow (approved):
Looks good!

PR #3101, opened 2026-09-01
Hold repeatedly-failing approved maintenance actions

recharte:
## What

The failure-handling escape hatches for disruption control ([spec 009](https://github.com/openeverest/specs/blob/main/specs/009-provider-upgrades.md) §7.7). Completes the Track 2 runtime started in #3081/#3082.

> **Stacked on #3082** — base is `provider-upgrade-request-maintenance`; will be retargeted once that merges. Only the last commit is new.

**Circuit breaker.** The reconciler counts consecutive failed `Sync`s against each disruptive action that was approved in that pass. At the threshold (5) the action is **held despite its approval** — `RequestMaintenance` returns `false` and the action reappears on `status.pendingMaintenance` — so a crash-looping provider cannot repeatedly disrupt a database. The breaker state is surfaced on the `MaintenancePending` condition with the new `RetriesExhausted` reason. Backoff *between* attempts is deliberately not reimplemented: controller-runtime's rate-limited requeue already provides per-item exponential backoff.

**Clear-stuck is declarative** (settles spec open question 2): changing or clearing `spec.maintenance.approved` resets the counts — so "clear, then re-set the approval" retries a tripped action, with no imperative reset signal or annotation.

Scope/behavior notes for review:
- Only actions above `NonDisruptive` are counted; failures of Syncs with no approved disruptive action never trip anything.
- Counts are in-memory: a controller restart re-arms retries. Worst case that re-attempts an action the owner already approved, with requeue backoff still spacing attempts — accepted for v1 rather than persisting attempt counts in status.
- Successful `Sync` resets the Instance's counts.

## Verification

Unit tests: threshold boundary, approval-change reset, success reset, no-approved-actions never counts, per-Instance isolation, breaker-held despite approval, `RetriesExhausted` condition reason. `go test -race ./provider-runtime/...` passes; lint clean.

### DIFF
--- a/api/core/v1alpha1/instance_types.go
+++ b/api/core/v1alpha1/instance_types.go
@@ -715,6 +715,12 @@ const (
 	// held; everything the provider requested was within the Instance's
 	// standing tolerance or has been applied.
 	ReasonNoActionsPending = "NoActionsPending"
+
+	// ReasonRetriesExhausted indicates an approved disruptive action kept
+	// failing and the runtime stopped retrying it so a crash-looping
+	// provider cannot repeatedly disrupt the database. Clear and re-set
+	// spec.maintenance.approved to retry.
+	ReasonRetriesExhausted = "RetriesExhausted"
 )
 
 // Reasons for the DataSourceReady condition.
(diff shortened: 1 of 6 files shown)


chilagrow (approved):
👍

PR #3104, opened 2026-09-01
feat: plugins are good citizens following the rules

spron-in:
Addresses concerns in #3009 that Diogo raised making Plugins good citizens. 

Makes generic plugins "good citizens" by routing their discovery/context through the standard OpenAPI + RBAC handler chain instead of hand-registered routes, and moves everything under cluster-scoped paths.

## Breaking changes

Plugin API paths are now cluster-scoped. All `/v1/plugins/...` routes move under `/v1/clusters/{cluster}/plugins/...`:

```
GET /v1/plugins → GET /v1/clusters/{cluster}/plugins
GET /v1/plugins/context → GET /v1/clusters/{cluster}/plugin-context
bundle/proxy /v1/plugins/{name}/* → /v1/clusters/{cluster}/plugins/{name}/*
POST /v1/plugins/{name}/instance-config → POST /v1/clusters/{cluster}/plugins/{name}/instance-config
```
plugins RBAC resource is now cluster-scoped (was cluster+namespace). Policy objects change from namespace/name to cluster/name — existing plugin RBAC policies must be rewritten.

Plugins hardcoding `/v1/plugins/...` will break and must use `api.basePath` / relative paths.


internal/server/plugins.go
recharte:
```suggestion
		if ok, err := pp.enforcer.Enforce(sub, rbac.ResourcePlugins, rbac.ActionAll, rbac.ClusterObjectName(c.Param("cluster"), name)); err != nil {
```

This PR adds `plugins` to `AllResources` and moves `ResourcePlugins` into `ClusterScopedResources`, so `loadAdminPolicy` now emits `(role:admin, plugins, *, */*)` — a two-segment object. Here we are asking for object `"*"`, which is one segment. casbin matches with `doublestar.Match(p.obj, r.obj)` and `*` never crosses a `/`, so `Match("*/*", "*")` is false and the admin fallback never fires.

That's a regression this PR introduces. Before it, `plugins` wasn't in `AllResources` at all, admins were denied at the list endpoint, and the UI showed nothing. Now the list works for admins — it's on the new chain and builds proper `prod/sql-explorer` objects — so plugins appear, the bundle loads over the unauthenticated GET route, and the first `pluginApi.fetch()` POST lands here and 403s. Visibly broken plugins instead of absent ones.

internal/server/handlers/k8s/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+// full plugin proxy URL. Absolute URLs, data URIs, and already-resolved /v1/
+// paths are returned unchanged. Empty strings are returned as-is.
+func resolvePluginAssetPath(cluster, pluginName, assetPath string) string {
+	if assetPath == "" {
+		return ""
+	}
+	if strings.HasPrefix(assetPath, "http://") ||
+		strings.HasPrefix(assetPath, "https://") ||
+		strings.HasPrefix(assetPath, "data:") ||
+		strings.HasPrefix(assetPath, "/v1/") {
```
recharte: This check exists so an already-resolved URL isn't prefixed twice. `/v1/plugins/` used to be exactly that set; `/v1/` is the whole API surface.

The input is plugin-supplied (`spec.icon` on the `Plugin` CR), so a plugin can now put any API path here and the UI will render it as ``. Low severity — the API uses bearer tokens, so the browser's request 401s — but nothing legitimate needs the wider check.

```suggestion
		strings.HasPrefix(assetPath, pluginBasePath(cluster, pluginName)+"/") {
```

Worth adjusting the doc comment on line 100 to match ("already-resolved plugin asset paths").

ui/apps/everest/src/hooks/api/extension-catalog/useExtensionCatalog.utils.ts
```diff
@@ -38,26 +38,30 @@ const resolveVersion = (
...
+  rawIcon?: string
+): string | undefined => {
   if (!rawIcon) {
     return undefined;
   }
   if (rawIcon.startsWith('http://') || rawIcon.startsWith('https://')) {
     return undefined;
   }
-  if (rawIcon.startsWith('data:') || rawIcon.startsWith('/v1/plugins/')) {
+  if (rawIcon.startsWith('data:') || rawIcon.startsWith('/v1/')) {
```
recharte: Same change as `resolvePluginAssetPath`, but this one reads icons out of the plugin-hub catalog, which is an external project. The comment right above this function says the check is there "so a misconfigured or malicious catalog can't inject external image loads" — `/v1/` accepts anything under the API from that same untrusted source.

```suggestion
  if (rawIcon.startsWith('data:') || rawIcon.startsWith(`/v1/clusters/${clusterName}/plugins/`)) {
```

api/openapi/http-api.yaml
```diff
@@ -636,6 +638,68 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PluginDescriptorList'
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  '/clusters/{cluster}/plugins/context':
```
recharte: Echo prefers a static segment over `:name`, so `GET /v1/clusters/main/plugins/context` resolves to this endpoint and a plugin actually named `context` can never be proxied on GET — while its POST/PUT/DELETE still work, which is the confusing part.

Two other reasons to move it rather than accept the collision. `docs/process/generic-plugins-design.md` already calls this endpoint `GET /v1/plugin-context` (lines 1087, 1139, 1218), so the spec and the design doc currently disagree. And it's a computed view rather than a resource, which in this API means a non-resource path — `/permissions`, `/settings`, `/cluster-info` — not a sub-path of a collection.

```suggestion
  '/clusters/{cluster}/plugin-context':
```

`x-everest-resource-name: plugins` stays as-is, so RBAC is unchanged, and `make gen` picks up the new path. Nothing hand-written calls it today — only the generated client and TS types — so there's no caller to update.

recharte (changes requested):

recharte (approved):

PR #3108, opened 2026-09-01
fix(cli): validate --interval on instance status

### LINKED ISSUE #3107: everestctl instance status: --interval is accepted without --watch and silently coerced when invalid
### What is the issue?

`instance status --interval` documents itself as `"Poll interval (only valid with --watch)"`, but nothing enforces that, and an invalid value is silently replaced rather than rejected.

The identical constraint on the sibling flag *is* enforced — `instance create --timeout` without `--wait` fails cleanly. So the same rule is validated in one command and ignored in another.

**Expected**
- `--interval` without `--watch` → error, matching the help text and `--timeout`'s behaviour
- `--interval 0s` → error (a non-positive poll interval is meaningless)

**Actual**
- `--interval` without `--watch` is accepted and ignored
- `--interval 0s` is silently rewritten to 2s

`statusRun` (`commands/instance/status.go:83`) does no flag validation at all — it goes straight to `NewInstanceStatusRunner`. Compare `commands/instance/create.go:122`:

```go
if err := waitcmd.ValidateWaitFlags(createOpts.Wait, cmd.Flags().Changed(cli.FlagInstanceTimeout), createOpts.Timeout); err != nil {
	output.PrintError(err, logger.GetLogger(), createCfg.Pretty)
	os.Exit(waitcmd.ExitFailed)
}
```

The silent coercion is in `pkg/cli/instance/status.go`:

```go
interval := opts.Interval
if interval <= 0 {
	interval = 2 * time.Second
}
```

### Steps to Reproduce

`everestctl` built from the `v2.0.0-dev.2` tag (behaviour unchanged on `main`).

```bash
# 1. --interval without --watch — help says "only valid with --watch"
everestctl instance status --name x -n y --interval 5s

# 2. invalid interval
everestctl instance status --name x -n y --watch --interval 0s

# 3. contrast — the same constraint on --timeout is enforced
everestctl instance create --name x -n y --provider p --timeout 5m
```

### Relevant Logs / Output

_No response_

### OpenEverest Version

 v2.0.0-dev.2

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Add a `ValidateWatchFlags` twin alongside the existing `waitcmd.ValidateWaitFlags` (`pkg/cli/waitcmd/waitcmd.go:46-53`) and call it from `statusRun`, rejecting `--interval` without `--watch` and any non-positive interval. Then drop the `if interval <= 0` coercion in `pkg/cli/instance/status.go`.

Happy to open a PR.

### Additional Context

_No response_

daniarjabagin:
Fixes #3107

## Why

`instance status --interval` says it is "only valid with --watch", but nothing enforced it: without `--watch` the flag was accepted and ignored, and `--interval 0s` was silently rewritten to 2s. The sibling `instance create --timeout` already rejects the same two cases through `waitcmd.ValidateWaitFlags`, so the rule was checked in one command and skipped in another.

## What

- `waitcmd.ValidateWatchFlags` next to `ValidateWaitFlags`, rejecting `--interval` without `--watch` and any non-positive interval
- `statusRun` calls it before building the runner, printing the error and exiting 1, same as `createRun`
- the `if interval <= 0` coercion in the watch loop is gone: the value is validated before it gets there, so the loop uses `opts.Interval` as given
- table test for the new validator alongside the existing one

## Checked

`go test ./pkg/cli/waitcmd/ ./pkg/cli/instance/` passes; `golangci-lint run --new-from-rev=HEAD` on the touched packages reports 0 issues.

Built `everestctl` from this branch and ran the three cases from the issue:

```
$ everestctl instance status --name x -n y --interval 5s
❌ --interval is only valid together with --watch
(exit 1)

$ everestctl instance status --name x -n y --watch --interval 0s
❌ --interval must be a positive duration (use e.g. --interval 5s)
(exit 1)

$ everestctl instance status --name x -n y --watch --interval 5s
❌ no active context found in config — run 'everestctl auth login' first
```

The last one shows a valid combination getting past validation to the normal auth check.

### DIFF
--- a/commands/instance/status.go
+++ b/commands/instance/status.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openeverest/openeverest/v2/pkg/cli"
 	"github.com/openeverest/openeverest/v2/pkg/cli/config"
 	instancecli "github.com/openeverest/openeverest/v2/pkg/cli/instance"
+	"github.com/openeverest/openeverest/v2/pkg/cli/waitcmd"
 	"github.com/openeverest/openeverest/v2/pkg/logger"
 	"github.com/openeverest/openeverest/v2/pkg/output"
 )
@@ -83,6 +84,11 @@ func statusPreRun(cmd *cobra.Command, _ []string) { //nolint:revive
 func statusRun(cmd *cobra.Command, _ []string) { //nolint:revive
 	cfgPath, err := config.DefaultPath()
 	if err != nil {
+		output.PrintError(err, logger.GetLogger(), statusCfg.Pretty)
+		os.Exit(1)
+	}
+
+	if err := waitcmd.ValidateWatchFlags(statusOpts.Watch, cmd.Flags().Changed(cli.FlagInstanceInterval), statusOpts.Interval); err != nil {
 		output.PrintError(err, logger.GetLogger(), statusCfg.Pretty)
 		os.Exit(1)
 	}
--- a/pkg/cli/instance/status.go
+++ b/pkg/cli/instance/status.go
@@ -95,14 +95,9 @@ func (is *InstanceStatusRunner) watch(
 	opts StatusOptions,
 	lastOutput string,
 ) error {
-	interval := opts.Interval
-	if interval <= 0 {
-		interval = 2 * time.Second
-	}
-
 	poll := newInstancePoll(c, opts.Cluster, opts.Namespace, opts.Name)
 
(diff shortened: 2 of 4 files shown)


daniar-state:
Sorry about that — I missed that you wanted to open the PR yourself. Closing this so it's yours; feel free to reuse anything from the diff if it helps.

PR #3112, opened 2026-09-02
feat(api): make the backup storage PATCH a real JSON Merge Patch

### LINKED ISSUE #3020: Make `patchBackupStorage` a real JSON Merge Patch

`PATCH /clusters/{cluster}/namespaces/{namespace}/backup-storages/{name}` advertises a patch verb but implements a hand-rolled "non-zero fields win" overlay: decode the body into a full typed `BackupStorage`, fetch the current one, copy the fields that look set onto it, `Update`. Three consequences:

- **Nothing can be cleared.** String members are guarded by `!= ""`, so `bucket`, `region`, `endpointURL` and `credentialsSecretRef.name` can be changed but never unset. There is no way for a client to express deletion at all.
- **The merge is a manual field list.** Every member added to `BackupStorageS3Spec` has to be added to the overlay by hand, or a patch naming it returns 200 having silently ignored it.
- **The semantics are invented.** The endpoint accepts `application/json` and behaves like nothing in particular, so a client cannot reason about it from the content type.

Same treatment as #3019: accept `application/merge-patch+json` (RFC 7386), forward to the Kubernetes API as `types.MergePatchType` with `fieldValidation=Strict`, and delete the hand-rolled overlay.

Two things that need deciding rather than discovering:

- **The write-only credential flow has to survive.** `applyS3Credentials` reads `spec.s3.accessKeyId`/`secretAccessKey` off the object, writes them into the referenced Secret and blanks them. A merge patch forwarded straight through would bypass that entirely. The handler must intercept those two members from the patch document, apply them to the Secret, and strip them before forwarding.
- **`spec.type` is currently immutable by accident.** The overlay never reads it, so it cannot be changed through this endpoint. A real merge patch makes it writable again. #2994 leans on that property, so if it's wanted it has to become an explicit rule — a CEL rule on the CRD, or a rejected member in the handler — rather than an artefact of the merge code.

- **Leave it as-is.** Workable for today's callers, including #2994, but it rots: the overlay silently drops any field nobody remembers to add to it, and the failure mode is a successful-looking 200.

### Additional Context

Split out of #3019, which proposes the same for the Instance API. Does **not** block #2994 — the existing overlay covers every flag that issue exposes.

VijetaPriya47:
The endpoint advertised a patch verb but implemented a hand-rolled overlay: decode the body into a full BackupStorage, fetch the stored one, copy across every field that looked set, then Update. Under those semantics nothing could ever be cleared, because a string guarded by != "" cannot tell "leave this alone" from "unset this", and any member nobody remembered to add to the copy list was dropped from a request that still answered 200.

Accept application/merge-patch+json and forward the document to the Kubernetes API as types.MergePatchType, which applies it to the stored object and retries internally on conflict. A member absent from the patch keeps its value, a member set to null is removed and server side defaulting reapplies, and a patch carrying metadata.resourceVersion becomes a precondition that fails with 409 rather than quietly overwriting someone else's change.

Forward with fieldValidation=Strict. Without it the API server prunes members that are not in the CRD schema, so a misspelt path returns 200 having changed nothing, which is the exact failure a patch API exists to prevent.

A merge patch can name any member of the stored object, so the validation handler rejects status, metadata.ownerReferences, metadata.finalizers, metadata.name, metadata.namespace and the two actor annotations. Instances already carried that check, so it moves into a helper both endpoints call instead of being written twice.

spec.s3.accessKeyId and spec.s3.secretAccessKey are write-only: they belong in the Secret named by credentialsSecretRef and are never kept on the object. A document relayed as sent would persist them, so the handler lifts them out of the patch first, and writes the Secret only once the API server has accepted the rest, so a document that gets rejected cannot rotate live credentials. The patched object names the Secret, which means a patch that repoints credentialsSecretRef and rotates in the same document lands in the new one.

Supplying one half of the credential pair is a caller mistake, but it surfaced as a 500 on every verb that takes those fields. The pair check now lives in one place behind a sentinel the server maps to 400, so create, update and patch report it alike, and on patch it runs before anything is sent so a rejected pair changes nothing at all.

spec.type stays writable through the patch, deliberately. The type enum holds a single value and the field is required, so the only documents a caller can write are a no-op or a rejection, and a CEL rule would forbid nothing that is reachable.

The UI's edit call and the api-test helper now label their bodies with the new media type; an undeclared Content-Type answers 415.
Closes #3020


VijetaPriya47:
@recharte PTAL

ui/apps/everest/src/api/backupStorage.ts
```diff
@@ -89,7 +89,8 @@ export const editBackupStorageFn = async (
   const { name, namespace } = formData;
   const response = await api.patch(
     `clusters/${cluster}/namespaces/${namespace}/backup-storages/${name}`,
-    formValuesToCrdEdit(formData)
+    formValuesToCrdEdit(formData),
+    { headers: { 'Content-Type': 'application/merge-patch+json' } }
```
recharte, line 93: I think this header isn't reaching the wire, and since the old media type is no longer accepted, editing a backup storage from the UI now fails outright.
The e2e `Edit Backup Storage` failure shows in the API server logs:

```
"method":"PATCH","uri":"/v1/.../backup-storages/pr-set-bac-str-a6b9x3","status":415,"error":"header Content-Type has unexpected value \"application/json\""
```

@solovevayaroslavna can you help us out here? What's the correct way of changing the header from `application/json` to `application/merge-patch+json`.

solovevayaroslavna: @VijetaPriya47 
The 415 is caused by the e2e Playwright config, not the app and ui code.

The global `use.extraHTTPHeaders` forces `Content-Type: application/json` onto every browser request, and Playwright applies it on top of (overriding) the `application/merge-patch+json` that axios sets on the edit call:
https://github.com/openeverest/openeverest/blob/main/ui/apps/everest/.e2e/playwright.config.ts#L65-L68

So fix should be in the test config, can you try to drop the global `Content-Type`? It should help.

```suggestion
    extraHTTPHeaders: {
      Accept: 'application/json',
    },
```

recharte (approved):
@VijetaPriya47 everything looks good but please rebase and run `make gen` to fix the conflicts.

VijetaPriya47:
@recharte could you merge it?

PR #3117, opened 2026-09-02
Deliver events to subscribers in seq order

recharte:
`Hub.broadcast` assigned `evt.Seq` under `h.mu` but released the lock before fanning out to subscribers. With concurrent publishers the goroutine holding seq N+1 could win the send race against the one holding seq N, so `Subscription.Ch` delivered out of order.

That breaks the documented resume contract: a consumer that reconnects with the `Seq` of the last event it handled — which is exactly what the SSE endpoint does — advances past N and never replays it, losing the event permanently.

The fix fans out inside the same critical section that assigns the seq. Every send was already non-blocking (buffered channel plus a `default` drop case), so holding the lock across the loop cannot let a slow subscriber stall seq assignment for anyone else; `sub.close()` still runs after the unlock, since it is the one teardown step that does not touch the hub's state.

This is the root cause of the intermittent `TestHub_ConcurrentPublishReconnectNoGapsNoDuplicates` failure seen on #3112 (`expected exactly 1000 distinct events seen, got 999 - missing: [14]`) — the test was asserting a real property the hub did not hold, so the fix is in the hub, not the test. `go test -race -count=30 ./pkg/events/` passes.


chilagrow (approved):
Nice! Thanks!

PR #3118, opened 2026-09-02
fix: name the actual assignee and likely cause on delegated assign failure

### LINKED ISSUE #2839: assign bot reports the wrong reason and the wrong person when a maintainer assigns someone else

HamidKhan1001:
Fixes #2839

The bot's rejection message only made sense for self-assign. If a maintainer delegates with `/assign @someone`, it still says "I couldn't assign this to you" and tells them a maintainer will pick it up, which is weird when they're the maintainer.

#2841 already tried to fix this but got stuck on the branch rename (targets `v1.x`, so the workflow can't even run) plus a DCO issue. Redoing it fresh against `main`.

This is a draft PR for now, keeping it small, and thats why its simple too.

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -125,7 +125,11 @@ jobs:
             // confirm rather than trusting the call to have taken effect.
             const { data: after } = await github.rest.issues.get({ owner, repo, issue_number });
             if (!(after.assignees || []).some((a) => a.login === target)) {
-              await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              if (target === commenter) {
+                await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              } else {
+                await reply(`@${commenter} I couldn't assign this to @${target} — GitHub rejected the assignment. This is usually because @${target} hasn't commented on this issue yet; GitHub only accepts assignees who are collaborators or have already participated in the thread. Ask them to comment here first, then try \`/assign @${target}\` again.`);
+              }
               return;
             }
 


HamidKhan1001:
Checked both branches manually (self-assign and delegated) by running the message-building logic standalone with mock commenter/target values — both produce the right text. No existing test harness for these inline workflow scripts in this repo, so not adding one for a change this size.

Marking ready.

julismo:
Hey @HamidKhan1001, I went back through my earlier PR (#2841) and compared it with this diff. There is one bit of Diogo's feedback worth carrying over: asking the target to leave a comment and then asking the maintainer to retry still takes two steps. The target can instead post `/assign` on its own line, which both adds their participation and runs the self-assignment flow.

He also asked for the message to address the target and name who tried to assign them, so the notification makes sense to someone who hasn't interacted with the issue. That discussion is here: https://github.com/openeverest/openeverest/pull/2841#discussion_r3744458156

Something like this would cover both points:

> @target, @commenter tried to assign this issue to you, but GitHub didn't apply the assignment. This can happen if you aren't a collaborator and haven't commented on this issue yet. To try taking it yourself, post `/assign` on its own line in a new comment here.

The existing self-assignment message can stay as it is. Saying "try" also leaves room for the existing checks: the issue still needs to be open and unassigned, and the open-issue limit still applies.

I followed up with a local check of the complete embedded workflow script at `d9d0eb5`, using mocked GitHub API responses. All 18 checks passed, including both rejection paths, a successful self-assignment retry, the maintainer retry, the issue limit, closed/already-assigned issues, and unchanged behavior outside the delegated failure message. This checks the script logic, not a live GitHub Actions run.

One nuance worth making explicit: your two-step route can work too. A maintainer's retry bypasses the open-issue limit, whereas the target's own `/assign` is subject to it. So the single-action suggestion is about following the earlier review and simplifying the normal recovery path; it isn't a guarantee of assignment or a claim that the two-step route is broken.

GitHub documents commenters as eligible assignees: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/assigning-issues-and-pull-requests-to-other-github-users. The failure handler only knows the assignment did not stick, so keeping the cause conditional matters. An API request that throws an error never reaches this message.

Your PR already targets `main` and passes DCO, which covers the two remaining blockers on mine. Hope the earlier review context and these checks help get this over the line.

.github/workflows/assign.yml
```diff
@@ -125,7 +125,11 @@ jobs:
             // confirm rather than trusting the call to have taken effect.
             const { data: after } = await github.rest.issues.get({ owner, repo, issue_number });
             if (!(after.assignees || []).some((a) => a.login === target)) {
-              await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              if (target === commenter) {
+                await reply(`@${commenter} I couldn't assign this to you — GitHub rejected the assignment. A maintainer will pick this up shortly.`);
+              } else {
+                await reply(`@${commenter} I couldn't assign this to @${target} — GitHub rejected the assignment. This is usually because @${target} hasn't commented on this issue yet; GitHub only accepts assignees who are collaborators or have already participated in the thread. Ask them to comment here first, then try \`/assign @${target}\` again.`);
```
recharte, line 131: Agreed with @julismo's suggestion above — please take it. #2839 asked for the delegated message to address the target rather than the commenter, and `/assign` gets them there in one action instead of two.

recharte (changes requested):
Hey @HamidKhan1001, thanks for picking this one up. One note inline.

recharte:
@julismo thanks for coming back to this, and for flagging the cap asymmetry — a maintainer's retry bypassing the open-issue limit is a real gap in what I asked for on #2841.

HamidKhan1001:
Updated the delegated-assign message to match what @recharte  and @julismo(Thank you for suggesting the changes. Appreciate it) suggested. It now pings the target directly, names who tried to assign them, and points at posting /assign as the one-step fix instead of the old comment-then-retry flow.

Have a look at it when you are free. Thank you!

recharte (approved):
Thanks @HamidKhan1001 🚀

PR #3119, opened 2026-09-02
Make VMAgent remoteWrite order deterministic

recharte:
## What

Fixes the CI flake that failed `Controller Integration Tests (test-integration-monitoring)` on #3070 (and can hit any PR): the order-sensitive `monitoringconfig-k8s` chainsaw assertion fails whenever the informer cache returns the MonitoringConfigs reversed.

The flake is a symptom of a real controller bug, not a test problem: `genVMAgentSpec` emits `spec.remoteWrite` in the cache's non-deterministic list order, and the spec is applied via `controllerutil.CreateOrUpdate`, whose DeepEqual diff is order-sensitive — so every order flip is a real `Update`: a spurious VMAgent generation bump, a VictoriaMetrics operator reconcile, and a vmagent config reload. vmagent also tracks its per-endpoint persistent queues by remote-write *position*, so reorders across reloads reshuffle buffered samples between endpoints.

Fix: sort by namespace/name before building the spec — same set of MonitoringConfigs ⇒ same VMAgent spec. Unit test asserts both input orders produce identical `remoteWrite` entries.

## Notes

Found investigating #3070's CI (that PR never touches the monitoring path); re-run its failed job after this merges.


chilagrow (approved):
Thank you!

PR #3122, opened 2026-09-02
fix(auth): preserve post-login redirect target across session bootstrap

### LINKED ISSUE #3105: Post-login redirect should return to originally requested page

When I open a direct link to a specific page (e.g. `/someurl`) while not authenticated, I get redirected to the login page as expected. However, after entering credentials and logging in, I land on the main dashboard (`/`) instead of the original `/someurl`.

After successful authentication, the user should be redirected to the originally requested URL. The typical pattern is to save the original path in a cookie or query parameter (e.g. `?redirect_to=/someurl`) before redirecting to the login page, and use it for post-login navigation.

### To Reproduce

1. Ensure you are logged out.
2. Open a direct link to any non-root page, e.g. `https://{OPENEVEREST}/someurl`.
3. Enter credentials on the login page.
4. Observe you are redirected to `/` (main page) instead of `/someurl`.

### Expected behavior

After successful login, the user should be taken back to the page they originally tried to access.

### Alternative solutions

_No response_

### Additional context

_No response_

codebysana:
## What this fixes

After logging in from a deep link (e.g. `/databases`), users were being
redirected to `/` (dashboard) instead of the page they originally requested.

## Root cause

`ProtectedRoute` correctly saves the originally requested path via
`setRedirectRoute()` when `authStatus` is `'unknown'`. However, the
background session bootstrap check (which runs on every page load to
validate/restore the session) called `setLogoutStatus()` when no valid
token was found and that function unconditionally cleared the saved
redirect via `setRedirect(null)`, wiping out the path that was just saved
a moment earlier.

## Fix

`setLogoutStatus` now takes an optional `clearRedirect` flag (default
`false`). The redirect is only cleared when the user explicitly logs out
(`logout()`), not when the automatic/background session check determines
the user isn't authenticated. This lets the originally requested path
survive the bootstrap check and be used after a successful login.

## How to test

1. Log out.
2. Navigate directly to a real protected route (e.g. `/databases`).
3. Confirm you're redirected to `/login`.
4. Log in.
5. Confirm you land back on `/databases` instead of `/`.

Closes #3105


spron-in:
@codebysana 
I'm ok to have it in v1, but why not fix it in main?

codebysana:
> @codebysana I'm ok to have it in v1, but why not fix it in main?

I read the branch table backwards and figured v1.x was the one to target. Didn't realize main had already become the active dev branch after the swap. I'll retarget this PR to main and make sure the fix still applies cleanly there.
Thanks for flagging it, and sorry for the confusion!

codebysana:
@spron-in, Retargeted this to `main` as requested. I also verified that the fix still applies cleanly, and all local tests are passing. It’s ready for review whenever you get a chance!

PR #3126, opened 2026-09-03
Add reconciler for `BackupImport` CR for provider managed `BackupClasses`

chilagrow:
This is a part 1 of several pieces to support creating an instance directly from backup in s3.
* Part 1: https://github.com/openeverest/openeverest/pull/3126
* Part 2: https://github.com/openeverest/provider-percona-server-mongodb/pull/94
* Part 3: https://github.com/openeverest/provider-percona-server-mongodb/pull/96
* Part 4: https://github.com/openeverest/provider-percona-server-mongodb/pull/97
* Part 5: https://github.com/openeverest/openeverest/pull/3138

## What's done

This PR reconciles `BackupImport` CR in provider managed `BackupClasses`.

It adds `ImportBackup` interface for providers to implement and dedup on Backups containing duplicate path and creates `Backup` CRs.

Refs https://github.com/openeverest/openeverest/issues/2471.


provider-runtime/reconciler/import.go
```diff
@@ -0,0 +1,259 @@
...
+
+	ours, err := resolveImportOwnership(ctx, r.client, imp, r.providerName)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !ours {
+		return reconcile.Result{}, nil
+	}
+
+	// Succeeded and Failed are terminal states, so we don't re-run the import.
```
chilagrow, line 70: The provider managed import reconciler does not use `Pending` and `Running` states, only `Succeeded`, `Failed` and `Error` state. Given job based import is not implemented yet, I will remove unused state in https://github.com/openeverest/openeverest/pull/3063 for clarity.

provider-runtime/controller/import.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	backupv1alpha1 "github.com/openeverest/openeverest/v2/api/backup/v1alpha1"
+)
+
+// BackupImportNameLabel is the label key stamped on Backup CRs created by an
+// import, carrying the name of the originating BackupImport.
+const BackupImportNameLabel = "backupImportName"
```
recharte: This label key ships on every imported Backup and is public API forever, so it's worth getting right now: bare keys are conventionally reserved for user labels, and this package already prefixes its automation-set label (`ProviderLabel = "core.openeverest.io/provider"`).

```suggestion
const BackupImportNameLabel = "backup.openeverest.io/backup-import"
```

Related: label values cap at 63 characters while a BackupImport name can be 253, so stamping `imp.Name` as the value makes every Create fail for long import names and the import loops in `Error`. Probably fine to just note the limit in the const doc, unless you'd rather truncate.

provider-runtime/reconciler/import.go
```diff
@@ -0,0 +1,259 @@
...
+
+	created, err := r.createBackups(ctx, imp, toCreate)
+	if err != nil {
+		imp.Status.CreatedCount = int32(created) //nolint:gosec // count is bounded by discovered
+		return r.updateErrorStatus(ctx, imp, fmt.Errorf("failed to create backups: %w", err))
+	}
+
+	// Discovery is complete (State is empty or Succeeded). The runtime resolves
+	// an empty state to Succeeded once the imported Backups are created.
+	imp.Status.State = backupv1alpha1.BackupImportStateSucceeded
```
recharte: `applyImportResult` resolves every non-`Failed` state to `Succeeded`, so a provider that returns `State: Error` with a nil error is marked `Succeeded` — terminal, never retried — even though the enum documents `Error` as "transient; the controller may retry". Could we switch on the expected states explicitly and route `Error` (and anything unexpected) through `updateErrorStatus` so it retries like a returned error?

provider-runtime/reconciler/import.go
```diff
@@ -0,0 +1,259 @@
...
+	imp *backupv1alpha1.BackupImport,
+	backups []*backupv1alpha1.Backup,
+) (int, error) {
+	count := 0
+	for _, backup := range backups {
+		if backup.Labels == nil {
+			backup.Labels = map[string]string{}
+		}
+
+		backup.Labels[controller.BackupImportNameLabel] = imp.Name
```
recharte, line 233: The dedup list is scoped to `imp.Namespace`, but nothing forces the created Backups into it — a provider that sets a different namespace escapes dedup silently, and one that sets none fails on Create. The PSMDB implementation does the right thing, but the runtime owns this invariant; one line next to the label makes it hold for every provider:

```suggestion
		backup.Namespace = imp.Namespace
		backup.Labels[controller.BackupImportNameLabel] = imp.Name
```

provider-runtime/controller/interface.go
```diff
@@ -345,6 +345,21 @@ type BackupMirror interface {
...
+
+// BackupImporter is an optional interface that providers implement to import
+// restorable backups already sitting in a BackupStorage and turn them into
+// Backup CRs.
+type BackupImporter interface {
+	// ImportBackups lists and parses the BackupStorage and returns the
+	// discovered Backups to create along with the resulting import state. A
+	// returned error is treated as transient (the runtime retries); a Failed
+	// State in the result is treated as terminal.
+	ImportBackups(ctx context.Context, imp *backupv1alpha1.BackupImport, storage *backupv1alpha1.BackupStorage) (BackupImportExecutionStatus, error)
```
recharte, line 362: Two BackupImports pointing at the same storage reconcile concurrently (different request keys), and both can pass `dedupBackups` before either creates. At that point the `IsAlreadyExists` guard is the only thing preventing duplicate Backup CRs for the same path — and it only holds if providers derive the Backup name deterministically from the storage path. PSMDB does (`importBackupName(storageRef, path)`), but nothing asks for it. Please make it part of the contract here, e.g. "returned Backups must be named deterministically from their storage path so concurrent imports dedupe on create".

provider-runtime/reconciler/import.go
recharte, line 1: The dedup logic is the heart of this PR and it's cheap to pin with a fake client — the package already does exactly this (`cascade_delete_test.go`, `maintenance_flush_test.go`). Cases I'd want covered: a discovered backup whose (storage, path) already exists in the namespace is skipped; two discovered backups sharing a path create one CR; `Succeeded`/`Failed` status short-circuits the reconcile; a storage fetch failure lands in `State: Error` with the message set.

provider-runtime/reconciler/import.go
```diff
@@ -0,0 +1,259 @@
...
+
+// applyImportResult records a terminal failure or creates the discovered
+// Backups and marks the import Succeeded.
+func (r *backupImportReconciler) applyImportResult(
+	ctx context.Context,
+	imp *backupv1alpha1.BackupImport,
+	result controller.BackupImportExecutionStatus,
+) (reconcile.Result, error) {
+	// The provider observed a terminal, non-retryable condition. Record it and
+	// stop; the import is not retried until the spec changes.
```
recharte: `BackupImportSpec` is CEL-immutable (`self == oldSelf`), so "until the spec changes" can't happen — the only way to re-run is deleting and recreating the CR. Small one, but this comment is what the next reader will trust:

```suggestion
	// The provider observed a terminal, non-retryable condition. Record it and
	// stop; the spec is immutable, so the import re-runs only if the CR is recreated.
```

provider-runtime/controller/import.go
```diff
@@ -0,0 +1,110 @@
...
+// s3Credentials reads the S3 access key and secret key from the Secret
+// referenced by the BackupStorage. It returns empty strings when no
+// credentials secret is referenced.
+func s3Credentials(
+	ctx context.Context,
+	c client.Client,
+	storage *backupv1alpha1.BackupStorage,
+) (string, string, error) {
+	ref := storage.Spec.S3.CredentialsSecretRef.Name
+	if ref == "" {
```
recharte: `CredentialsSecretRef` is `Required` with `MinLength=1` on the name, so this branch is unreachable — and if it ever ran, `NewStaticCredentialsProvider("", "", "")` fails every request with the SDK's opaque "static credentials are empty". I'd drop the branch. The misconfiguration users will actually hit is a secret missing the two keys — erroring on that here would give a much clearer message than the SDK's.

provider-runtime/controller/import.go
chilagrow: Thanks updated. As for the label length, I've added note on the label.

provider-runtime/reconciler/import.go
chilagrow: Thanks for capturing this, my oversight, I've updated to capture the states in swithc

provider-runtime/reconciler/import.go
chilagrow: That's true, thanks 🙏

provider-runtime/controller/interface.go
chilagrow: I've added a comment here about naming used for dedulication.

provider-runtime/reconciler/import.go
chilagrow: Thanks I added some tests in addition to cases you mentioned 🙏

provider-runtime/reconciler/import.go
chilagrow: Thanks updated the comment

provider-runtime/controller/import.go
chilagrow: Thanks removed branch 🙏

provider-runtime/controller/import.go
recharte: Thanks! One knock-on from the rename: `pkg/common.BackupImportNameLabel` (from #3063) still holds the old key `"backupImportName"`, so the module now has two same-named constants with different values. It has no consumers yet, so deleting it or pointing it at the new key is cheap now — the first consumer to import the wrong one will silently query a key nobody stamps.

Tiny wording thing on the new doc line: it's the label *value* (the BackupImport name) that's capped at 63 characters, the key itself is fine as is.

provider-runtime/reconciler/import.go
recharte: Thanks — the switch reads much better. One edge the `default` branch changes: a provider returning the zero value (`State: ""`) now loops in Error with "importer returned unknown state", where before it resolved to Succeeded. Fine contract, but `BackupImportExecutionStatus.State`'s doc doesn't say the field is required — could you add a line there ("State must be set; an empty state is retried as an error"), or handle `""` explicitly in the switch?

chilagrow: Yeah, let me handle this 🙏

chilagrow: I added `BackupImportExecutionStatus.State`'s interface definition about the requirement of setting field.

provider-runtime/controller/import.go
chilagrow: The sole key is now defined in `pkg/common/constants.go`. Rewording applied too 🙏

Issue #2304, opened 2026-05-26
SSO login does not work with Auth0


gaubej:
### What is the issue?

Description
OIDC integration with Auth0 fails after successful login. Authentication completes and the redirect works, but the session cannot be established.

IdP Configuration
Auth0 (latest)
OAuth2/OIDC provider configured with PKCE

Scopes: openid, email, profile
Redirect URI set correctly (<EVEREST_URL>/login-callback)
Login itself succeeds — the user authenticates against Auth0 and is redirected back to Everest — so client ID, issuer URL, and redirect URIs are all correct.

Root Cause
Everest's session manager attempts to parse the access token as a JWT and extract the jti claim.
Auth0 issues opaque (random string) access tokens for the authorization code flow, not JWTs.
In order to get a proper access token, Everest must provide an audience parameters.
After that, 'jti' claims is not present by default in Auth0, and for example in other product like Dex. It causes problem, because they are following this list of standard claim (https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)

Expected Behaviour
Everest should :
- Provide audience in order to get an id_token for session identity
- Use standard claims, and no need to add custom claims like 'jti' or 'uti' (https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)
or
- Document that provider that can't emit non-opaque JWT token can't be supported
- Document that provider that can't add custom claims can't be supported (for example Dex)

### Steps to Reproduce

Steps to Reproduce

Install OpenEverest 1.15.2
Configure an Auth0 OAuth2/OIDC provider per the SSO docs
Configure configMap everest-settings with OIDC configuration
Attempt to log in via SSO
Observe successful redirect but failed session creation in logs

### Relevant Logs / Output

JWT problems

```
{"time":"2026-05-25T23:16:18.823349911Z","id":"c9447465b1b44fe30f6afc3d5ee25194","remote_ip":"91.162.58.80","host":"dbaas.infra.connectwork.fr","method":"GET","uri":"/static/roboto-latin-400-normal-DXyFPIdK.woff2","user_agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36 Edg/148.0.0.0","status":200,"error":"","latency":270453,"latency_human":"270.453µs","bytes_in":0,"bytes_out":18536}
[...]

PR #3128, opened 2026-09-03
Sso/add opaque and fix redirect

### LINKED ISSUE #1904: SSO login does not work with Authentik
## Description

OIDC integration with Authentik fails after successful login. Authentication completes and the redirect works, but the session cannot be established.

## IdP Configuration

- Authentik (latest)
- OAuth2/OIDC provider configured with PKCE
- Signing key assigned (RSA)
- Scopes: openid, email, profile
- Redirect URI set correctly (`<EVEREST_URL>/login-callback`)

Login itself succeeds — the user authenticates against Authentik and is redirected back to Everest — so client ID, issuer URL, and redirect URIs are all correct.

## Root Cause

Everest's session manager (`session/manager.go:215`) attempts to parse the access token as a JWT and extract the `jti` claim. Authentik issues opaque (random string) access tokens for the authorization code flow, not JWTs. Only the `id_token` is a signed JWT. This is compliant with the OIDC spec — the spec does not require access tokens to be JWTs.

Authentik has no setting to switch access tokens to JWT format for authorization code grants. This means Everest's OIDC integration is currently incompatible with Authentik (and likely any other IdP that issues opaque access tokens).

## Expected Behaviour

Everest should either:

1. Use the `id_token` (which is always a JWT per the OIDC spec) for session identity and `jti` extraction, or
2. Support opaque access tokens via the IdP's token introspection endpoint (`/application/o/introspect/`), or
3. Document that the IdP must issue JWT access tokens as a requirement for OIDC integration.

## Steps to Reproduce

1. Install OpenEverest 1.13.0
2. Configure an Authentik OAuth2/OIDC provider per the SSO docs
3. Run `everestctl settings oidc configure --issuer-url=<url> --client-id=<id>`
4. Attempt to log in via SSO
5. Observe successful redirect but failed session creation in logs

## Source

Original forum post: https://forums.percona.com/t/sso-login-does-not-work-with-authentik/40326

spron-in:
Fixes #1904 #2304
Superseds #2051

What it does: adds `POST /v1/session/sso` endpoint that validates any OIDC access token (JWT or opaque) via the IdP UserInfo endpoint and mints a short-lived Everest JWT; rewires the frontend to exchange the token and hardens the jwtDecode calls that caused the blank /login-callback.


pkg/session/manager_test.go
```diff
@@ -149,6 +156,45 @@ func TestExtractUsername(t *testing.T) {
...
+	key, err := rsa.GenerateKey(rand.Reader, testRSAKeyBits)
+	require.NoError(t, err)
+	mgr := &Manager{signingKey: key}
+
+	const (
+		subject = "oidc-subject-uuid"
+		id      = "9d1c1f98-a479-41e3-8939-c7cb3e049a"
+		issuer  = "https://idp.example.com/application/o/everest/"
+		email   = "user@example.com"
+	)
```
chilagrow: Making it more idiomatic in go.
```suggestion
	subject := "oidc-subject-uuid"
	id      := "9d1c1f98-a479-41e3-8939-c7cb3e049a"
	issuer  := "https://idp.example.com/application/o/everest/"
	email   := "user@example.com"
```

pkg/oidc/userinfo.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, userInfoURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create userinfo request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+accessToken)
+
+	client := &http.Client{Timeout: defaultHTTPClientTimeout}
+	resp, err := client.Do(req)
```
chilagrow: Could this call be bottleneck? Do we need caching?

internal/server/session.go
```diff
@@ -29,11 +29,13 @@ import (
 	"github.com/percona/everest/api"
 	"github.com/percona/everest/pkg/accounts"
 	"github.com/percona/everest/pkg/common"
+	"github.com/percona/everest/pkg/oidc"
 )
 
 const (
 	jwtSubjectTml    = "%s:%s" // username:capability
 	jwtDefaultExpiry = time.Hour * 24
+	jwtSSOExpiry     = time.Hour
```
chilagrow: This is a mismatch to the token lifetime for built-in session of 24 hours above. Also I think this means after IdP disables the user, it's valid for 1 hour. Is that expected?

spron-in: It is actually valid. 
The OIDC token is validated against the IdP's UserInfo endpoint only once, at exchange time. After that, the Everest-signed JWT is self-contained and validated locally - Everest never re-checks the IdP per request. 

It is the expected trade-off of caching identity in an Everest JWT rather than calling the rate-limited UserInfo endpoint on every request.

spron-in:
@chilagrow actually your suggestion to remove the `const` conflicts with gofmt check :) Reverted it.

chilagrow (dismissed):

solovevayaroslavna:
1) Makes sense to validate once and not hit UserInfo every request — but since the blocklist only fills on logout, an IdP-side disable isn't caught and the session stays alive for up to 1h. Is that lag ok for us?
2) Now this branch always logs out on an expired Everest token and never tries a silent renew like the old path did — so won't SSO users get kicked out roughly every hour even while their IdP session is still valid?

spron-in:
@solovevayaroslavna thanks!
1. for now I reduced the expiry to 15 mins. To get to instant expiry we need an architecture change.
2. renewal - yeah, good catch. also 401 handler now attempts to renew first before logging out.

pkg/oidc/userinfo.go
```diff
@@ -0,0 +1,74 @@
...
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// UserInfo contains the subset of claims returned by the OIDC provider's UserInfo endpoint.
+type UserInfo struct {
```
solovevayaroslavna, line 29: The exchanged Everest JWT only carries `sub` + `email`, so `name`/`preferred_username` are dropped from the session identity. Before this change the browser stored the raw OIDC access token, and `UserIcon` resolved the display name as `preferred_username || name || email || sub`. After the exchange, `preferred_username` and `name` are gone, so the app-bar falls back to `email` — or to the raw `sub` (a UUID) when the provider doesn't return `email` from UserInfo (email scope not granted / not exposed). That's a visible regression for SSO users who previously saw their username or full name.

Since UserInfo already returns these when `profile`/`email` scopes are present, could we carry them through? Add them here and propagate into `SSOClaims` + `CreateSSO` so `UserIcon` keeps resolving a human-readable name:

```go
type UserInfo struct {
    Subject           string `json:"sub"`
    Email             string `json:"email"`
    Name              string `json:"name"`
    PreferredUsername string `json:"preferred_username"`
}
```

PR #3135, opened 2026-09-04
feat: add everestctl instance update

### LINKED ISSUE #2993: Add `everestctl instance update` subcommand

VijetaPriya47:
Instances could be created, listed, watched and deleted from the terminal but never changed, so scaling replicas or growing storage meant reaching for the UI or editing the CR by hand, which skips Everest's own auth and RBAC. This fills in the last missing verb on the instance group.

Changes are expressed the way instance create already does them, with --set dot paths and a -f values file. The overrides go out as a single JSON Merge Patch and the server applies them, so members the patch does not name keep their value, a member set to null is removed, and lists are replaced wholesale. Sending one PATCH rather than a read-modify-write leaves the resourceVersion precondition, the conflict window and the retry loop on the API server, which is the only place that can do them correctly, and it drops the truncation that a typed client older than the server would otherwise write back.

That also decides what the CLI does not check. A misspelt field is caught by strict field validation and removing a required field is caught by the CRD, so both are relayed rather than duplicated here. What the server cannot catch is a misspelt component name, because spec.components is a map and any key is structurally valid, so that check stays and runs against the provider the instance already references. It fails open, and now says so out loud, since a lookup failure must not quietly pass for a check that ran.

--dry-run reads the current values and prints the paths the patch names with what they would become, because "only these fields change" is the whole contract of a patch, where a diff of two entire specs buries it. A path already at the value being set is labelled rather than drawn as an arrow to itself, so a --set that would do nothing is visible before it is sent.

--set foo=null now parses as a real null instead of the string "null". A values file already produced one for a YAML null, so the two spellings of the same override disagreed; this is also what lets update remove a field.

--wait is deliberately absent. Gating success on the rollout needs a generation that InstanceStatus does not carry yet.
Closes #2993


VijetaPriya47:
@recharte could you please take a look once?

pkg/cli/instance/create.go
```diff
@@ -390,6 +385,24 @@ func validateTopology(topology string, prov *client.Provider) error {
 
 // validateComponents only checks --set paths starting with "components.".
```
recharte, line 386: This is the other half of #3016's component-name comment: fixed on the update side, still open here. `patchedComponents` reads the merged overrides, so `-f` and `--set` are both covered by `update`. Create still parses `opts.Set` strings, so `instance create -f values.yaml` with `components: {engien: {replicas: 5}}` sails through — and the server won't catch it either, since `spec.components` is a map.

Your refactor already did the hard part. If the call site up at the top of `Run` moves below `buildSpecOverrides` and uses the two helpers you just extracted, this whole function goes away:

```go
	// Merge order: preset < -f file < --set flags.
	specOverrides, err := buildSpecOverrides(opts.ValuesFile, opts.Set)
	if err != nil {
		return err
	}
	if names := patchedComponents(specOverrides); len(names) > 0 {
		if err := validateComponentNames(names, prov, resolvedTopology); err != nil {
			return err
		}
	}
```

Before the preset merge, so it only checks names the user actually supplied.

commands/instance/update.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+func init() {
+	updateCmd.Flags().StringVar(&updateOpts.Name, cli.FlagInstanceName, "", "Instance name (required)")
+	updateCmd.Flags().StringVarP(&updateOpts.Namespace, cli.FlagInstanceNamespace, "n", "", "Namespace the instance is in (required)")
+	updateCmd.Flags().StringVar(&updateOpts.Cluster, cli.FlagInstanceCluster, "main", "Cluster name")
+	updateCmd.Flags().StringVar(&updateOpts.Context, cli.FlagInstanceContext, "", "Context to use (default: current context)")
+	updateCmd.Flags().StringVarP(&updateOpts.ValuesFile, cli.FlagInstanceFile, "f", "", "Path to a YAML file with spec overrides (--set takes precedence)")
+	updateCmd.Flags().StringArrayVar(&updateOpts.Set, cli.FlagInstanceSet, nil, "Set a spec field: --set components.engine.replicas=3 or --set backup.enabled=true (repeatable)")
+	updateCmd.Flags().BoolVar(&updateOpts.DryRun, cli.FlagInstanceDryRun, false, "Print the fields the patch would change, without writing")
```
recharte, line 83: The change list is the right thing to print, and it catches what the server can't — wrong value, wrong component, a no-op, a `-f` file about to replace a list wholesale. But it never reaches the server, so it can't catch a misspelt field: `--set components.engine.replicass=5` previews as `components.engine.replicass: <unset> -> 5`, exactly like setting a field that isn't set yet, and the real run then fails on strict validation.

Fixing that properly needs `?dryRun=All` on the API, which isn't this PR. For here, just don't let the flag promise more than it does:

```suggestion
	updateCmd.Flags().BoolVar(&updateOpts.DryRun, cli.FlagInstanceDryRun, false, "Print the fields the patch would change, without writing or validating")
```

I'm opening a separate issue for the server-side dry run.

commands/instance/update.go
```diff
@@ -0,0 +1,110 @@
...
+The name and namespace flags are required. Changes are expressed with --set
+dot-notation paths and/or a -f values file, using the same merge rules as
+'instance create' (--set wins over -f). At least one of --set/-f is required.
+
+The overrides are sent as a single JSON Merge Patch (RFC 7386), which the
+server applies:
+
+  - Every field not named by --set/-f keeps its current value, so -f is a
+    merge, not a desired-state replacement: it never prunes.
+  - Lists are replaced wholesale, never merged element by element. A partial
```
recharte, line 46: The doc half of #3016's list-indexing comment landed, and it's the right line to have. The other half didn't — `deepSet` still splits on `.` only, so `--set backup.storages[0].pitr.enabled=true` builds a literal `"storages[0]"` key.

Better than before, since the server now rejects it naming `spec.backup.storages[0]` rather than telling you to go hunting for a typo. But it still reads as a schema problem instead of "indexing isn't supported here", and `storages` and `schedules` are exactly where people will try it.

Catching `[` in a path segment in `parseSetFlags` and failing with "--set does not support list indices; set the whole list with -f" closes it, and covers `instance create` at the same time. Follow-up is fine if you'd rather keep this PR to the patch mechanism — just say which.

VijetaPriya47: hey @recharte I was thinking if we could support editing list indices also? how do you think we should handle , we can do it by cli, by first fetching the list and then changing that item. but it could introduce race condition and the second way is to send it from the server side. this is what I was able to think of? do you have any other idea in mind?

recharte: Honestly I don't have a settled answer — but I don't think it's (a). Fetching the list and sending it back is the read-modify-write we rebuilt this command to remove.

Server-side is the right instinct: #3019 flagged JSON Patch, and Server-Side Apply is probably nicer since it matches elements by key rather than by position — which is what people mean anyway.

That makes it more of a types question than a CLI one, though. SSA only merges by key if the list declares one, and all of ours are atomic. `schedules` could take `+listType=map` / `+listMapKey=name` as-is; `storages` couldn't, since its identity is nested inside `storageRef`. Might be worth revisiting those types now — Kubernetes warns against flipping a list from atomic to map later, so it's cheapest while we're still `v1alpha1`.

I'll open issues for both. For this PR I'd still just like the honest failure — detect `[` in a path segment and say "--set does not support list indices; set the whole list with -f".

pkg/cli/instance/create.go
recharte: Reopening — I think this got swept up when you resolved the batch. `create.go` hasn't changed since, so `instance create -f values.yaml` with `components: {engien: {replicas: 5}}` still goes through unchecked, and the server won't catch it either since `spec.components` is a map.

Still just the two helpers you extracted, moved below the merge at line 143. Happy for it to be a follow-up if you'd rather keep this PR to `update` — just say so rather than closing it.

recharte:
@VijetaPriya47 I reopened the 1st thread and left a new comment on the 3rd thread, please have a look.

VijetaPriya47:
@recharte resolved it

recharte (approved):
Looks good 🚀 thanks

PR #3139, opened 2026-09-07
feat(api): add update command for backup storage management

### LINKED ISSUE #2994: Add `everestctl backupstorage update` subcommand

Backup storages can be created, listed, and deleted from the terminal, but not modified. The most common day-2 need — rotating S3 credentials — currently means deleting and recreating the storage (blocked while instances reference it) or editing Secrets by hand with kubectl.

Add a `backupstorage update` subcommand. Required flags: `--name`, `--namespace`, plus at least one field flag.

Deliberately **only the safe-to-edit fields** are exposed:
- `--access-key-id` / `--secret-access-key` — credential rotation via the spec's write-only `s3.accessKeyId`/`s3.secretAccessKey` fields (the webhook persists them into the referenced Secret and clears them from the object), using the same input mechanism as `backupstorage create`.
- `--credentials-secret` — re-point `s3.credentialsSecretRef` to a different existing Secret; mutually exclusive with the credential pair.
- `--verify-tls` / `--force-path-style` — connection settings.

The identity coordinates (`type`, `s3.bucket`, `s3.region`, `s3.endpointURL`) are **not** exposed, even though the API does not currently enforce their immutability: re-pointing them on a storage that already holds backups orphans every artifact referenced through it (Backup CRs reference the storage by name; engines registered the old coordinates). Re-homing a storage is create-new + delete-old, not an update. The help text states this and the reasoning.

Uses the existing `UpdateBackupStorage` endpoint via read-modify-write; a 409 `resourceVersion` conflict is re-fetched and reapplied once. A 404 is mapped to `backup storage %q not found in namespace %q`. Follows existing CLI conventions: `--json`/`--verbose`, `--cluster`, `--context`.

```
# Rotate credentials
everestctl backupstorage update --name my-s3 --namespace everest \
  --access-key-id AKIA... --secret-access-key ...

# Point at an externally managed Secret instead
everestctl backupstorage update --name my-s3 --namespace everest \
  --credentials-secret my-rotated-creds

# Toggle TLS verification for a self-signed MinIO
everestctl backupstorage update --name my-minio --namespace everest --verify-tls=false
```

- `kubectl edit backupstorage` / editing the Secret directly — bypasses Everest API authentication/RBAC and the write-only credential flow.
- Exposing all spec fields via `--set` like `instance update` — rejected: the schema is small and fixed, and generic paths would silently invite the dangerous identity edits described above.

### Additional Context

Completes create/list/update/delete for the `backupstorage` group. If identity-field immutability is later enforced server-side (CEL or validation handler — arguably it should be), the CLI's stance is already aligned.

VijetaPriya47:
Rotating S3 credentials currently means deleting and recreating the storage (blocked while instances reference it) or hand-editing the credentials Secret with kubectl, which bypasses Everest's RBAC and its write-only credential handling.

Add `backup-storage update` with --access-key-id/--credentials-secret (mutually exclusive) for credential rotation, and --verify-tls/ --force-path-style for connection settings. Changes are sent as a single JSON merge patch, so identity fields (type, bucket, region, endpoint-url) are never named in the request and stay untouched even though the API doesn't reject such a patch itself; re-homing a storage that already holds backups is create-new + delete-old, not an update, so those fields intentionally have no flags here.

Closes #2994


VijetaPriya47:
@recharte could you please test it out once?

recharte (approved):
Well done 💪

PR #3153, opened 2026-09-08
fix(cli): reject invalid --interval/--watch combos on instance status

### LINKED ISSUE #3107: everestctl instance status: --interval is accepted without --watch and silently coerced when invalid

shivansh-source:
instance status --interval documented itself as "only valid with --watch" but nothing enforced that, and an invalid value was silently rewritten instead of rejected. The sibling flag pair (--timeout/--wait on instance create) already validates this correctly, so this brings status in line with it.

--interval without --watch → now errors, instead of being silently ignored
--interval 0s (or any non-positive value) → now errors, instead of being silently rewritten to 2s
Extracted the shared validation logic into waitcmd.ValidatePollFlags, reusing the same pattern as waitcmd.ValidateWaitFlags
Confirmed the --watch/-w example next to --interval in the command help, as requested in review
Added TestValidatePollFlags in pkg/cli/waitcmd/waitcmd_test.go covering: interval without watch, watch with zero interval, watch with a valid interval
Fixes #3107


shivansh-source:
@chilagrow PTAL  thank you !

PR #3163, opened 2026-09-10
fix(version): gracefully handle unparseable version strings in IsRC and IsDev

Norway-02:
### Summary
`version.IsRC(v)` and `version.IsDev(v)` in `pkg/version/version.go` previously used `version.Must(version.NewVersion(v))`, which panics when provided with unparseable or non-semver version strings (e.g. `"latest"`, `"invalid"`, or empty inputs).

This PR updates `IsRC` and `IsDev` to handle version parsing errors gracefully by returning `false` instead of panicking. Unit tests have been added to cover invalid and empty version strings.

### Changes
- `pkg/version/version.go`: Check for `err != nil` from `version.NewVersion(v)` and return `false`.
- `pkg/version/version_test.go`: Added test cases for `"invalid"` and `""` inputs to `TestIsRC` and `TestIsDev`.


Norway-02:
cc @spron-in @recharte @chilagrow - please take a look when you have a moment. Thanks!

PR #3167, opened 2026-09-12
fix(common): add nil check in NewSupportedVersion to prevent panic

Norway-02:
### Summary
`common.NewSupportedVersion(meta)` in `pkg/common/version.go` evaluates `meta.GetSupported()` directly. If `meta` is `nil` (e.g. when network fetches fail or version metadata is missing), calling `meta.GetSupported()` causes an unhandled nil pointer dereference panic, crashing the CLI tool during `everestctl install` or `everestctl upgrade`.

This PR adds a `nil` check at the start of `NewSupportedVersion` to return an explicit error instead of panicking.

### Changes
- `pkg/common/version.go`: Return an error if `meta == nil`.
- `pkg/common/version_test.go`: Added `TestNewSupportedVersion` unit tests covering `nil` metadata handling, valid constraint parsing, and invalid constraint errors.

### DIFF
--- a/pkg/common/version.go
+++ b/pkg/common/version.go
@@ -73,6 +73,10 @@ func CheckConstraint[V version](v V, c string) bool {
 
 // NewSupportedVersion returns a new SupportedVersion struct.
 func NewSupportedVersion(meta *versionpb.MetadataVersion) (*SupportedVersion, error) {
+	if meta == nil {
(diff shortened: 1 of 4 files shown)


Norway-02:
cc @spron-in @recharte @chilagrow - please review when you get a chance. Thanks!

PR #3179, opened 2026-09-15
fix(common): gracefully handle invalid versions and constraints in CheckConstraint

Norway-02:
### Summary
`common.CheckConstraint(v, c)` in `pkg/common/version.go` previously used `goversion.Must(goversion.NewVersion(val))` and `panic(err)`, which panics when provided with unparseable or non-semver version strings or malformed constraint strings.

This PR updates `CheckConstraint` to handle version parsing and constraint parsing errors gracefully by returning `false` instead of panicking.

### Changes
- `pkg/common/version.go`: Check for `err != nil` from `goversion.NewVersion(val)` and `goversion.NewConstraint(c)` and return `false`.


Norway-02:
cc @spron-in @recharte @chilagrow - please take a look when you get a chance. Thanks!

PR #3180, opened 2026-09-15
fix(server): add HTTP timeout and non-OK logging for telemetry

### LINKED ISSUE #2203: telemetry: add request timeout and log response body on non-200

shivansh-source:
`(*EverestServer).report` sent telemetry via `http.DefaultClient`, which has no
timeout, so a slow or unreachable endpoint could block the telemetry job
indefinitely. On a non-200 it logged only the status code.

## Changes

- Outbound telemetry uses an `http.Client` with a 30s timeout.
- The POST is extracted into `postTelemetryPayload`, which takes the client as a
  parameter so the non-200 path is unit-testable.
- Non-200 responses return an error and log the status with a body snippet
  capped at 4KB via `io.LimitReader`, plus a warning when that cap is hit.
- On 200 the body is drained into `io.Discard` so the connection can be reused.

## Testing

`internal/server/telemetry_test.go` uses `httptest.Server` and zap's `observer`:
500 with an over-limit body, 400 with a short body, 200, transport error, and a
guard on the client timeout.

`go test ./internal/server/` and `go vet` pass. I could not run `golangci-lint`
locally (installed binary is go1.25, repo targets 1.26.4).

Supersedes #2225. Fixes #2203


shivansh-source:
@recharte PTAL

PR #3204, opened 2026-09-17
feat(rbac): enhance restore operations with RBAC enforcement and cluster support

### LINKED ISSUE #3189: RBAC: restore endpoints enforce nothing; backups/restores must be keyed by owning instance
### What is the issue?

Two gaps on the backup/restore surface, per spec 013 §4.5 (openeverest/specs#22). Both have working v1 counterparts to port.

**1. The flat restore endpoints check nothing.** `internal/server/handlers/rbac/restore.go:24-35` (branch `main`) is pure pass-through:

```go
func (h *rbacHandler) CreateRestore(ctx context.Context, restore *backupv1alpha1.Restore) (*backupv1alpha1.Restore, error) {
	return h.next.CreateRestore(ctx, restore)
}
```

Any authenticated user can create, get, and delete restores — and a restore overwrites the target instance's data. The handler interface also lacks the cluster parameter, so even a correct check has nothing to build the object from. (The nested list at `instance_restores.go:28-39` does enforce, so the gap is only the flat routes.)

**2. Backups are keyed by their own name instead of the owning instance's.** `backup.go:27-47` builds objects from the backup's name:

```go
object := rbac.ClusterNamespacedObjectName(cluster, namespace, name)
```

Backup names are usually generated (`generateName`), so nobody can write a grant that names them. The spec keys both backups and restores by the owning instance.

**v1 reference** — `internal/server/handlers/rbac/database_cluster_restore.go` (branch `v1.x`) does both things right, in every method: it reads the restore first when the route only carries the restore's name, then enforces on `restore.Spec.DBClusterName` (Get: lines 72-78, Delete: lines 60-66, Create from the request body: lines 48-51). That read-first-then-check shape is exactly the spec's fetch-then-enforce carve-out.

### Steps to Reproduce

1. Authenticate as a user with no grants at all.
2. `POST /v1/clusters/{c}/namespaces/{ns}/restores` targeting any instance.
3. The restore is created. It should return 403.

### Relevant Logs / Output

_No response_

### OpenEverest Version

main (v2 Developer Preview)

### Installation Method

Helm

### Kubernetes Distribution

any

### Environment Details

_No response_

### Possible Solution

Port the v1 pattern with the v2 object shape (`cluster/namespace/instance`):

- Add the cluster parameter to the restore handler interface (all three layers: rbac, validation, k8s).
- `CreateRestore`: enforce `restores:create` on the target instance from the request body — v1 `database_cluster_restore.go:48-51`.
- `GetRestore`/`DeleteRestore` and `GetBackup`/`DeleteBackup`: fetch the object first (with the server's identity), take the owning instance from its spec, then enforce — v1 lines 60-78. The fetched object is used only for the check: never return it or mention it in the error, and map both "not found" and "denied" to the same 403 so names cannot be probed.
- `CreateBackup` (`backup.go:36-38`): key on `backup.Spec.InstanceRef.Name` instead of `backup.GetName()`.
- Update `backup_test.go`/`instance_restores_test.go` and add flat-route tests asserting the 403s.

### Additional Context

Part of the spec 013 enforcement work (openeverest/specs#22). The restore derived checks (`restores:create` on the target plus `backups:read` on the source backup's instance — v1's `enforceDBRestore`, `database_cluster_restore.go:96`) are tracked in #3187 and should land on top of this wiring.

VijetaPriya47:
closes #3189


VijetaPriya47:
@recharte could you please check it out.

VijetaPriya47:

VijetaPriya47:
@recharte could you check it out why it's failing?

internal/server/handlers/rbac/backup.go
```diff
@@ -17,33 +17,63 @@ package rbac
...
 func (h *rbacHandler) GetBackup(ctx context.Context, cluster, namespace, name string) (*backupv1alpha1.Backup, error) {
-	object := rbac.ClusterNamespacedObjectName(cluster, namespace, name)
+	backup, err := h.next.GetBackup(ctx, cluster, namespace, name)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil, ErrInsufficientPermissions
+		}
+		return nil, fmt.Errorf("GetBackup failed: %w", err)
+	}
+	object := rbac.ClusterNamespacedObjectName(cluster, namespace, backupInstanceName(backup))
```
recharte, line 50: The flat routes now key on the owning instance, but `ListInstanceBackups` (`instance_backup.go:35`) and `ListInstanceRestores` (`instance_restores.go:35`) still key on each item's own name:

```go
object := rbac.ClusterNamespacedObjectName(cluster, b.GetNamespace(), b.GetName())
```

So the two routes for the same object now disagree. With `p, role:x, backups, read, prod/ns1/inst-1`, reading one backup works but listing that instance's backups returns `[]` — and backup names are generated, so no grant can ever match them.

Both routes already receive the instance name, so one check before the loop is enough:

```go
object := rbac.ClusterNamespacedObjectName(cluster, namespace, instance)
if err := h.enforce(ctx, rbac.ResourceBackups, rbac.ActionRead, object); err != nil {
	return nil, err
}
return h.next.ListInstanceBackups(ctx, cluster, namespace, instance)
```

(Commenting here because `instance_backup.go` isn't in this diff.)

internal/server/handlers/rbac/restore.go
```diff
@@ -16,21 +16,55 @@ package rbac
...
+	}
+	return restore, nil
 }
 
-// CreateRestore creates a new restore.
-func (h *rbacHandler) CreateRestore(ctx context.Context, restore *backupv1alpha1.Restore) (*backupv1alpha1.Restore, error) {
-	return h.next.CreateRestore(ctx, restore)
+// CreateRestore creates a new restore, gated by RBAC on the instance it targets.
+func (h *rbacHandler) CreateRestore(ctx context.Context, cluster string, restore *backupv1alpha1.Restore) (*backupv1alpha1.Restore, error) {
+	object := rbac.ClusterNamespacedObjectName(cluster, restore.GetNamespace(), restore.Spec.InstanceRef.Name)
```
recharte, line 48: `RestoreSpec.InstanceRef` is a value, not a pointer, so a body with no `instanceRef` decodes to the zero value and this builds `prod/ns1/` — which `globMatch` matches against any wildcard grant in the namespace. RBAC is the first handler in the chain (`everest.go:391`), so `validateRestoreRefs` hasn't rejected it yet.

Not an escalation today, since an empty name only matches grants that were already namespace-wide, but the check is silently meaningless for that request. An authz handler with no object to check should fail closed:

```suggestion
	if restore.Spec.InstanceRef.Name == "" {
		return nil, ErrInsufficientPermissions
	}
	object := rbac.ClusterNamespacedObjectName(cluster, restore.GetNamespace(), restore.Spec.InstanceRef.Name)
```

Worth a test case alongside it — the existing `namespace wildcard` case passes either way.

recharte (changes requested):
Thanks @VijetaPriya47, overall looks good.

The failing `Check` isn't your fault: our workflow uses a reviewdog reporter name that doesn't exist, and it only breaks fork PRs. Fix is in #3245 — merge `main` once that lands and it should go green.

One thing to sort out: the flat routes now key on the instance, but the nested list routes still key on each object's own name, so someone who can read a backup gets an empty list for that instance. Left a comment inline, plus a smaller note on `CreateRestore`.

PR #3220, opened 2026-09-20
fix(controller): guard against nil PMM spec in genVMAgentSpec

Norway-02:
## What

In `genVMAgentSpec` (`internal/controller/monitoring/monitoringconfig_controller.go`), when iterating over `MonitoringConfig` items matching `PMMMonitoringType`, the code directly dereferenced `mc.Spec.PMM.URL` and `mc.Spec.PMM.VerifyTLS` without checking if `mc.Spec.PMM` is nil.

Because `PMM *PMMMonitoringSpec` is an optional pointer (`+optional`, `omitempty`), any `MonitoringConfig` resource with `type: pmm` and an omitted or nil `pmm` block causes a nil pointer dereference panic during reconciliation:

```
```

This crashes the reconciliation loop for all monitoring configurations in the cluster. Everywhere else in the controller (e.g. lines 234, 272, 632), `mc.Spec.PMM` is guarded before dereferencing.

This change adds an explicit nil check to safely ignore configs with an unpopulated `pmm` spec when generating VMAgent remote writes.

## Verification

- `gofmt -w internal/controller/monitoring/monitoringconfig_controller.go`
- `go vet ./internal/controller/monitoring/...`
- `go test -v -cover ./internal/controller/monitoring/...`
- `go tool golangci-lint run --new-from-merge-base=origin/main ./internal/controller/monitoring/... --concurrency=2`


Norway-02:
cc @recharte - please take a look when you have a moment. Thanks!

PR #3222, opened 2026-09-21
[BREAKING CHANGE] v2 time based `retention` policy

AdityaPimpalkar:
Ref #3072

The breaking change includes removing `Instance.spec.backup.storages[].schedules[].retentionCopies` in favour of `Instance.spec.backup.storages[].schedules[].retention`. 

Obviously this will affect existing providers which have backup enabled, but I think it's worth the effort, otherwise we'll have to introduce obscure and confusing translation logic to translate `retentionCopies` count -> `retention` object.

I am skipping the UI schema in this because the PR already too big for review but it will be done in the follow up PR 👍 

Feedback is appreciated!


provider-runtime/controller/backup_validation.go
```diff
@@ -162,6 +183,93 @@ func ValidateInstanceBackupAgainstClass(in *corev1alpha1.Instance, bc *backupv1a
...
+// Instance:
+//   - retention is required
+//   - type/count/duration shape matches the CRD rules (defense in depth)
+//   - when the BackupClass declares supportedRetentionTypes, the schedule's
+//     retention.type must appear in that list
+//
+// Empty supportedRetentionTypes means both count and time are allowed. A nil
+// or Job-mode BackupClass skips the capability check but still validates
+// shape. Safe to call with nil inputs.
+func ValidateInstanceBackupRetention(in *corev1alpha1.Instance, bc *backupv1alpha1.BackupClass) error {
```
recharte, line 196: With `supportedRetentionTypes` gone, please remove `ValidateInstanceBackupRetention` too, along with its sentinels, reasons, reconciler mapping and tests. The CRD's enum, pattern and CEL rules already reject everything the shape checks cover before this code runs.

api/backup/v1alpha1/backupclass_types.go
```diff
@@ -157,8 +157,31 @@ type ProviderManagedSpec struct {
...
+
+	// SupportedRetentionTypes lists the schedule retention modes this class
+	// accepts (count and/or time). Empty means both are allowed. The runtime
+	// rejects Instance schedules whose retention.type is not in this list;
+	// providers declare the subset their engine supports (e.g. PSMDB/PXC:
+	// count; CNPG/Barman: time).
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=2
+	SupportedRetentionTypes []ScheduleRetentionType `json:"supportedRetentionTypes,omitempty"`
```
recharte, line 169: Could we leave `supportedRetentionTypes` out of this PR? I'd like backup capabilities to follow `TopologyComponent.supportedFields` (#3057): one OpenAPI v3 schema over `InstanceBackupSpec` on the class, rather than a list per capability. That's more than this PR should carry, and I'd rather not ship a field we'd then remove. I'll open a follow-up issue for it.

Until then, a provider that can't do time-based retention rejects it in its own `Validate()`, the same way PXC already rejects more than one PITR-enabled storage. Dropping the field also drops `backupv1alpha1.ScheduleRetentionType`, which only mirrors the core enum.

api/core/v1alpha1/instance_types.go
```diff
@@ -296,12 +335,11 @@ type InstanceBackupSchedule struct {
...
-	// schedule. Zero (or unset) means "keep all". Negative values are
-	// rejected.
-	// +kubebuilder:validation:Minimum=0
-	// +optional
-	RetentionCopies int32 `json:"retentionCopies,omitempty"`
+	// Retention configures count-based or time-based backup retention for
+	// this schedule. Required on every schedule; use type=count with
+	// count unset or 0 to keep all backups.
+	// +kubebuilder:validation:Required
+	Retention *BackupScheduleRetention `json:"retention"`
```
recharte, line 342: Could we make `retention` optional, with nil meaning keep all? That's what an unset `retentionCopies` meant, so schedules that don't care about retention don't have to mention it.

```go
	// Retention configures count-based or time-based backup retention for
	// this schedule. Unset keeps all backups.
	// +optional
	Retention *BackupScheduleRetention `json:"retention,omitempty"`
```

Then `count` can be required with `Minimum=1` when `type` is `count`, so there's only one way to say keep all.

recharte (changes requested):
Hey @AdityaPimpalkar, thanks for picking this up, and for raising the breaking change up front! Dropping `retentionCopies` is fine by me: v2 is a Developer Preview, and translating count to time would be worse to live with.

One thing blocks: FE Checks and FE E2E are red. The regenerated TS types drop `retentionCopies`, but ~20 files under `ui/apps/everest` still use it, so the UI part can't wait for the follow-up PR. Could you make the minimal change here (map the existing copies field to `retention: { type: 'count', count }` and back) and leave the time-based form for later?

The `Check` failure isn't yours. It's a reviewdog reporter bug on fork PRs, fixed on `main` in #3245. Please rebase so lint and the codegen drift check actually run. Locally, `go tool golangci-lint run --new-from-merge-base=origin/main` finds two issues in the new lines: a `cyclop` one that goes away with `ValidateInstanceBackupRetention`, and a `godot` one at `instance_types.go:282`.

